PR #1, opened 2023-07-22
added wagmi-->ethers signer conversion for sdk compatibility

Hugo0:
Added this code:
image

Works now:
image

SDK consumes an ethers signer object for now. AND it has to be ethersv6.

Not sure how to make the sdk compatible with wagmi walletclient, ethersv5, web3js, viem, etc out of the box. Any ideas appreciated :)

Further docs:
- https://wagmi.sh/react/ethers-adapters
- https://github.com/wagmi-dev/wagmi/blob/056b7fd8418c2764c818c94e6d7ef605b89b6c44/docs/pages/core/ethers-adapters.en-US.mdx#L118-L119

Additional comments:
- i have a lot of hiccups with socket timing out, but it might be the bad internet at ethcc
- i tried a few times and the link generation only worked when I selected polygon network on metamask AND on the ui. Seemed a bit finnicky/clunky to get that right. Also, it didn't let me do it in goerli even though i have goerli eth in my wallet.
- full flow doesn't work, leaving that to you. link is only console logged for now

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "@web3modal/react": "^2.7.0",
     "autoprefixer": "^10.4.14",
     "encoding": "^0.1.13",
+    "ethers": "^6.6.4",
     "jotai": "^2.2.2",
     "lokijs": "^1.5.12",
     "next": "13.4.10",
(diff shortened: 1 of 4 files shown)


borcherd:
Weird vercel build errors, build does succeed locally so ignoring this for now.

borcherd (approved):

PR #3, opened 2023-08-06
hotfix: sdk version update

Hugo0:
Spent some time refactoring how gas is handled in the sdk. Latest version is v0.0.61
I tested out with this repo and:
- native token sending works
- erc20 token sending works

Notable errors:
- usdc & some other erc20s fail. This is because tokenDecimals needs to be passed to the createLink() function, else it assumes it's 18 decimals, and since usdc only has 6 bad things happen :(

### Sidenotes

There is a dependency conflict between `@uauth/js` and `@uauth/web3modal`:
```
$ npm install
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: peanut-ui@0.1.0
npm ERR! Found: @uauth/js@2.8.0
npm ERR! node_modules/@uauth/js
npm ERR!   @uauth/js@"^2.8.0" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer @uauth/js@"2.7.0" from @uauth/web3modal@2.7.0
npm ERR! node_modules/@uauth/web3modal
npm ERR!   @uauth/web3modal@"^2.7.0" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR! 
npm ERR! 
npm ERR! For a full report see:
npm ERR! /home/hugo/.npm/_logs/2023-08-05T23_47_31_873Z-eresolve-report.txt

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/hugo/.npm/_logs/2023-08-05T23_47_31_873Z-debug-0.log
```
Have to use `npm install --force` to install the dependencies.

Also:
Getting a lot of NPM warnings when installing
image

### Feedback on Flow:
Some things I noted whilst trying to get the gas thing working:
- doesn't remember connection between sessions
- when i set stuff up (ie selecting goerli and selecting a token), after connecting it overwrites what I set up
- on page refresh I have to connect again?
- dashboard link seemed to be broken 
- "transaction speedup" on metamask doesn't work. Not sure how to handle this (maybe in the sdk? How?)
- page refresh whilst transaction is mining should give alert or be interrupted (don't wanna lose link!)
- connect modal is mobile modal, no? - not desktop?
- have to pass in tokenDecimals to sdk when calling createLink() (initial.view.tsx). Else it will fail for any token that doesn't have 18 decimals
- waiting for api.socket.tech seems to make refreshes/page changes hang a lot - maybe it's not necessary to wait for it?

### Further considerations
There seems to be an occasional error with `app-index.js:31 Failed to fetch gas price from provider: TypeError: provider.getFeeData is not a function`, but I haven't been able to debug it so far. Only happens sometimes.


borcherd:
I've implemented the tokenDetails and works as expected now. Also made some updates regarding your feedback:

- Does remember walletconnection between sessions and on refresh now. No need to reconnect everytime, it autoconnects now.
- I've removed the Uauth dependencies, they arent needed anymore. Thanks for telling me this, I always use yarn install and this didn't give me any errors. The warnings are just warnings caused by peer dependencies from certain packages that we use, nothing to worry about there. 
- Removed the overwriting of the chain, token and amount when the wallet get connected. This was relevant when we only showed the balances the user had, not anymore. 
- I haven't made the dashboard page yet, so that link does indeed not work yet. 
- Good call on the alert when refreshing so the link doesnt get lots. I've added in a 'are you sure you want to leave this site' alert when the user refreshes.
- The walletConnect modal is not a mobile version, that's just the way their modal looks. The mobile version is pretty similar but is a drawer coming fro the bottom of the screen.
- I'll have a look at the socket.tech api calls and at the txn speedup in metamask


I am still getting the same `transaction_underpriced' error when I try to send matic on polygon mainnet using metamask, like I sent you in the video. 
Also getting a new error when I try to send usdc on polygon mainnet (the image below). Happening on 
Screenshot 2023-08-06 at 14 55 32

Hugo0:
Huge!

I'll pull and try to debug the gas price issue. Not sure why it's sometimes failing, works with Node and other type of providers. Perhaps walletconnect related? I'll see

Weird you still get the tx underpriced error, that should be fixed now. Will investigate :monocle_face:

Hugo0:
Added fix for gas approval. Tested a few times and seemed to work. Can you try?

### Secondary considerations
- through testing a few times it was waiting for rpc.walletconnect.com for a long time (e.g. 30 secs or more). Not sure if artefact of my walletconnection or smth local.
- when connecting for first time (or since last session), it should probably autoswitch selected network to whatever the active network is in the users wallet
image

Hugo0:
Just a few other small things (mostly cosmetic):

- when a token isn't selected, the button is effed:
image

- network selection and token selection is just plain text - can add the token Icon and chain Icon. Chain details sdk object includes the following field:
"icon": [
            {
                "url": "ipfs://QmdwQDr6vmBtXmK2TmknkEuZNoaDqTasFdZdu3DRw8b2wt",
                "width": 1000,
                "height": 1628,
                "format": "png"
            }
        ],
Can load the icons from there (replace ipfs by an ipfs provider, e.g. https://ipfs.io/ipfs/QmdwQDr6vmBtXmK2TmknkEuZNoaDqTasFdZdu3DRw8b2wt)

- I get these scrolling bars on my screen: 
image

Any idea what's causing them?

- On mobile, the SDK banner is way too big compared to navbar
image

- also on mobile, the peanut guy is kinda just awkwardly hovering there:
image

borcherd:
I think its just something with my metamask, I'm still having the same errors that the tx is underpriced, but only with metamask and not on any other wallet, so I think it's good. (the errors dont occur when I change the gas to smt else)

- I'll have a look at the rpc.walletconnect.com, but haven't had any problems with that before.
- If we do the autoswitching to the connected network in the wallet, we'll have to reset the form (token and amount), else we'll get some weird combinations. Is that okay or would you rather leave it as is? 
- True, I thought I had fixed the weird looking button but clearly not lol. -> done
- I'll add in the chain logo and token logo in the dropdowns -> with the select object that we are using, you cannot add an image to each option. I'll have a look on how I can do this
- Not sure why you are getting those scrollbars, but I'll have a look. (don't have them, neither on chrome, safari or mobile)-> found the issue, is fixed now :) 
-  I'll make the changes on mobile too -> done

borcherd:
I've found a different library to use as a select dropdown, but not every chainDetail item has an icon object. Could you have a look? Also, is there a preference/recommendation for an ipfs provider? Don't have too much experience with ipfs yet.
Also, if we do the autoswitching to the connected network in the wallet, we'll have to reset the form (token and amount), else we'll get some weird combinations. Is that okay or would you rather leave it as is?

 @Hugo0

borcherd:
I'm not sure if we should add in the logo of the token on the dropdown. if the chain is mainnet i.e, the logo's of close to 4000 tokens need to be downloaded. The website is already pretty heavy because of the static token list, I think adding in the logo's of the tokens might be too much. We can do the logo's of the chains though since they are way less.

Hugo0:
> Also, if we do the autoswitching to the connected network in the wallet, we'll have to reset the form (token and amount), else we'll get some weird combinations. Is that okay or would you rather leave it as is?

- do we need a library for this? Can't we create a dropdown ourselves as a component?

- ipfs: you can just link take the ipfs content id (CID) and direct to the ipfs.io gateway like this:  
`QmdwQDr6vmBtXmK2TmknkEuZNoaDqTasFdZdu3DRw8b2wt` --> https://ipfs.io/ipfs/QmdwQDr6vmBtXmK2TmknkEuZNoaDqTasFdZdu3DRw8b2wt
That should link to the image icon

> Also, if we do the autoswitching to the connected network in the wallet, we'll have to reset the form (token and amount), else we'll get some weird combinations. Is that okay or would you rather leave it as is?

- why not have smth like a flag "hasFormBeenTouched". If the user has changed anything in the form (e.g. set the amount already, or changed the network), then we don't touch it. But if they are just connecting, then we can autoswitch to their current network.

> I'm not sure if we should add in the logo of the token on the dropdown. if the chain is mainnet i.e, the logo's of close to 4000 tokens need to be downloaded. The website is already pretty heavy because of the static token list, I think adding in the logo's of the tokens might be too much. We can do the logo's of the chains though since they are way less.

- Fair point, wondering how we could do this better. For a future mature implementation, i think it's definitely needed and just looks much better for the users. E.g. see uniswap token select interface:
image

Ideally, we'd lazy load the token icons, and only have them be requested once the user scrolls down or searches and they're in view.

PR #16, opened 2023-08-24
Milkomeda lp

0xkkonrad:
added milkomeda LP. should be launched, but could get a few small styling fixes 
1. reducing top gap (the weird blue space, no idea where it comes from?)


borcherd:
Merging this directly to main, will merge main into develop afterwards

PR #43, opened 2023-10-10
Develop

borcherd:
AASA file and improved 404


borcherd:
what has been done: 
- cleaning
- universal linking files: IOS and Android (AASA and assetLinks.json)
- transferWtf page
- added linea

PR #110, opened 2024-01-24
Web3inbox fix 500

thegodfatherbeeearth:
1. Don't return 500 error in all calls to the frontend's /api/notify
2. Send web3inbox notifications on x-chain & multilink claiming

### DIFF
--- a/src/app/api/notify/route.ts
+++ b/src/app/api/notify/route.ts
@@ -27,12 +27,18 @@ export async function POST(request: NextRequest, response: NextResponse) {
             },
             body: JSON.stringify(notificationPayload),
         })
+
+        return NextResponse.json({
+            status: 200,
+        })
     } catch (error: any) {
         console.error('Error occured while sending notification:', error)
-        throw new Error('Error occured while sending notification')
-    }
 
-    return NextResponse.json({
-        status: 200,
-    })
+        return NextResponse.json({
+            status: 500,
+            body: {
+                error: 'Error occured while sending notification',
+            },
+        })
+    }
 }
--- a/src/app/api/notify/route.ts
+++ b/src/app/api/notify/route.ts
@@ -1,7 +1,7 @@
 import { NextRequest, NextResponse } from 'next/server'
 const projectId = process.env.WC_PROJECT_ID
 if (!projectId) {
-    throw new Error('You need to provide NEXT_PUBLIC_PROJECT_ID env variable')
+    throw new Error('You need to provide WC_PROJECT_ID env variable')
 }
 
 export async function POST(request: NextRequest, response: NextResponse) {
@@ -28,6 +28,11 @@ export async function POST(request: NextRequest, response: NextResponse) {
             body: JSON.stringify(notificationPayload),
         })
     } catch (error: any) {
+        console.error('Error occured while sending notification:', error)
         throw new Error('Error occured while sending notification')
     }
+
+    return NextResponse.json({
+        status: 200,
+    })
 }
--- a/src/components/claim/multilinkViews/multilinkSuccess.view.tsx
+++ b/src/components/claim/multilinkViews/multilinkSuccess.view.tsx
@@ -7,9 +7,11 @@ import * as global_components from '@/components/global'
 import * as utils from '@/utils'
 import dropdown_svg from '@/assets/dropdown.svg'
 import { useRouter } from 'next/navigation'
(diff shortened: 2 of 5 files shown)


src/app/api/notify/route.ts
```diff
@@ -28,6 +28,11 @@ export async function POST(request: NextRequest, response: NextResponse) {
             body: JSON.stringify(notificationPayload),
         })
     } catch (error: any) {
+        console.error('Error occured while sending notification:', error)
         throw new Error('Error occured while sending notification')
     }
+
+    return NextResponse.json({
+        status: 200,
+    })
```
Hugo0: Why is this necessary? Because fetch() needs a 200 response else it raises a 500 err? (just trying to understand)

src/utils/notifications.utils.ts
```diff
@@ -5,8 +5,26 @@ interface INotification {
...
+export const sendNotification = async (
+    senderAddress: string,
+    recipientAddress: string | undefined,
+    chainName: string
+) => {
+    console.log('sendNotification', senderAddress)
+    const accounts = [`eip155:1:${senderAddress}` ?? '']
+    const notification: INotification = {
+        title: 'Peanut Protocol',
+        body: `Your link has been claimed on ${chainName} by ${recipientAddress ?? ''}`,
```
Hugo0, line 17: This is not part of this PR, but imo I'd also add the USD amount / tokenamount to the message.

"your 10 usdc link ...."

Imagine if you sent multiple links, how does this tell you which one it is?

src/components/claim/views/success.view.tsx
```diff
@@ -35,25 +35,12 @@ export function ClaimSuccessView({ txHash, claimDetails, senderAddress }: _const
...
-            body: `Your link has been claimed on ${chainName} by ${address ?? ''}`,
-            url: undefined,
-            type: '2aee6e5f-091d-444e-96cd-868ba2ddd0e7',
-        }
-        utils.sendNotification({
-            notification,
-            accounts,
-        })
+    const sendNotification = async () => {
+        const chainName = chainDetails.find((detail) => detail.chainId === claimDetails[0].chainId)?.name
```
Hugo0, line 42: Given that chainDetails doesn't update on success view, and this line is repeated: `chainDetails.find((detail) => detail.chainId === claimDetails[0].chainId)?.name`

Why not make it a const?

src/components/claim/multilinkViews/multilinkSuccess.view.tsx
```diff
@@ -23,8 +25,14 @@ export function multilinkSuccessView({ txHash, claimDetails }: _consts.IClaimScr
 
     useEffect(() => {
         router.prefetch('/send')
+        sendNotification()
     }, [])
 
+    const sendNotification = async () => {
+        const chainName = chainDetails.find((detail) => detail.chainId === claimDetails[0].chainId)?.name
```
Hugo0, line 32: sweet! thanks for thinking of multilinks in regards to web3inbox! good thinking

src/utils/notifications.utils.ts
Hugo0: Also, good on splitting this out, since we have multiple success views. Sweet.

src/app/api/notify/route.ts
thegodfatherbeeearth: I think it's how nextjs works, not related to `fetch`. Function must return something successful, otherwise next thinks something went wrong and returns a 500 to the caller

src/utils/notifications.utils.ts
thegodfatherbeeearth: Yeah, agreed! I was also thinking about this. Created a backlog task with owner = Bo

src/components/claim/views/success.view.tsx
thegodfatherbeeearth: Honestly, dunno. From what I've found, chainDetails in UI is always used this way: `const [chainDetails] = useAtom(store.defaultChainDetailsAtom)`, and I thought it's better to not make any changes to the approach without Bo.

src/app/api/notify/route.ts
borcherd: Nextjs wants all api calls to happen from the api folder. This is handles as an actual 'next api', so also expects an next api response. Thanks for introducing this alexey

PR #147, opened 2024-02-10
Develop

Hugo0:


src/app/create-giga-packet/page.tsx
borcherd, line 1: The title and description shouldnt be hardcoded here imo, will always say mantle rn

src/components/dashboard/dashboard.tsx
borcherd, line 1: Should also update the dashboard entry UI then to display smt different than 0

src/components/packet/packet.tsx
```diff
@@ -39,83 +41,47 @@ export function Packet() {
         }))
     }
 
-    async function fetchLeaderboardInfo(link: string) {
```
borcherd, line 42: Not sure why this got removed, will add it back in. If we do this on the success/empty screen itself, it'd just spawn in after being rendered, thats why it should be done here.

src/components/packet/views/packetSucces.view.tsx
```diff
@@ -7,30 +7,45 @@ import * as utils from '@/utils'
...
+                baseUrl: `${consts.next_proxy_url}/get-raffle-leaderboard`,
+                APIKey: 'doesnt-matter',
+            }).then(setLeaderboardInfo)
+        }
     }, [])
 
+    const userLeaderboard = leaderboardInfo?.find((user) => user.address == address)
+    console.log(userLeaderboard)
+    console.log(raffleInfo)
+    console.log(raffleClaimedInfo)
```
borcherd, line 39: Bad job bo, remove this before u commit

src/consts/general.consts.ts
```diff
@@ -10,4 +10,7 @@ export const ipfsProviderArray = [
     'https://gw3.io/ipfs/',
 ]
 
-export const peanut_api_url = 'https://api.peanut.to'
+// TODO: get from env
+// export const peanut_api_url = 'https://api.peanut.to'
+export const peanut_api_url = 'https://api.staging.peanut.to/'
```
borcherd, line 15: remove from here, only define in one place

borcherd (commented):
(nothing major, just added comments for me so i know what to fix)

PR #159, opened 2024-02-11
Use getUserRaffleStatus in UI

thegodfatherbeeearth:
Integrate latest sdk changes to use getUserRaffleStatus function instead of hasAddressParticipatedInRaffle.

Needed to make things faster!


Hugo0 (commented):
lgtm :+1: 
@borcherd plz okay

PR #167, opened 2024-02-12
Token selection dropdown

thegodfatherbeeearth:
Adds a token selection dropdown that automatically fetches the list of available tokens.


thegodfatherbeeearth:
Canceling as it's cool, but an overkill :)

PR #183, opened 2024-02-13
bugfix

borcherd:


src/components/giga-packet/gigaPacket.tsx
```diff
@@ -1009,11 +1015,18 @@ export function GigaPacket() {
                         // prepare the transactions. It is calculated based on the number of slots and the max transactions per block
                         for (let index = 0; index <= quotient; index++) {
                             const numberofLinks = index != quotient ? _consts.MAX_TRANSACTIONS_PER_BLOCK : remainder
+
+                            let lastTx = false
+
+                            if (index == quotient - 1 && remainder == 0) {
```
thegodfatherbeeearth, line 1021: If remainder is zero, won't it mean that we are at the point where no transactions need to be made? (i.e. too late)

PR #217, opened 2024-02-29
Replace socket with mobula

borcherd:


borcherd:
Bo to make utils out of fetchTokenDetails

PR #237, opened 2024-03-13
bugfixing sesh

borcherd:
Changelog: 
- Bump sdk version: updated tokenList and getDefaultProvider error handler
- Updated mobula api call: only requesting chainIds that we both support (mobula & peanut) and not requesting pnl
- fixed inconsistent token price fetching: only fetching when chain is supported by mobula (mainly for not fetching on testnets) and only setting the tokenprice when the call isnt outdated (quick token switching)
- claiming to ens name fixed
- some error message updates
- some copy changes
- some aesthetic ui fixed: no more floating card on mobile, no border on app in header on hover, smaller copy button, zindex changes on tokendenomination switch icon & modal
- minimumamount of 0,00001 when sending. -> Temp fix, is task in backlog to do this in the SDK
- remove unused stuff (ens name fetching, caused some unwanted waittimes)
- fixed duplicate addresses in tokendetails
- small lp changes: smaller logos on mobile and stats copy changes
- retry functionality for mobula walletbalance calls
- ...


src/store/store.tsx
```diff
@@ -226,40 +226,62 @@ export function Store({ children }: { children: React.ReactNode }) {
...
+                            const mobulaResponseJson = await mobulaResponse.json()
 
-                const mappedUserBalances = mapToUserBalances(mobulaResponseJson.data).sort(
-                    (a: interfaces.IUserBalance, b: interfaces.IUserBalance) => {
-                        if (a.chainId === b.chainId) {
-                            if (a.address.toLowerCase() === '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee') return -1
-                            if (b.address.toLowerCase() === '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee') return 1
+                            const mappedUserBalances = mapToUserBalances(mobulaResponseJson.data).sort((a, b) => {
+                                if (a.chainId === b.chainId) {
+                                    if (a.address.toLowerCase() === '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee')
```
thegodfatherbeeearth, line 258: Is it possible that `0x00..00` address will be used for eth? Should we check for it as well?

src/store/store.tsx
```diff
@@ -226,40 +226,62 @@ export function Store({ children }: { children: React.ReactNode }) {
...
-                        address,
-                    }),
-                })
+                const chainIds = defaultChainDetails
+                    .filter((detail) => detail.mainnet)
+                    .filter((detail) => supportedMobulaChains.map((chain) => chain.chainId).includes(detail.chainId))
+                    .map((detail) => detail.chainId)
+                    .join(',')
+
+                let attempts = 0
```
thegodfatherbeeearth, line 235: In the future I would encapsulate the retry logic in a dedicated place and reuse for everything. Maybe even put it in the sdk and reuse for both ui and the api

src/components/send/sendViews/initial.view.tsx
```diff
@@ -1275,7 +1318,7 @@ export function SendInitialView({ onNextScreen, setClaimLink, setTxHash, setChai
                                               ))
                                             : tokenList.map((token) => (
                                                   <div
-                                                      key={token.address}
+                                                      key={token.address + Math.random()}
```
thegodfatherbeeearth, line 1321: 🤨 Why to do random here?

src/components/send/sendViews/initial.view.tsx
```diff
@@ -1261,6 +1301,9 @@ export function SendInitialView({ onNextScreen, setClaimLink, setTxHash, setChai
                                                               src={token.logo}
                                                               className="h-6 bg-white"
                                                               loading="eager"
+                                                              onError={(e: any) => {
```
thegodfatherbeeearth, line 1304: What's the purpose of dis?

src/components/send/sendViews/initial.view.tsx
```diff
@@ -709,32 +739,36 @@ export function SendInitialView({ onNextScreen, setClaimLink, setTxHash, setChai
...
+            // if the user is connected and the forms hasnt been touched by the user, we set the chain to the current chain and the token to native token on that chain
             sendForm.setValue(
                 'token',
                 chainDetails.find((chain) => chain.chainId == currentChain.id)?.nativeCurrency.symbol ?? ''
             )
             sendForm.setValue('chainId', currentChain.id.toString())
-        } else if (chainsToShow.length > 0 && !formHasBeenTouched) {
-            sendForm.setValue('chainId', chainsToShow[0].chainId)
-            sendForm.setValue('token', chainsToShow[0].nativeCurrency.symbol)
+        } else if (userBalances.length > 0 && !userBalances.some((balance) => balance.chainId == formwatch.chainId)) {
```
thegodfatherbeeearth, line 762: I would personally question this entire logic of automatically switching to some chain. The problem is that the response from the mobula api arrives after a certain delay, which means that this switch also happens after a certain delay.

It has happened to me multiple times that I went to the app, saw the default of chain X (let's say Ethereum) being selected, I quickly change it to the chain that I need (let's say Arbitrum), but then automatically switches back to Ethereum, which is suuuuuuper confusing

src/components/claimLink/claim.tsx
```diff
@@ -174,7 +174,7 @@ export function Claim() {
                     setLinkState('ALREADY_CLAIMED')
                 } else {
                     const _tokenprice = await utils.fetchTokenPrice(linkDetails.tokenAddress, linkDetails.chainId)
-                    setTokenPrice(_tokenprice)
+                    setTokenPrice(_tokenprice?.price.toString())
```
thegodfatherbeeearth, line 177: What was the issue here before?

src/components/send/sendViews/initial.view.tsx
borcherd: There are some duplicate tokens in the tokenDetails which make for duplicate rows with the same key. The random value makes the key different. Basically just to suppress  the error in the console

src/components/send/sendViews/initial.view.tsx
borcherd: Ipfs sometimes ratelimits, this handles the error silently

src/components/claimLink/claim.tsx
borcherd: No issue, just changes in the utils function

src/components/send/sendViews/initial.view.tsx
borcherd: Fair point, will make new entry in the notion doc and address this in the next pr

src/store/store.tsx
borcherd: hmm yes, dont see why it should be in the sdk but could just be a utils function. Will make backlog task

src/store/store.tsx
borcherd: Fair point, will fix. This was because socket.tech (prev balance api) returned as 0xeee always, but should change it for mobula. tyy

PR #352, opened 2024-08-09
chore: replaced copy for meta preview

Hugo0:
@borcherd 

1. is src/components/Create/Link/Confirm.view.tsx and src/components/Create/Link/Input.view.tsx okay with this change?
2. Why do we have the the same code in 10+ places? Should we maybe standardize this or is it a bad idea? 
image
 

## Summary by CodeRabbit

- **New Features**
	- Updated page descriptions across multiple sections to emphasize "Text Tokens" instead of "Send crypto via link," indicating a shift in focus and branding.
	- Enhanced clarity in user interface components related to link creation, aligning terminology with new branding.

- **Bug Fixes**
	- Improved user instructions for link creation to enhance understanding and streamline the process.
	- Expanded support information to include EU accounts in the account type descriptions.


borcherd:
> 1. is src/components/Create/Link/Confirm.view.tsx and src/components/Create/Link/Input.view.tsx okay with this change?
> 2. Why do we have the the same code in 10+ places? Should we maybe standardize this or is it a bad idea?

Metadata should be set specifically per page, but yes we can create consts out of those values. @Hugo0

Hugo0:
Well then maybe correct solution is to change the page Metadata instead? And add specific titles to all our pages

PR #356, opened 2024-08-12
Profiles qa implementation

borcherd:
## Summary by CodeRabbit

- **New Features**
  - Enhanced error handling for username updates, providing specific feedback when a username conflict occurs.
  - The `Profile` component now dynamically generates referral lists and adapts to window size, improving user interaction.
  - The `TextEdit` component reflects updates to the initial text for a more responsive editing experience.

- **Bug Fixes**
  - Improved modal visibility controls to prevent unintended displays based on context.

- **Chores**
  - Updated avatar generation logic in the `Profile` component for a fresh visual representation.


src/app/api/peanut/user/update-user/route.ts
```diff
@@ -37,6 +37,15 @@ export async function POST(request: NextRequest) {
 
         const data = await response.json()
 
+        if (data.message === 'Username already exists') {
```
Hugo0: concern: this seems super brittle!!!
- if the error message ever changes in api, then this won't work. Even if we capitalize differently, this won't work!

Hugo0: Why do we set custom status instead of just forwarding the response status? e.g. if 404 we set 404. better to just forward it no?

Same with if username already exists, i think better to handle through status codes and by forwarding the response message

Hugo0: side nitpick: I think a high level comment for a function like this makes sense. As in, "this function does this and this. If this edge case, then this and this."

src/components/Global/TextEdit/index.tsx
```diff
@@ -1,17 +1,23 @@
...
+    const { user } = useAuth()
 
     useEffect(() => {
         if (isEditing && inputRef.current) {
             inputRef.current.focus()
         }
     }, [isEditing])
 
+    useEffect(() => {
+        setText(user?.user?.username || initialText)
```
Hugo0: nit: is this component just for username? in that case I'd rename setText, initialText to reflect that (e.g. setUsernameText etc)

src/components/Profile/Components/MobileTableComponent.tsx
```diff
@@ -131,36 +133,19 @@ export const MobileTableComponent = ({
...
-                            className="flex h-12 w-full items-center gap-2 px-4 text-h8 text-sm font-bold transition-colors last:mb-0 hover:bg-n-3/10 disabled:cursor-not-allowed disabled:bg-n-4 disabled:hover:bg-n-4/90 dark:hover:bg-white/20 "
-                        >
-                            Delete
-                        </div>
-                    </>
                 ) : (
-                    type === 'accounts' && (
-                        <div
-                            onClick={() => {
-                                console.log('Delete') // TODO: implement delete
```
Hugo0, line 158: thought: we'll have to implement delete down the line (when we implement linking accounts

src/components/Profile/index.tsx
```diff
@@ -2,7 +2,7 @@
 import Icon from '../Global/Icon'
 
 import { createAvatar } from '@dicebear/core'
-import { avataaarsNeutral } from '@dicebear/collection'
+import { funEmoji } from '@dicebear/collection'
```
Hugo0: TODO: let's use identicons from dicebear. The problem with a specific artstyle like these cartoon avatars is that they clash a lot with current peanut colors and it makes for an uncohesive whole

image

src/components/Profile/index.tsx
```diff
@@ -71,6 +70,23 @@ export const Profile = () => {
...
+
+    useEffect(() => {
+        calculateItemsPerPage()
+        window.addEventListener('resize', calculateItemsPerPage)
+
+        return () => {
+            window.removeEventListener('resize', calculateItemsPerPage)
+        }
+    }, [])
+
```
Hugo0: Thought: I think this isn't a _great_ approach, because:

- if user zooms, it changes how many items it shows

However, i can't think of anything significantly better

One suggestion though: add some buffer
```suggestion
    // Calculate the number of items that can be displayed on the page
    const calculateItemsPerPage = () => {
    const itemHeight = 100
    const availableHeight = window.innerHeight - 300
    const calculatedItems = Math.floor(availableHeight / itemHeight)
    // Only update if the change is significant
    if (Math.abs(calculatedItems - itemsPerPage) > 1) {
        setItemsPerPage(Math.max(calculatedItems, 1))
    }
}
    useEffect(() => {
        calculateItemsPerPage()
        window.addEventListener('resize', calculateItemsPerPage)
        return () => {
            window.removeEventListener('resize', calculateItemsPerPage)
        }
    }, [])

```

src/components/Profile/index.tsx
```diff
@@ -409,18 +425,41 @@ export const Profile = () => {
...
+                                                    user.pointsPerReferral?.find((ref) =>
+                                                        utils.compareTokenAddresses(
+                                                            ref.address,
+                                                            referral.account_identifier
+                                                        )
+                                                    )?.points ?? 0
+                                                )
+                                            )
+                                        }, 0)}
+                                    </label>
```
Hugo0, line 473: Nit: this seems a bit complex/confusing and i can't tell from a first glance what it is for. Would be useful to have a 1 line comment explaining what the label is for

src/context/authContext.tsx
```diff
@@ -62,15 +62,19 @@ export const AuthProvider = ({ children }: { children: ReactNode }) => {
...
-                }
-            } else {
-                throw new Error(response.statusText)
+            if (response.status === 409) {
+                toast.close(toastIdRef.current)
+                toastIdRef.current = toast({
+                    status: 'error',
+                    title: 'Username already exists',
+                }) as ToastId
+
```
Hugo0: Nit: same as before, i think we should be able to just reuse the message from the api instead of duplicating it in frontend

Hugo0 (approved):
Added some minor code quality comments. Please resolve and then merge, no further review necessary

src/components/Profile/index.tsx
borcherd: I thought this was what you suggested, having x amount per screen size? Sure i'll add buffer

src/app/api/peanut/user/update-user/route.ts
borcherd: In that case it requires api changes. Currently the api returns the following: 

status: 400
statusText: Bad Request
data.message: Username already exists. 

@pawell24 can you have a look in the api to return a decent error?

src/components/Profile/index.tsx
borcherd: IMO its not that complicated, just reducing the points in an array to get the total points. 

Also, dont like putting comments in html code, makes it v messy

PR #393, opened 2024-09-30
fix: fetchRouteRaw containing scientific notion string number

nezz0746:
A scientific number breaks the cashout flow. Using BigInt to make sure the Number is being formatted correctly to be sent to the SDK

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved precision in token amount calculations by switching to `BigInt`.
	- Enhanced error handling with clearer console messages for route fetching issues.


Hugo0 (approved):
Yeah always use Bigint math where possible. If you find other places in the code where we're not doing that, pls update too

merging

PR #404, opened 2024-10-02
fix(request): two bugs when preparing pay transaction

jjramirezn:
1. Because a case sensitive comparison two token addresses were treated as different when they should be equal

2. Because the token selector and the request component both depend on
selectedTokenAddress and selectedChainID, they will try to execute
their effects at the same time. This causes the request component to use
incorrect token price data. This commit fixes that by adding a
tokenPriceCompleted state to the token selector context, which is set to
true once the token price has been fetched.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new property, `isTokenPriceFetchingComplete`, to improve token price fetching and transaction fee estimation.
	- Enhanced address comparison logic by replacing `compareTokenAddresses` with `areTokenAddressesEqual` for improved accuracy in token selection across multiple components.

- **Bug Fixes**
	- Improved error messages for clearer user feedback during the claim process and token selection.

- **Documentation**
	- Updated utility functions for managing Ethereum address validation and local storage operations.

### DIFF
--- a/src/components/Request/Pay/Views/Initial.view.tsx
+++ b/src/components/Request/Pay/Views/Initial.view.tsx
@@ -28,7 +28,12 @@ export const InitialView = ({
     const { switchChainAsync } = useSwitchChain()
     const { open } = useWeb3Modal()
     const { setLoadingState, loadingState, isLoading } = useContext(context.loadingStateContext)
-    const { selectedChainID, selectedTokenAddress, selectedTokenDecimals } = useContext(context.tokenSelectorContext)
+    const {
+        selectedChainID,
+        selectedTokenAddress,
+        selectedTokenDecimals,
+        tokenPriceCompleted
+    } = useContext(context.tokenSelectorContext)
     const [errorState, setErrorState] = useState<{
         showError: boolean
         errorMessage: string
@@ -86,8 +91,19 @@ export const InitialView = ({
                 setTxFee('0')
             }
         }
+
+        // wait for token selector to fetch token price, both effects depend on
+        // selectedTokenAddress and selectedChainID, but we depend on that
+        // effect being completed first
+        if (!tokenPriceCompleted) return
+
         estimateTxFee()
-    }, [selectedTokenAddress, selectedChainID])
+    }, [
+        selectedTokenAddress,
+        selectedChainID,
+        selectedTokenDecimals,
+        tokenPriceCompleted
+    ])
 
     const handleConnectWallet = async () => {
         open()
--- a/src/components/Request/Pay/Views/Initial.view.tsx
+++ b/src/components/Request/Pay/Views/Initial.view.tsx
@@ -55,7 +55,10 @@ export const InitialView = ({
     useEffect(() => {
         const estimateTxFee = async () => {
             setLinkState(RequestStatus.LOADING)
-            if (selectedChainID === requestLinkData.chainId && selectedTokenAddress === requestLinkData.tokenAddress) {
+            if (
+                selectedChainID === requestLinkData.chainId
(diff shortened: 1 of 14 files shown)


src/components/Request/Pay/Views/Initial.view.tsx
```diff
@@ -55,7 +60,10 @@ export const InitialView = ({
     useEffect(() => {
         const estimateTxFee = async () => {
             setLinkState(RequestStatus.LOADING)
-            if (selectedChainID === requestLinkData.chainId && selectedTokenAddress === requestLinkData.tokenAddress) {
+            if (
+                selectedChainID === requestLinkData.chainId
+                && utils.compareTokenAddresses(selectedTokenAddress, requestLinkData.tokenAddress)
```
Hugo0: thought: this should prob be called `areTokenAddressesEqual` or equivalent or `isSameTokenAddress`

src/components/Request/Pay/Views/Initial.view.tsx
```diff
@@ -83,8 +91,19 @@ export const InitialView = ({
...
+        if (!tokenPriceCompleted) return
+
         estimateTxFee()
-    }, [selectedTokenAddress, selectedChainID])
+    }, [
+        selectedTokenAddress,
+        selectedChainID,
+        selectedTokenDecimals,
+        tokenPriceCompleted
+    ])
```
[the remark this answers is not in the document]
Hugo0: This seems sensible, no @jjramirezn ?

src/context/tokenSelector.context.tsx
```diff
@@ -21,6 +21,7 @@ export const tokenSelectorContext = createContext({
     refetchXchainRoute: false as boolean,
     setRefetchXchainRoute: (value: boolean) => {},
     resetTokenContextProvider: () => {},
+    tokenPriceCompleted: false as boolean,
```
Hugo0: nit: for clearness of understanding I'd prob rename this `isTokenPriceFetchingComplete`.

But it seems like an dangerous pattern to me

Hugo0 (approved):
nice - good description on PR. Approved, feel free to merge

src/context/tokenSelector.context.tsx
jjramirezn: I also was having doubts with this pattern, but I'm not sure of what refactor could I do to avoid the race condition between the two effects

jjramirezn: renamed

jjramirezn: renamed

src/components/Request/Pay/Views/Initial.view.tsx
jjramirezn: I agree, renamed

src/components/Request/Pay/Views/Initial.view.tsx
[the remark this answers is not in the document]
jjramirezn: Seems sensible, implemented

src/context/tokenSelector.context.tsx
```diff
@@ -109,6 +118,7 @@ export const TokenContextProvider = ({ children }: { children: React.ReactNode }
             setSelectedTokenAddress(prefs.tokenAddress)
             setSelectedChainID(prefs.chainId)
             setSelectedTokenDecimals(prefs.decimals)
+            setTokenPriceFetchingComplete(true)
```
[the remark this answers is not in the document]
jjramirezn: token price fetch complete indicates that we have token data available, so If that data comes from prefs we also set is as true

PR #406, opened 2024-10-03
fix: make FAQ's menu z-index  bigger

nezz0746:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the visibility of the FAQ menu by adjusting its stacking order, ensuring it appears above other elements.


Hugo0 (approved):

Hugo0:
Holistic thinking: is this the only toggle menu we have?

nezz0746:
@Hugo0 A few, but not seeing any that are anywhere near the header

PR #419, opened 2024-10-05
fix: close IFrameWrapper closing feat

nezz0746:
The IFrame own closing button was clashing with the modal closing button, that's probably why it was set to "hidden". But there was no way clear way to close it other than clicking out. Added a close button at the bottom
Screenshot 2024-10-05 at 10 33 16
Screenshot 2024-10-05 at 10 33 36

### DIFF
--- a/src/components/Global/IframeWrapper/index.tsx
+++ b/src/components/Global/IframeWrapper/index.tsx
@@ -32,22 +32,32 @@ const IframeWrapper = ({
         <Modal
             visible={visible}
             onClose={onClose}
-            classWrap="w-full max-w-2xl"
+            classWrap="w-full max-w-2xl border-none sm:border"
             classOverlay="bg-black bg-opacity-50"
             video={false}
+            /**
+             * Making sure the Iframe showing on top of the Crisp Chat widget
+             * which has z-index of 1000000
+             */
+            className="z-[1000001]"
             classButtonClose="hidden"
         >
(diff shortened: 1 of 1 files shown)


nezz0746:
Screenshot 2024-10-05 at 10 28 08
Same for TOS

Hugo0:
Looks good on desktop, but since we should develop mobile first:

image
image

1. The Iframe KYC  card comes from the bottom on mobile, so it is weird to add the close button at the bottom - Can we try to put it on the top?

2. can we get rid of the padding? Maybe it looks bad though. But if we can and it doesn't look bad, it'd be great, then users have more real estate on  mobile

src/components/Global/IframeWrapper/index.tsx
```diff
@@ -34,17 +32,37 @@ const IframeWrapper = ({
         <Modal
             visible={visible}
             onClose={onClose}
-            classWrap="w-full max-w-2xl"
+            classWrap="w-full max-w-2xl border-none sm:border"
             classOverlay="bg-black bg-opacity-50"
             video={false}
+            /**
+             * Making sure the Iframe showing on top of the Crisp Chat widget
+             * which has z-index of 1000000
```
Hugo0, line 40: ok we're getting into ridiculous z-index ranges ahahhhahahahha

Hugo0 (approved):

PR #423, opened 2024-10-07
fix: Propagate error when bridge account already exists for IBAN

panosfilianos:
Also adds example of Respone interface type

### DIFF
--- a/src/app/api/bridge/external-account/create-external-account/route.ts
+++ b/src/app/api/bridge/external-account/create-external-account/route.ts
@@ -68,6 +68,26 @@ export async function POST(request: NextRequest) {
         const data = await response.json()
 
         if (!response.ok) {
+            if (data.code && data.code == 'duplicate_external_account') {
+                // in case the bridge account already exists
+                //
+                // sending back error responses is not currently common across app 
+                // in how we deliver errors from backend -> frontend
+                // sends back an object like:
+                // {
+                //     id: '4c537540-80bf-41dd-a528-dbe39a4',
+                //     code: 'duplicate_external_account',
+                //     message: 'An external account with the same information has already been added for this customer'
+                //   }
+                //
+                // TODO: standardize error responses across backend
+                return new NextResponse(JSON.stringify(data), {
+                    status: response.status,
+                    headers: {
+                        'Content-Type': 'application/json',
+                    },
+                })
+            }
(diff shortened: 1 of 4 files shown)


src/app/api/bridge/external-account/create-external-account/route.ts
```diff
@@ -68,6 +68,26 @@ export async function POST(request: NextRequest) {
...
+                // sending back error responses is not currently common across app 
+                // in how we deliver errors from backend -> frontend
+                // sends back an object like:
+                // {
+                //     id: '4c537540-80bf-41dd-a528-dbe39a4',
+                //     code: 'duplicate_external_account',
+                //     message: 'An external account with the same information has already been added for this customer'
+                //   }
+                //
+                // TODO: standardize error responses across backend
```
Hugo0, line 83: praise: nice comment!

src/components/Global/LinkAccountComponent/index.tsx
```diff
@@ -193,7 +206,7 @@ export const GlobaLinkAccountComponent = ({ accountNumber, onCompleted }: IGloba
             onCompleted ? onCompleted() : setCompletedLinking(true)
         } catch (error) {
             console.error('Error during the submission process:', error)
-            setErrorState({ showError: true, errorMessage: 'An error occurred. Please try again later' })
+            setErrorState({ showError: true, errorMessage:  String(error)})
```
Hugo0, line 209: suggestion: Shouldn't we use error.message? String(error) is gonna be gibberish

src/utils/cashout.utils.ts
```diff
@@ -187,13 +187,36 @@ export async function createExternalAccount(
             }),
         })
 
+        const data = await response.json()
+
         if (!response.ok) {
+            if (data.code && data.code == 'duplicate_external_account') {
```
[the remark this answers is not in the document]
Hugo0: nit: this is a nitpick and wont change anth here, but generally good advice

Hugo0 (approved):
@panosfilianos

src/components/Global/LinkAccountComponent/index.tsx
panosfilianos: There is string repr that makes the conversion for the Error type

PR #425, opened 2024-10-07
feat: reload page when clicking on create request

jjramirezn:
To permit easy creation of new request after creating one

### DIFF
--- a/src/components/Global/Header/index.tsx
+++ b/src/components/Global/Header/index.tsx
@@ -228,7 +228,12 @@ const MenuLinks = () => {
             <NavLink href={'/send'} onClick={handleClick}>
                 <Text display="block">send</Text>
             </NavLink>
-            <NavLink href={'/request/create'}>
+            <NavLink
+                href={'/request/create'}
+                onClick={() => {
+                    if (window.location.pathname == '/request/create') window.location.reload()
+                }}
+            >
                 <Text display="block" className="flex items-center">
                     request
                 </Text>
--- a/src/components/Global/Header/index.tsx
+++ b/src/components/Global/Header/index.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import React, { useEffect, useState } from 'react'
+import React, { useEffect, useState, useCallback } from 'react'
 import { Box, Flex, Text, Stack, Collapse, useDisclosure } from '@chakra-ui/react'
 import { useLottie, LottieOptions } from 'lottie-react'
 import Link from 'next/link'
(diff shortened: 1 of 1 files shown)


src/components/Global/Header/index.tsx
```diff
@@ -228,7 +228,12 @@ const MenuLinks = () => {
             <NavLink href={'/send'} onClick={handleClick}>
                 <Text display="block">send</Text>
             </NavLink>
-            <NavLink href={'/request/create'}>
+            <NavLink
+                href={'/request/create'}
+                onClick={() => {
+                    if (window.location.pathname == '/request/create') window.location.reload()
+                }}
+            >
```
Hugo0: TODO: nice - I think this is the desired behavior. For consistency, can we have the same behaviour across all pages? So send, request, cashout.

Generally when submitting a PR we should think holistically - else we get fragmented behaviour. Even though the ticket was submitted only for request links navbar, its valid across other items, so the issue fix shouldn't just tackle the complaint, but ideally tackle whole issue holistically.  And if not, it should at least make mention of this wider issue

a very similar thing was raised in this PR from @panosfilianos: https://github.com/peanutprotocol/peanut-ui/pull/426#pullrequestreview-2356706285

Hugo0 (changes requested):

src/components/Global/Header/index.tsx
jjramirezn: great point, implementing

src/components/Global/Header/index.tsx
```diff
@@ -115,6 +115,37 @@ const MenuToggle = ({ toggle, isOpen }: { toggle: () => void; isOpen: boolean })
...
+            if (window?.location.pathname === route) {
+                // Force a hard reload of the current page
+                window.location.reload()
+            } else {
+                // For different routes, use router.push()
+                router.push(route)
+            }
+        },
+        [router, route]
+    )
```
Hugo0, line 133: praise: nice! Way cleaner :)

src/components/Global/Header/index.tsx
```diff
@@ -115,6 +115,37 @@ const MenuToggle = ({ toggle, isOpen }: { toggle: () => void; isOpen: boolean })
     )
 }
 
+const MenuLink = ({ route, title, isBeta = false }: { route: string; title: string; isBeta?: boolean }) => {
```
Hugo0, line 118: thought: in the future we might also want some other properties (e.g. isNew or hasNotification or similar)

Hugo0 (approved):
v nice!

PR #400, opened 2024-10-01
refactor: merge Success views for offramping flows (link claim, casho…

panosfilianos:
…ut) to separate component view

- Creates a separate component view under Offramping
- Defines interfaces for needed props
- Cleans up typo error in Confirm flow (and unnecessary .toLowerCase())

## Release Notes

- **New Features**
	- Introduced `OfframpSuccessView` to display transaction success messages and details.
	- Added `offrampType` property to the `Cashout` component for better handling of cashout transactions.
	- Enhanced `OfframpConfirmView` to support cross-chain transaction details and improved error handling.

- **Improvements**
	- Updated success screen mapping to use the new `OfframpSuccessView`.
	- Refined error messaging and user feedback in the cashout confirmation process.
	- Streamlined handling of liquidation addresses and transaction fees.
	- Centralized management of constants related to the Optimism network for better maintainability.
	- Improved flexibility in success screen properties with new interface definitions.
	- Removed the deprecated `ConfirmClaimLinkIbanView` component to simplify the codebase.

- **Documentation**
	- Added new interfaces and enums for better structure and flexibility in the offramp process.

PR #426, opened 2024-10-08
Refactor/ternary confirm view and abstract fee explainers

panosfilianos:
This refers to https://github.com/peanutprotocol/peanut-ui/pull/400#discussion_r1785961248

### DIFF
--- a/src/components/Offramp/Confirm.view.tsx
+++ b/src/components/Offramp/Confirm.view.tsx
@@ -79,19 +79,14 @@ export const OfframpConfirmView = ({
     // type: 'iban' or other
     // TODO: standardize this type
     let accountType = user?.accounts.find((account) => account.account_identifier === offrampForm.recipient)?.account_type
-
-    let fee  = 0;
-
-    let feeExplainer = ''
-    if (offrampType == OfframpType.CASHOUT) {
-        fee = accountType === 'iban' ? OFFRAMP_IBAN_FEE_USD : OFFRAMP_NON_IBAN_FEE_USD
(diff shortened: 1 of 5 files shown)


src/components/Offramp/Confirm.view.tsx
```diff
@@ -68,6 +68,45 @@ export const OfframpConfirmView = ({
...
     // state and context vars for claim link offramp
 
+    //////////////////////
+    // utility JSX vars
+
+    // type: 'iban' or other
+    // TODO: standardize this type
+    let accountType = user?.accounts.find((account) => account.account_identifier === offrampForm.recipient)?.account_type
+
+    // TODO: remove hardcoding fee amounts from here and refactor to consts file
```
Hugo0: praise: good todo comment.

These should prob be pulled from `Cashout.consts.ts`

Hugo0 (commented):
I think it makes little sense to refactor this in confirm view but not in success view (where we also have these fees)

Imo: extend this PR to also cover confirm success view (and ideally move values to consts file)

wdyt @panosfilianos ?

src/components/Offramp/Confirm.view.tsx
panosfilianos: fair

Hugo0 (changes requested):
TODO: also do in success view

panosfilianos:
This branch has a file conflict w/ `develop`. This is due to renaming the consts file from `cashout.utils.ts` to `offramp.utils.ts` to retain naming convention from other parts of the codebase. Once this has passed review, I will merge `develop` in the current branch and resolve the conflict.

PR #427, opened 2024-10-08
fix: Change Total to You will receive to make it clear UX

panosfilianos:


Hugo0:
This should also be changed in the success view no?

Hugo0 (changes requested):
TODO: also do in success view

PR #430, opened 2024-10-08
feat(token-input): show dollar sign for stablecoins

jjramirezn:
On all views that use the token input, now it shows a dollar sign for stablecoins


Hugo0 (approved):
nice!

PR #432, opened 2024-10-09
feat(request): pre-load requested pair on pay

jjramirezn:
Now when accessing a request link to pay, the selected token/chain pair will be the one that the requester set. The payer can pay directly in this or change it to use cross-chain

### DIFF
--- a/src/components/Request/Pay/Views/Initial.view.tsx
+++ b/src/components/Request/Pay/Views/Initial.view.tsx
@@ -144,6 +144,10 @@ export const InitialView = ({
         }
     }, [requestLinkData, tokenPriceData])
 
+    useEffect(() => {
+        resetTokenAndChain()
+    }, [])
+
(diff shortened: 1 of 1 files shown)


src/components/Request/Pay/Views/Initial.view.tsx
```diff
@@ -144,6 +144,10 @@ export const InitialView = ({
         }
     }, [requestLinkData, tokenPriceData])
 
+    useEffect(() => {
+        resetTokenAndChain()
+    }, [])
+
```
Hugo0, line 150: Style: Code wise, it's gonna be really difficult to parse in 6 months WHY we have this code here. But there's a good reason for it!

For code like this, I think adding a small comment above explaining its purpose is the right thing to do

Hugo0 (approved):

PR #434, opened 2024-10-10
fix(request): correctly identify native tokens

jjramirezn:
When fetching the price token we were inadvertently changing the token address, so then we did not recognized it as native. Fixed that

Also did some refactoring around the address zero comparison which also made unnecessary the individual utils file for Request/Pay


src/components/Request/Pay/Views/Initial.view.tsx
```diff
@@ -9,13 +9,19 @@ import Icon from '@/components/Global/Icon'
...
-import { ADDRESS_ZERO, EPeanutLinkType, RequestStatus } from '../utils'
 
 const ERR_NO_ROUTE = 'No route found to pay in this chain and token'
 
+enum RequestStatus {
+    LOADING = 'LOADING',
+    CLAIM = 'CLAIM',
+    NOT_CONNECTED = 'NOT_CONNECTED',
+    NOT_FOUND = 'NOT_FOUND',
+}
```
Hugo0, line 23: Question: Is there an explainer somewhere with these statuses? How is claim a status?

src/components/Request/Pay/Views/Initial.view.tsx
```diff
@@ -9,13 +9,19 @@ import Icon from '@/components/Global/Icon'
 import MoreInfo from '@/components/Global/MoreInfo'
 import * as consts from '@/constants'
 import { useCreateLink } from '@/components/Create/useCreateLink'
-import { peanut } from '@squirrel-labs/peanut-sdk'
+import { peanut, interfaces } from '@squirrel-labs/peanut-sdk'
 import TokenSelector from '@/components/Global/TokenSelector/TokenSelector'
 import { switchNetwork as switchNetworkUtil } from '@/utils/general.utils'
-import { ADDRESS_ZERO, EPeanutLinkType, RequestStatus } from '../utils'
```
Hugo0, line 15: praise: nice removing imports

src/utils/general.utils.ts
```diff
@@ -325,14 +325,18 @@ export const isTestnetChain = (chainId: string) => {
 
 export const areTokenAddressesEqual = (address1: string, address2: string): boolean => {
     if (address1.toLowerCase() === '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE'.toLocaleLowerCase())
-        address1 = '0x0000000000000000000000000000000000000000'
+        address1 = ethers.constants.AddressZero
     if (address2.toLowerCase() === '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE'.toLocaleLowerCase())
-        address2 = '0x0000000000000000000000000000000000000000'
+        address2 = ethers.constants.AddressZero
```
Hugo0, line 330: nit: actually i slightly prefer the previous one because it's one less dependency on ethers and pretty self-evident

src/utils/general.utils.ts
```diff
@@ -325,14 +325,18 @@ export const isTestnetChain = (chainId: string) => {
...
     // By using ethers.getAddress we are safe from different cases
     // and other address formatting
     return ethers.utils.getAddress(address1) === ethers.utils.getAddress(address2)
 }
 
+export const isAddressZero = (address: string): boolean => {
+    return areTokenAddressesEqual(address, ethers.constants.AddressZero)
+}
+
```
Hugo0, line 339: nice

src/utils/fetch.utils.ts
```diff
@@ -50,10 +50,6 @@ export const fetchTokenPrice = async (
     host?: string
 ): Promise<ITokenPriceData | undefined> => {
     try {
-        if (tokenAddress.toLowerCase() == '0x0000000000000000000000000000000000000000') {
```
Hugo0, line 53: terrible, not even strict equality sign :crying_cat_face:

src/components/Request/Pay/Views/Initial.view.tsx
```diff
@@ -168,6 +173,7 @@ export const InitialView = ({
     }, [requestLinkData, tokenPriceData])
 
     useEffect(() => {
+        // Load the token chain pair from the request link data
```
Hugo0, line 176: <3

Hugo0 (approved):

src/components/Request/Pay/Views/Initial.view.tsx
jjramirezn: no explainer, and is only used in this view. I agree could use better names

src/utils/general.utils.ts
jjramirezn: makes sense!

PR #437, opened 2024-10-14
fix(request): store correct destination hash for xchain

jjramirezn:
We only have the destination hash after we fetch it from squid. We were incorrectly storing the origin hash as destination hash

### DIFF
--- a/src/components/Request/Pay/Views/Success.view.tsx
+++ b/src/components/Request/Pay/Views/Success.view.tsx
@@ -5,9 +5,12 @@ import * as utils from '@/utils'
 import { useContext, useEffect, useMemo, useState } from 'react'
 import { fetchDestinationChain } from '@/components/utils/utils'
 import * as context from '@/context'
+import { peanut } from '@squirrel-labs/peanut-sdk'
+import { useAccount } from 'wagmi'
 
-export const SuccessView = ({ transactionHash, requestLinkData }: _consts.IPayScreenProps) => {
-    const { selectedChainID } = useContext(context.tokenSelectorContext)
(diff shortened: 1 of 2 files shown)


src/components/Request/Pay/Views/Success.view.tsx
```diff
@@ -18,10 +21,32 @@ export const SuccessView = ({ transactionHash, requestLinkData }: _consts.IPaySc
     const explorerUrlAxelarWithTx = 'https://axelarscan.io/gmp/' + transactionHash
 
     useEffect(() => {
-        if (transactionHash) {
+        if (explorerUrlDestChainWithTxHash) {
+            peanut.submitRequestLinkFulfillment({
```
Hugo0, line 25: thought: i wonder if we can combine this with the x-chain claim functionality - seems like a good place for DRY

src/components/Request/Pay/Views/Success.view.tsx
```diff
@@ -18,10 +21,32 @@ export const SuccessView = ({ transactionHash, requestLinkData }: _consts.IPaySc
...
+                chainId: requestLinkData.chainId,
+                hash: explorerUrlDestChainWithTxHash.transactionId,
```
Hugo0, line 27: Question: Just so I understand this, does the requestLinkFulfillment only get submitted once the destination chain tx is through? So, if the user closes the page, it would never submit the fulfillment right?

Hugo0 (approved):

src/components/Request/Pay/Views/Success.view.tsx
jjramirezn: Correct. I think the way to go around this is submit the origin transaction to the backend (but not mark as paid) and then with the destination chain tx we can do as today (if the user did not close the page) or in the backend we can check

PR #442, opened 2024-10-15
refactor: Change linkState types in enums for more robust checking

panosfilianos:
- Changes linkState types from string choice to enum
- This allows for type checking with enum members vs simple strings which to not have unexpected values we need to thoroughly typecheck everywhere

### DIFF
--- a/src/components/Request/Pay/Pay.consts.ts
+++ b/src/components/Request/Pay/Pay.consts.ts
@@ -9,7 +9,13 @@ export interface IPayScreenState {
     idx: number
 }
 
-export type IRequestLinkState = 'LOADING' | 'READY_TO_PAY' | 'ALREADY_PAID' | 'ERROR' | 'CANCELED'
+export enum IRequestLinkState {
+    LOADING = 'LOADING',
+    READY_TO_PAY = 'READY_TO_PAY',
+    ALREADY_PAID = 'ALREADY_PAID',
+    ERROR = 'ERROR',
+    CANCELED = 'CANCELED'
+}
 
 export const INIT_VIEW_STATE: IPayScreenState = {
     screen: 'INITIAL',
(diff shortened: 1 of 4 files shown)


src/components/Request/Pay/Pay.consts.ts
```diff
@@ -9,7 +9,13 @@ export interface IPayScreenState {
...
 }
 
-export type IRequestLinkState = 'LOADING' | 'READY_TO_PAY' | 'ALREADY_PAID' | 'ERROR' | 'CANCELED'
+export enum IRequestLinkState {
+    LOADING = 'LOADING',
+    READY_TO_PAY = 'READY_TO_PAY',
+    ALREADY_PAID = 'ALREADY_PAID',
+    ERROR = 'ERROR',
+    CANCELED = 'CANCELED'
+}
```
Hugo0, line 18: This now conflicts with https://github.com/peanutprotocol/peanut-ui/pull/443

@jjramirezn can you review this?

panosfilianos: the only comment from my side, is that I find enums preferable because you can check against type, not string - this way check strings are in one place only.
@jjramirezn wdyt?

panosfilianos:
PR motivation added. This really is a nitpicking, opinionated PR - so wouldn't spend much time if @jjramirezn strongly disagrees.

src/components/Request/Pay/Pay.consts.ts
jjramirezn: For things like state I like enum vs a string type. I don't think there is any conflict #443 is merged to develop already (and the changes I made were over a particular view)

jjramirezn (approved):
Looks good feel free to merge

PR #443, opened 2024-10-15
fix: request pay issues

jjramirezn:
- Added a spinner to the request pay success view when doing cross chain payments to show that the payment is on route. Resolve when we have the txHash for the destination chain
- "No route found" error no longer flashes before fetching route when changing token/chain
- No longer trying to get axelar tx for same chain swaps
- Added link for shortened addresses
- In request pay, now showing amounts better formated

Also
- Refactored the state handling and flow of the request pay which helped on implementing the above

### DIFF
--- a/src/components/Global/AddressLink/index.tsx
+++ b/src/components/Global/AddressLink/index.tsx
@@ -1,0 +1,28 @@
+import Link from 'next/link'
+import { useEffect, useState } from 'react'
+
+import * as utils from '@/utils'
+
+const AddressLink = ({ address }: { address: string }) => {
+    const [url, setUrl] = useState<string>('')
+    useEffect(() => {
+        if (!address) return
+        if (address.endsWith('.eth')) {
+            utils.resolveFromEnsName(address).then((resolvedAddress) => {
+                if (!resolvedAddress) return
+                setUrl(`https://debank.com/profile/${resolvedAddress}`)
+            })
+            return
+        }
+        setUrl(`https://debank.com/profile/${address}`)
+    }, [address])
+    return url ? (
+        <Link className="cursor-pointer underline" href={url} target="_blank">
+            {utils.printableAddress(address)}
+        </Link>
+    ) : (
+        <span>{utils.printableAddress(address)}</span>
+    )
+}
+
+export default AddressLink
--- a/src/components/Request/Pay/Views/Initial.view.tsx
+++ b/src/components/Request/Pay/Views/Initial.view.tsx
@@ -12,14 +12,40 @@ import { useCreateLink } from '@/components/Create/useCreateLink'
 import { peanut, interfaces } from '@squirrel-labs/peanut-sdk'
 import TokenSelector from '@/components/Global/TokenSelector/TokenSelector'
 import { switchNetwork as switchNetworkUtil } from '@/utils/general.utils'
+import { type ITokenPriceData } from '@/interfaces'
 
 const ERR_NO_ROUTE = 'No route found to pay in this chain and token'
 
-enum RequestStatus {
+enum ViewState {
+    INITIAL = 'INITIAL',
     LOADING = 'LOADING',
-    CLAIM = 'CLAIM',
-    NOT_CONNECTED = 'NOT_CONNECTED',
-    NOT_FOUND = 'NOT_FOUND',
+    READY_TO_PAY = 'READY_TO_PAY',
+    ERROR = 'ERROR',
+}
+
+async function createXChainUnsignedTx({
+    tokenData,
+    requestLink,
+    senderAddress,
+}: {
+    tokenData: ITokenPriceData
+    requestLink: Awaited<ReturnType<typeof peanut.getRequestLinkDetails>>
+    senderAddress: string
+}) {
+    const xchainUnsignedTxs = await peanut.prepareXchainRequestFulfillmentTransaction({
+        fromToken: tokenData.address,
+        fromChainId: tokenData.chainId,
+        senderAddress,
+        link: requestLink.link,
+        squidRouterUrl: 'https://apiplus.squidrouter.com/v2/route',
+        apiUrl: '/api/proxy/get',
+        provider: await peanut.getDefaultProvider(tokenData.chainId),
+        tokenType: utils.isAddressZero(tokenData.address)
+            ? interfaces.EPeanutLinkType.native
+            : interfaces.EPeanutLinkType.erc20,
+        fromTokenDecimals: tokenData.decimals as number,
+    })
+    return xchainUnsignedTxs
 }
 
 export const InitialView = ({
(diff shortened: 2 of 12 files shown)


src/components/Global/AddressLink/index.tsx
```diff
@@ -0,0 +1,28 @@
+import Link from 'next/link'
+import { useEffect, useState } from 'react'
+
+import * as utils from '@/utils'
+
+const AddressLink = ({ address }: { address: string }) => {
```
Hugo0, line 6: Thought: this is a great utility. What's your thoughts on using this throughout the app?

Hugo0: EDIT: Saw that we're also using  this in the profile dashboard view. This is great! Looks so much better.

Probably worth also adding to the claimlink flow?

src/components/Request/Pay/Views/Initial.view.tsx
```diff
@@ -12,14 +13,40 @@ import { useCreateLink } from '@/components/Create/useCreateLink'
...
+    INITIAL = 'INITIAL',
     LOADING = 'LOADING',
-    CLAIM = 'CLAIM',
-    NOT_CONNECTED = 'NOT_CONNECTED',
-    NOT_FOUND = 'NOT_FOUND',
+    READY_TO_PAY = 'READY_TO_PAY',
+    ERROR = 'ERROR',
+}
+
+async function createXChainUnsignedTx({
```
Hugo0, line 27: thought: this seems like an unnecessary function no? it literally just wraps an sdk function

src/components/Request/Pay/Views/Initial.view.tsx
```diff
@@ -12,14 +13,40 @@ import { useCreateLink } from '@/components/Create/useCreateLink'
...
-enum RequestStatus {
+enum ViewState {
+    INITIAL = 'INITIAL',
     LOADING = 'LOADING',
-    CLAIM = 'CLAIM',
-    NOT_CONNECTED = 'NOT_CONNECTED',
-    NOT_FOUND = 'NOT_FOUND',
+    READY_TO_PAY = 'READY_TO_PAY',
+    ERROR = 'ERROR',
+}
```
Hugo0, line 25: I think this is helpful

src/utils/general.utils.ts
```diff
@@ -235,6 +235,12 @@ export function formatAmount(amount: number) {
     return amount.toFixed(2)
 }
 
+export function formatAmountWithSignificantDigits(amount: number, significantDigits: number): string {
+    let fractionDigits = Math.floor(Math.log10(1 / amount)) + significantDigits
+    fractionDigits = fractionDigits < 0 ? 0 : fractionDigits
+    return amount.toFixed(fractionDigits)
+}
```
Hugo0, line 242: Thought: i'm slightly worried this function will have bad behaviour. Also, maybe we should add basic unit tests in UI codebase for functions like these?

Hugo0 (approved):
Good Job overall - particularly like the AddressLink component

I feel like the request payment views are quite complex and have a lot of state and state dependencies that can go wrong and are hard to reason about - any idea on how we can make this simpler?

src/utils/general.utils.ts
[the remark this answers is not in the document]
jjramirezn: very thorough, you are correct and will have into consideration for future programming

src/components/Request/Pay/Views/Success.view.tsx
```diff
@@ -9,15 +11,20 @@ import { peanut } from '@squirrel-labs/peanut-sdk'
...
     const sourceUrlWithTx = useMemo(
         () => `${utils.getExplorerUrl(selectedChainID)}/tx/${transactionHash}`,
         [transactionHash, selectedChainID]
     )
-    const [explorerUrlDestChainWithTxHash, setExplorerUrlDestChainWithTxHash] = useState<
-        { transactionId: string; transactionUrl: string } | undefined
-    >(undefined)
+    const isXChain = useMemo(() => {
+        return requestLinkData.chainId !== selectedChainID
+    }, [requestLinkData, selectedChainID])
```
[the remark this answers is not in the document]
jjramirezn: I think you are right, will have into consideration for future

src/components/Request/Pay/Views/Success.view.tsx
```diff
@@ -38,21 +45,59 @@ export const SuccessView = ({ transactionHash, requestLinkData, tokenPriceData }
...
+                },
+                link: requestLinkData.link,
+            })
+        }
+    }, [isXChain, requestLinkData, transactionHash, address, selectedTokenAddress])
+
+    useEffect(() => {
+        if (isXChain) {
+            setLoadingState('Awaiting route fulfillment')
             fetchDestinationChain(transactionHash, setExplorerUrlDestChainWithTxHash)
```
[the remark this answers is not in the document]
jjramirezn: We want to execute once so empty deps is ok

src/components/Request/Pay/Views/Success.view.tsx
```diff
@@ -38,21 +45,59 @@ export const SuccessView = ({ transactionHash, requestLinkData, tokenPriceData }
...
+            utils.saveRequestLinkFulfillmentToLocalStorage({
+                details: {
+                    ...requestLinkData,
+                    destinationChainFulfillmentHash: transactionHash,
+                    createdAt: new Date().toISOString(),
+                },
+                link: requestLinkData.link,
+            })
+        }
+    }, [isXChain, requestLinkData, transactionHash, address, selectedTokenAddress])
```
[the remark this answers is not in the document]
jjramirezn: hmmmm i think you may be right

src/components/Request/Pay/Views/Initial.view.tsx
```diff
@@ -244,7 +290,11 @@ export const InitialView = ({
                     await switchNetwork(selectedTokenData!.chainId)
                 }
                 setLoadingState('Sign in wallet')
-                const xchainUnsignedTxs = await createXChainUnsignedTx()
+                const xchainUnsignedTxs = await createXChainUnsignedTx({
+                    tokenData: selectedTokenData!,
+                    requestLink: requestLinkData,
+                    senderAddress: address ?? '',
+                })
```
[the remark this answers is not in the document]
jjramirezn: yes yes yes yes, but we have to take into consideration that the routes available only live for 20 seconds

src/components/Request/Pay/Views/Initial.view.tsx
jjramirezn: agree!

src/utils/general.utils.ts
jjramirezn: you had me at unit tests, and yes, it doesnt handle negative significant digits for example or other edgecases

src/components/Global/AddressLink/index.tsx
jjramirezn: yes to all

jjramirezn:
State management and flow is something that have to be done once and then forget, I agree that right now is far from ideal. Maybe mapping the states and transitions and then implementing that would help, due to the reactive nature of React (lol) special attention have to be paid to the transition

Here is an example only for one view (the pay one) and without labelling the transitions (that's the tricky part)
```mermaid
stateDiagram-v2
    [*] --> Initial
    Initial --> Loading
    Loading --> Error
    Loading --> ReadyToClaim
    Error --> Loading
    ReadyToClaim --> Loading
    ReadyToClaim --> [*]
```

PR #447, opened 2024-10-16
fix(request): allow route even if fees are less that a cent

jjramirezn:
If we are at this point we have a route, the feeEstimation is only to show and in this case is less than a cent. Removing unneeded check that was causing to reject a valid route


src/components/Request/Pay/Views/Initial.view.tsx
```diff
@@ -134,15 +134,9 @@ export const InitialView = ({
                 })
                 const { feeEstimation, estimatedFromAmount } = txData
                 setEstimatedFromValue(estimatedFromAmount)
-                if (Number(feeEstimation) > 0) {
```
Hugo0, line 137: xd so dumb

PR #448, opened 2024-10-16
fix(token-selector): dont ask for connected account to fetch token data

jjramirezn:
We were having undesirable behavior on the TokenAmountInput where the user was disconnected, the usd <-> token toggle was not loading. This PR fixes that by fetching token data even if there is no wallet connected (we don't need any of the wallet data for this)


Hugo0:
Completely agreed on this!

Hugo0 (approved):
agreed, nice.

PR #450, opened 2024-10-16
feat(request): simpler create success screen

jjramirezn:
The info is already in the previous screen and in the link itself, also changed "Request a paymet" to more on brand "Yay!"


Hugo0 (approved):
love PRs where we remove stuff <3

PR #466, opened 2024-10-21
[ISSUE-201] fix(token-selector): show empty message when no token match filter

jjramirezn:
Token selector was fake loading forever if there was a wallet connected and no token matched the filter, now we show the empty list message, we no longer need to check for hasFetchedBalances because we will always have token to show (the sdk ones)


Hugo0 (commented):
nice!

Hugo0 (approved):
nice, i was concerned about tokenselection elements changing after api call but it's pretty quick

PR #468, opened 2024-10-22
Disable sorting

Hugo0:


jjramirezn (approved):
thought: we should create a task to fix this and asses priority

PR #479, opened 2024-10-24
[TASK-6259] chore: redirect claim to main version

jjramirezn:
No reason to use legacy claim (that also has less features)


Hugo0 (approved):

Hugo0:
nice1

PR #484, opened 2024-10-25
[ISSUE-252] refactor(cashout): use ValidatedInput for account number

jjramirezn:
Added the ValidatedInput component to check if the account number is valid and show feedback to the user before they try to proceed

### DIFF
--- a/src/components/Cashout/Components/Initial.view.tsx
+++ b/src/components/Cashout/Components/Initial.view.tsx
@@ -61,9 +61,9 @@ export const InitialCashoutView = ({
     const [_tokenValue, _setTokenValue] = useState<string | undefined>(
         inputDenomination === 'TOKEN' ? tokenValue : usdValue
     )
-    const [accountNumber, setAccountNumber] = useState<string>('')
-    const [isValidAccountNumber, setIsValidAccountNumber] = useState<boolean>(false)
-    const [isValidatingAccountNumber, setIsValidatingAccountNumber] = useState<boolean>(false)
+    const [bankAccountNumber, setBankAccountNumber] = useState<string>('')
+    const [isValidBankAccountNumber, setIsValidBankAccountNumber] = useState<boolean>(false)
+    const [isValidatingBankAccountNumber, setIsValidatingBankAccountNumber] = useState<boolean>(false)
 
     const { prepareCreateLinkWrapper } = useCreateLink()
 
@@ -85,16 +85,16 @@ export const InitialCashoutView = ({
     const isDisabled = useMemo(() => {
         return (
             !_tokenValue ||
-            !isValidAccountNumber ||
-            isValidatingAccountNumber ||
+            !isValidBankAccountNumber ||
+            isValidatingBankAccountNumber ||
             !xchainAllowed ||
             !!isBelowMinLimit ||
             !!isExceedingMaxLimit
         )
     }, [
         _tokenValue,
-        isValidAccountNumber,
-        isValidatingAccountNumber,
+        isValidBankAccountNumber,
+        isValidatingBankAccountNumber,
         xchainAllowed,
         isBelowMinLimit,
         isExceedingMaxLimit,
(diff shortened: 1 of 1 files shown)


src/components/Cashout/Components/Initial.view.tsx
```diff
@@ -281,35 +262,30 @@ export const InitialCashoutView = ({
...
                                                     onClick={() => {
                                                         if (!xchainAllowed) return
-                                                        if (selectedBankAccount === account.account_identifier) {
-                                                            setSelectedBankAccount(undefined)
-                                                        } else {
-                                                            setSelectedBankAccount(account.account_identifier)
-                                                            setActiveInput('selectedBankAccount')
-                                                        }
+                                                        setAccountNumber(account.account_identifier)
                                                     }}
```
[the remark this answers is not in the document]
jjramirezn: those account should not generate error states, and if so the ValidatedInput component would clear any error message

src/components/Cashout/Components/Initial.view.tsx
```diff
@@ -61,6 +61,9 @@ export const InitialCashoutView = ({
     const [_tokenValue, _setTokenValue] = useState<string | undefined>(
         inputDenomination === 'TOKEN' ? tokenValue : usdValue
     )
+    const [accountNumber, setAccountNumber] = useState<string>('')
```
panosfilianos: issue (blocking): what is this accountNumber? can we comment? (i.e. bankAccountNumber)
the same for the additions below

src/components/Cashout/Components/Initial.view.tsx
```diff
@@ -281,35 +262,30 @@ export const InitialCashoutView = ({
...
                                                     </div>
                                                     <div className="flex w-6 justify-center">
-                                                        {selectedBankAccount === account.account_identifier && (
+                                                        {accountNumber === account.account_identifier && (
                                                             <button
                                                                 className="text-lg text-black"
                                                                 onClick={(e) => {
                                                                     e.stopPropagation()
-                                                                    setSelectedBankAccount(undefined)
+                                                                    setAccountNumber('')
```
panosfilianos: thought (non-blocking): imo it's better to have this typed as `<string | undefined>` bc we can run ifs like `if(accountNumber)` more cleanly. An empty string is falsy, but it just feels cleaner to me.

panosfilianos (changes requested):
I think a small naming change would be helpful here. This comes from me playing around w/ wallets currently and seeing there can be many hooks/modules w/ same naming (i.e. account) and it becomes harder/ slower to disambiguate.

wdyt?

src/components/Cashout/Components/Initial.view.tsx
jjramirezn: good idea, implementing

src/components/Cashout/Components/Initial.view.tsx
jjramirezn: In the end [inputs need a defined value](https://react.dev/reference/react-dom/components/input#im-getting-an-error-a-component-is-changing-an-uncontrolled-input-to-be-controlled) I prefer to have the state of bankAccountNumber and the value on the input to always be the same (as opposed to the state being undefined and the input be empty string)

src/components/Cashout/Components/Initial.view.tsx
```diff
@@ -86,36 +85,32 @@ export const InitialCashoutView = ({
...
         )
-    }, [_tokenValue, selectedBankAccount, newBankAccount, xchainAllowed, isBelowMinLimit, isExceedingMaxLimit])
+    }, [
+        _tokenValue,
+        isValidBankAccountNumber,
+        isValidatingBankAccountNumber,
+        xchainAllowed,
+        isBelowMinLimit,
+        isExceedingMaxLimit,
+    ])
```
nezz0746, line 101: nit: booleans like these variables usually don't need useMemo, you're not getting any performance improvement. Also it includes potential bugs if you're missing a variable in the dependency array the the compute is broken

nezz0746 (approved):

src/components/Cashout/Components/Initial.view.tsx
panosfilianos: agreed

panosfilianos (approved):

PR #485, opened 2024-10-25
[ISSUE-256] fix(request): fix token amount for request

jjramirezn:
Request were being created with usd value as token amount


panosfilianos (approved):
neato

PR #487, opened 2024-10-25
[ISSUE-169] fix: us link account form

jjramirezn:
US account details form was impossible to advance due to wrong error handling and clearing, this is now fixed. Also modified a little the handling of errors for create-external-account, because on error the API doesn't return a json so it was always failing when trying to parse it

### DIFF
--- a/src/app/api/bridge/external-account/create-external-account/route.ts
+++ b/src/app/api/bridge/external-account/create-external-account/route.ts
@@ -65,31 +65,40 @@ export async function POST(request: NextRequest) {
             body: JSON.stringify(body),
         })
 
-        const data = await response.json()
-
         if (!response.ok) {
-            if (data.code && data.code == 'duplicate_external_account') {
-                // in case the bridge account already exists
-                //
-                // sending back error responses is not currently common across app 
-                // in how we deliver errors from backend -> frontend
-                // sends back an object like:
-                // {
-                //     id: '4c537540-80bf-41dd-a528-dbe39a4',
-                //     code: 'duplicate_external_account',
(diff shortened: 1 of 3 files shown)


src/app/api/bridge/external-account/create-external-account/route.ts
```diff
@@ -65,31 +65,40 @@ export async function POST(request: NextRequest) {
...
+                        headers: {
+                            'Content-Type': 'application/json',
+                        },
+                    })
+                }
+            } catch (e) {
+                console.error('Error creating external account', e)
             }
-            throw new Error(`HTTP error! status: ${response.status}`)
+            return new NextResponse('', {
```
panosfilianos, line 94: question (non-blocking): is an empty data field here enough to handle all errors to function it returns it too? wdyt?

panosfilianos (approved):

src/app/api/bridge/external-account/create-external-account/route.ts
panosfilianos: as an additional note that's visible in the TODOs in these diffs: we need a std way to handle these errors and propagate that error state in the whole UI. I'm currently creating a prototype of what a solution could look like in `peanut-wallet`.

tl;dr: don't waste too much time on this

jjramirezn: the empty data field is not enough and I love that you are doing something about the error state, it is needed

panosfilianos: cool!
this is not the scope to fix that and I have a Task and TODOs around that.

PR #491, opened 2024-10-28
[TASK-6558] prevent e in token input

nezz0746:


jjramirezn:
suggestion(blocking): it's a good opportunity to prevent '+' and '-' also (they are valid in decimal mode too)

jjramirezn (changes requested):
suggestion(blocking): it's a good opportunity to prevent '+' and '-' also (they are valid in decimal mode too)

src/components/Global/TokenAmountInput/index.tsx
```diff
@@ -78,6 +78,9 @@ const TokenAmountInput = ({ className, tokenValue, setTokenValue, onSubmit }: To
                             e.preventDefault()
                             if (onSubmit) onSubmit()
                         }
+                        if (['e', '+', '-'].includes(e.key.toLowerCase())) {
```
jjramirezn, line 81: praise: nice!

jjramirezn (approved):

PR #498, opened 2024-10-30
[TASK-6625 & 6622] fix kyc iframe dimensions & cose mechanism

nezz0746:


src/components/Global/IframeWrapper/index.tsx
```diff
@@ -28,11 +29,28 @@ const IframeWrapper = ({
         }
     }, [onClose])
 
+    const areYouSurePromptRow = (
+        <div className='flex flex-col text-center sm:text-left sm:flex-row justify-between sm:items-center gap-4 p-2 sm:p-0 w-full'>
+            <p className="text-sm ml-1">Are you sure ? your KYC progress will be lost.</p>
```
jjramirezn: issue(blocking): message is specific to KYC flow, component is used in other flows

In particular, this message is confusing (and wrong) to get on the accept TOS flow (that must be closed manually also).



suggestion: I would add a confirmationMessage optional prop and show that message if exists or just direclty close if not

src/components/Global/IframeWrapper/index.tsx
```diff
@@ -1,17 +1,18 @@
...
     visible,
     onClose,
-}: {
-    src: string
-    style?: React.CSSProperties
-    visible: boolean
-    onClose: () => void
-}) => {
+}: IframeWrapperProps) => {
+    const [areYouSure, setAreYouSure] = useState(false)
```
jjramirezn: nitpick (non-blocking): better naming

areYouSure is a little confusing, at first I though it contained the answer of an user, but this is a flag to show confirmation or not, maybe somethin like showConfirmationMessage or something similar would be better

jjramirezn (changes requested):
praise: nice adding the confirmation functionality, I think is useful for KYC 

resolve blocking comment and ready to merge

src/components/Global/IframeWrapper/index.tsx
nezz0746: Yes, component is only used in TOS & KYC but I agree

jjramirezn: I would rather not have confirmation on the TOS, because as part of normal flow the user must close the modal after acepting, and there is no progress to be lost (either they have accepted or not, is not too inconvinient to the user to close like in the kwy case). For example, here I completed and now want to close the modal but I'm afraid to "lose progress"

jjramirezn (approved):

PR #501, opened 2024-10-30
[TASK-6450] add register/login functions in setup flow & home page

nezz0746:
### Changes
- Calling register in setup to create passkey in format {handle}.peanut.wallet
- Calling **handleLogin** in homepage to have kernelClient be current connected client
- Small type fixes
- UI work on homepage

### DIFF
--- a/src/app/(mobile-ui)/home/page.tsx
+++ b/src/app/(mobile-ui)/home/page.tsx
@@ -7,7 +7,7 @@ import React from 'react'
 import PeanutWalletIcon from '@/assets/icons/peanut-wallet.png'
 import Icon from '@/components/Global/Icon'
 import { motion, useAnimation } from 'framer-motion'
-import { useRef, useEffect, useState } from 'react'
+import { useRef, useEffect } from 'react'
 import useAvatar from '@/hooks/useAvatar'
 import classNames from 'classnames'
 import { HomeLink } from '@/components/Home/HomeLink'
@@ -21,19 +21,21 @@ const cardMargin = 16
 
 const Home = () => {
     const controls = useAnimation()
-    const { address, handleLogin, isLoggingIn } = useZeroDev()
-    const [selectedWalletIndex, setSelectedWalletIndex] = useState(0)
+    const { handleLogin, isLoggingIn } = useZeroDev()
 
     const carouselRef = useRef<HTMLDivElement>(null)
 
-    const { wallets } = useWallet()
+    const { wallets, selectedWallet, setSelectedWallet } = useWallet()
 
-    const { uri: avatarURI } = useAvatar(wallets[selectedWalletIndex] ? wallets[selectedWalletIndex].address : 'i am sad bc i dont have peanut')
+    const { uri: avatarURI } = useAvatar(selectedWallet ? selectedWallet.address : 'i am sad bc i dont have peanut')
 
-    const currentWallet = wallets[selectedWalletIndex]
-    const isConnectWallet = currentWallet?.account === address
+
+    const rawIndex = wallets.findIndex(wallet => wallet.address === selectedWallet?.address);
+    const selectedWalletIndex = rawIndex === -1 ? 0 : rawIndex
     const hasWallets = wallets.length > 0
+    const isConnectWallet = selectedWallet?.connected;
 
+    console.log({ selectedWallet, isConnectWallet })
 
     useEffect(() => {
         controls.start({
@@ -43,7 +45,7 @@ const Home = () => {
     }, [selectedWalletIndex, controls])
 
     const handleCardClick = (index: number) => {
-        setSelectedWalletIndex(index)
+        setSelectedWallet(wallets[index])
     }
 
     return (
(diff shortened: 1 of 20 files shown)


src/app/(mobile-ui)/home/page.tsx
```diff
@@ -53,76 +42,57 @@ const Home = () => {
...
-                setWalletErrorStrings([...walletErrorStrings])
-
-                // TODO: differentiate between kinds of errors
-                setConnectProviderButtonShown(true)
-            } else {
-                throw error
-            }
-        }
+    const handleCardClick = (index: number) => {
+        setSelectedWallet(wallets[index])
```
panosfilianos, line 46: issue (non-blocking): we need to add logic around this selection for BYOW (eg. the provider is not connected to the right wallet)

src/components/0_Bruddle/Toast.tsx
```diff
@@ -0,0 +1,125 @@
...
+  type?: ToastType;
+  duration?: number;
+  position?: ToastPosition;
+}
+
+interface ToastMessage extends ToastOptions {
+  id: number;
+}
+
+interface ToastContextType {
```
panosfilianos, line 21: thought: ideally these would go to a .consts file

src/components/Setup/Setup.helpers.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+interface PasskeyStore {
+    passkeys: Passkey[]
+}
+
+/**
+ * TEMPORARY: This is a temporary storage solution for storing passkeys locally to move forward with UI development.
+ */
+export class PasskeyStorage {
```
panosfilianos, line 15: question: do we expect to delete this file?

src/components/Setup/Views/Handle.provider.view.tsx
panosfilianos, line 1: question: it seems like we should delete this file. Am I missing something?

src/components/Setup/Views/Welcome.tsx
```diff
@@ -1,20 +1,36 @@
...
             setTimeout(() => {
+                if (handle.length <= 3) {
+                    reject(new Error('Handle must be at least 3 characters long'))
+                }
                 resolve()
             }, 1000)
         })
     }
 
+    const businessLogic = async () => {
```
panosfilianos: thought: this should probably be renamed

src/context/authContext.tsx
```diff
@@ -64,16 +62,6 @@ export const AuthProvider = ({ children }: { children: ReactNode }) => {
...
-    // TODO: this needs to be moved elsewhere (i.e. possible walletContext), was
-    // here for testing - POSSIBLY be removed bc the order is reversed:
-    // you first login w/ passkeys, so flow below would never happen
-    useEffect(() => {
-        if (kernelClientAddress != null && isKernelClientReady) {
-            // set PW as active wallet
-            setupWalletsAfterLogin()
-        }
-    }, [kernelClientAddress, isKernelClientReady])
-
```
panosfilianos, line 76: note: my upcoming PR will introduce similar functionality

src/context/walletContext/walletContext.tsx
```diff
@@ -40,252 +27,107 @@ const WalletContext = createContext<WalletContextType | undefined>(undefined)
...
-    //
-
-    // handles checks:
-    // PW:
-    //      - kernel not ready
-    // BYOW:
-    //      - wagmi not connected
-    //      - wagmi connected to wrong wallet
-    // throws WalletError in case of error which is an 'undefined' return
-    const checkActivateWallet = async (wallet: interfaces.IWallet): Promise<interfaces.IWallet | undefined> => {
```
panosfilianos, line 117: issue (non-blocking): this is the functionality we need for BYOWs. wdyt?

panosfilianos (commented):
Overall great PR. Very partial about the walletContext.ts but I think it's a major improvement. Let's get these things answered and let's merge.

src/app/(mobile-ui)/home/page.tsx
nezz0746: for connection status, kernelClient (PW) & wagmi context (BYOB) hold the source of truth so shouldn't mater when doing this selection. You can select a non connected wallet.

src/components/0_Bruddle/Toast.tsx
nezz0746: Tbh If it's not a large amount of code, I usually keep non exported types & interfaces in the file they're used, keeps branching & reading easier. But will not die on that hill.

src/components/Setup/Setup.helpers.ts
nezz0746: Hopefully, once we replaced it with backend accounts

src/components/Setup/Views/Handle.provider.view.tsx
nezz0746: Was trying go straight to feature here, updated what was needed. But kept as is in case you wanted to use.

src/context/walletContext/walletContext.tsx
nezz0746: I think here you couldn't select as active if it wasn't connected. I've unlocked this possibility & let users select the wallet they want, and sign in in a second time.

PR #502, opened 2024-10-30
Merge develop into peanut-wallet

jjramirezn:
Both branches drifted quite a lot, merging now to have more time to deal with any bugs before release. Ideally we should periodically do this as develop changes to have fewer conflicts


Hugo0:
100% agree

PR #509, opened 2024-11-04
feat: Fetch wallets from accounts

panosfilianos:
## Adds
- Logic to properly auth after register (note: does **not** work on reload)
- Logic to add PW to user accounts
- Logic to get wallets from accounts (refreshes on `user.accounts` refresh)
- `AccountType` from backend to frontend
- `ChainType` newly created here
- Update Account type with `chain` and `AccountType`

## Removes
- `handle` from `IWallet` (using `user.user.username` from `useAuth()` context)

### DIFF
--- a/src/context/walletContext/walletContext.tsx
+++ b/src/context/walletContext/walletContext.tsx
@@ -67,8 +67,8 @@ export const WalletProvider = ({ children }: { children: ReactNode }) => {
 
     ////// Wallets
     const { data: wallets } = useQuery<interfaces.IWallet[]>({
-        queryKey: ['wallets', user?.user.userId],
-        queryFn: () => {
+        queryKey: ['wallets', user?.accounts],
+        queryFn: async () => {
             const processedWallets = user?.accounts
                 .filter((account) => Object.values(interfaces.WalletProviderType).includes(account.account_type))
                 .map((account) => ({
--- a/src/context/walletContext/walletContext.tsx
+++ b/src/context/walletContext/walletContext.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { createContext, useContext, useEffect, useState, ReactNode, useMemo, useCallback } from 'react'
+import { createContext, useContext, useEffect, useState, ReactNode, useCallback } from 'react'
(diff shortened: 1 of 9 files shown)


src/context/walletContext/walletContext.tsx
```diff
@@ -43,43 +44,25 @@ export const WalletProvider = ({ children }: { children: ReactNode }) => {
...
-            ]
+            const processedWallets = user?.accounts.filter(
+                account => Object.values(interfaces.WalletProviderType).includes(account.account_type)
+            ).map(account=> ({
+                walletProviderType: account.account_type,
+                protocolType: account.chain,
+                address: account.account_identifier,
+                connected: false
+            }))
+            return processedWallets ? processedWallets : []
```
nezz0746, line 83: questions:
- Can we keep kernelClientAddress as to be non-breaking to current flow ? (Also maybe merge both localStorage accounts & fetched ones in the response)
- Does `account.account_type` is the same type as walletProviderType ?
- Is `handle` missing ?

Maybe we can store this PR until we get the right data into the account type ?

panosfilianos: Some questions to make sure I understand:
1. Can you explain a bit deeper on how `kernelClientAddress` is influencing this? My understanding is that that wallets should be populated only when the user accounts are populated/ depopulated (auth/logout). Happy to hop on a call to solve this quickly.
2. Correct me if I'm wrong but i think the user will be fetched every time the page is reloaded and the accounts will be populated. The same way we cleared walletContext and decided to have single sources of truth, I think the best idea here would be to have a single source of truth via `user.accounts`.  Additionally, when a new BYOW is added we add to DB and `user.accounts`. Wdyt?
3. That's the expectation with `account.account_type`
4. I have not included it (don't see the need to). Wdyt?

nezz0746: Apologies for late reply !

- Yes kernelClientAddress was just to update that query when someone registered a new wallet and got redirected to home screen.
- I see it's been added, perfect
- Will need it to be displayed in the profile, but we can grab this from the user probably since for now it's one to one relationship with PWs

PR #514, opened 2024-11-05
fix: should be clearer for US users how to enter their bank account

Hugo0:
US bank accounts have 9 digit routing number and then account number. This isn;t clear for users currently and we had 5 complaints

### DIFF
--- a/src/components/Cashout/Components/Initial.view.tsx
+++ b/src/components/Cashout/Components/Initial.view.tsx
@@ -324,7 +324,7 @@ export const InitialCashoutView = ({
                                         setErrorState({
                                             showError: true,
                                             errorMessage:
-                                                'Invalid bank account. Please make sure your account is supported',
+                                                'Invalid bank account. For US bank accounts, enter your bank routing number followed by your account number.',
                                         })
                                     } else {
                                         setErrorState({
--- a/src/components/Cashout/Components/Initial.view.tsx
+++ b/src/components/Cashout/Components/Initial.view.tsx
@@ -324,7 +324,8 @@ export const InitialCashoutView = ({
                                         setErrorState({
                                             showError: true,
                                             errorMessage:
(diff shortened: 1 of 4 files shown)


src/components/Cashout/Components/Initial.view.tsx
```diff
@@ -324,7 +324,7 @@ export const InitialCashoutView = ({
                                         setErrorState({
                                             showError: true,
                                             errorMessage:
-                                                'Invalid bank account. Please make sure your account is supported',
+                                                'Invalid bank account. For US bank accounts, enter your bank routing number followed by your account number.',
```
panosfilianos: thought: maybe we should add an example on the error and info messages. Something like:
Input: `1112223330001234567`
where:
- Routing Number (9 digits): example `111222333`
- Account Number (8-17 digits): example `0001234567`
wdyt?

src/components/Global/GeneralRecipientInput/index.tsx
```diff
@@ -34,7 +34,9 @@ const GeneralRecipientInput = ({ placeholder, recipient, onUpdate, className }:
                 type = 'iban'
                 isValid = await utils.validateBankAccount(recipient)
                 if (!isValid) errorMessage.current = 'Invalid IBAN, country not supported'
-            } else if (/^[0-9]{6,17}$/.test(recipient)) {
+            } else if (/^[0-9]{6,26}$/.test(recipient)) {
+                // routing number: 9 digits
+                // account number: 8-12 digits (can go up to 17)
```
panosfilianos, line 39: thought: These checks may prove not to be enough down the line (as more geographies are added, maybe there is overlap in format detection). For now, I'm good with this.

panosfilianos (approved):
This works. Nice, quick PR!
This is approved, but I'll give some time to address comment (either push addition or resolve).

PR #515, opened 2024-11-05
[TASK-6753] feat: check username availability

jjramirezn:
- Add a HEAD proxy passthrough (api url in client side is not standarized, maybe we should add a PUBLIC env variable, still proxy endpoint is useful)
- Check for username availability in setup
- Use bruddle component in ValidatedInput

### DIFF
--- a/src/components/Setup/Views/Welcome.tsx
+++ b/src/components/Setup/Views/Welcome.tsx
@@ -1,49 +1,52 @@
 import { Button } from '@/components/0_Bruddle'
-import BaseInput from '@/components/0_Bruddle/BaseInput'
 import { useToast } from '@/components/0_Bruddle/Toast'
 import { useSetupFlow } from '@/components/Setup/context/SetupFlowContext'
 import { useState } from 'react'
+import ValidatedInput from '@/components/Global/ValidatedInput'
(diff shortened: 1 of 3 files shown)


src/components/Setup/Views/Welcome.tsx
```diff
@@ -1,49 +1,52 @@
...
+    const checkHandleValidity = async (handle: string): Promise<boolean> => {
+        if (handle.length < 4) {
+            toast.error('Handle must be at least 4 characters long')
             return false
         }
+        const res = await fetch(`${next_proxy_url}/get/users/username/${handle}`, {
+            method: 'HEAD', // we only need the status code no body
+        })
+        // 404 on user means no user exist with this handle
+        const isHandleTaken = 404 !== res.status
```
panosfilianos: thought (def non-blocking): I think this check should test for 404 not if it is not 404. The idea here is that in the future we may do any number of things in the api-ts and we may end up with other error codes (eg. `429`s or `500` in case of our error etc.)

based on the above we'd have a handle taken result which wouldn't be accurate.

panosfilianos (approved):
Neat, short PR.
Check my `thought` below. If you want, push a fix. Otherwise fft merge.

PR #519, opened 2024-11-07
feat: call fetchUser at the right place for register

panosfilianos:
Removes old `useEffect` hook and calls `fetchUser() `after `handleRegister()` on register page


jjramirezn (approved):

jjramirezn:
Approved, Login and page reload will be added in another PR

PR #521, opened 2024-11-11
Fix/cashout bugs6

Hugo0:
A **lot** of changes.

- addressLink used everywhere
- recipientInput improved
- recent account suggestions added
- tooltips fixed
- various cashout bugs fixed

TODO: no way of both suggesting IBAN AND US accounts. Need to rework this flow in the future to also suggest browser us accounts if possible.


src/app/cashout/page.tsx
```diff
@@ -1,4 +1,4 @@
-import * as components from '@/components'
+import { Cashout } from '@/components'
```
nezz0746, line 1: I like

src/components/Cashout/Cashout.tsx
```diff
@@ -92,6 +92,43 @@ export const Cashout = ({}) => {
         fetchAndSetCrossChainDetails()
     }, [])
 
+    // For testing - check if URL has screen param
+    useEffect(() => {
```
nezz0746, line 96: Should this stil be in here if it was for testing ?

src/components/Global/AddressLink/index.tsx
```diff
@@ -1,27 +1,37 @@
...
         setUrl(`https://debank.com/profile/${address}`)
-    }, [address])
+
+        // Update display: prefer ENS name, fallback to shortened address
+        if (ensName) {
+            setDisplayAddress(ensName)
+        } else {
+            setDisplayAddress(utils.printableAddress(address))
+        }
+    }, [address, ensName])
```
nezz0746, line 27: nit: I feel like there's no real need for effects or state here. We're just displaying conditionally either the ens is present or not

src/components/Cashout/Cashout.tsx
Hugo0: yes! Its useful
peanut.to/cashout?screen=confirm

No reason not to, way easier  for future testing

src/components/Global/AddressLink/index.tsx
Hugo0: no cuz we do:

0x --> fetch if ens exists --> ens

nezz0746 (approved):

PR #523, opened 2024-11-12
feat: add /update-user call

nezz0746:
## Changes

- Update user call for contact setup screen
- UI fixes


src/context/walletContext/zeroDevContext.context.tsx
Hugo0, line 1: thought: this diff massively sucks

we should always format our code properly :(

need gh action for this

package.json
```diff
@@ -25,7 +25,7 @@
         "@dicebear/core": "^9.2.2",
         "@headlessui/react": "^1.7.19",
         "@headlessui/tailwindcss": "^0.2.1",
-        "@hookform/resolvers": "^3.9.1",
```
Hugo0, line 28: Q: why are we installing an older version?

package.json
```diff
@@ -83,7 +83,7 @@
         "viem": "2.17.1",
         "wagmi": "2.12.25",
         "web3": "^1.10.4",
-        "zod": "^3.23.8"
```
Hugo0, line 86: Q: what is this new dependency?

Hugo0 (changes requested):

package.json
nezz0746: Was when trying to resolve the error with zod, but I guess can bump now

package.json
nezz0746: It's an object validation library, was the one breaking vercel because of some obscure compatibility issues just used another one to fix

src/context/walletContext/zeroDevContext.context.tsx
nezz0746: Oh you're right, yes. I had to disable my prettier formatter to not include massive diffs on some files. We need to take care of this yes

src/app/(mobile-ui)/home/page.tsx
Hugo0, line 1: image
Q: i think this was already discussed, but what was the purpose of these (....) names?

src/app/(mobile-ui)/layout.tsx
```diff
@@ -132,34 +139,33 @@ const Layout = ({ children }: { children: React.ReactNode }) => {
...
-                                'text-purple-1': pathName === tab.href,
-                            })}
-                        >
-                            <NavIcons name={tab.icon} size={30} />
-                        </Link>
-                    )
-                })}
-            </div>
+                        return (
+                            <Link
```
Hugo0, line 156: Thought: I see a lot of Link classes, all with inline Tailwind classnames. Can we abstract this away and standardize it?

src/components/Setup/Views/ContactInfo.tsx
```diff
@@ -1,21 +1,63 @@
...
+                method: 'POST',
+                headers: {
+                    'Content-Type': 'application/json',
+                },
+                body: JSON.stringify({
+                    email: isEmail ? contact : undefined,
+                    telegram: !isEmail ? contact : undefined,
+                    userId,
+                }),
+            })
```
Hugo0, line 43: Q: where's the authentication here? do we not pass in jwt?

src/components/Setup/Views/ContactInfo.tsx
```diff
@@ -1,21 +1,63 @@
...
+            console.error('Error updating user', error)
+            toast.error('Failed to update contact info')
+        },
+        onSuccess: () => {
+            handleNext()
+        },
+    })
     const methods = useForm<ContactFormData>({
-        resolver: zodResolver(contactSchema),
+        resolver: yupResolver(contactSchema),
```
Hugo0, line 60: Q: what is yup and zod :laughing:

src/components/Setup/Views/ContactInfo.tsx
```diff
@@ -41,9 +82,12 @@ const ContactInfo = () => {
         <FormProvider {...methods}>
             <form onSubmit={onSubmit}>
                 <div className="flex h-full flex-col justify-end gap-2">
-                    <Field placeholder="Email" {...register('email')} autoComplete="off" />
-                    <Field placeholder="Telegram" {...register('telegram')} />
-                    <Button onClick={onSubmit} type="submit">
+                    <Field
+                        placeholder="squirrel-love@peanuts.com or @sqluv"
```
Hugo0: TODO: copy change like in notion

src/components/Setup/Views/InstallPWA.tsx
Hugo0, line 1: Q: On install pwa view - its not clear what the user is supposed to do after they install? what happens after? do we auto open the pwa? or do they still press skip?

Hugo0, line 1: Thought: we might want them to install PWA in the future by prompting them form the home page if they skipped. Is setup/views/installpwa the best place for this component?

src/hooks/usePWAStatus.ts
Hugo0, line 1: Q: Scenario: Bob has PWA installed but is using peanut in browser. Can we detect this? And tell him to use PWA (or deeplink him there)?

src/utils/general.utils.ts
```diff
@@ -4,12 +4,20 @@ import peanut from '@squirrel-labs/peanut-sdk'
 import { ethers } from 'ethers'
 import chroma from 'chroma-js'
 
+export const colorMap = {
+    lavender: '#90A8ED',
+    pink: '#FF90E7',
+    green: '#98E9AB',
+    yellow: '#FFC900',
+}
```
Hugo0, line 12: nooooooooooooooooooooooooooo

We have these defined in tailwind!

Hugo0 (changes requested):

src/app/(mobile-ui)/home/page.tsx
nezz0746: These are route groups that share a common layout, it's a NextJS router convention that allows you to easily share layout & code without changing the routing

src/app/(mobile-ui)/layout.tsx
nezz0746: These are bottom tabs for the mobile UI they're not used anywhere else, seems fine as is no ? Can refactor the component yes for sure, but tailwind will stay

src/components/Setup/Views/ContactInfo.tsx
nezz0746: We do ! It's being passed in the /udpate-user route to our server

src/components/Setup/Views/ContactInfo.tsx
nezz0746: I've left a comment above ! They're object validation utility libraries that make it easier to validate form. Check them out https://zod.dev/ can also be used in other contexts as well.

PS: I gave up on zod because I couldn't solve the prod vercel issue that came from a obscure compatibility issue 💀 but they do the same job

src/hooks/usePWAStatus.ts
nezz0746: Don't think we can read what Bob has installed but could have it stored ourselves and have a recommendation style prompt later on that would be nice 💡

src/utils/general.utils.ts
nezz0746: Actually if you look below I just wanted not to break the method if we use it again in the future.

In most apps you use your colors in different contexts. Wether it be in a function that creates new color, in a web manifest, in a tailwind.config.js file, or even sometimes in backend content generating methods.

Will most likely end up having them in a colors object anyway !

src/app/(mobile-ui)/layout.tsx
Hugo0: Generally if we have the same code repeated in 4+ places it just makes sense to standardize

This is non-blocking but it would've been cleaner if we did this

src/components/Setup/Views/ContactInfo.tsx
Hugo0: ok!

src/components/Setup/Views/ContactInfo.tsx
```diff
@@ -1,21 +1,63 @@
...
+
+            const response = await fetch('/api/peanut/user/update-user', {
+                method: 'POST',
+                headers: {
+                    'Content-Type': 'application/json',
+                },
+                body: JSON.stringify({
+                    email: isEmail ? contact : undefined,
+                    telegram: !isEmail ? contact : undefined,
+                    userId,
```
Hugo0, line 41: nice

src/utils/general.utils.ts
Hugo0: But can't we define these ONCE and then pull from there? I don't want to have our colors defined in a bunch of places. We have them defined  in tailwind.config.js. Why can't we pull from there?

Hugo0 (approved):

PR #530, opened 2024-11-16
[TASK-6301] refactor: use base url env var

jjramirezn:
Don't trust request headers, use our own variable


Hugo0 (approved):

Hugo0:
@jjramirezn is this also useful in main?

PR #533, opened 2024-11-18
[TASK-6981] (main) fix: disable autocorrect on address input

jjramirezn:


src/components/Global/RecipientInput/index.tsx
```diff
@@ -20,6 +20,9 @@ const RecipientInput = ({ placeholder, value, setValue, onEnter }: RecipientInpu
                 type="text"
                 placeholder={placeholder}
                 spellCheck="false"
+                autoCorrect="off"
+                autoCapitalize="off"
+                autoComplete="off"
```
Hugo0, line 22: Q, blocking: @jjramirezn why do we need this in both **RecipientInput** and **ValidatedInput**?

jjramirezn: yes, recipient input also will contain addresses and ENS although not validated the issue can be present there too

Hugo0 (approved):

PR #539, opened 2024-11-20
[TASK-6955] feat: populate token selector with squid supported token

jjramirezn:
## Main focus

Instead of relaying only on our sdk list of supported chain and tokens, the token selector is now populated mainly from the supported tokens from squid, in the flows of link and request creation we also use the sdk list because we want users to be able to create those even if those wont be able to be fulfilled crosschain

## Other changes
- Cache the list of squid supported tokens for 12 hours, we don't expect this to change much
- On request pay dont show the token selector if the token doesnt support xchain
- Lazy load the images on the token selector (the lists are bigger now with all the squid tokens)  

## Should also fix
- [TASK-6241]
- [TASK-6859]
- [TASK-6581]
- [TASK-6580]
- [TASK-6568]


jjramirezn:
Added copy update https://www.notion.so/peanutprotocol/Bug-xchain-copy-wrong-ti-130838117579802d8cadcbdb75057e2e?pvs=4

jjramirezn:
Added fix for balance not showing https://www.notion.so/peanutprotocol/Bug-Can-get-into-bad-balance-state-on-tokenSelector-12e83811757980febfdbf6a9b62b46ba?pvs=4

src/app/actions/squid.ts
```diff
@@ -0,0 +1,40 @@
+'use server'
+
+import { getSquidChains, getSquidTokens } from '@squirrel-labs/peanut-sdk'
```
Hugo0, line 3: Thought: we might want to integrate across.to or a tron bridge in the future. Might want to add an abstraction layer here (ie not SquidChains, but X-chain supported chains)

src/components/Global/TokenSelector/TokenSelector.tsx
```diff
@@ -5,151 +5,206 @@ import Modal from '../Modal'
...
 import { useBalance } from '@/hooks/useBalance'
-import { peanutTokenDetails, supportedPeanutChains } from '@/constants'
+import { supportedPeanutChains, peanutTokenDetails } from '@/constants'
 import * as context from '@/context'
-import * as utils from '@/utils'
-import * as consts from '@/constants'
+import { areTokenAddressesEqual, formatTokenAmount } from '@/utils'
 import { AdvancedTokenSelectorButton } from './Components'
-import { IToken, IUserBalance } from '@/interfaces'
+import { IUserBalance, IToken } from '@/interfaces'
```
Hugo0, line 12: lol? how did this happen? Automated formatting?

Hugo0 (changes requested):
Needs changes, review here: https://www.notion.so/peanutprotocol/Review-Bug-TokenSelector-should-not-show-non-xchain-supported-tokens-in-request-pay-and-claimlink-c-14483811757980be9e92e8fdaf8d4588

Hugo0 (approved):
huge ux win

PR #548, opened 2024-11-27
Bypass gas estimation for Safe wallet connector

ardaerturk:
- Added Safe wallet detection to skip gas estimation
- Safe uses Account Abstraction (AA) wallet where gas is sponsored/handled by Safe
- AA wallets don't need to hold native tokens for gas, making standard gas estimation irrelevant
- Returns default gas values (0) when Safe detected to prevent estimation errors
- Normal gas estimation flow maintained for regular wallets


ardaerturk (commented):
resolved ai bot suggestions

Hugo0 (approved):
lgtm! <3

PR #560, opened 2024-12-04
Fix container breaking for link input view

kushagrasarathe:
hey @Hugo0 I overrode the `max-width` class for the create component since the card class was being used on other places too and fixing it directly in `tailwind config` declaration would affect the ui on other places too.

For reference the below two screenshots show the current fix implementation

The first screenshot shows how a normal email will look like
Screenshot 2024-12-04 at 10 33 00 PM

The second screenshot shows how it would look if the email is comparatively bigger, so it will be truncated 
Screenshot 2024-12-04 at 10 33 14 PM

Thoughts?

## Summary by CodeRabbit

- **New Features**
	- Enhanced styling for the Create component, improving visual appeal.
	- Simplified layout and structure for the CreateLinkInputView component, enhancing readability.
- **Bug Fixes**
	- Adjusted error message display for improved clarity.
- **Chores**
	- Minor adjustments to the conditional rendering for better responsiveness to different transaction types.

### DIFF
--- a/src/components/Create/Create.tsx
+++ b/src/components/Create/Create.tsx
@@ -121,7 +121,7 @@ export const Create = () => {
     }, [address])
 
     return (
-        <div className="card">
+        <div className="card max-w-lg">
             {createElement(_consts.CREATE_SCREEN_MAP[step.screen].comp, {
                 onPrev: handleOnPrev,
                 onNext: handleOnNext,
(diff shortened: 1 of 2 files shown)


Hugo0:
@kushagrasarathe i think the better fix here would be to deal with the issue in the card class! In fact, I think ideally we'd move away from card being a tailwind class and have it be a react component

Hugo0 (changes requested):
Imo fix should be more thorough, by moving card from tailwind class to making it a react component

kushagrasarathe:
> Imo fix should be more thorough, by moving card from tailwind class to making it a react component

hey @Hugo0, this sounds like a good idea, I'll get into this and try to structure a new reusable card component

Hugo0 (approved):

PR #627, opened 2025-01-15
[TASK-8097] feat: order wallet cards

jjramirezn:
First peanut wallet, then external wallets by date created

## Summary by CodeRabbit

- **New Features**
  - Enhanced wallet sorting mechanism to prioritize Peanut Wallets at the top of the list
  - Wallets now sorted by type and creation date for improved user experience


Hugo0 (approved):
Logic looks good. Can you double check that we also use this in the wallet header, not just in home screen cards?

jjramirezn:
Wallet header uses the same wallet list from the wallet context so it should also be ordered

PR #654, opened 2025-01-30
fix: hide wallet icons for disconnected wallets

kushagrasarathe:
- save connector details to db
- hide wallet icons for disconnected wallets

fixes TASK-8398

## Summary by CodeRabbit

- **New Features**
	- Enhanced wallet icon display logic to provide more accurate and context-aware wallet images.
	- Improved wallet connection functionality with more detailed wallet information.

- **Improvements**
	- Updated wallet creation process to support more flexible wallet details.
	- Added utility function to retrieve user account information.
	- Refined wallet connection hook to include additional wallet metadata.

- **Technical Enhancements**
	- Optimized memoization for wallet-related components.
	- Improved dependency tracking for wallet image and connection states.

### DIFF
--- a/src/components/Global/WalletHeader/index.tsx
+++ b/src/components/Global/WalletHeader/index.tsx
@@ -80,7 +80,7 @@ const WalletHeader = ({ className, disabled }: WalletHeaderProps) => {
                 disabled={disabled}
                 variant="yellow"
                 className={twMerge(
-                    'bg-yellow-4/80 flex h-auto w-fit min-w-32 items-center justify-between py-2',
+                    'flex h-auto w-fit min-w-32 items-center justify-between bg-yellow-4/80 py-2',
                     className
                 )}
                 onClick={() => setShowModal(true)}
@@ -254,7 +254,7 @@ const WalletEntryCard = ({ wallet, isActive, onClick }: WalletEntryCardProps) =>
 // add new wallet component, triggers web3modal
 const AddNewWallet = ({ onClick }: { onClick: () => void }) => (
     <Card onClick={onClick}>
-        <Card.Content className="bg-purple-5 flex min-h-16 w-full cursor-pointer items-center justify-center gap-3 px-4 py-3 hover:bg-opacity-90">
+        <Card.Content className="flex min-h-16 w-full cursor-pointer items-center justify-center gap-3 bg-purple-5 px-4 py-3 hover:bg-opacity-90">
             <div className="flex size-7 items-center justify-center rounded-full border border-n-1">
                 <Icon name="plus" fill="black" className="h-7 w-7" />
             </div>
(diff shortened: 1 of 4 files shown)


src/components/Global/WalletHeader/index.tsx
```diff
@@ -164,8 +164,11 @@ const WalletEntryCard = ({ wallet, isActive, onClick }: WalletEntryCardProps) =>
 
     // get wallet icon to display
     const walletImage = useMemo(() => {
-        return isPeanutWallet ? PeanutWalletIcon : wallet?.connector?.iconUrl || PeanutWalletIcon
-    }, [isPeanutWallet, wallet])
+        if (wallet.walletProviderType === WalletProviderType.PEANUT) {
```
jjramirezn: suggestion: remove isPeanutWallet (or turn it into a memo if used in several places)

good catch here! but now we have in line 155 the isPeanutWallet that may be causing issues elsewhere I would turn that into a memo and use it here

src/components/Global/WalletHeader/index.tsx
```diff
@@ -164,8 +164,11 @@ const WalletEntryCard = ({ wallet, isActive, onClick }: WalletEntryCardProps) =>
...
+            return PeanutWalletIcon
+        }
+        return isConnected ? wallet.connector?.iconUrl || PeanutWalletIcon : PeanutWalletIcon
```
jjramirezn, line 170: question(blocking): why do we show peanut icon for disconnected? why not their own icon

src/components/Home/WalletCard.tsx
```diff
@@ -74,14 +74,14 @@ export function WalletCard({ type, onClick, ...props }: WalletCardProps) {
...
-            }).toDataUri()
-        )
-    }, [wallet])
+
+        const avatar = createAvatar(identicon, {
+            seed: wallet.address,
+            size: 128,
+        }).toDataUri()
+
+        return isConnected ? wallet.connector?.iconUrl || avatar : avatar
```
jjramirezn, line 83: same question as before

src/components/Global/WalletHeader/index.tsx
kushagrasarathe: aah thanks for pointing out, I missed it, this might have been messed up due merge conflicts with another pr, I should be using isPeanutWallet only and memoizing it, fixing it 🫡

src/components/Global/WalletHeader/index.tsx
kushagrasarathe: @jjramirezn  check TASK-8398 , acc to it, we dont' want to show icons for disconnected wallet, that's why updated it like this, so on header, to maintain good ui, we showing peanut icon, while on wallet cards, we show a avatar based on address

jjramirezn (approved):

PR #661, opened 2025-02-04
fix: network switching error when trying to pay a req

kushagrasarathe:
- fixes TASK-8547 : network switching error

## Summary by CodeRabbit

- **New Features**
  - Enhanced wallet connectivity: The system now dynamically selects the appropriate network based on the connected wallet type, ensuring that users are automatically directed to the correct blockchain network. This update improves overall wallet management, providing a more seamless and reliable experience when switching between different wallet options.


Hugo0 (approved):
Good catch, oversight bug. I think useWallet would benefit from a docstring

PR #662, opened 2025-02-04
fix: clear button and chevron overlap on address input

kushagrasarathe:
fixes TASK-8619 where clear btn and chevron icon would overlap on the address input


Hugo0 (approved):
hard to mentally evaluate CSS but lgtm

PR #663, opened 2025-02-04
fix: add loading peanut for history page 

kushagrasarathe:
- fixes TASK-8459 where sad peanut would appear instead of loading state on history page
- also fixes ui inconsistencies on cashout

## Summary by CodeRabbit

- **New Features**
  - Introduced a new loading indicator to inform users when data is being fetched.
  - Updated the cashout status page with clearer visuals, refined layouts, and improved navigation options.
  - Launched an accessible spinner component for a more consistent loading experience across the app.

- **Style**
  - Enhanced the layout for list items, allowing content to display more flexibly.


Hugo0 (approved):
lgtm. We also have a similar loading problem in other places; e.g. in /home it first shows 0 wallets and then it shows all wallets. Instead ideally we should have a loading anim

PR #665, opened 2025-02-04
fix: refetch wallet balance after claim/req-pay/send flow completion

kushagrasarathe:
- fixes TASK-8586
- request pay, claim funds and send funds flow now properly updates wallet balances

## Summary by CodeRabbit

- **New Features**
	- Wallet balance now refreshes automatically after completing transactions (claims, link creation, and payments), ensuring your account always shows the latest information.

- **Bug Fixes**
	- Improved how the correct wallet address is used during balance updates.
	- Removed an unnecessary loading screen in the payment success view for a smoother transaction experience.
	- Adjusted loading state management in the InitialView component for better user experience.

### DIFF
--- a/src/components/Claim/useClaimLink.tsx
+++ b/src/components/Claim/useClaimLink.tsx
@@ -11,7 +11,7 @@ import { useWallet } from '@/hooks/wallet/useWallet'
 import * as utils from '@/utils'
 
 export const useClaimLink = () => {
-    const { chain: currentChain } = useWallet()
+    const { chain: currentChain, refetchBalances } = useWallet()
     const { switchChainAsync } = useSwitchChain()
 
     const { setLoadingState } = useContext(context.loadingStateContext)
@@ -25,6 +25,9 @@ export const useClaimLink = () => {
                 baseUrl: `${consts.next_proxy_url}/claim-v2`,
                 APIKey: 'doesnt-matter',
             })
+
+            // refetch wallet balance after successful claim
+            await refetchBalances(address)
 
             return claimTx.transactionHash ?? claimTx.txHash ?? claimTx.hash ?? claimTx.tx_hash ?? ''
         } catch (error) {
--- a/src/components/Claim/useClaimLink.tsx
+++ b/src/components/Claim/useClaimLink.tsx
@@ -27,7 +27,7 @@ export const useClaimLink = () => {
             })
 
             // refetch wallet balance after successful claim
-            await refetchBalances(address)
+            refetchBalances(address)
 
             return claimTx.transactionHash ?? claimTx.txHash ?? claimTx.hash ?? claimTx.tx_hash ?? ''
         } catch (error) {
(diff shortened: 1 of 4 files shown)


src/components/Claim/useClaimLink.tsx
```diff
@@ -26,6 +26,9 @@ export const useClaimLink = () => {
                 APIKey: 'doesnt-matter',
             })
 
+            // refetch wallet balance after successful claim
+            await refetchBalances(address)
```
Hugo0: Q: do we need await here?

src/components/Create/useCreateLink.tsx
```diff
@@ -761,7 +761,8 @@ export const useCreateLink = () => {
                 transaction: type === 'deposit' ? response && response.unsignedTxs[0] : undefined,
             })
 
-            await refetchBalances(address ?? '')
+            // refetch wallet balance after successful link creation
+            await refetchBalances(selectedWallet?.address || '')
```
Hugo0: Same

src/components/Request/Pay/Views/Initial.view.tsx
```diff
@@ -335,6 +333,9 @@ export const InitialView = ({
 
                 setTransactionHash(hash ?? '')
                 onNext()
+
+                // update wallet balance after successful req payment
+                await refetchBalances(address)
```
Hugo0: - same
- flag: repetitive code. Would be good to at least evaluate a more holistic approach (even if actually implementing it is high effort and we go with the more hacky short term solution)

src/components/Request/Pay/Views/Success.view.tsx
```diff
@@ -79,20 +78,22 @@ export const SuccessView = ({ transactionHash, requestLinkData, tokenPriceData }
...
-                    <span className="sr-only">{loadingState}</span>
-                </div>
 
-                <label className="text-h8 font-bold ">
-                    Funds are on their way to <AddressLink address={requestLinkData.recipientAddress} />!
-                </label>
-            </div>
-        )
-    }
+    // todo: manage loading state properly, currently it keeps loading forever
```
Hugo0, line 82: ?

Hugo0: This is pending on request v2?

Hugo0 (approved):

src/components/Request/Pay/Views/Success.view.tsx
kushagrasarathe: the loading state was buggy, it kept loading forever, so to smoothen the ux, removed it for a while, yes, in req v2 will have it improved

src/components/Request/Pay/Views/Initial.view.tsx
kushagrasarathe: not sure what you mean by repetitive code here, but we are already reusing the refetchBalances fn from wallet hook, and these are three different actions that are being done where refetchBalances is used, create/pay/claim flows, where we need to refetch the balance

src/components/Claim/useClaimLink.tsx
kushagrasarathe: good catch, probably not, removed it

PR #667, opened 2025-02-04
fix: update error msg wen user reject tx + fix refund link render

kushagrasarathe:
fixes TASK-8635 for prod

> Note: pr for wallet goes separately 🫡 

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced error messaging to ensure that users receive clearer and more processed error information.
  - Refined the display of refund messages so they appear only when all necessary details are available.


Hugo0:
@kushagrasarathe nit: althought 2 prs are needed, i dont think we need 2 pr reviews for this. can just merge directly

PR #669, opened 2025-02-06
feat : req v2 ui and integrations

kushagrasarathe:
- added req v2 parser
- integrated new api routes for req/charge/payment creation
- status views based on charge and request id's

fixes TASK-8486
fixes TASK-7032
fixes TASK-8032
fixes TASK-8889
fixes TASK-8890

## Summary by CodeRabbit

- **New Features**
	- Enhanced the payment workflow with new components for payment history, payment form, payment confirmation, and payment status views, providing real-time updates and clear transaction details.
	- Introduced a guest login modal, simplifying authentication and improving the overall user experience.
	- Added a user-friendly error handling component for displaying payment-related errors.
	- Implemented a new loading indicator and improved interface elements for better navigation and responsiveness across devices.
	- Added a new `PaymentInfoRow` component for displaying payment details in a structured format.
	- Introduced `ValidationErrorView` to inform users of invalid payment URLs with clear instructions.
	- Added new error classes for enhanced validation error handling.
	- Introduced a `PaymentLayout` component for improved structure in payment-related pages.

- **Bug Fixes**
	- Resolved various validation errors and improved error messaging for better user guidance during payment processes.


src/lib/validation/recipient.ts
```diff
@@ -0,0 +1,56 @@
...
+            if (!isValidPeanutUsername) {
+                throw new RecipientValidationError('Invalid Peanut username')
+            }
+
+            return recipient
+
+        default:
+            throw new RecipientValidationError('Invalid recipient type')
+    }
+}
```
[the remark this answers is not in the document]
jjramirezn: I agree with this comment but don't think is blocking

src/components/Global/Icon/index.tsx
```diff
@@ -208,13 +211,11 @@ const icons = {
     wallet: 'M4.398 4.286c-.947 0-1.714.767-1.714 1.714v12c0 .947.767 1.714 1.714 1.714h10.674c.947 0 1.714-.767 1.714-1.714v-12c0-.947-.767-1.714-1.714-1.714H4.398zm8.024 4.8h3.733c.947 0 1.714.767 1.714 1.714v2.4c0 .947-.767 1.714-1.714 1.714h-3.162c-1.42 0-2.571-1.151-2.571-2.571v-.686c0-1.42 1.151-2.571 2.571-2.571z',
 }
 
-export type IconNameType = keyof typeof icons
-
 type IconProps = {
     className?: string
-    name: IconNameType
+    name: any
```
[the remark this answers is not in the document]
jjramirezn: Agree, any bad

src/app/(mobile-ui)/[...recipient]/error.tsx
```diff
@@ -0,0 +1,26 @@
...
+                </Card.Description>
+            </Card.Header>
+            <Card.Content>
+                <Button onClick={reset} variant="purple">
+                    Try again
+                </Button>
+            </Card.Content>
+        </Card>
+    )
+}
```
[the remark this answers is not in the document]
Comment (jjramirezn) [reply src/app/(mobile-ui)/[...recipient]/error.tsx]: @kushagrasarathe is this true? blocking if so I agree with changing the name

src/lib/validation/resolvers/recipient-resolver.ts
```diff
@@ -0,0 +1,36 @@
...
+        }
+        return resolved
+    }
+
+    // check if its a Peanut username
+    const isPeanutUsername = await verifyPeanutUsername(recipient)
+    if (isPeanutUsername) {
+        // todo: move to env or constants
+        // append .testvc.eth to resolve Peanut username
+        const peanutEns = `${recipient}.testvc.eth`
```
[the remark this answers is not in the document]
jjramirezn: @kushagrasarathe blocking agree, use env variable

src/app/(mobile-ui)/[...recipient]/page.tsx
```diff
@@ -0,0 +1,220 @@
...
+                        let nameToResolve = parsedURL.recipient
+
+                        // remove chain part if present (after @)
+                        if (nameToResolve.includes('@')) {
+                            nameToResolve = nameToResolve.split('@')[0]
+                        }
+
+                        // if username has no dots, treat as native peanut username
+                        if (!nameToResolve.includes('.')) {
+                            // todo: move to env
```
Comment (jjramirezn) [inline src/app/(mobile-ui)/[...recipient]/page.tsx]: chore(blocking): do the todo

src/components/Global/Icon/index.tsx
jjramirezn: comment: any = bad

src/app/(mobile-ui)/[...recipient]/page.tsx
```diff
@@ -0,0 +1,220 @@
...
+                }
+
+                if (!recipientAddress) {
+                    throw new Error('No valid recipient address')
+                }
+
+                // fetch requests using the resolved address
+                const fetchedRequest = await requestsApi.search({
+                    recipient: recipientAddress,
+                    // todo: add other filters here
```
Comment (jjramirezn) [inline src/app/(mobile-ui)/[...recipient]/page.tsx]: chore(blocking): do this todo

jjramirezn (changes requested):
Very big PR, confidence in my review quality: low

Some blocking comments and also guest flow is very important

jjramirezn (approved):
Approved the changes. Nothing wrong pops out. We will test once merged

src/app/(mobile-ui)/layout.tsx
```diff
@@ -24,11 +18,7 @@ const publicPathRegex = /^\/(request\/pay|claim)/
 const Layout = ({ children }: { children: React.ReactNode }) => {
     const pathName = usePathname()
     const [isReady, setIsReady] = useState(false)
-    const { signInModal, selectExternalWallet } = useWallet()
-    const web3Modal = useAppKit()
-    const { user, isFetchingUser } = useAuth()
-    const { handleLogin, isLoggingIn } = useZeroDev()
-    const toast = useToast()
+    const { user } = useAuth()
```
Hugo0, line 21: well this is a nice simplifying diff

Hugo0, line 21: well this is a nice simplifying diff

src/app/(mobile-ui)/layout.tsx
```diff
@@ -105,55 +95,7 @@ const Layout = ({ children }: { children: React.ReactNode }) => {
...
-                                    toast.error('Error connecting wallet')
-                                })
-                                .finally(signInModal.close)
-                        }}
-                    >
-                        Connect External Wallet
-                    </Button>
-                </div>
-            </Modal>
+            <GuestLoginModal />
```
Hugo0, line 98: nice!

src/app/[...recipient]/error.tsx
```diff
@@ -0,0 +1,26 @@
...
+    useEffect(() => {
+        console.error(error)
+    }, [error])
+
+    return (
+        <Card className="shadow-primary-4">
+            <Card.Header>
+                <Card.Title>Something went wrong!</Card.Title>
+                <Card.Description>
+                    {error.message || 'An error occurred while loading the payment page.'}
```
Comment (Hugo0) [inline src/app/[...recipient]/error.tsx:16]: nit: is "the payment page" always accurate? Might be overly specific? or no?

src/app/[...recipient]/layout.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import classNames from 'classnames'
+import { twMerge } from 'tailwind-merge'
+
+export default function PaymentLayout({ children }: { children: React.ReactNode }) {
+    return (
+        <div className="flex h-[100dvh] w-full bg-background">
+            {/* Wrapper div for desktop layout */}
+            <div className="flex w-full">
+                {/* Sidebar - Fixed on desktop */}
```
Comment (Hugo0) [inline src/app/[...recipient]/layout.tsx:16]: love the comments

src/app/[...recipient]/loading.tsx
Comment (Hugo0) [inline src/app/[...recipient]/loading.tsx:1]: do we need this component? why not use PeanutLoading directly?

src/app/[...recipient]/page.tsx
```diff
@@ -0,0 +1,236 @@
...
+import { requestsApi } from '@/services/requests'
+import { resolveFromEnsName } from '@/utils'
+import { useRouter, useSearchParams } from 'next/navigation'
+import { useEffect, useMemo, useState } from 'react'
+import { twMerge } from 'tailwind-merge'
+import { isAddress } from 'viem'
+
+export default function PaymentPage({ params }: { params: { recipient: string[] } }) {
+    const dispatch = useAppDispatch()
+    const { currentView, attachmentOptions, resolvedAddress, requestDetails } = usePaymentStore()
```
Comment (Hugo0) [inline src/app/[...recipient]/page.tsx]: thoughts on redux vs react context? super out of scope for this PR, but maybe we should revisit this at some point

src/app/[...recipient]/page.tsx
```diff
@@ -0,0 +1,236 @@
...
+    const chargeId = searchParams.get('chargeId')
+    const requestId = searchParams.get('id')
+
+    // avoid re-parsing the URL on every render
+    const parsedURL = useMemo(() => {
+        try {
+            const parsed = parsePaymentURL(params.recipient)
+
+            // if chain is specified but token not, set USDC as default
+            if (parsed.chain && !parsed.token) {
```
Comment (Hugo0) [inline src/app/[...recipient]/page.tsx]: again, great comments

src/app/[...recipient]/page.tsx
```diff
@@ -0,0 +1,236 @@
...
+            </div>
+        )
+    }
+
+    if (isLoadingCharge || (chargeId && !currentView)) {
+        return <PeanutLoading />
+    }
+
+    if (isLoadingRequests && !chargeId) {
+        return <PeanutLoading />
```
Comment (Hugo0) [inline src/app/[...recipient]/page.tsx:144]: nit: idk  why we have loading.tsx if we use peanutloading directly?

src/app/[...recipient]/page.tsx
```diff
@@ -0,0 +1,236 @@
...
+    }
```
Comment (Hugo0) [inline src/app/[...recipient]/page.tsx]: 1. these 2 statements can be combined
2. might be better to have an explicit isLoadingRequest or isLoadingCharge.  abd not have difficult to parse `|| (chargeId && !currentView)` which is hard to understand without a comment

src/components/Global/FeeDescription/index.tsx
Hugo0, line 1: why is this part of the PR?

src/components/Global/FlowHeader/index.tsx
```diff
@@ -12,7 +12,8 @@ interface FlowHeaderProps {
 const FlowHeader = ({ onPrev, disableBackBtn, disableWalletHeader = false }: FlowHeaderProps) => {
     const { user } = useAuth()
 
-    if (!user) return null
+    // todo: check if this condition removal is affecting other flows
```
Hugo0: ?

src/components/Global/GuestLoginModal/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+    const { handleLogin, isLoggingIn } = useZeroDev()
+    const toast = useToast()
+
+    return (
+        <Modal
+            visible={signInModal.visible}
+            onClose={() => {
+                signInModal.close()
+            }}
+            title={'Sign In with your Peanut Wallet'}
```
Hugo0: nit: lowercase in

src/components/Global/GuestLoginModal/index.tsx
Hugo0, line 1: Do we need this modal? Can't we just redirect to setup? Or is this valuable to have to not have to redirect?

Hugo0: If we want to keep user in line, wouldnt a redirect to setup with a post-login redirect work better? less code duplication

src/components/Payment/History/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+export default function PaymentHistory({ recipient, history }: PaymentHistoryProps) {
+    if (history.length === 0 || !recipient) return null
+
+    const getTransactionType = (type: HistoryEntryType, userRole: HistoryUserRole) => {
+        if (type === 'CLAIM') return 'Link Received'
+        if (type === 'REQUEST' && userRole === 'SENDER') return 'Money Requested'
+        if (type === 'REQUEST' && userRole === 'RECIPIENT') return 'Request paid'
+        if (type === 'CASHOUT') return 'Cash Out'
+        return type
+    }
```
Hugo0: nit: this is never used

src/components/Payment/History/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    return (
+        <div className="space-y-3 border-b border-b-black">
+            <div className="text-base font-semibold">
+                Payment history to <span className="underline">{recipient}</span>
```
Hugo0: Nit: we should have a universal component for identities/accounts that's not just underline: `<span className="underline">{`

src/components/Payment/History/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+                            secondaryInfo={{
+                                mainText: `$${Number(entry.amount).toFixed(2)}`,
+                            }}
+                            metadata={{
+                                recipientAddress: entry.senderAccount
+                                    ? `Paid By: ${entry.senderAccount.username || printableAddress(entry.senderAccount.identifier)} | Status: ${entry.status}`
+                                    : `Status: ${entry.status}`,
+                                subText: new Date(entry.timestamp).toLocaleDateString(),
+                                disableEnsResolution: true,
+                            }}
```
Hugo0, line 43: Q: why disable ensResolution? We should resolve the addresses if possible! If we have an ens name for an address, rather show that than the address itself

src/components/Payment/PaymentForm/index.tsx
```diff
@@ -0,0 +1,361 @@
...
+    const searchParams = useSearchParams()
+    const requestId = searchParams.get('id')
+    const { isConnected: isExternalWalletConnected } = useAccount()
+    const isConnected = isExternalWalletConnected || isPeanutWallet
+
+    const tokenInfo = useMemo(() => SUPPORTED_TOKENS[token?.toUpperCase() || ''], [token])
+
+    const urlChainName = chain && getReadableChainName(chain?.toString())
+
+    const showPeanutWalletWarning =
```
Hugo0: nit: this is unused

src/components/Payment/PaymentForm/index.tsx
```diff
@@ -0,0 +1,361 @@
...
+        // set token from URL if present
+        if (token) {
+            const upperToken = token.toUpperCase()
+            const chainId = Number(selectedChainID)
+
+            // handle native ETH
+            if (upperToken === 'ETH') {
+                // Set native token address
+                setSelectedTokenAddress('0x0000000000000000000000000000000000000000')
+            } else {
```
Hugo0: Q: what about native tokens on other chains? e.g. POL on polygon

Hugo0: Why do we need special behaviour for native eth, cant we get it from indexing supportedtokens w/ upperToken?

src/components/Payment/PaymentForm/index.tsx
```diff
@@ -0,0 +1,361 @@
...
+            }
+
+            if (chainId) {
+                setSelectedChainID(chainId)
+            }
+        }
+
+        // set token from URL if present
+        if (token) {
+            const upperToken = token.toUpperCase()
```
Hugo0: nit: why upperCase QQ

we only use upperToken for string comparison. just explicitly lowercase (or uppercase) both sides during the string comparison

src/components/Payment/PaymentForm/index.tsx
```diff
@@ -0,0 +1,361 @@
...
+                throw new Error('Request details not found')
+            }
+
+            const createChargeRequestPayload: CreateChargeRequest = {
+                pricing_type: 'fixed_price',
+                local_price: {
+                    amount: tokenValue,
+                    currency: 'USD',
+                },
+                baseUrl: window.location.origin,
```
Hugo0, line 113: @jjramirezn we also have a NEXT PUBLIC BASE URL variable. Why did we introduce it? can't we use window.location.origin everywhere? and if not, shouldnt this use  the env var?

src/services/charges.ts
```diff
@@ -0,0 +1,65 @@
...
+            },
+            body: JSON.stringify(data),
+        })
+
+        if (!response.ok) {
+            throw new Error(`Failed to create charge: ${response.statusText}`)
+        }
+
+        return response.json()
+    },
```
Hugo0, line 19: Q: Do we need this code for every single call? Can't we DRY a bit here?

Also I'd rather just raise the error message instead of adding a useless "Failed to create charge" infront

src/components/Payment/Views/Confirm.payment.view.tsx
```diff
@@ -0,0 +1,476 @@
...
+    const { attachmentOptions, urlParams, error, chargeDetails, resolvedAddress } = usePaymentStore()
+    const { selectedChainID, selectedTokenData } = useContext(tokenSelectorContext)
+
+    const searchParams = useSearchParams()
+    const chargeId = searchParams.get('chargeId')
+    const [tokenPriceData, setTokenPriceData] = useState<ITokenPriceData | undefined>(undefined)
+    const [isLoading, setIsLoading] = useState<boolean>(false)
+    const { sendTransactions, checkUserHasEnoughBalance } = useCreateLink()
+    // todo: use redux store for this
+    const [transactionHash, setTransactionHash] = useState<string>('')
```
Hugo0: nit: some unused vars here

src/components/Payment/Views/Error.validation.view.tsx
```diff
@@ -0,0 +1,38 @@
...
+                </h3>
+            </div>
+            <Button
+                onClick={() => {
+                    if (!!user) {
+                        router.push('/home')
+                    } else {
+                        router.push('/signin')
+                    }
+                }}
```
Hugo0: /signin leads to a website that says "invalid payment URL" - whats going on here

src/lib/url-parser/constants/tokens.ts
```diff
@@ -0,0 +1,52 @@
...
+            [mainnet.id]: '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48', // mainnet
+            [optimism.id]: '0x0b2c639c533813f4aa9d7837caf62653d097ff85', // optimism
+            [polygon.id]: '0x3c499c542cef5e3811e1192ce70d8cc03d5c3359', // polygon
+            [arbitrum.id]: '0xaf88d065e77c8cC2239327C5EDb3A432268e5831', // arbitrum
+            [base.id]: '0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913', // base
+            // todo: add more chain addresses
+        },
+        minAmount: '0.01',
+        maxAmount: '1000000',
+    },
```
Hugo0: Issue, Blocking: This is like the 4th place (or more) we now define tokens & chains in (we have in sdk, in bridge supported tokens, in tokenselector, in app, prob in api too, and prob other places too including where we handle squid stuff ). I'm very afraid of so much repetition and how it makes maintainance and debugging much harder. Please give it a think if we can combine this with other places where we maintain tokens and chains, or if we really need a separate registry

src/lib/url-parser/constants/tokens.ts
```diff
@@ -0,0 +1,52 @@
...
+        symbol: 'USDT',
+        name: 'Tether USD',
+        decimals: 6,
+        addresses: {
+            [mainnet.id]: '0xdac17f958d2ee523a2206206994597c13d831ec7',
+            [polygon.id]: '0x3813e82e6f7098b9583FC0F33a962D02018B6803',
+            [arbitrum.id]: '0xfd086bc7cd5c481dcc9c85ebe478a1c0b69fcbb9',
+            [bsc.id]: '0x55d398326f99059ff775485246999027b3197955',
+        },
+        minAmount: '0.01',
```
Hugo0: Blocking Q: Why the arbitrary minimum amount of 1 cent? And for eth, why the mega arbitrary 20cents minimum amount?

src/lib/url-parser/constants/tokens.ts
Hugo0, line 1: What is the purpose of this file? Do we only support these tokens/chains for request? Is the todo: add more chain addresses actionable? When is that scoped in?

src/lib/url-parser/parsers/address.parser.ts
Hugo0, line 1: If this file references the erc-7828 implementation, I think adding this as a comment would be very helpful

src/lib/url-parser/parsers/address.parser.ts
```diff
@@ -0,0 +1,55 @@
...
+        return null
+    }
+
+    // split the input into user and chain parts
+    const [user, chain] = parts
+
+    // validate both parts exist
+    if (!user || !chain) {
+        return null
+    }
```
Hugo0: Q, nit: Why not just return user in this case? also is this user or is this address?

e.g. `0x125125512@` or `kushagra.eth@` would qualify here. I think its slightly better to return kushagra.eth as the parsed param, and null for chain, instead of null for everything

src/lib/url-parser/parsers/address.parser.ts
```diff
@@ -0,0 +1,55 @@
...
+        chain,
+    }
+}
+
+// helper function to validate user
+function isValidUser(user: string): boolean {
+    // check for ENS names first (including those starting with 0x)
+    if (user.endsWith('.eth')) {
+        return ENS_REGEX.test(user)
+    }
```
Hugo0: Issue, blocking: as previously mentioned, ens names dont have to end with `.eth` - please use justaname resolver for validating ens names

src/lib/url-parser/parsers/address.parser.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    // for addresses, must be a valid Ethereum address
+    if (user.startsWith('0x')) {
+        return isAddress(user)
+    }
+
+    // for ens names, must match regex
+    return ENS_REGEX.test(user)
```
Hugo0: nit: these comments are overly verbose

src/lib/validation/constants/index.ts
```diff
@@ -0,0 +1,8 @@
+// L1 TLD list based on supported chains
+export const TLDS = ['eth', 'arbitrum', 'optimism', 'base', 'polygon'] as const
+
```
Hugo0: Issue: same comment as with tokens/chains earlier: do we REALLY need yet another place where we define chains? Please please please lets combine these, this is a breeding ground for unintended behaviour and trouble debugging down the line. The simpler, the better, and this PR already adds 3000 lines of code

src/lib/url-parser/parsers/address.parser.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    // return the parsed parts
+    return {
+        user,
+        chain,
+    }
+}
+
+// helper function to validate user
+function isValidUser(user: string): boolean {
```
Hugo0: nit: the function name "isValidUser" is potentially confusing imo. We resolve:

- peanut user names (no .TLD needed)
- raw addresses
- arbitrary .eth names

All of these resolve to a _single_ _address_. I think `isValidAddress` or `resolvesToAddress` would be better names. This would also mean that we would have to rename the other existing function that we have, `isAddress`, to a more specific name like `isEVMAddress` like suggested in a previous PR review

src/lib/url-parser/types/token.ts
```diff
@@ -0,0 +1,9 @@
+export interface TokenInfo {
+    symbol: string
+    name: string
+    decimals: number
+    addresses: Record<number, string> // chainId -> address mapping
+    logoURI?: string
+    minAmount?: string
+    maxAmount?: string
+}
```
Hugo0: nit: Im worried about repetition again - we already have types for tokens defined in other places. Why do we need to define again? and what's the reasoning behind having min/max?

src/lib/validation/resolvers/chain-resolver.ts
```diff
@@ -0,0 +1,178 @@
...
+}
+
+interface ResolvedChain {
+    id: string
+    name: string
+    isTestnet: boolean
+}
+
+export function resolveChain(chainIdentifier: string): ResolvedChain {
+    // case 1: chain id (0x[a-fA-F0-9]{1,64})
```
Hugo0: nit: good to specify the particular eip that defines chainid namespace, if any

src/lib/validation/resolvers/chain-resolver.ts
```diff
@@ -0,0 +1,178 @@
...
+    const chainName = getChainName(numericChainId.toString())
+    if (chainName) return chainName
+
+    throw new Error(`Unknown chain ID: ${chainId}`)
+}
+
+// normalize chain names using CHAIN_NAME_VARIANTS mapping
+export function normalizeChainName(chainName: string): string {
+    const normalized = chainName.toLowerCase().trim()
+    // check if the normalized string matches any of the variants
```
Hugo0, line 39: mini-nit: useless comment imo

src/lib/validation/resolvers/chain-resolver.ts
```diff
@@ -0,0 +1,178 @@
...
+    // fallback to chain name from utils if not found in variants
+    const chainName = getChainName(numericChainId.toString())
+    if (chainName) return chainName
+
+    throw new Error(`Unknown chain ID: ${chainId}`)
+}
+
+// normalize chain names using CHAIN_NAME_VARIANTS mapping
+export function normalizeChainName(chainName: string): string {
+    const normalized = chainName.toLowerCase().trim()
```
Hugo0, line 38: we trim and lowercase (good) but we don't remove spaces.

Q: What about spaces in the request link? is the expectation someone would write `pay/kushagra@ethereum%20mainnet`?

src/lib/validation/resolvers/chain-resolver.ts
```diff
@@ -0,0 +1,178 @@
...
+    // if no match found, return the original normalized string
+    return normalized
+}
+
+// validate if chain supports a specific token
+export function validateChainTokenSupport(chainId: string, tokenSymbol: string): boolean {
+    const token = SUPPORTED_TOKENS[tokenSymbol.toUpperCase()]
+    if (!token) return false
+
+    return !!token.addresses[Number(chainId)]
```
Hugo0: nit: 
either shorten it like this:
`return token && !!token.addresses[Number(chainId)]`

or, ideally, write it in a way that is easy to parse

```
if (!token) {
return false;
}
if (!token.addresses[Number(chainId)]) {
return false;
}
return true;
```

This last way is more verbose, but easier to understand when reading

src/lib/validation/resolvers/recipient-resolver.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export async function resolveRecipientToAddress(recipient: string): Promise<string> {
+    // if already an address, return it
+    if (isAddress(recipient)) {
+        return recipient
+    }
+
+    // todo: add support for other ENS TLDs
+    // if its an ENS name (ends with .eth)
+    if (recipient.endsWith('.eth')) {
```
Hugo0: again, .eth not only ens tld

src/lib/url-parser/parser.ts
```diff
@@ -0,0 +1,184 @@
...
+import { isAddress } from 'viem'
+import { CHAIN_ID_REGEX } from '../validation/constants'
+import { resolveChainId } from '../validation/resolvers/chain-resolver'
+import { SUPPORTED_TOKENS } from './constants/tokens'
+import { AmountValidationError, ChainValidationError, RecipientValidationError, TokenValidationError } from './errors'
+import { parseChainSpecificAddress } from './parsers/address.parser'
+import { ParsedURL, RecipientType } from './types/payment'
+
+export function detectRecipientType(recipient: string): RecipientType {
+    if (recipient.endsWith('.eth')) return 'ENS'
```
Hugo0: eth not only ens tld, see other commetns

src/lib/url-parser/parser.ts
```diff
@@ -0,0 +1,184 @@
...
+                return baseResult
+            }
+
+            // for all other cases, throw the chain validation error
+            throw new ChainValidationError(error.message)
+        }
+    }
+
+    // check for ENS names
+    if (recipient.endsWith('.eth')) {
```
Hugo0: ^

src/lib/validation/resolvers/recipient-resolver.ts
```diff
@@ -0,0 +1,37 @@
+import { JUSTANAME_ENS } from '@/constants'
+import { resolveFromEnsName } from '@/utils'
+import { isAddress } from 'viem'
+import { verifyPeanutUsername } from '../recipient'
+
+export async function resolveRecipientToAddress(recipient: string): Promise<string> {
```
Hugo0: nit: here we're resolving recipient to address, but we also have a function called `isValidUser` that is very similar (but with a very different name!). Can we consolidate?

src/lib/validation/resolvers/recipient-resolver.ts
```diff
@@ -0,0 +1,37 @@
...
+    // if already an address, return it
```
Hugo0: overly verbose comment, this code is very clear

src/lib/validation/resolvers/recipient-resolver.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (isAddress(recipient)) {
+        return recipient
+    }
+
+    // todo: add support for other ENS TLDs
```
Hugo0: just use justaname resolver

src/lib/validation/resolvers/recipient-resolver.ts
Hugo0: if it has a dot in the name, try to resolve it with justaname - we discussed thsi

.env.example
```diff
@@ -1,3 +1,5 @@
+export NEXT_PUBLIC_JUSTANAME_ENS_DOMAIN=""
```
Hugo0: nit: better to keep vars semantically grouped (e.g. all NEXT_PUBLIC ones together)

src/services/charges.ts
```diff
@@ -0,0 +1,65 @@
+import { PEANUT_API_URL } from '@/constants'
+import { CreateChargeRequest, PaymentCreationResponse, TCharge, TRequestChargeResponse } from './services.types'
+
+export const chargesApi = {
+    create: async (data: CreateChargeRequest): Promise<TCharge> => {
+        const response = await fetch(`/api/proxy/charges`, {
```
Hugo0, line 6: issue, nonblocking: consistency with proxy routes. here you use /api/proxy/charges, but in a different place you use `next_proxy_url`

src/lib/validation/amount.ts
```diff
@@ -0,0 +1,50 @@
...
+        try {
+            const value = parseUnits(amount, token?.decimals || 18)
+            return {
+                value: value.toString(),
+                formatted: amount,
+            }
+        } catch (error) {
+            throw new AmountValidationError('Invalid amount precision')
+        }
+    }
```
Hugo0: Q: rather throw than round down?

src/lib/validation/chain.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Chain } from 'viem'
+import { ChainValidationError } from '../url-parser/errors'
+import { resolveChain } from './resolvers/chain-resolver'
+
+export function validateChain(chainIdentifier: string | number): Chain {
+    try {
+        // convert number to hex string
+        const identifier = typeof chainIdentifier === 'number' ? `0x${chainIdentifier.toString(16)}` : chainIdentifier
+
+        // resolve chain
```
Hugo0: these are very very very very overly verbose comments.

// resolve chain
.... resolvechain()

if you use AI to generate this, please clean up afterwards, don't force a PR review when code is like this

src/lib/validation/resolvers/chain-resolver.ts
```diff
@@ -0,0 +1,178 @@
...
+        supportsNativeEth: true,
+        chainId: base.id.toString(),
+    },
+}
+
+interface ResolvedChain {
+    id: string
+    name: string
+    isTestnet: boolean
+}
```
Hugo0: Issue, blocking: why do we need a `ResolvedChain` type? Why not just use a generic chain type that we use throughout the app? And even if we made a new type, why call it resolvedChain and not type: chain or smth?

please please please lets keep repetition to a minimum. This PR adds 3000 LoC and I worry that a lot of it is unnecessary and surface for bugs

Hugo0: We only use this type in _one_ place. If that's the case, I'd rather just be explicit about the types in the object, instead of giving it a name like ResolvedChain. Now when i go on chain.ts and hover over the return type of `resolveChain` it says it returns `ResolvedChain` type, which tells me absolutely nothing, and i have to go check out what ResolvedChain type is, which results in bad devx

src/lib/validation/resolvers/chain-resolver.ts
```diff
@@ -0,0 +1,178 @@
...
+        chainId: optimism.id.toString(),
+    },
+    arbitrum: {
+        variants: ['arbitrum', 'arbitrum one', 'arbitrum mainnet', 'arb', 'arbitrum-one'],
+        supportsNativeEth: true,
+        chainId: arbitrum.id.toString(),
+    },
+    polygon: {
+        variants: ['polygon', 'polygon pos', 'matic', 'polygon mainnet'],
+        supportsNativeEth: false,
```
Hugo0: Q, issue: what is the logic behind supportsNativeEth? Why do we need this? Seems like a potential footgun.

Also as mentioned earlier, we define chains and tokens in sooooooooooo many places. Walletconnect chains, mobula chains, now these chains, its a complete mess. Please make an effort to consolidate if possible

src/lib/validation/recipient.ts
```diff
@@ -0,0 +1,56 @@
...
+    } catch (err) {
+        return false
+    }
+}
+
+export async function validateRecipient(recipient: string, type: RecipientType): Promise<string> {
+    switch (type) {
+        case 'ENS':
+            // check if the recipient is a valid ENS name
+            if (!recipient.endsWith('.eth')) {
```
Hugo0: Two issues:
1. as mentioned earlier, more ens TLDs than just .eth
2. why do we need image.png so many places where we check for .eth?
- `src/lib/url-parser` twice
- `address.parser.ts`
- `recipient.ts`
- `chain-resolver.ts` (this one is potentially understandable)
- `recipient-resolver.ts`

This seems like insane repetition to me and i don't think we need it

src/lib/validation/token.ts
```diff
@@ -0,0 +1,40 @@
...
+import { SUPPORTED_TOKENS } from '../url-parser/constants/tokens'
+import { TokenValidationError } from '../url-parser/errors'
+import { TokenInfo } from '../url-parser/types/token'
+import { normalizeChainName, supportsNativeEth } from './resolvers/chain-resolver'
+
+export interface ValidatedToken {
+    info: TokenInfo
+    address?: string // optional because ETH is native
+    isNative: boolean
+}
```
Hugo0: issue: same as resolvedChain, this type definition seems nonsensical to me

src/lib/validation/token.ts
```diff
@@ -0,0 +1,40 @@
+import { Chain } from 'viem'
+import { SUPPORTED_TOKENS } from '../url-parser/constants/tokens'
+import { TokenValidationError } from '../url-parser/errors'
+import { TokenInfo } from '../url-parser/types/token'
+import { normalizeChainName, supportsNativeEth } from './resolvers/chain-resolver'
+
+export interface ValidatedToken {
+    info: TokenInfo
+    address?: string // optional because ETH is native
```
Hugo0: praise: there's a comment explaining why address is optional, which is good, we should always comment edge cases that are hard to understand. And eth in this case is an edge case. However the more elegant method is to just deal with it so that you catch edge cases - in other cases we've handled the native gas token by giving it the 0x00 address

src/lib/validation/token.ts
```diff
@@ -0,0 +1,40 @@
...
+    const normalizedSymbol = tokenSymbol.toUpperCase()
+    const tokenInfo = SUPPORTED_TOKENS[normalizedSymbol]
+
+    if (!tokenInfo) {
+        throw new TokenValidationError(`Unsupported token: ${tokenSymbol}`)
+    }
+
+    if (chain) {
+        const normalizedChainName = normalizeChainName(chain.name)
+        const isNative = normalizedSymbol === 'ETH' && supportsNativeEth(normalizedChainName)
```
Hugo0: nit: When doing string comparisons, its a good practice to lowercase/uppercase explicitly during the comparison. In this case it says "normalizedSymbol" but if im debugging im not sure, from just one glance, if normalized means its been uppercased or lowercased (alongside trimming etc)

src/redux/types/payment.types.ts
```diff
@@ -0,0 +1,18 @@
+import { IAttachmentOptions } from '@/components/Create/Create.consts'
+import { ParsedURL } from '@/lib/url-parser/types/payment'
+import { PaymentCreationResponse, TCharge, TRequestChargeResponse, TRequestResponse } from '@/services/services.types'
+
+export type TPaymentView = 'INITIAL' | 'CONFIRM' | 'SUCCESS'
+
+export interface IPaymentState {
+    currentView: 'INITIAL' | 'CONFIRM' | 'SUCCESS'
```
Hugo0: Q: what about payment state where you just reload the page to check the payment status? would that be success?

src/utils/__tests__/url-parser.test.ts
```diff
@@ -0,0 +1,175 @@
...
+
+        it('should parse ENS name starting with 0x', () => {
+            const result = parsePaymentURL(['0xdhruv.eth@arbitrum'])
+            expect(result).toEqual({
+                recipient: '0xdhruv.eth',
+                recipientType: 'ENS',
+                chain: '42161',
+            })
+        })
+    })
```
Hugo0: nit: can generate more test cases than just arbitrum

src/utils/__tests__/url-parser.test.ts
Hugo0, line 1: Issue, blocking: no tests for peanut usernames!

Also, test format seems not ideal imo. We should test the full string format as a user would write it - not already with extracted URL params. I would love to test specifically strings like this:

`<baseURL>/konrad@optimism/10usdc`
`/konrad/10`
`konrad/usdc`
`konrad@base/usdc`
`konrad.eth/0.1/eth`

Also, 3000 LoC PR but only 150 LoC for tests seems skewed to me.

Hugo0 (changes requested):
+3600 Lines of Code is a lot, and I worry we're repeating instead of reusing code that we have for other functions, thus  increasing surface for bugs and making debugging and further maintainance much harder.

src/app/[...recipient]/error.tsx
Comment (kushagrasarathe) [reply src/app/[...recipient]/error.tsx:16]: imo it is good, since it's a page where I can pay users or pay a request created by a user, it makes more sense to keep it like this, do you have any other suggestions ?

src/app/[...recipient]/loading.tsx
Comment (kushagrasarathe) [reply src/app/[...recipient]/loading.tsx:1]: added this to smoothen the ux a bit, it's next.js loading for a particular route and helps show a loading state while all the components in that route are being loaded.

src/app/[...recipient]/page.tsx
Comment (kushagrasarathe) [reply src/app/[...recipient]/page.tsx]: wdym? revisit as in only keep use one thing out of two, if that's what you mean then yes thats the plan, rn i'm using redux for every new feature we build, ideas is to migrate old ones too to redux gradually, sofar it hasn't worked well due to time constraints tho

src/app/[...recipient]/page.tsx
Comment (kushagrasarathe) [reply src/app/[...recipient]/page.tsx:144]: the transition b/w different views (Initial, confirm) wasn't smooth wen reqs were being fetched, so showing these loading states for those, while loading.tsx is for next route segments

src/app/[...recipient]/page.tsx
Comment (kushagrasarathe) [reply src/app/[...recipient]/page.tsx]: adding a comment explaing

src/components/Global/FeeDescription/index.tsx
kushagrasarathe: included them to use in the confirm view, going forward, we will follow similar ui structure and reuse these components, mainly `InfoRow`

src/components/Global/FlowHeader/index.tsx
kushagrasarathe: it's a note, needed to test this change for other flows, forgot to remove the comment, mb :*(

src/components/Global/GuestLoginModal/index.tsx
kushagrasarathe: actually this is good imo, the way i see it is, from ux pov, if i'm on payment page, i want to pay a req, and i'm not logged in, using this modal i've tow options, either i login to my peanut user using my passkey or i connect a external wallet to pay that request

src/components/Payment/History/index.tsx
kushagrasarathe: agree, there's a bug with that, i need to parse the recipient id separately in this case since in this case, otherwise ListItemView would try to parse the weird string like `To: 0cad.... | Status: NEW`, resulting in error, i've this thing tracked on my end, but considering the other issues in hand, had to push this back for a while

src/redux/types/payment.types.ts
kushagrasarathe: yes, should have renamed to STATUS, poor code management by me :*(

src/services/charges.ts
kushagrasarathe: wdym, i did not get this

> Do we need this code for every single call?

do you mean, that content type, body, method type specification to be reused somehow ?

src/utils/__tests__/url-parser.test.ts
```diff
@@ -0,0 +1,396 @@
...
+
+        it('should parse ENS with chain and token', async () => {
+            const result = await parsePaymentURL(['vitalik.eth@ethereum', '1ETH'])
+            expect.objectContaining({
+                recipient: expect.objectContaining({
+                    identifier: 'vitalik.eth',
+                    recipientType: 'ENS',
+                    resolvedAddress: '0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045',
+                }),
+                chain: expect.objectContaining({ chainId: 1 }),
```
Hugo0, line 220: why is chainid number? we represent chainids as strings

src/utils/__tests__/url-parser.test.ts
```diff
@@ -0,0 +1,396 @@
...
+        },
+        verifyPeanutUsername: (handle: string) => {
+            const validUsernames = ['kusharc', 'satoshi']
+            return Promise.resolve(validUsernames.includes(handle))
+        },
+        getRecipientType: (recipient: string) => {
+            if (recipient.includes('.')) return 'ENS'
+            if (recipient.startsWith('0x')) return 'ADDRESS'
+            return 'USERNAME'
+        },
```
Hugo0: 1. this mock function is better than our real function lmao

2. why do we need a mock function? cant we import getrecipientype?

src/services/charges.ts
Hugo0: yes, i meant that. it's a very nitpick issue though and maybe bad approach

src/lib/url-parser/parser.consts.ts
Hugo0, line 1: 1. why spaces? 
2. this has to be massively expanded no?

Suggestion: we already have different chain names from our diff chain definition/sources (mobula, squid, reown, wagmi, etc). we can just combine them and match on chainid

src/app/[...recipient]/error.tsx
Comment (Hugo0) [reply src/app/[...recipient]/error.tsx:16]: oki!

src/lib/url-parser/parser.ts
```diff
@@ -0,0 +1,166 @@
...
+    amountString = amountString.trim()
+
+    // handle empty string
+    if (!amountString) {
+        return {}
+    }
+
+    // match number part (including decimals) followed by token part
+    // this regx will match patterns like "0.1usdc", "100eth", "1.5pol"
+    const match = amountString.match(/^(\d*\.?\d*)([a-zA-Z]+)$/)
```
Hugo0, line 21: test for this regex?
should ideally have a parset.test.ts file right next to this file, in the same order structure, that tests this regex and parseAmountAndToken

src/lib/url-parser/parser.ts
```diff
@@ -0,0 +1,166 @@
...
+
+        // handle empty chain part after @
+        if (!chain || chain.trim() === '') {
+            chain = undefined
+        }
+    }
+
+    const recipientDetails = await validateAndResolveRecipient(recipient)
+
+    // get all squid chains and tokens
```
Hugo0: 1. useless comment
2. two awaits? this sucks

consider if we need two awaits, makes code slow. If we really need, you can parallelize them with promise all

src/lib/url-parser/parser.ts
```diff
@@ -0,0 +1,166 @@
...
+        throw new RecipientValidationError('Invalid URL format: No recipient specified')
+    }
+
+    // decode the first segment to handle URL encoding
+    const firstSegment = decodeURIComponent(segments[0])
+    let recipient = firstSegment
+    let chain: string | undefined = undefined
+    let chainDetails: (interfaces.ISquidChain & { tokens: interfaces.ISquidToken[] }) | undefined
+
+    // handle multiple @ symbols - take the last one as chain identifier
```
Hugo0, line 54: good comment. but what's an example where this could be the case?

src/lib/url-parser/parser.ts
```diff
@@ -0,0 +1,166 @@
...
+            )
+            return {
+                recipient: recipientDetails,
+                token: tokenDetails,
+                chain: chainDetails,
+                ...(validatedAmount && { amount: validatedAmount.amount }),
+            }
+        }
+
+        // if only amount specified, return with current chain
```
Hugo0, line 135: where is current chain coming from?

Hugo0: should the response here even include a chain in that case? wouldnt it be better to include no chain in the response, and for the caller to decide what to do with the missing chain?

src/utils/general.utils.ts
```diff
@@ -347,18 +347,32 @@ export async function resolveFromEnsNameAndProviderUrl(
 }
 
 export async function resolveFromEnsName(ensName: string): Promise<string | undefined> {
-    let resolvedAddress: string | undefined
+    const mainProviderUrl = 'https://mainnet.infura.io/v3/' + infuraApiKey
```
Hugo0: shouldnt infuraapikey be a const and thus uppercase?

kushagrasarathe: it is a const, updating it to be uppercase, this was like this from before ig

src/lib/url-parser/parser.ts
kushagrasarathe: - current chain is the chain details from squid data, based on url param
- ahahha, good catch, mb, should be undefined actually

src/lib/url-parser/parser.ts
kushagrasarathe: i'll add a test case for this 🫡

src/lib/url-parser/parser.ts
kushagrasarathe: as discussed, will put these in a new separate pr

src/utils/__tests__/url-parser.test.ts
kushagrasarathe: refactored, added it during debugging, forgot to remove 🫠

src/utils/__tests__/url-parser.test.ts
kushagrasarathe: actually treating all the params as string for payments, so for consistency kept it like this 👀

Hugo0 (approved):
APPROVED

src/components/Payment/PaymentForm/index.tsx
jjramirezn: This should use the env var for consistency. We added NEXT_PUBLIC_BASE_URL because somethings would break on preview and also to have it available on the next server side

src/lib/validation/resolvers/chain-resolver.ts
jjramirezn: I think we should expect or accept spaces in these links. The Idea is they should be human readeable and also human writeable

src/services/charges.ts
jjramirezn: agree, also next_proxy_url should be UPPER_CASE

PR #679, opened 2025-02-10
fix: sending funds using native tokens fails with error, even wen enough balance

kushagrasarathe:
fixes TASK-8767

## Summary by CodeRabbit

- **New Features**
	- Enhanced transaction processing to estimate and include associated fees when verifying available wallet funds.
	- Improved error messages now clarify that users must have sufficient balance to cover both the token amount and the transaction fees.
- **Bug Fixes**
	- Streamlined error handling for native token transactions by removing redundant balance checks.

### DIFF
--- a/src/components/Create/Link/Input.view.tsx
+++ b/src/components/Create/Link/Input.view.tsx
@@ -130,7 +130,7 @@ export const CreateLinkInputView = ({
                     setLoadingState('Loading')
                     await checkUserHasEnoughBalance({
                         tokenValue: tokenValue,
-                        gasAmount: maxGasAmount * 1.1, // Add 10% buffer for gas price fluctuations
+                        gasAmount: maxGasAmount,
                     })
                 }
             } else {
--- a/src/components/Create/Link/Input.view.tsx
+++ b/src/components/Create/Link/Input.view.tsx
@@ -100,8 +100,43 @@ export const CreateLinkInputView = ({
                 showError: false,
                 errorMessage: '',
             })
-            setLoadingState('Asserting values')
-            await checkUserHasEnoughBalance({ tokenValue: tokenValue })
+
+            // for native token, estimate gas first before checking balance
+            if (isNativeCurrency(selectedTokenAddress)) {
+                setLoadingState('Loading')
+                const linkDetails = generateLinkDetails({
+                    tokenValue: tokenValue,
+                    envInfo: environmentInfo,
+                    walletType: walletType,
+                })
+
+                const password = await generatePassword()
+                const prepareDepositTxsResponse = await prepareDepositTxs({
+                    _linkDetails: linkDetails,
+                    _password: password,
+                })
+
+                if (prepareDepositTxsResponse) {
+                    const { feeOptions } = await estimateGasFee({
+                        chainId: selectedChainID,
+                        preparedTx: prepareDepositTxsResponse.unsignedTxs[0],
+                    })
+
+                    // calculate gas amount in ETH
+                    const gasLimit = BigInt(feeOptions.gasLimit)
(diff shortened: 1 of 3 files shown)


Hugo0 (commented):
idg it, this doesnt touch request flow?

> fix: paying req links and sending funds with native tokens fails with error #679

src/utils/sdkErrorHandler.utils.tsx
```diff
@@ -82,7 +82,7 @@ export const ErrorHandler = (error: any) => {
                 .toString()
                 .includes('Please ensure that you have sufficient balance of the token you are trying to send')
         ) {
-            return 'Please ensure that you have sufficient balance of requested token.'
+            return 'Please ensure that you have sufficient balance of the token you are trying to send including gas fees.'
```
Hugo0: nit: missing comma or pareentheses

src/components/Create/Link/Input.view.tsx
```diff
@@ -100,8 +100,43 @@ export const CreateLinkInputView = ({
...
+                    _password: password,
+                })
+
+                if (prepareDepositTxsResponse) {
+                    const { feeOptions } = await estimateGasFee({
+                        chainId: selectedChainID,
+                        preparedTx: prepareDepositTxsResponse.unsignedTxs[0],
+                    })
+
+                    // calculate gas amount in ETH
```
Hugo0: nit" in native token **

src/components/Create/useCreateLink.tsx
```diff
@@ -53,7 +54,14 @@ export const useCreateLink = () => {
                     selectedChainID,
                     selectedTokenAddress
                 )?.amount
-                if (!balanceAmount || (balanceAmount && balanceAmount < Number(tokenValue))) {
+
+                // consider gas fees in the balance check for native/non-stable tokens
+                const effectiveAmount =
```
Hugo0: nit: effectiveAmount could be more specific since we have multiple amounts here

suggestion: totalNativeTokenAmount or effectiveNativeTokenAmount

src/components/Create/useCreateLink.tsx
```diff
@@ -53,7 +54,14 @@ export const useCreateLink = () => {
...
+                    isNativeCurrency(selectedTokenAddress) && gasAmount
+                        ? Number(tokenValue) + gasAmount
+                        : Number(tokenValue)
+
+                if (!balanceAmount || (balanceAmount && balanceAmount < effectiveAmount)) {
                     throw new Error(
                         'Please ensure that you have sufficient balance of the token you are trying to send'
```
Hugo0, line 66: Q: is this the correct error message if no balanceAmount from the balanceByToken call? what if the balance provider just doesn't find that token

src/utils/sdkErrorHandler.utils.tsx
```diff
@@ -82,7 +82,7 @@ export const ErrorHandler = (error: any) => {
                 .toString()
                 .includes('Please ensure that you have sufficient balance of the token you are trying to send')
         ) {
-            return 'Please ensure that you have sufficient balance of requested token.'
```
Hugo0, line 85: i have very mixed feelings about this file

Hugo0 (approved):
The task name & description is very very confusing. PR looks good though

src/components/Create/useCreateLink.tsx
kushagrasarathe: i think yes, the balanceAmount is being calculated based on accounts balance, and selected token, if balance provider doesn't find that token in the account, then its highly likely the case that user doesnt have balance for that token, in which case this error is fine imo

PR #689, opened 2025-02-12
fix: claim flow state updates lead to claiming funds on xChain in peanut wallet

kushagrasarathe:
fixes TASK-8881, where on the claim flow, switching wallets back and forth using wallet header would lead to claiming funds on sent network in peanut wallet even tho should only have happened with USDC on Arb

also fixes state update issue in the recipient input when requesting funds and switching from peanut wallet to external wallet

## Summary by CodeRabbit

- **New Features**
	- Wallet and recipient handling has been improved. Now, when the wallet connection or type changes, related selections and validations reset automatically.
	- Button labels and error messages have been refined, offering clearer instructions and feedback to ensure a smoother user experience.

### DIFF
--- a/src/components/Claim/Link/Initial.view.tsx
+++ b/src/components/Claim/Link/Initial.view.tsx
@@ -407,14 +407,23 @@ export const InitialClaimLinkView = ({
             return
         }
 
-        // reset recipient when wallet type changes or when selected wallet changes
+        // reset recipient and chain selection when wallet type changes or when selected wallet changes
         if (selectedWallet) {
-            // reset recipient
+            // reset states
             setRecipient({ name: undefined, address: '' })
             setIsValidRecipient(false)
+            setSelectedRoute(null)
+            setHasFetchedRoute(false)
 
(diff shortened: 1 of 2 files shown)


src/components/Claim/Link/Initial.view.tsx
```diff
@@ -626,12 +625,12 @@ export const InitialClaimLinkView = ({
                                 // imperative to ensure address here is fetched by useWallet
                                 if (!isConnected && recipient.address.length === 0) {
                                     handleConnectWallet()
-                                } else if ((hasFetchedRoute && selectedRoute) || recipient.address !== address) {
-                                    if (recipientType === 'iban' || recipientType === 'us') {
-                                        handleIbanRecipient()
```
jjramirezn, line 631: question(blocking): is it ok to remove handleIbanRecipient?

kushagrasarathe: aaah, nooo, thanks for flagging, forgot to re-add it, doing it now

PR #698, opened 2025-02-19
[TASK-9019] fix: dont trim handle with space

jjramirezn:
This was causing the validator to hang because the change in value was not being registered

## Summary by CodeRabbit

- **Refactor**
  - Updated the signup input handling so that user entries are now converted to lowercase without trimming whitespace. This change affects how handles are processed during sign-up.


kushagrasarathe (approved):
lgtm

PR #720, opened 2025-03-05
fix: req v2 token/balance realted bugs

kushagrasarathe:
- fixes TASK-9434 :  now showing USD values 
- fixes TASK-9455 : fractional decimals error
- fixes TASK-9314 : show 2 significant digits after decimal
- fixes TASK-9454 : issue with reading token symbol 
- fixes TASK-9436 : hard refresh page if clicked on sidebar link and on same path
- fixes TASK-9467 : default to arbitrum in case token is specified but chain is not, in the url
- fixes TASK-9535 : invalid token address error when only chain and recipient were specified in the url, now defaults to usdc token selection in the token selector

## Summary by CodeRabbit

- **New Features**
  - The payment form now fetches token prices in real time and displays the corresponding USD value, providing clearer transaction details.
  - The confirmation view has been updated to format token amounts and recipient addresses more clearly.
  - Navigation links refresh the page when active, ensuring the latest content is shown.
  - Expanded support now recognizes additional chain name variants, including Binance Smart Chain, for improved payment URL processing.

- **Bug Fixes**
  - Improved input validation and error handling, ensuring smoother and more reliable payment transactions.


src/components/Payment/PaymentForm/index.tsx
```diff
@@ -93,7 +105,7 @@ export const PaymentForm = ({ recipient, amount, token, chain }: ParsedURL) => {
             case 'ADDRESS':
                 let tokenSymbol = selectedTokenData?.symbol ?? getTokenSymbol(recipientTokenAddress, recipientChainId)
                 if (!tokenSymbol) {
-                    throw new Error('Failed to get token symbol')
```
jjramirezn, line 96: question(blocking): why? in this case we might have a tokenAddress a token and inform USDC when it's not

src/components/Payment/PaymentForm/index.tsx
```diff
@@ -116,23 +128,26 @@ export const PaymentForm = ({ recipient, amount, token, chain }: ParsedURL) => {
...
-                return tokenDecimals
+                if (recipientTokenAddress) {
+                    const decimals = getTokenDecimals(recipientTokenAddress, recipientChainId)
+                    if (decimals !== undefined) {
+                        return decimals
+                    }
+                }
+                return 6
             default:
-                throw new Error('Invalid recipient type')
```
jjramirezn, line 126: same q as above

src/components/Payment/PaymentForm/index.tsx
```diff
@@ -167,6 +192,45 @@ export const PaymentForm = ({ recipient, amount, token, chain }: ParsedURL) => {
         }
     }, [isPeanutWallet])
 
+    // fetche token price
+    useEffect(() => {
```
jjramirezn, line 196: question: why is this needed and why the current fetching in the tokenselectorcontext is not working?

also typo in comment

src/components/Payment/PaymentForm/index.tsx
```diff
@@ -289,14 +365,21 @@ export const PaymentForm = ({ recipient, amount, token, chain }: ParsedURL) => {
     const renderRequestedPaymentDetails = () => {
         if (!requestDetails || !requestDetails.tokenAmount || !requestDetails.tokenSymbol) return null
 
+        const tokenAmount = parseFloat(requestDetails.tokenAmount)
+        const tokenUsdValue =
+            requestedTokenPrice && tokenAmount > 0 ? formatAmount(tokenAmount * requestedTokenPrice) : ''
+
+        const tokenSymbol = requestDetails.tokenSymbol || token?.symbol
+
+        const dispalyAmount = tokenUsdValue
```
jjramirezn: typo

src/components/Payment/Views/Confirm.payment.view.tsx
```diff
@@ -139,31 +142,36 @@ export default function ConfirmPaymentView() {
...
-                        recipientAddress: chargeDetails.requestLink.recipientAddress,
-                        chainId: chargeDetails.chainId,
-                        tokenAmount: chargeDetails.tokenAmount,
-                        tokenAddress: chargeDetails.tokenAddress,
-                        tokenDecimals: chargeDetails.tokenDecimals,
-                        tokenType: chargeDetails.tokenType,
-                    },
-                    address
-                )
+                try {
```
jjramirezn: try/catch does nothing

src/lib/url-parser/parser.ts
```diff
@@ -129,8 +129,9 @@ export async function parsePaymentURL(
             }
         }
         if (token) {
-            if (!chainDetails) {
-                return { parsedUrl: null, error: { message: EParseUrlError.INVALID_CHAIN, recipient } }
+            if (!chainDetails && !isPeanutRecipient) {
+                // default to arb even for non-USERNAME recipients if no chain is specified
```
jjramirezn, line 133: dont know if this behavior is desired, I think in this case we should be explicit and have chain

jjramirezn (commented):
Reviewed and left some questions

src/components/Payment/PaymentForm/index.tsx
kushagrasarathe: this was leading to error when going directly to `p.me/kushagrasarathe.eth` if token was not defined in url, i added usdc as default to avoid this error, and users now have option to select token/chain in ui

image

src/components/Payment/PaymentForm/index.tsx
kushagrasarathe: leading to bad ux, token decimals were not being properly found and set, ideally instead of throwing errors for these scenarios, i'd prefer showing the ui and let users select token/chain, thats why this solution

src/components/Payment/PaymentForm/index.tsx
kushagrasarathe: @jjramirezn the token selector context works when a token/chain is selected in it, but for cases like when peanut wallet is selected or on the first load, the requested token price aren't fetched, and also the price would switch if i switch token in the selector, so to specifically fetch token price details of the requested tokens, added this call

src/lib/url-parser/parser.ts
kushagrasarathe: defaulted it to arb, to avoid client side errors, users anyways have the option to change it from token selector

src/components/Payment/Views/Confirm.payment.view.tsx
kushagrasarathe: noted, missed it, was trying somethings, removing it, thanks

src/components/Payment/PaymentForm/index.tsx
jjramirezn: this is not used anywhere now, also slice in redux

jjramirezn (approved):
Approved 

But before merging please delete the userEffect in the comment and the setPaymentValues in redux

PR #765, opened 2025-03-24
fix(qr-scanner): reset payment status on qr scan

jjramirezn:
We were having the issue that when we scanned after paying, we would go to the status page of the last payment done. Reseting the status on scan fixes this.

## Summary by CodeRabbit

- **Refactor**
  - Streamlined QR code payment processing by consolidating the state update and navigation logic.
  - Ensures that payment actions and navigation occur in a cohesive, sequential manner for improved transaction flow.


kushagrasarathe (approved):
lgtm 🫡

PR #786, opened 2025-03-31
fix: use justaname batch api call for resolving to ens

kushagrasarathe:
- now using justaname's batch `usePrimaryNameBatch` hook to resolve to ens names
- contributes to TASK-10131
- also fixes TASK-9912 : address resolution in link previews

## Summary by CodeRabbit

- **New Features**
  - Wallet displays now show more recognizable names by retrieving and displaying ENS names for wallet addresses. Users will see a primary name when available, with the address serving as a fallback.
  
- **Refactor**
  - Updated the wallet display components to streamline data flow and consistently pass the primary name to wallet cards.
  - Simplified logic for retrieving primary names in the wallet display components.


Hugo0 (approved):
good initiative!

PR #799, opened 2025-04-11
feat: show qr on req create and payment history sections

kushagrasarathe:
contributes to TASK-10347 and also contributes to TASK-10349

## Summary by CodeRabbit

- **New Features**
  - Introduced a streamlined interface for creating token request links with optional attachments, enhanced sharing options, and dynamic QR code display.

- **Enhancements**
  - Improved modal behavior for transaction details with clearer sharing and download options.
  - Added loading state support with visual feedback to the QR code component.
  - Updated navigation header for a more compact and refined look.

- **Bug Fixes**
  - Removed duplicate imports in the QR code drawer component.

- **Refactor**
  - Consolidated the request creation flow into a single component, removing obsolete multi-step views and related exports.

- **Chores**
  - Removed the "Receive" button from the wallet actions on the home page.


src/components/Global/QRBottomDrawer/index.tsx
```diff
@@ -21,7 +22,7 @@ const QRBottomDrawer = ({ url, collapsedTitle, expandedTitle, text, buttonText }
             collapsedHeight={23}
             halfHeight={75}
             expandedHeight={95}
-            isOpen={true}
+            isOpen={open ? open : true}
```
jjramirezn: issue: This doesn't behave as intended. if open is false then true will be pass to isOpen

QRBottomDrawer should have an isOpen (better than `open`) prop that defaults to true and then pass that prop to BottomDrawer

src/components/Global/ListItemView/index.tsx
```diff
@@ -197,6 +205,18 @@ export const ListItemView = ({ id, variant, primaryInfo, secondaryInfo, metadata
                             Check Status
                         </a>
                     )}
+                    {transactionDetails?.type === 'Request Link' && !transactionDetails?.txHash && (
+                        <div className="p-6">
+                            <QRCodeWrapper url={transactionDetails.link!} />
```
jjramirezn, line 139: issue: pending send links also need to be shareable

src/components/Request/Create/Views/Initial.view.tsx
```diff
@@ -55,6 +55,19 @@ export const InitialView = ({
         (inputDenomination === 'TOKEN' ? tokenValue : usdValue) ?? ''
     )
 
+    const qrCodeLink = useMemo(() => {
```
jjramirezn: suggestion: getRequestLink utils function should be used here

https://github.com/peanutprotocol/peanut-ui/blob/peanut-wallet/src/utils/general.utils.ts#L1033

jjramirezn (changes requested):
QR works!!!

Need some changes

src/components/Request/Create/Views/Initial.view.tsx
kushagrasarathe: @jjramirezn  i think this is fine we will need whole req details if i use getRequestLink fn, with this implementation the qr updates quickly when user changes wallets, chain, token or amount

jjramirezn:
I think we can implement the screen closer to the hifi already. I even think what we can remove external wallets from the request flow, that way we don't have to try to shoehorn the external wallet there. One think that may change (and I did not note when reviewing) Is that we may still may need the confirm button to actually store the memo/attachment if the user add those

signal-2025-04-11-170804_002

jjramirezn:
praise: It looks goooood 

Two things:

1. There is a debouncing issue when modifying value when we have memo, instead of 10 the request is generated with 1

https://github.com/user-attachments/assets/b6e78eaa-a0d3-47be-b8c1-2c66aa648b68


2. We should hide the receive button for external wallets to avoid confusion
signal-2025-04-14-223819_003

jjramirezn (changes requested):
Address comment

PR #804, opened 2025-04-16
fix: req creation using old req amount

kushagrasarathe:
fixes TASK-10535

## Summary by CodeRabbit

- **Bug Fixes**
  - Refined request handling to exclude empty token amounts, ensuring accurate payment data when fetching recipient requests.

### DIFF
--- a/src/app/[...recipient]/page.tsx
+++ b/src/app/[...recipient]/page.tsx
@@ -99,7 +99,12 @@ export default function PaymentPage({ params }: { params: { recipient: string[] 
 
                 //  conditional request params
                 const requestParams: any = { recipient: recipientIdentifier }
-                if (parsedPaymentData.amount) requestParams.tokenAmount = parsedPaymentData.amount
+
+                // include amount only if it is explicitly provided in the URL
+                if (parsedPaymentData.amount && parsedPaymentData.amount !== '') {
+                    requestParams.tokenAmount = parsedPaymentData.amount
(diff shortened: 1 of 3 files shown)


src/components/Global/BottomDrawer/index.tsx
```diff
@@ -86,7 +86,7 @@ const BottomDrawer: React.FC<BottomDrawerProps> = ({
         if (contentRef.current && sheetRef.current) {
             // Small delay to ensure content is rendered
             setTimeout(() => {
-                if (sheetRef.current) {
+                if (contentRef.current && sheetRef.current) {
```
jjramirezn: not needed, pull from peanut-wallet-dev already fixed

src/app/[...recipient]/page.tsx
```diff
@@ -99,7 +99,12 @@ export default function PaymentPage({ params }: { params: { recipient: string[]
 
                 //  conditional request params
                 const requestParams: any = { recipient: recipientIdentifier }
-                if (parsedPaymentData.amount) requestParams.tokenAmount = parsedPaymentData.amount
+
+                // include amount only if it is explicitly provided in the URL
+                if (parsedPaymentData.amount && parsedPaymentData.amount !== '') {
```
Comment (jjramirezn) [inline src/app/[...recipient]/page.tsx:104]: bad, we may get amount from request or from charge (and that is more important than amount in case the url is malformed)

PR #806, opened 2025-04-17
feat: public profile ui

kushagrasarathe:
contributes to TASK-10531

## Summary by CodeRabbit

- **New Features**
  - Introduced a public profile view for users, displaying profile details and recent transaction activity.
  - Added a "Send" button on public profiles, allowing users to initiate payments directly from the profile page.
  - Added a new user icon for enhanced visual cues.

- **Improvements**
  - Enhanced navigation header with support for custom actions.
  - Improved transaction card display to better handle cases without a status.

- **Bug Fixes**
  - Corrected empty state messaging and actions for users with no transaction history.


src/app/[...recipient]/page.tsx
```diff
@@ -145,6 +153,34 @@ export default function PaymentPage({ params }: { params: { recipient: string[]
...
+        !chargeId &&
+        !requestId &&
+        !showPaymentView
+    ) {
+        const username = parsedPaymentData.recipient.identifier
+
+        return (
+            <PublicProfile
+                username={username}
+                fullName={username} // todo: replace with actual full name
```
Comment (jjramirezn) [inline src/app/[...recipient]/page.tsx]: todo: do todo

user service has the getByUsername method

src/components/Global/Icons/user-plus.tsx
jjramirezn, line 1: question: why do we use a new component here instead of using components/Global/Icon

src/components/Profile/components/PublicProfile.tsx
```diff
@@ -0,0 +1,232 @@
...
+                                <h2 className="text-lg font-bold">Transactions</h2>
+                                {/* todo: fix href */}
+                                <Link href={`/transactions?user=${username}`}>
+                                    <div className="flex items-center">
+                                        <Icon name="chevron-up" size={16} fill="black" className="rotate-90" />
+                                    </div>
+                                </Link>
+                            </div>
+
+                            {/* Example transaction - todo: replace with actual transaction data */}
```
jjramirezn, line 145: todo: do todo

Wont be the same as in the history project, for now we can show what we already show. At least until doing history project

src/components/Profile/components/PublicProfile.tsx
```diff
@@ -0,0 +1,232 @@
...
+                            </Card>
+                        </div>
+                    )}
+
+                    {/* Transactions List - Only shown if user has transactions */}
+                    {!!hasTransactions && (
+                        <div className="space-y-2">
+                            <div className="flex items-center justify-between">
+                                <h2 className="text-lg font-bold">Transactions</h2>
+                                {/* todo: fix href */}
```
jjramirezn: question: what is there to fix?

jjramirezn (changes requested):
Comments

Also this PR is to contacts-ui while contacts-ui has a open PR for peanut-wallet-dev
Also there are merge conflicts

src/components/Global/Icons/user-plus.tsx
kushagrasarathe: answered on discord

src/components/Profile/components/PublicProfile.tsx
kushagrasarathe: addressed in other pr

src/components/Profile/components/PublicProfile.tsx
kushagrasarathe: need to fix the href, it's wrong rn, will do

src/app/[...recipient]/page.tsx
Comment (kushagrasarathe) [reply src/app/[...recipient]/page.tsx]: @jjramirezn  the `getByUsername` method only returns username and accounts of a user from api, not full name and complete details of the user. For now going with username only, but we need to change this in backend

src/components/Profile/components/PublicProfile.tsx
kushagrasarathe: correction, these are anyways not visible as the prop from parent is being passed as zero rn, these are user specific txns and would be done in history project it self. So for now users will only see empty state

jjramirezn (approved):
approved, solve mergeconflicts and merge

jjramirezn (approved):
lgtm

Hugo0 (approved):

PR #872, opened 2025-05-22
fix: pinta payments on polygon

kushagrasarathe:
contributes to TASK-11259

### DIFF
--- a/src/context/kernelClient.context.tsx
+++ b/src/context/kernelClient.context.tsx
@@ -78,7 +78,13 @@ export const createKernelClientForChain = async <C extends Chain>(
         bundlerTransport: http(bundlerUrl),
         pollingInterval: 500,
         userOperation:
-            // only hardcode gas values for arbitrum cuz its using ultra relay as provider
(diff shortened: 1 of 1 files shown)


src/context/kernelClient.context.tsx
```diff
@@ -75,17 +75,21 @@ export const createKernelClientForChain = async <C extends Chain>(
...
-            // better performance: https://docs.zerodev.app/sdk/core-api/sponsor-gas#ultrarelay
-            estimateFeesPerGas: async ({ bundlerClient: _ }) => {
-                return {
-                    maxFeePerGas: BigInt(0),
-                    maxPriorityFeePerGas: BigInt(0),
-                }
-            },
-        },
+        userOperation:
+            // only hardcode gas values for arbitrum cuz its using ultra relay as provider
```
Hugo0: nit: explain here why polygon not.

Also, what latency affect does that have on polygon? why hardcode on arb> comment not clear

Hugo0 (approved):

PR #880, opened 2025-05-29
feat: add transactionType key for charge creation 

kushagrasarathe:
- contributes to TASK-10750

### DIFF
--- a/src/app/(mobile-ui)/history/page.tsx
+++ b/src/app/(mobile-ui)/history/page.tsx
@@ -1,6 +1,7 @@
 'use client'
 
 import { CardPosition } from '@/components/Global/Card'
+import EmptyState from '@/components/Global/EmptyStates/EmptyState'
 import NoDataEmptyState from '@/components/Global/EmptyStates/NoDataEmptyState'
 import NavHeader from '@/components/Global/NavHeader'
 import PeanutLoading from '@/components/Global/PeanutLoading'
@@ -65,7 +66,12 @@ const HistoryPage = () => {
     if (isError) {
         console.error(error)
         Sentry.captureException(error)
-        return <div className="w-full py-4 text-center">Error loading history: {error?.message}</div>
+        return (
+            <div className="mx-auto mt-6 w-full space-y-3 md:max-w-2xl">
+                <h2 className="text-base font-bold">Transactions</h2>{' '}
+                <EmptyState icon="alert" title="Error loading transactions!" description="Please try again later" />
+            </div>
+        )
     }
 
     if (allEntries.length === 0) {
--- a/src/app/(mobile-ui)/home/page.tsx
+++ b/src/app/(mobile-ui)/home/page.tsx
@@ -16,10 +16,12 @@ import { useAuth } from '@/context/authContext'
 import { useWallet } from '@/hooks/wallet/useWallet'
 import { useUserStore, useWalletStore } from '@/redux/hooks'
 import { formatExtendedNumber, getUserPreferences, printableUsdc, updateUserPreferences } from '@/utils'
+import { useDisconnect } from '@reown/appkit/react'
 import Image from 'next/image'
 import Link from 'next/link'
 import { useEffect, useMemo, useState } from 'react'
 import { twMerge } from 'tailwind-merge'
+import { useAccount } from 'wagmi'
 
 export default function Home() {
     const { balance, address, isFetchingBalance, isFetchingRewardBalance } = useWallet()
@@ -29,6 +31,8 @@ export default function Home() {
         const prefs = getUserPreferences()
         return prefs?.balanceHidden ?? false
     })
+    const { isConnected: isWagmiConnected } = useAccount()
+    const { disconnect: disconnectWagmi } = useDisconnect()
 
     const { isFetchingUser, addAccount } = useAuth()
     const { user } = useUserStore()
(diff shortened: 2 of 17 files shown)


src/app/(mobile-ui)/history/page.tsx
```diff
@@ -65,7 +66,12 @@ const HistoryPage = () => {
     if (isError) {
         console.error(error)
         Sentry.captureException(error)
-        return <div className="w-full py-4 text-center">Error loading history: {error?.message}</div>
+        return (
+            <div className="mx-auto mt-6 w-full space-y-3 md:max-w-2xl">
+                <h2 className="text-base font-bold">Transactions</h2>{' '}
+                <EmptyState icon="alert" title="Error loading transactions!" description="Please try again later" />
```
jjramirezn, line 72: praise: much better

src/app/(mobile-ui)/home/page.tsx
```diff
@@ -62,6 +66,13 @@ export default function Home() {
         }
     }, [user, address])
 
+    // always reset external wallet connection on home page
```
jjramirezn, line 69: question: what is the reasoning for this

src/components/Home/HomeHistory.tsx
jjramirezn, line 1: praise: better states all around!

src/components/Payment/Views/Confirm.payment.view.tsx
```diff
@@ -153,6 +153,7 @@ export default function ConfirmPaymentView({
             currency,
             currencyAmount,
             isAddMoneyFlow: !!isAddMoneyFlow,
+            transactionType: isAddMoneyFlow ? 'DEPOSIT' : 'REQUEST',
```
jjramirezn, line 156: question: is it always request?

I know this is used in several flows. I would be ok with leaving it undefined if we're not sure. If we are sure then is better to set it

src/components/Payment/PaymentForm/index.tsx
```diff
@@ -376,6 +376,7 @@ export const PaymentForm = ({
             currency,
             currencyAmount,
             isAddMoneyFlow: !!isAddMoneyFlow,
+            transactionType: isAddMoneyFlow ? 'DEPOSIT' : isDirectPay ? 'DIRECT_SEND' : 'REQUEST',
```
jjramirezn, line 379: question: is it always request?

I know this is used in several flows. I would be ok with leaving it undefined if we're not sure. If we are sure then is better to set it

src/components/TransactionDetails/TransactionAvatarBadge.tsx
```diff
@@ -57,9 +60,10 @@ const TransactionAvatarBadge: React.FC<TransactionAvatarBadgeProps> = ({
         case 'add':
             displayIconName = 'wallet-outline'
             displayInitials = undefined
-            calculatedBgColor = context === 'card' ? AVATAR_TEXT_DARK : AVATAR_WALLET_BG
-            iconFillColor = context === 'card' ? AVATAR_TEXT_LIGHT : AVATAR_TEXT_DARK
-            textColor = context === 'card' ? AVATAR_TEXT_LIGHT : AVATAR_TEXT_DARK
+            // todo: revisit wen more deposit types are added
```
jjramirezn, line 63: question: why? and what is the result on the history?

src/hooks/useTransactionHistory.ts
```diff
@@ -158,23 +158,31 @@ export function useTransactionHistory({
                             tokenAddress: entry.tokenAddress as Hash,
                             chainId: entry.chainId,
                         })
-                        usdAmount = formatUnits(BigInt(entry.amount), tokenDetails?.decimals ?? 6)
-                        tokenSymbol = tokenDetails?.symbol ?? ''
+                        usdAmount = entry.amount.toString()
```
jjramirezn: question: is this right? 

Did you test this? amount in sendlink comes as wei amount

jjramirezn (commented):
Please answer comments

src/components/TransactionDetails/TransactionAvatarBadge.tsx
kushagrasarathe: @jjramirezn  for deposit there will be more methods in future, ex, google pay, apple etc etc, and for now we only have wallet withdraw for which icons dont need customisation, so for future reference left a comment

src/components/Payment/PaymentForm/index.tsx
kushagrasarathe: @jjramirezn  in this flow yes, direct pay and deposit are handled separately, rest any other flow is request related

src/app/(mobile-ui)/home/page.tsx
kushagrasarathe: @jjramirezn this is to avoid wallet req payment from external wallet when user is logged in and tries to pay a direct request, check this bug [here](https://www.notion.so/peanutprotocol/Staging-errors-201838117579801b9819d384bda84048?pvs=4#20283811757980fcb4c6fc4c35415e34)

src/components/Payment/Views/Confirm.payment.view.tsx
kushagrasarathe: see reply in other comment

src/hooks/useTransactionHistory.ts
kushagrasarathe: hmm, i was getting errors for deposit and withdraw histories, lemme check for send link

kushagrasarathe: @jjramirezn pushed a fix for this

PR #897, opened 2025-06-05
Feat/request link preview

FacuBozzi:
Social Preview for request links. 

Ex: `peanut.me/bozzitesting/233USDC?chargeId=f087162d-253a-4805-a889-7f7f8c0489d1`
Sending that URL in Whatsapp will load a dynamically generated social preview with the design made by Manu.

Removed `getPreviewUrl` function, I think this was the old generic (and static) svg social preview.

Obviously, you can't see the social preview on the `peanut.me` domain since it isn't deployed, but I tested using `ngrok` to expose the URL and they were generated correctly.

Next step is the send/claim links, but I have quite a few blockers there and I don't want to stay stuck.


src/app/api/og/route.tsx
```diff
@@ -0,0 +1,71 @@
...
+    const [knerdFilled, knerdOutline, montserratMedium, montserratSemibold] = await Promise.all([
+        getFont('knerd-filled.ttf'),
+        getFont('knerd-outline.ttf'),
+        getFont('montserrat-medium.ttf'),
+        getFont('montserrat-semibold.ttf'),
+    ])
+
+    const { searchParams } = new URL(req.url)
+    const type = (searchParams.get('type') ?? 'generic') as 'send' | 'request' | 'generic'
+    const username = searchParams.get('username') ?? 'Peanut'
```
jjramirezn: thought: this would generate "Peanut is requesting/sending you"

Which is weird, in what cases do we expect username to be empty? Do we need to handle it with a generic image or something similar

src/app/api/og/route.tsx
```diff
@@ -0,0 +1,71 @@
+import { ImageResponse } from '@vercel/og'
```
jjramirezn: question: any reason to use @vercel/og instead of @next/og ?

I know in the end next/og uses vercel/og but we already import it and we're using next so if it works would make sense to use it

jjramirezn (commented):
Congratz on first PR. Very good PR, clear and understandable

Check my comments and answer/do any change if needed

Also check coderabbit comments, what we do with those (because it is a 50/50 with useful catches vs nonsense) is go one by one and answering shortly and resolving if no action needed or modifying if they found something important. Use your own sense its AI I'll make mistakes some times

src/app/api/og/route.tsx
FacuBozzi: Makes sense. next/og is best for this case. On it.

src/app/api/og/route.tsx
FacuBozzi: removing fallback to Peanut since url will never not-have a username. If it does, the link would be broken and the page will 404

package.json
```diff
@@ -43,6 +43,7 @@
         "@tanstack/react-query": "5.8.4",
         "@typeform/embed-react": "^3.20.0",
         "@vercel/analytics": "^1.4.1",
+        "@vercel/og": "^0.6.8",
```
jjramirezn: chore: remove this import

we wont use @vercel/og

jjramirezn (approved):
Approved, please do the chore comment and don't merge until we do today's release (this is staged for next one)

package.json
FacuBozzi: yeah, forgot to uninstall the package. Done

src/app/[...recipient]/layout.tsx
```diff
@@ -77,12 +74,15 @@ export async function generateMetadata({ params }: any) {
             icon: '/logo-favicon.png',
         },
         openGraph: {
-            images: [{ url: previewUrl }],
+            title,
+            description: 'Seamless payment infrastructure for sending and receiving digital assets.',
```
Comment (Hugo0) [inline src/app/[...recipient]/layout.tsx:78]: Nit: this copy sucks, where is this from?  @FacuBozzi 

image

PR #900, opened 2025-06-09
feat(claim): claim link preview + add getNoPubKey api

FacuBozzi:
Last Social Preview PR. This time for claim links.

Ex: peanut.me/claim?c=42161&v=v4.3&i=4123#p=08ElxE4TpCKSj1ym
Sending that URL in Whatsapp will load a dynamically generated social preview with the design made by Manu.

Tested using ngrok to expose the URL and they were generated correctly.

- Add new getNoPubKey method to sendLinks service
- Improve claim page metadata generation with parallel requests
- Update OG image generation to use new endpoint and parameters
- Fix React key warning in WalletNavigation component

### DIFF
--- a/src/services/sendLinks.ts
+++ b/src/services/sendLinks.ts
@@ -175,6 +175,7 @@ export const sendLinksApi = {
         return data
     },
 
+    // like the get function, but no public key needed to get TX details
     getNoPubKey: async (link: string): Promise<SendLink> => {
         const params = getParamsFromLink(link)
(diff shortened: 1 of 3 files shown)


src/services/sendLinks.ts
```diff
@@ -175,6 +175,19 @@ export const sendLinksApi = {
         return data
     },
 
+    getNoPubKey: async (link: string): Promise<SendLink> => {
```
Hugo0, line 179: nit: add comment on why this fn is needed, if its not mega obvious

FacuBozzi: on it

PR #901, opened 2025-06-09
fix(withdraw): remove Peanut username support from recipient input

FacuBozzi:
Update recipient input validation to reject Peanut usernames for withdrawals and adjust placeholder text accordingly. This change ensures users can only enter valid addresses (banks) or ENS names when withdrawing funds.


src/components/Global/GeneralRecipientInput/index.tsx
```diff
@@ -58,6 +58,14 @@ const GeneralRecipientInput = ({
             } else {
                 try {
                     const validation = await validateAndResolveRecipient(trimmedInput)
+
+                    // Only accept ENS accounts, reject usernames
+                    if (validation.recipientType.toLowerCase() === 'username') {
```
kushagrasarathe: @FacuBozzi wrong approach, recipient input is used in other parts of the app, and we want usernames to be enabled on those other flows, example direct request, you can try requesting funds in guest flow using `/request/<username>` and enter where you want to request funds. So disabling it directly at root of the component is a bad choice. 

Also the error msg imo seems too long, was this taken from the mockups?
 
Screenshot 2025-06-10 at 12 00 40 PM

kushagrasarathe (changes requested):
@FacuBozzi blocking, please check comment

Screenshot 2025-06-10 at 12 01 21 PM

src/components/Global/GeneralRecipientInput/index.tsx
FacuBozzi: you're totally right. I started the approach by adding a prop to this specific input/component. Don't know why I changed it. On it.

FacuBozzi:
@kushagrasarathe
Fixed peanut username validation input in Withdrawals by adding prop level validation

Regarding pfp:

> check line 84, 85 in InitialWithdrawView, and the conditions in PeanutActionDetailsCard from 84 to 101 line


fixed it this way. Initially it worked by just adding the condition `transactionType === 'WITHDRAW’,` but since you mentioned also checking against `recipientType` being the username, I also added that in the validation since it also works and maybe prevents something from breaking that I’m not aware of

src/utils/tailwind-colors.ts
kushagrasarathe, line 1: blocking: why introduce a new utils file? we already have colors defined in `tailwind.config.js`

src/components/Global/GeneralRecipientInput/index.tsx
```diff
@@ -58,6 +60,13 @@ const GeneralRecipientInput = ({
             } else {
                 try {
                     const validation = await validateAndResolveRecipient(trimmedInput)
+
+                    // Only accept ENS accounts, reject usernames
+                    if (isWithdrawal && validation.recipientType.toLowerCase() === 'username') {
```
kushagrasarathe, line 65: nice

src/components/Profile/AvatarWithBadge.tsx
kushagrasarathe, line 1: blocking: changes done in this file are un-necessary, its affecting the transaction card colors, without these changes, it works fine without these changes

image

kushagrasarathe (changes requested):
please check comments, blocking

kushagrasarathe (approved):
noice, lgtm

PR #908, opened 2025-06-12
Bugfix/receipt timestamp

FacuBozzi:
When a user created a request, the transaction was timestamped with the date and time of the request creation. However, when the payer completed the payment, the receipt showed the original request timestamp instead of the actual date and time of the payment.

This PR updates the receipt to display the actual date and time when the payment is made, not when the request was created. Both on normal receipts and in the box drawer receipt when the payment is accepted.


jjramirezn (approved):
Looking good

PR #913, opened 2025-06-16
fixed receipt social preview being the same as request preview

FacuBozzi:
Social preview was the same for a request link than for a receipt link with transaction details. This PR fixes that logic and renders a preview for the receipt. Right now this new social preview for the receipt is kind of ugly but i will make a ticket to make it prettier (like Manu said) for next sprint. Right now its functional and I have more important tickets i believe.

Next sprint's ticket will also have instructions to update the claim social preview to be the same as the request social preview (but with 'claiming' instead of 'requesting', obviously)

### DIFF
--- a/src/app/[...recipient]/layout.tsx
+++ b/src/app/[...recipient]/layout.tsx
@@ -52,7 +52,13 @@ export async function generateMetadata({ params }: any) {
     const ogUrl = new URL(`${siteUrl}/api/og`)
     ogUrl.searchParams.set('type', 'request')
     ogUrl.searchParams.set('username', recipient)
-    ogUrl.searchParams.set('amount', String(amount))
+    if (amount) {
+        // if its amount, make request/claim social preview, otherwise make receipt social preview
+        ogUrl.searchParams.set('amount', String(amount))
+    } else {
+        //receipt social preview
+        ogUrl.searchParams.set('isReceipt', 'true')
+    }
 
     if (!siteUrl) {
         console.error('Error: Unable to determine site origin')
--- a/src/components/og/ReceiptCardOG.tsx
+++ b/src/components/og/ReceiptCardOG.tsx
@@ -1,0 +1,120 @@
+import { PaymentLink } from '@/interfaces'
+
+function usernamePxWidth(name: string) {
+    const charPx = 0.6 * 80 // ≈48 px per glyph
+    return Math.round(name.length * charPx) + 40 // +40 padding
+}
+
+export function ReceiptCardOG({
+    link,
+    iconSrc,
+    logoSrc,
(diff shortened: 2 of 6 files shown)


src/app/[...recipient]/layout.tsx
```diff
@@ -52,7 +52,13 @@ export async function generateMetadata({ params }: any) {
     const ogUrl = new URL(`${siteUrl}/api/og`)
     ogUrl.searchParams.set('type', 'request')
     ogUrl.searchParams.set('username', recipient)
-    ogUrl.searchParams.set('amount', String(amount))
+    if (amount) {
```
Comment (kushagrasarathe) [inline src/app/[...recipient]/layout.tsx:55]: noice

src/components/og/ReceiptCardOG.tsx
```diff
@@ -0,0 +1,120 @@
+import { PaymentLink } from '@/interfaces'
+
+function usernamePxWidth(name: string) {
```
kushagrasarathe, line 3: qn (blocking): why is this needed? Usernames already have a limit of 12 characters, and for addresses, we can use the printableAdrress utility fn we have, you can search for it in the utility file

kushagrasarathe (changes requested):
@FacuBozzi left a blocking comment, also please address codderrabbit comments if applicable

src/components/og/ReceiptCardOG.tsx
FacuBozzi: sorry, this is an implementation of a fix coderabbit suggested and i just now saw your comment about coderabbit. I was implementing the coderabbit diff for avoiding username overflow. I was unaware of the 12 character limit and the printableAddress function. Should I roll this commit back?

kushagrasarathe: @FacuBozzi yes if that works without issues, no worries about the limit, you havent had exposure to thatbpart of code ig so that's totally fine.

kushagrasarathe: @FacuBozzi  sorry i meant to remove the new fn `usernamePxWidth`, sorry for misunderstanding, should have been more clear. Lets remove the  new fn for usernames, and use the `printableAddress` for address types, rest i think the 12 characters for a username will easily fit in the card

kushagrasarathe (changes requested):
see commment, sorry if it was not clear earlier :*(

src/components/og/ReceiptCardOG.tsx
kushagrasarathe: discussed on discord, all good, no further action item

kushagrasarathe (approved):
LFG 🚀

PR #920, opened 2025-06-18
logically fixed the error, but unable to test it

FacuBozzi:
Fxing the [squid TX](https://www.notion.so/peanutprotocol/Review-Bug-squid-tx-appears-when-depositing-or-fulfilling-a-request-21683811757980c4a6a6e87ce59f886a) issue. Logically I think it should work, but sadly I can't test it for technical reasons that we should address. Please test it Jota and let me know if it works.


jjramirezn (approved):
Short and sweet. Also tested

PR #923, opened 2025-06-19
fixed 2 transactions when making a request link with comment/attachment

FacuBozzi:
this PR solves [this issue](https://www.notion.so/peanutprotocol/Bug-request-via-link-with-comment-and-attachment-is-creating-2-txs-in-the-history-206838117579806a92edd1fcb7068301)

The original ticket says that the problem is related to websockets but it's not. The websocket issue was solved in [this PR](https://github.com/peanutprotocol/peanut-ui/pull/920). 

This issue was due to the fact the Link payments were generated whenever the comment/attachment input status was updated, so if a user adds a comment, then adds another word to the comment, then attachs a document, 3 payment links would be generated.

This PR makes it so that payment link is generated only when a user clicks the "Generate Link" button, so only 1 payment link is generated and picked up by the websocket.

TODO: when the comment field is updated, the generate link still makes the loading animation even if it doesn't make the payment link, this doesn't affect functionality but it would be nicer if only made the loading animation when the user clicks the generate link button and the payment actually generates. Making a ticket for next sprint for this.

### DIFF
--- a/src/components/Request/link/views/Create.request.link.view.tsx
+++ b/src/components/Request/link/views/Create.request.link.view.tsx
@@ -215,7 +215,8 @@ export const CreateRequestLinkView = () => {
 
     const generateLink = useCallback(async () => {
         if (generatedLink) return generatedLink
-        if (Number(tokenValue) === 0) return qrCodeLink
+        if (Number(tokenValue) === 0 && !hasAttachment) return qrCodeLink
+
         setIsCreatingLink(true)
         const link = await createRequestLink({
             recipientAddress,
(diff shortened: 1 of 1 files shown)


jjramirezn (commented):
This is not ideal, because the idea is that the users can share the request inmediatly by showing the QR code. Not sure what the best solution would be.

One thing that should not happen is that the link is regenerated when clicking sharing. Don't know what the current behavior is

jjramirezn:
I think the solution here would be that after the creation of the request if we modify the comment or the attachment we update the same request instead of creating a new one

jjramirezn (changes requested):
See last comment

PR #929, opened 2025-06-25
[TASK-12442] feat: Details on onramp deposits

jjramirezn:


src/components/Global/Icons/chevron-down.tsx
kushagrasarathe, line 1: thought: we you can use `chevron-up` icon and rotate it 90 degree, no need to add a new icon imo

src/components/TransactionDetails/TransactionDetailsDrawer.tsx
```diff
@@ -285,6 +289,261 @@ export const TransactionDetailsReceipt = ({
                         />
                     )}
 
+                    {/* Onramp deposit instructions for bridge_onramp transactions */}
```
kushagrasarathe, line 292: thought: with different categories of `TransactionDetailsReceipt`, i think it would be better to further abstract rows and render them, ex, create a separate component for onramp details in drawer

src/components/Global/Icons/chevron-down.tsx
jjramirezn: Lol you're right

src/components/TransactionDetails/TransactionDetailsDrawer.tsx
```diff
@@ -285,6 +289,261 @@ export const TransactionDetailsReceipt = ({
...
+                    {transaction.direction === 'bank_deposit' &&
+                        transaction.status === 'pending' &&
+                        transaction.extraDataForDrawer?.depositInstructions &&
+                        transaction.extraDataForDrawer.depositInstructions.bank_name && (
+                            <>
+                                <PaymentInfoRow
+                                    label={
+                                        <div className="flex items-center gap-1">
+                                            <span>Deposit Message</span>
+                                            <MoreInfo text="Make sure you enter this exact message as the transfer concept or description. If it's not included, the deposit can't be processed." />
```
kushagrasarathe, line 302: thought, we want this message to be clearly visible to users, idk how many users will actually read this by opening a tooltip (haven't gone through designs tho)

kushagrasarathe (changes requested):
trying to add money, redirects to withdraw flow

https://github.com/user-attachments/assets/737b7d30-5a5e-41de-aedb-c7001d07493b

kushagrasarathe (approved):
aah nvm, withdraw will go in diff pr, rest lgtm, left some comments, up to you to itemize or leave as it is

src/components/TransactionDetails/TransactionDetailsDrawer.tsx
jjramirezn: Yeah, its in the designs. In this case is part of the details. But in the main flow, there is a pop up before showing any transaction data

PR #943, opened 2025-07-01
[TASK-11884] request via link with comment and attachment is creating 2 txs in the history

FacuBozzi:
Fix for [this issue](https://www.notion.so/peanutprotocol/Bug-request-via-link-with-comment-and-attachment-is-creating-2-txs-in-the-history-206838117579806a92edd1fcb7068301). This PR implements a PATCH request implementation (i also opened a backend PR for this) that lets the user update the request link's details when editing the comments/files instead of creating a new one (previously a POST request) for every modification. 

Basically:
Creation -> POST
Every update/modification after that -> PATCH

### IMPORTANT
The problem is that when creating a request payment link, the amount is shared also in the url (ex: /jota/10usdc), so updating the request link after its creation would have a discrepancy between the actual amount and the amount in the URL.

For now, the solution is that the user enters an amount and once the request link is created, the amount input gets disabled and they can only update the comments/file. I think this solution makes sense actually.


src/app/api/proxy/withFormData/requests/[id]/route.ts
```diff
@@ -0,0 +1,28 @@
+import { PEANUT_API_URL } from '@/constants'
```
Comment (kushagrasarathe) [inline src/app/api/proxy/withFormData/requests/[id]/route.ts]: @FacuBozzi  qn (blocking), create new route? we use either server action or service, there are already service methods created for charges and requests

Comment (jjramirezn) [reply src/app/api/proxy/withFormData/requests/[id]/route.ts]: I believe this is for supporting attachments, check how we do the create one. Don't know if that's legacy or needed. I thing is legacy, did you try to call directly from the service @FacuBozzi ?

src/components/Request/link/views/Create.request.link.view.tsx
```diff
@@ -336,6 +404,9 @@ export const CreateRequestLinkView = () => {
         }
     }, [_tokenValue])
 
+    // handle updates to existing requests when debounced attachment options change
```
jjramirezn: thought: this comment seems unnecessary

tailwind.config.js
```diff
@@ -544,4 +544,4 @@ module.exports = {
             filter: ['hover'],
         },
     },
-}
+};
```
jjramirezn: nit: just let it as it was (we dont use semi-colon in the codebase)

jjramirezn (commented):
Generally looks good, main thing to address is kushagra's comment. I also added  a bit more context

src/app/api/proxy/withFormData/[...slug]/route.ts
```diff
@@ -14,11 +14,18 @@ export async function POST(request: NextRequest) {
     formData.forEach((value, key) => {
         apiFormData.append(key, value)
     })
+
+    const apiKey = process.env.PEANUT_API_KEY
+    if (!apiKey) {
```
Comment (jjramirezn) [inline src/app/api/proxy/withFormData/[...slug]/route.ts:19]: praise: good!!!

src/components/Request/link/views/Create.request.link.view.tsx
```diff
@@ -209,11 +236,47 @@ export const CreateRequestLinkView = () => {
...
+
+                // Wait for the update to complete
+                while (needsUpdate || isUpdatingOnBlur) {
+                    await new Promise((resolve) => setTimeout(resolve, 50))
+                }
+            }
+
+            // Return the existing link
+            return generatedLink || ''
+        }
```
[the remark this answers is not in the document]
jjramirezn: @FacuBozzi I aggree with coderabbit here, why the while? wouldnt using react callbacks or effect be better here?

jjramirezn (changes requested):
Check comment from coderabbit

jjramirezn (dismissed):

src/hooks/useDebounce.ts
kushagrasarathe, line 1: ohh god, thank you, this was very much needed, @FacuBozzi  can you also expand the scope to use this in the signup input too? and anywhere else where we need debounce? also please make sure to test the flows properly just to make sure nothing breaks

jjramirezn (approved):

src/components/Request/link/views/Create.request.link.view.tsx
```diff
@@ -159,272 +175,220 @@ export const CreateRequestLinkView = () => {
...
-    ])
-
-    // check for token value debouncing
-    const isDebouncing =
-        (hasAttachment &&
-            attachmentOptions?.message &&
-            (!debouncedAttachmentOptions?.message ||
-                debouncedAttachmentOptions.message !== attachmentOptions.message)) ||
-        (hasAttachment && _tokenValue !== debouncedTokenValue)
+    }, [isConnected, address, setSelectedChainID, setSelectedTokenAddress])
```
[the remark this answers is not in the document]
FacuBozzi: no

PR #948, opened 2025-07-03
feat: add alachemy as fallback rpc url

kushagrasarathe:
fixes TASK-7669 : add alchemy as rpc fallback provider


src/constants/general.consts.ts
```diff
@@ -1,25 +1,63 @@
...
+        `https://base-sepolia.infura.io/v3/${INFURA_API_KEY}`,
+        `https://base-sepolia.g.alchemy.com/v2/${ALCHEMY_API_KEY}`,
+    ].filter(
+        (url) => (INFURA_API_KEY && url.endsWith(INFURA_API_KEY)) || (ALCHEMY_API_KEY && url.endsWith(ALCHEMY_API_KEY))
+    ),
     // Infura is returning weird estimations for BSC @2025-05-14
     //[bsc.id]: `https://bsc-mainnet.infura.io/v3/${INFURA_API_KEY}`,
-    [bsc.id]: 'https://bsc-dataseed.bnbchain.org',
-    [scroll.id]: `https://scroll-mainnet.infura.io/v3/${INFURA_API_KEY}`,
+    [bsc.id]: ['https://bsc-dataseed.bnbchain.org'],
```
Hugo0: no infura/alchemy on bnb?

src/utils/general.utils.ts
```diff
@@ -956,9 +957,14 @@ export async function fetchTokenSymbol(tokenAddress: string, chainId: string): P
...
         try {
+            const provider = await getSDKProvider({ chainId })
+            if (!provider) {
+                console.error(`Failed to get provider for chain ID ${chainId}`)
+                return undefined
+            }
             const contract = await peanut.getTokenContractDetails({
                 address: tokenAddress,
-                provider: await peanut.getDefaultProvider(chainId),
+                provider: provider,
```
Hugo0, line 967: Q, blocking: why are we using getDefaultProvider here (these are all public) instead of the infura/alchemy ones?

src/utils/sdk.utils.ts
```diff
@@ -0,0 +1,40 @@
...
+            Sentry.captureException(error)
+            console.error(`Failed to get default provider for chain ID ${chainId}`, error)
+            return undefined
+        }
+    }
+
+    for (const url of urls) {
+        try {
+            const provider = new providers.JsonRpcProvider(url)
+            await provider.getNetwork() // check if the provider is valid
```
Hugo0: Q, non-blocking: What does this do? is it actually useful? Does it introduce any latency? is getNetwork an http round trip?

Performance issue, blocking: we have many places where we use RPCs. How long does getSDKProvider run? Is it an apicall round trip? If so, can we cache it so it only takes time on the first instantiation?

Hugo0: Issue, blocking: 

1. I don't think provider.getNetwork() actually validates that the provider is working
2. this only seems useful during instantiation of provider. If provider is instantiated but fails _later_, this doesn't do much. It doesn't doesn't really act as a JIT fallback

Hugo0 (changes requested):

src/constants/general.consts.ts
jjramirezn: I did this change, because infura was not returning correct estimates, was very weird and was solved by using another provider. Don't know about alchemy, but probably with this task is a good time to check infura again and add alchemy

src/utils/sdk.utils.ts
kushagrasarathe: - its to check the live-ness, this would help incase provider obj is created but fails on the first use, this would help with debuging, and for latency, yes, it adds lil bit

- second one valid points, i'll update this, missed these initally

src/constants/general.consts.ts
kushagrasarathe: @jjramirezn  checked, infura still being weird with bnb

src/utils/sdk.utils.ts
kushagrasarathe: addreseed in latest commit

src/utils/general.utils.ts
kushagrasarathe: @Hugo0  check discord thread

Hugo0 (dismissed):

jjramirezn (approved):

PR #949, opened 2025-07-04
[TASK-12678] fix: coral withdraw fixes

jjramirezn:
- Stop refreshing the page after successful payment
- Better error message for unsupported token pairs

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -26,6 +26,10 @@ export WC_PROJECT_ID=""
 export GA_KEY=""
 export SOCKET_API_KEY=""
 
+# SQUID
+export SQUID_API_URL=""
+export SQUID_INTEGRATOR_ID=""
+
 # Passkey envs
 export NEXT_PUBLIC_ZERO_DEV_PASSKEY_PROJECT_ID=""
 export NEXT_PUBLIC_ZERO_DEV_BUNDLER_URL=""
--- a/src/app/(mobile-ui)/withdraw/crypto/page.tsx
+++ b/src/app/(mobile-ui)/withdraw/crypto/page.tsx
@@ -75,10 +75,9 @@ export default function WithdrawCryptoPage() {
             router.push('/withdraw')
             return
         }
-        dispatch(paymentActions.setChargeDetails(null))
         clearErrors()
-        setCurrentView('INITIAL')
-    }, [amountToWithdraw, router, dispatch, setAmountToWithdraw, setCurrentView])
+        dispatch(paymentActions.setChargeDetails(null))
+    }, [amountToWithdraw])
 
     useEffect(() => {
         setPaymentError(paymentErrorFromHook)
@@ -205,19 +204,6 @@ export default function WithdrawCryptoPage() {
 
         if (result.success && result.txHash) {
             setCurrentView('STATUS')
-
-            // reset the entire withdraw flow after successful payment
-            setTimeout(() => {
-                setAmountToWithdraw('')
-                setWithdrawData(null)
-                setCurrentView('INITIAL')
-
-                // clear any errors
-                clearErrors()
-
-                // clear charge details
-                dispatch(paymentActions.setChargeDetails(null))
-            }, 3000) // wait 3 seconds to show success status before resetting
         } else {
             console.error('Withdrawal execution failed:', result.error)
             const errMsg = result.error || 'Withdrawal processing failed.'
--- a/src/app/(mobile-ui)/withdraw/crypto/page.tsx
+++ b/src/app/(mobile-ui)/withdraw/crypto/page.tsx
@@ -25,6 +25,7 @@ import { interfaces as peanutInterfaces } from '@squirrel-labs/peanut-sdk'
 import { PEANUT_WALLET_CHAIN, PEANUT_WALLET_TOKEN } from '@/constants'
 import { useRouter } from 'next/navigation'
 import { useCallback, useEffect, useMemo } from 'react'
+import { captureMessage } from '@sentry/nextjs'
 
 export default function WithdrawCryptoPage() {
     const router = useRouter()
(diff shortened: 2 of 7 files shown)


src/constants/general.consts.ts
```diff
@@ -6,7 +6,7 @@ export const peanutWalletIsInPreview = true
 
 export const INFURA_API_KEY = process.env.NEXT_PUBLIC_INFURA_API_KEY
 
-export const SQUID_INTEGRATOR_ID = '11CBA45B-5EE9-4331-B146-48CCD7ED4C7C'
+export const SQUID_INTEGRATOR_ID = process.env.SQUID_INTEGRATOR_ID!
```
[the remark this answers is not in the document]
jjramirezn: you are right, adding

.env.example
```diff
@@ -26,6 +26,10 @@ export WC_PROJECT_ID=""
 export GA_KEY=""
 export SOCKET_API_KEY=""
 
+# SQUID
+export SQUID_API_URL=""
+export SQUID_INTEGRATOR_ID=""
```
Hugo0, line 31: Q: integrator id in FE? Shouldn't squid route be fetched from backend?

Slightly more latency but better architecture imo

src/app/(mobile-ui)/withdraw/crypto/page.tsx
```diff
@@ -75,10 +75,9 @@ export default function WithdrawCryptoPage() {
             router.push('/withdraw')
             return
         }
-        dispatch(paymentActions.setChargeDetails(null))
         clearErrors()
-        setCurrentView('INITIAL')
-    }, [amountToWithdraw, router, dispatch, setAmountToWithdraw, setCurrentView])
+        dispatch(paymentActions.setChargeDetails(null))
```
Hugo0, line 80: huh! lg

Hugo0: any idea why we were resettng page before? doesnt make sense to me

src/app/(mobile-ui)/withdraw/page.tsx
```diff
@@ -93,6 +94,12 @@ export default function WithdrawPage() {
         [setRawTokenAmount]
     )
 
+    // Clean state
+    useEffect(() => {
+        setAmountToWithdraw('')
+        setWithdrawData(null)
+    }, [])
+
```
Hugo0, line 102: good!

Thought: we have a **lot** of places where we don't reset state properly. It is a common source of bugs. Any idea on how we can do this better & more foolproof?

src/components/Payment/Views/Confirm.payment.view.tsx
```diff
@@ -344,7 +344,7 @@ export default function ConfirmPaymentView({
 
         // For peanut wallet flows, only RFQ routes are allowed
         if (xChainRoute.type === 'swap') {
-            return 'This route requires external wallet payment. Peanut Wallet only supports RFQ (Request for Quote) routes.'
+            return 'This token pair is not supported'
```
Hugo0: > route requires external wallet payment. Peanut Wallet only supports RFQ (Request for Quote) routes.
this is such a technical error. I think the new one is better.

Thought, potentially blocking: What happens if rfq is down for e.g. polygon usdt? Lets say its only temporarily down. I think the message should be different there - ie it should indicate that it being down is only temporary (try again later...)

src/constants/general.consts.ts
Hugo0, line 9: thank god

src/hooks/usePaymentInitiator.ts
```diff
@@ -250,12 +250,14 @@ export const usePaymentInitiator = () => {
...
                     setIsCalculatingFees(true)
                     setEstimatedGasCost(
-                        await estimateTransactionCostUsd(
-                            tx.unsignedTx.from! as Address,
-                            tx.unsignedTx.to! as Address,
-                            tx.unsignedTx.data! as Hex,
-                            selectedChainID
-                        )
+                        isPeanutWallet
+                            ? 0
```
Hugo0, line 254: neat

Hugo0 (dismissed):

src/components/Payment/Views/Confirm.payment.view.tsx
jjramirezn: We have no way of differentiating one case from the other and don't know how much later is later. But I'm open to copy suggestions

src/app/(mobile-ui)/withdraw/page.tsx
jjramirezn: Ideally, state is tied to the flow, this is needed because the context is for all the app. Maybe the withdraw flow context is ok to only declare in the withdraw page.

src/app/(mobile-ui)/withdraw/crypto/page.tsx
jjramirezn: Introduced trying to fix state issues: https://www.notion.so/peanutprotocol/Bug-Status-is-not-being-refreshed-after-completing-a-withdraw-2068381175798056af35ed5e6bfc798d

.env.example
jjramirezn: Thats a good question, I'm actually ok having this logic in next backend instead of backend-backend (yay modern web development!) Unless we need to xchain route from the backend or need something with DB in that case I'm ok with moving it to BE

Hugo0: I missed it was next BE - that's all good!

In future we will want to have our own orchestration module - we'll have to wrap squid & other providers to optimize speed/bps. Food for thought

src/app/(mobile-ui)/withdraw/crypto/page.tsx
Hugo0: nice, thanks for linking. I commented on notion

src/components/Payment/Views/Confirm.payment.view.tsx
Hugo0: Ok - short term i recommend we change to

'No route found for this token pair. You can try with a different token pair, or try again later'

if it fits in the screen. Also, would be good to have metrics of this; can we throw a sentry warn or smth?

jjramirezn: Done and done

Hugo0 (dismissed):

Hugo0 (approved):

PR #950, opened 2025-07-04
fix: centre align views on mobile

kushagrasarathe:
fixes TASK-11922 : centre align views


Hugo0 (approved):
I see no blockers. 

Nit: something seems bad if we have to individually style with h-full min-h-[inherit] in 9 places. Can/should we abstract away?

PR #951, opened 2025-07-04
[TASK-12653] removed white spaces before processing to BIC

FacuBozzi:
couldnt reproduce Hugo's error to be honest but maybe it was due to the fact that the IBAN's blank spaces were not removed before processing (like the package's example, so this should be merged since its a proper implementation of the library)

works good for me (tried multiple countries) but maybe this fixes it for Hugo. Let's do QA and see if its solved, otherwise Hugo and Konrad will have to provide more details on how to reproduce


Hugo0 (commented):
Does this work w/ my linked ibans? See discord thread

Hugo0 (approved):

PR #952, opened 2025-07-04
[TASK-12542] added formatIban function for correct display

FacuBozzi:
quick fix :)


Hugo0 (changes requested):
change lgtm. However, flagging, i dont think this is the only place in the code we show IBAN. Please review if this fix needs to be applied to other places before merging.

src/components/TransactionDetails/TransactionDetailsDrawer.tsx
```diff
@@ -10,6 +10,7 @@ import { useUserStore } from '@/redux/hooks'
 import { chargesApi } from '@/services/charges'
 import { sendLinksApi } from '@/services/sendLinks'
 import { formatAmount, formatDate, getInitialsFromName } from '@/utils'
+import { formatIban } from '@/utils/general.utils'
```
Hugo0, line 13: nit: not in scope of PR but i took a look at formatIban and i think its insufficient / quite basic

Hugo0 (approved):
lgtm

PR #955, opened 2025-07-05
[TASK-12682] fix: show token selector for guest flow

jjramirezn:


src/components/Payment/PaymentForm/index.tsx
```diff
@@ -635,7 +635,13 @@ export const PaymentForm = ({
                     currency={currency}
                 />
 
-                {!chain && isConnected && !isAddMoneyFlow && (
+                {/*
+                    Url request flow (peanut.me/<address>)
```
Hugo0, line 639: nit: Guest* request flow

All our requests are URLs

Praise: good comment!

src/components/Payment/PaymentForm/index.tsx
```diff
@@ -635,7 +635,13 @@ export const PaymentForm = ({
...
+                    If we are paying from peanut wallet we only need to
+                    select a token if it's not included in the url
+                    From other wallets we always need to select a token
+                */}
+                {!(chain && isPeanutWallet) && isConnected && !isAddMoneyFlow && (
```
Hugo0, line 644: thought: again, im worried about this code pattern. Very hard to reason about

("so... NOT chain & peanut wallet, AND it has to be connected, AND it is not the addmoney flow?")

Hugo0 (approved):
not smoke tested. Logic looks fine if a bit confusing

src/components/Payment/PaymentForm/index.tsx
jjramirezn: Not only guests use this flow, also peanut users.

jjramirezn: Semantic requests is the descriptor we have

src/components/Payment/PaymentForm/index.tsx
jjramirezn: Yeah, that's why I thought it needed a comment

PR #961, opened 2025-07-08
[TASK-12855] fix: gas estimation for external wallets

jjramirezn:
We were re-rendering unnecesarily the prepare transaction, also now we catch error if any

Also deleted bunch of code, that is not used and was deleted before (maybe during a merge it found it way back to life?)

### DIFF
--- a/src/app/actions/clients.ts
+++ b/src/app/actions/clients.ts
@@ -1,21 +1,12 @@
 'use server'
-import { unstable_cache } from 'next/cache'
-import type { PublicClient, Chain, Hash } from 'viem'
-import { createPublicClient, http, extractChain, maxUint256, keccak256, encodeAbiParameters, numberToHex } from 'viem'
+import type { PublicClient, Chain } from 'viem'
+import { createPublicClient, http, extractChain } from 'viem'
 import * as chains from 'viem/chains'
-import { jsonStringify } from '@/utils'
 
 import { PUBLIC_CLIENTS_BY_CHAIN, infuraRpcUrls } from '@/constants'
 
 const allChains = Object.values(chains)
 export type ChainId = (typeof allChains)[number]['id']
-
-export type FeeOptions = {
-    gas: bigint
-    maxFeePerGas: bigint
-    maxPriorityFeePerGas: bigint
-    error?: string | null
-}
 
 export const getPublicClient = async (chainId: ChainId): Promise<PublicClient> => {
     let client: PublicClient | undefined = PUBLIC_CLIENTS_BY_CHAIN[chainId]?.client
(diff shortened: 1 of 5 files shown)


src/hooks/usePaymentInitiator.ts
```diff
@@ -160,22 +158,19 @@ export const usePaymentInitiator = () => {
...
-            const actualFromChainId = fromChainId ?? selectedChainID
-
-            if (
-                !selectedTokenData ||
-                (!peanutWalletAddress && !wagmiAddress) ||
-                !actualFromTokenAddress ||
-                !actualFromChainId
-            ) {
+        }) => {
+            if (!peanutWalletAddress && !wagmiAddress) {
```
Hugo0, line 173: nit: wagmiAddress is a bad name imo (talks about the tool, not the concept). ExternalWalletAddress maybe

src/hooks/usePaymentInitiator.ts
```diff
@@ -187,13 +182,12 @@ export const usePaymentInitiator = () => {
...
 
             setEstimatedGasCost(undefined)
-            setFeeOptions([])
 
             setIsPreparingTx(true)
 
             try {
-                const _isXChain = actualFromChainId !== chargeDetails.chainId
-                const _diffTokens = !areEvmAddressesEqual(actualFromTokenAddress, chargeDetails.tokenAddress)
+                const _isXChain = from.chainId !== chargeDetails.chainId
```
Hugo0, line 189: style nit: for readability, I prefer this as:

`                const _isXChain = (from.chainId !== chargeDetails.chainId)`

src/hooks/usePaymentInitiator.ts
```diff
@@ -249,16 +242,21 @@ export const usePaymentInitiator = () => {
...
-                                  tx.unsignedTx.from! as Address,
-                                  tx.unsignedTx.to! as Address,
-                                  tx.unsignedTx.data! as Hex,
-                                  selectedChainID
-                              )
-                    )
+                    let gasCost = 0
+                    if (!isPeanutWallet) {
+                        try {
+                            gasCost = await estimateTransactionCostUsd(
```
Hugo0, line 248: nit: transaction, tx -  better to consolidate to one (tx imo)

src/hooks/usePaymentInitiator.ts
```diff
@@ -249,16 +242,21 @@ export const usePaymentInitiator = () => {
...
-                        isPeanutWallet
-                            ? 0
-                            : await estimateTransactionCostUsd(
-                                  tx.unsignedTx.from! as Address,
-                                  tx.unsignedTx.to! as Address,
-                                  tx.unsignedTx.data! as Hex,
-                                  selectedChainID
-                              )
-                    )
+                    let gasCost = 0
```
Hugo0, line 245: Q: good idea to init as 0? do we even need to init it?

if we init it, 1 gwei gas cost might be better default vallue

src/hooks/usePaymentInitiator.ts
Hugo0, line 248: thought: RPC calls can fail. @kushagrasarathe made a PR to have backup provider (alchemy + infura) - but ideally _every call_ has a fallback - not the provider as a concept

src/hooks/usePaymentInitiator.ts
```diff
@@ -249,16 +242,21 @@ export const usePaymentInitiator = () => {
...
+                        try {
+                            gasCost = await estimateTransactionCostUsd(
+                                tx.unsignedTx.from! as Address,
+                                tx.unsignedTx.to! as Address,
+                                tx.unsignedTx.data! as Hex,
+                                chargeDetails.chainId
+                            )
+                        } catch (error) {
+                            captureException(error)
+                            setIsFeeEstimationError(true)
```
Hugo0, line 256: mininit: english - `setIsFeeEstimationError` sounds a bit awkard.  FeeEstimationError or hasFeeEstimationError

src/app/actions/clients.ts
```diff
@@ -27,75 +18,3 @@ export const getPublicClient = async (chainId: ChainId): Promise<PublicClient> =
...
-
-export type PreparedTx = {
-    account: Hash
-    from?: Hash
-    erc20Token?: Hash
-    data: Hash
-    to: Hash
-    value: string
-}
-export const getFeeOptions = unstable_cache(
```
Hugo0, line 39: Q: i'm a bit conufsed here, this was pretty recent code? Why remove now?

Hugo0 (approved):
tiny nits

Flag: 
- didn't smoke test
- I don't deeply understand this part of the code

src/app/actions/clients.ts
jjramirezn: This is an old one, that we used to have compatibility with the sdk. The new is in src/app/actions/tokens.ts

I swear I removed this when creating the one in tokens.ts but I'm not sure now

src/hooks/usePaymentInitiator.ts
jjramirezn: Maybe the var name is misleading, this is the transaction cost. And is inited as 0 because is 0 for peanutwallet transfer (this calculation is done only to show not to do any transaction)

src/hooks/usePaymentInitiator.ts
jjramirezn: lol gotta love automatic creating react states

PR #962, opened 2025-07-08
[TASK-12867] fixing dots not being enabled in ENS?

FacuBozzi:
Please read Notion ticket description (and the comments in the Notion task)


jjramirezn:
Closed because, we don't actually know the reason for this. I propose we scope investigation for this and a better solution

PR #973, opened 2025-07-10
[TASK-12866] fix: ui comments

jjramirezn:
Comments in task


FacuBozzi (approved):
havent smoke tested but PR looks good lfg

PR #974, opened 2025-07-11
[TASK-12771] Feat/new landing page

FacuBozzi:
new landing page


Hugo0 (dismissed):

index.ts
Hugo0, line 1: why do we need a top level index.ts??

src/components/LandingPage/businessIntegrate.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react'
+import Image from 'next/image'
+import { Star } from '@/assets'
+import scribble from '@/assets/scribble.svg'
+import peanutMeans from '@/assets/illustrations/peanut-means.svg'
+
+// Define the background color as a constant
+const businessBgColor = '#90A8ED'
```
kushagrasarathe, line 8: blocking: @FacuBozzi `90A8ED` is already defined in tailwind config

src/components/LandingPage/marquee.tsx
```diff
@@ -0,0 +1,24 @@
+import { HandThumbsUp } from '@/assets'
```
kushagrasarathe, line 1: qn: i'm not sure why we have a new marquee component when we could directly use the marquee wrapper?

kushagrasarathe (changes requested):

src/components/LandingPage/businessIntegrate.tsx
FacuBozzi: dammit should have just used bg-secondary-1

Hugo0 (changes requested):
Merging out of urgency but this has several needed changes and this doesn't pass standard for code that should be in production

Code Issues:
- 9k LoC json file incorrectly added to diff (should be moved back to where it was)
- new index.ts TOPLEVEL file is incorrect
- Coderabbit comments are correct and should be addressed (e.g. hardcoded imports)
- should not hardcode colors
- there's a lot of repetitive code for anim handling

Design also has various issues as pointed out by manu in [this thread](https://discord.com/channels/972435984954302464/1391740284139737148/1394825373602287726)

PR #976, opened 2025-07-11
[TASK-12312] Bugfix/ens error withdrawing

FacuBozzi:


src/components/Global/GeneralRecipientInput/index.tsx
```diff
@@ -61,17 +61,22 @@ const GeneralRecipientInput = ({
                 try {
                     const validation = await validateAndResolveRecipient(trimmedInput)
 
-                    // Only accept ENS accounts, reject usernames
+                    // For withdrawals, reject usernames and show ENS error
                     if (isWithdrawal && validation.recipientType.toLowerCase() === 'username') {
-                        errorMessage.current = 'Peanut usernames are not supported for withdrawals.'
+                        errorMessage.current = 'ENS name not found'
```
jjramirezn: question: why does the error talks about ens but we validate for type === username ?

FacuBozzi: good one! fixed the logic

jjramirezn (approved):

PR #983, opened 2025-07-15
fix: slider tap bug logic

kushagrasarathe:
fixes TASK-13071


jjramirezn (changes requested):
The user can click anywhere without sliding. Is weird (and if they click on the right it will approve)


https://github.com/user-attachments/assets/91eb1da2-67c8-44ca-be3f-9d5cefbaccc6

jjramirezn (approved):

PR #988, opened 2025-07-16
[TASK-13006] fix: balance warning modal

jjramirezn:
- Slider instead of button
- Expiry for local storage instead of session storage
- Warning icon color


src/utils/general.utils.ts
```diff
@@ -94,12 +94,17 @@ export function jsonParse<T = any>(data: string): T {
     })
 }
 
-export const saveToLocalStorage = (key: string, data: any) => {
+export const saveToLocalStorage = (key: string, data: any, expirySeconds?: number) => {
```
FacuBozzi, line 97: nit: key should have its own type (string enum or string union) so that we have more visibility of what we're storing and accepting (especially as codebase grows and we use localStorage more and more)

FacuBozzi (approved):

src/utils/general.utils.ts
jjramirezn: Good idea

PR #1005, opened 2025-07-22
[TASK-13152] feat: add referral campaign modal

jjramirezn:


src/components/Home/FloatingReferralButton/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+}
+
+const FloatingReferralButton: React.FC<FloatingReferralButtonProps> = ({ onClick }) => {
+    return (
+        <button
+            onClick={onClick}
+            className="absolute left-[43%] top-[15%] z-50 animate-pulse cursor-pointer text-4xl transition-all duration-300 hover:scale-110 hover:animate-none"
+            aria-label="Open referral campaign"
+        >
+            💝
```
Hugo0, line 14: future todo: replace

src/components/Home/FloatingReferralButton/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+    onClick: () => void
+}
+
+const FloatingReferralButton: React.FC<FloatingReferralButtonProps> = ({ onClick }) => {
+    return (
+        <button
+            onClick={onClick}
+            className="absolute left-[43%] top-[15%] z-50 animate-pulse cursor-pointer text-4xl transition-all duration-300 hover:scale-110 hover:animate-none"
+            aria-label="Open referral campaign"
+        >
```
Hugo0, line 13: blocking: left 43% on desktop makes it be in the middle and looks broken:
image

Our users are not desktop first but we can't show them a broken experience; pls responsive here or hide

src/components/Home/ReferralCampaignModal/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+interface ReferralCampaignModalProps {
+    visible: boolean
+    onClose: () => void
+}
+
+const ReferralCampaignModal: React.FC<ReferralCampaignModalProps> = ({ visible, onClose }) => {
+    const router = useRouter()
+
+    const handleInviteFriends = () => {
+        router.push('/send?createLink=true')
```
Hugo0, line 17: comment: i think url navigation is just so much easier than other state management. i love this.

Hugo0: comment: would be great if we had amount in url too. e.g. prefill with $1 or smth.

Hugo0: also comment: back doesnt work properly

home --> referral --> send?createLink=true

now if u click back u dont go back to homme or referral, u go to /send

src/components/Send/views/SendRouter.view.tsx
```diff
@@ -2,21 +2,30 @@
...
         dispatch(sendFlowActions.resetSendFlow())
         setIsSendingByLink(true)
-    }
+    }, [dispatch])
+
+    // Check for createLink parameter on mount
+    useEffect(() => {
+        const createLink = searchParams.get('createLink')
+        if (createLink === 'true') {
+            handleLinkCardClick()
```
Hugo0: i think this is nice.

Ideallly i think `/send/link` would be more elegant and lead to better code probably

Hugo0 (changes requested):

src/components/Send/views/SendRouter.view.tsx
jjramirezn: agree, but was a bitter refactor

src/components/Home/FloatingReferralButton/index.tsx
jjramirezn: hid it

src/components/Home/ReferralCampaignModal/index.tsx
jjramirezn: fixed back

Hugo0 (approved):

PR #1006, opened 2025-07-22
Feat: Landing page changes

Zishan-7:
Contributes to TASK-13226 : Landing Page 1.1


src/app/layout.tsx
```diff
@@ -40,6 +40,11 @@ const knerdFilled = localFont({
     variable: '--font-knerd-filled',
 })
 
+const robotoFlexBold = localFont({
```
kushagrasarathe, line 43: @Zishan-7  qn, one is declared in tailwind config, another here, if this one is not working or used we should remove it

tailwind.config.js
```diff
@@ -112,6 +112,9 @@ module.exports = {
                     3: '#FF4A4A',
                 },
             },
+            screens: {
```
kushagrasarathe: @Zishan-7 qn (blocking): i'm concerned on this new breakpoint, rn we dont use xs breakpoint anywhere, but afaik, the default breakpoint will be overridden by this and might break ui on some places. Are you sure this change is not affecting anything anywhere? have you smoke tested all the flows? including large desktop views and mobile views>

kushagrasarathe (changes requested):
@Zishan-7  left some comments, lmk when responded/addressed with action items if any

tailwind.config.js
Zishan-7: This is not used anywhere, I forgot to remove it. I have removed it now

src/app/layout.tsx
Zishan-7: The one declare here is used and is working, check the `font-roboto-flex-extrabold` in `globals.css`. The one declared in tailwind config is not needed, I have removed it.

kushagrasarathe (dismissed):
lgtm, have only reviewed the code changes

Hugo0 (approved):

PR #1007, opened 2025-07-23
feat: guest claimed link history ui

kushagrasarathe:
contributes to TASK-13307 : handle history entry ui for bank account guest claimed links

Screenshot 2025-07-30 at 8 12 30 PM


src/components/TransactionDetails/TransactionDetailsDrawer.tsx
```diff
@@ -324,28 +329,37 @@ export const TransactionDetailsReceipt = ({
...
+                                value={
+                                    <div className="flex items-center gap-2">
+                                        <span>{transaction.id.toUpperCase()}</span>
+                                        <CopyToClipboard textToCopy={transaction.id.toUpperCase()} iconSize="4" />
+                                    </div>
+                                }
+                                hideBottomBorder={
+                                    !transaction.status && !transaction.memo && !transaction.attachmentUrl
+                                }
+                            />
```
Hugo0: thought: it would be super useful if you could send a ss of this with the PR submission. That way reviewer can quickly visualize this without having to spin up an instance etc

src/components/TransactionDetails/TransactionDetailsDrawer.tsx
```diff
@@ -324,28 +329,37 @@ export const TransactionDetailsReceipt = ({
...
-                    {transaction.id && transaction.direction === 'bank_withdraw' && (
-                        <PaymentInfoRow
-                            label="Transfer ID"
-                            value={
-                                <div className="flex items-center gap-2">
-                                    <span>{transaction.id.toUpperCase()}</span>
-                                    <CopyToClipboard textToCopy={transaction.id.toUpperCase()} iconSize="4" />
+                                    <span>
+                                        {isGuestBankClaim
+                                            ? `****${transaction.bankAccountDetails.identifier.slice(-4)}`
```
Hugo0: **potentially blocking: ** We'll have to see on this about how users feel, in principle I think it is fine to communicate bank accounts (this is normal for tradfi). 

But **FLAGGING HERE:** please please please always be cautious/flag if you're doing Frontend "security". Always assume frontend is compromised. If we have a data requirement but backend is passing the data, we're just masking it in Frontend, then that's very bad (FYI)

src/components/TransactionDetails/TransactionDetailsDrawer.tsx
```diff
@@ -324,28 +329,37 @@ export const TransactionDetailsReceipt = ({
...
+                                            iconSize="4"
+                                        />
+                                    )}
                                 </div>
                             }
                             hideBottomBorder={!transaction.status && !transaction.memo && !transaction.attachmentUrl}
                         />
                     )}
+                    {transaction.id &&
+                        (transaction.direction === 'bank_withdraw' || transaction.direction === 'guest_bank_claim') && (
```
Hugo0: nit: guest_bank_claim sounds more like a type than a direction

src/components/TransactionDetails/TransactionDetailsDrawer.tsx
```diff
@@ -324,28 +329,37 @@ export const TransactionDetailsReceipt = ({
...
+                            <PaymentInfoRow
+                                label="Transfer ID"
+                                value={
+                                    <div className="flex items-center gap-2">
+                                        <span>{transaction.id.toUpperCase()}</span>
+                                        <CopyToClipboard textToCopy={transaction.id.toUpperCase()} iconSize="4" />
```
Hugo0, line 379: praise: i like this component!

src/components/TransactionDetails/TransactionDetailsHeaderCard.tsx
```diff
@@ -18,6 +18,7 @@ export type TransactionDirection =
     | 'withdraw'
     | 'add'
     | 'bank_withdraw'
+    | 'guest_bank_claim'
     | 'bank_deposit'
```
Hugo0: naming thought: here we do it correctly, we don't mention  bridge specifically in name, unlike in backend PR.

We should have consistency between BE/FE names here

src/components/TransactionDetails/TransactionDetailsHeaderCard.tsx
```diff
@@ -89,6 +90,9 @@ const getTitle = (
             case 'bank_withdraw':
                 titleText = `${status === 'completed' ? 'Withdrew' : 'Withdrawing'} to ${displayName}`
                 break
+            case 'guest_bank_claim':
+                titleText = displayName
```
Hugo0, line 94: Q: just the DIsplayName? is that correct?

src/components/TransactionDetails/transactionTransformer.ts
```diff
@@ -212,6 +212,12 @@ export function mapTransactionDataForDrawer(entry: HistoryEntry): MappedTransact
             nameForDetails = 'Bank Account'
             isPeerActuallyUser = false
             break
+        case EHistoryEntryType.GUEST_BANK_CLAIM:
+            direction = 'guest_bank_claim'
+            transactionCardType = 'bank_withdraw'
+            nameForDetails = 'Claimed to Bank'
+            isPeerActuallyUser = false
```
Hugo0, line 219: Q: what does this var do? why Peer? Why not name Receiver or Claimer?

src/components/TransactionDetails/transactionTransformer.ts
```diff
@@ -239,7 +245,11 @@ export function mapTransactionDataForDrawer(entry: HistoryEntry): MappedTransact
     }
 
     // map the raw status string to the defined ui status types
-    if (entry.type === EHistoryEntryType.BRIDGE_OFFRAMP || entry.type === EHistoryEntryType.BRIDGE_ONRAMP) {
+    if (
+        entry.type === EHistoryEntryType.BRIDGE_OFFRAMP ||
+        entry.type === EHistoryEntryType.BRIDGE_ONRAMP ||
```
Hugo0, line 250: nit: naming inconsistency. Bridge in some, bridge not in others

Hugo0 (dismissed):

src/components/TransactionDetails/TransactionDetailsHeaderCard.tsx
kushagrasarathe: yes, we show claimed to bank account for guest claims, manu was out so this is something i came up with, will discuss with him and see if any changes are required

src/components/TransactionDetails/transactionTransformer.ts
kushagrasarathe: ahh this is very old implementation to check if the sender is also a user or not, agree, naming here could be better

Hugo0 (dismissed):

jjramirezn (dismissed):

Zishan-7 (approved):

PR #1022, opened 2025-07-26
Feat/status

Hugo0:
we need to know status of all our services - this creates endpoints for all our services + upstream providers.

missing: bridge & manteca


Zishan-7 (approved):
LGTM

PR #1028, opened 2025-07-28
Fix: Add crypto showing incorrect view on refresh

Zishan-7:
Contributes to TASK-13290

### DIFF
--- a/src/app/[...recipient]/client.tsx
+++ b/src/app/[...recipient]/client.tsx
@@ -322,7 +322,10 @@ export default function PaymentPage({ recipient, flow = 'request_pay' }: Props) 
             return
         }
 
-        dispatch(paymentActions.setView('STATUS'))
+        // show status view only if fulfillment payment is successful
+        if (chargeDetails?.fulfillmentPayment?.status === 'SUCCESSFUL') {
+            dispatch(paymentActions.setView('STATUS'))
+        }
 
         // only open transaction details drawer if not add money flow
         if (!isAddMoneyFlow) {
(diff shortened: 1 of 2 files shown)


kushagrasarathe (changes requested):
@Zishan-7  changes looks good, but one issue, see recording below.
i created a charge, then refreshed, i see the amount view (correct), but now when i click review it resets the form

two options here, imo:
- either persist the charge details and initiate a new payment for that same charge, without resetting the form (this is ideal)
- or completely dismiss the old charge on refresh (or state clear scenarios) and create a new one

more in favour of first one as that is ideal for charge spec

https://github.com/user-attachments/assets/98d5fe91-9232-46df-ad4a-7d469a6d04e2

Zishan-7:
> @Zishan-7 changes looks good, but one issue, see recording below. i created a charge, then refreshed, i see the amount view (correct), but now when i click review it resets the form
> 
> two options here, imo:
> 
> * either persist the charge details and initiate a new payment for that same charge, without resetting the form (this is ideal)
> * or completely dismiss the old charge on refresh (or state clear scenarios) and create a new one
> 
> more in favour of first one as that is ideal for charge spec
> 
>  Screen.Recording.2025-07-29.at.4.20.15.PM.mov

Should be fixed. Now if you refresh, you will land on the confirm view again and can continue your payment.

kushagrasarathe (changes requested):
@Zishan-7  still one more issue, showing confirm view directly is okay but is not the best choice cuz, i selected usdc base in token selector then on confirm view refreshed the page, but now the confirm view takes usdc on arb as default cuz token selector state is lost.

in this particular case we want to avoid creating new charge (already handled on line 118), so we should show the initial view with details filled from the charge and let users select the chain/token they want to add moni from

https://github.com/user-attachments/assets/9590d114-1b56-4d95-a905-d8e28b8717e1

Zishan-7:
@kushagrasarathe should be fixed now

kushagrasarathe (approved):
smoke test passed, lgtm 🫡

PR #1029, opened 2025-07-29
Fix KYC issues

Zishan-7:
Contributes to TASK-12602

### DIFF
--- a/src/components/Global/IframeWrapper/index.tsx
+++ b/src/components/Global/IframeWrapper/index.tsx
@@ -95,6 +95,7 @@ const IframeWrapper = ({ src, visible, onClose, closeConfirmMessage }: IFrameWra
                 </div>
                 <div className="h-full w-full flex-grow overflow-scroll">
                     <iframe
+                        key={src}
                         src={src}
                         allow="camera;"
                         style={{ width: '100%', height: '100%', border: 'none' }}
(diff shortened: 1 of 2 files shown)


src/hooks/useKycFlow.ts
```diff
@@ -126,20 +126,16 @@ export const useKycFlow = ({ onKycSuccess, flow }: UseKycFlowOptions = {}) => {
...
-                        visible: true,
-                        closeConfirmMessage: 'Are you sure? Your KYC progress will be lost.',
-                    })
-                }, 100)
+
+                setIframeOptions({
+                    src: kycUrl,
+                    visible: true,
+                    closeConfirmMessage: 'Are you sure? Your KYC progress will be lost.',
+                })
```
Zishan-7, line 137: @kushagrasarathe  Removed the setTimeout here, I think you added it because the iframe wasnt updating the KYC link without the delay. I think the TOS acceptance problem comes out of here only, 100ms maybe sometimes is not enough time for the UI to close the modal, and thats why even with a successful webhook it does not redirect to the next link

src/components/Global/IframeWrapper/index.tsx
```diff
@@ -95,6 +95,7 @@ const IframeWrapper = ({ src, visible, onClose, closeConfirmMessage }: IFrameWra
                 </div>
                 <div className="h-full w-full flex-grow overflow-scroll">
                     <iframe
+                        key={src}
```
Zishan-7, line 98: Added key here to make sure the iframe re-renders if the link changes

kushagrasarathe (approved):
tested, works fine 🔥

PR #1030, opened 2025-07-29
fix: more and more bugssssss

kushagrasarathe:
- fixes TASK-13448 : always show $ value and not token value in receipts and success views
- also fixes TASK-13450 : broken back navigation on withdraw and links-v2 guest claim flow


Zishan-7 (changes requested):
The changes look good, just one bug- 

I sent `$0.12` to andrevalen.eth on polygon in POL but when I click on the activity entry it shows `$0.45 ($0.12 == 0.45POL)`


Screenshot 2025-07-29 at 8 45 21 PM

Zishan-7 (approved):
lgtm 🔥

PR #1040, opened 2025-07-31
fix: refetch wallet balance

kushagrasarathe:
- fixes TASK-13447 : refetch wallet balance when direct req is paid
- also deleted some dead code


Zishan-7 (approved):
Working smoothly! 🔥

PR #1047, opened 2025-07-31
[TASK-13143] fix(cross-chain): improve route estimation

jjramirezn:
- Use linear interpolation to set realistic bounds

### DIFF
--- a/src/components/Claim/Link/Initial.view.tsx
+++ b/src/components/Claim/Link/Initial.view.tsx
@@ -18,7 +18,6 @@ import {
     PINTA_WALLET_CHAIN,
     PINTA_WALLET_TOKEN,
     ROUTE_NOT_FOUND_ERROR,
-    SQUID_API_URL,
 } from '@/constants'
 import { TRANSACTIONS } from '@/constants/query.consts'
 import { loadingStateContext, tokenSelectorContext } from '@/context'
@@ -37,11 +36,11 @@ import {
 } from '@/utils'
 import { NATIVE_TOKEN_ADDRESS, SQUID_ETH_ADDRESS } from '@/utils/token.utils'
 import * as Sentry from '@sentry/nextjs'
-import { getSquidRouteRaw } from '@squirrel-labs/peanut-sdk'
 import { useQueryClient } from '@tanstack/react-query'
 import { useRouter, useSearchParams } from 'next/navigation'
 import { useCallback, useContext, useEffect, useMemo, useState } from 'react'
 import { formatUnits } from 'viem'
+import type { Address } from 'viem'
 import { IClaimScreenProps } from '../Claim.consts'
 import useClaimLink from '../useClaimLink'
 import GuestActionList from '@/components/GuestActions/MethodList'
@@ -51,6 +50,7 @@ import { Slider } from '@/components/Slider'
 import Image from 'next/image'
 import { PEANUT_LOGO_BLACK, PEANUTMAN_LOGO } from '@/assets'
 import { BankFlowManager } from './views/BankFlowManager.view'
+import { type PeanutCrossChainRoute, getRoute } from '@/services/swap'
 
 export const InitialClaimLinkView = (props: IClaimScreenProps) => {
     const {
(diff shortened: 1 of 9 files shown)


src/services/swap.ts
```diff
@@ -300,82 +314,116 @@ async function findOptimalFromAmount(
...
+    const testResponse = await getSquidRouteRaw(
+        {
+            ...params,
+            fromAmount: initialFromAmount.toString(),
+        },
+        options
+    )
+
+    const actualReceived = BigInt(testResponse.route.estimate.toAmountMin)
+    console.log(
```
kushagrasarathe, line 343: thought, we have a lot of informative logs in app, but it's not good practice to leave logs in prod code, we should remove logs like these

src/services/swap.ts
```diff
@@ -300,82 +314,116 @@ async function findOptimalFromAmount(
...
+        'Initial test - fromAmount:',
+        initialFromAmount,
+        'receivedAmount:',
+        actualReceived,
+        'targetToAmount:',
+        targetToAmount
+    )
+
+    // Step 2: Check if initial test is already good enough
+    const initialOverage = calculateOverage(actualReceived, targetToAmount)
```
kushagrasarathe, line 353: nit: @jjramirezn  it's kind of unclear at first look why exactly do we need these changes, would be great if you could add some comments for future reference

src/services/swap.ts
```diff
@@ -300,82 +314,116 @@ async function findOptimalFromAmount(
...
+        console.log('Initial estimate is good enough, overage:', initialOverage)
+        return testResponse
+    }
+
+    // Step 3: Use linear interpolation to calculate realistic bounds
+    const ratio = Number(targetToAmount) / Number(actualReceived)
+    const linearEstimate = (initialFromAmount * BigInt(Math.floor(ratio * 1000))) / 1000n
+
+    // Set tight bounds around the linear estimate (±2% range)
+    const boundRange = linearEstimate / 50n // 2% of linear estimate
```
kushagrasarathe, line 364: qn (non-blocking): why do we want the range to be in ±2% range? can't it be dynamic?

src/services/swap.ts
```diff
@@ -300,82 +314,116 @@ async function findOptimalFromAmount(
...
+
+    // Step 2: Check if initial test is already good enough
+    const initialOverage = calculateOverage(actualReceived, targetToAmount)
+    if (initialOverage >= 0 && initialOverage <= maxOverage) {
+        console.log('Initial estimate is good enough, overage:', initialOverage)
+        return testResponse
+    }
+
+    // Step 3: Use linear interpolation to calculate realistic bounds
+    const ratio = Number(targetToAmount) / Number(actualReceived)
```
kushagrasarathe, line 360: qn (non-blocking): are there any edge cases where linear interpolation might fail?

kushagrasarathe (approved):
@jjramirezn  approved 🔥 
left few qns to understand few things, lmk when addressed

src/services/swap.ts
jjramirezn: I left these because they are in the next server side and are useful for debugging issues

src/services/swap.ts
jjramirezn: We assume that the linear will be good enough so the band can be small

src/services/swap.ts
jjramirezn: If the relation is non linear, we may have the same issue we had before with our range being outside of the real estimation but I believe that is less likely and in tests it has proven very accurate

src/components/Claim/Link/Initial.view.tsx
```diff
@@ -409,20 +415,26 @@ export const InitialClaimLinkView = (props: IClaimScreenProps) => {
...
+                    {
+                        from: {
+                            address: (claimLinkData.sender?.accounts[0].identifier ??
+                                claimLinkData.senderAddress) as Address,
+                            tokenAddress: fromToken as Address,
+                            chainId: claimLinkData.chainId,
+                        },
+                        to: {
+                            address:
+                                recipientType === 'us' || recipientType === 'iban'
```
kushagrasarathe, line 428: @jjramirezn  i dont think this check is needed anymore, afaik we do not accept iban/us account type claims directly in the input now for claims, we have separate flow for tha

PR #1086, opened 2025-08-12
Payment Initiator Refactor

jjramirezn:
Separate State and reusable UI for the different payment flows. 

Each flow logic and state must be separated from eachother and also reset properly on unmount (ie: the dev shoulndt have to remember to call a resetState() ) For this each payment flow will be its own hook, we are removing redux stores for state and storing directly in the hook (and on unmount it dissapear, that is intended). Also using tanstack query for some calls (we are using in other parts of the application and they are working well for us)

UI should not contain flow logic and as much as possible just receive the necesary props. 

After all this we should be able to have the same flows with the same UI, with less bug and developer overhead

For more information read: https://www.notion.so/peanutprotocol/PaymentForm-and-flow-refactor-workshop-22b838117579800c8074d649e6bc565f?source=copy_link


Hugo0:
closing

PR #1101, opened 2025-08-18
Fix back button not working on /profile

Zishan-7:
Contributes to TASK-13205

### DIFF
--- a/src/components/Profile/components/PublicProfile.tsx
+++ b/src/components/Profile/components/PublicProfile.tsx
@@ -60,10 +60,14 @@ const PublicProfile: React.FC<PublicProfileProps> = ({
                 ) : (
                     <NavHeader
                         onPrev={() => {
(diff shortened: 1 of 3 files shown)


src/components/Profile/components/PublicProfile.tsx
```diff
@@ -60,10 +60,14 @@ const PublicProfile: React.FC<PublicProfileProps> = ({
                 ) : (
                     <NavHeader
                         onPrev={() => {
-                            if (window.history.length > 1) {
+                            // Check if the referrer is from the same domain (internal navigation)
+                            const isInternalReferrer =
```
kushagrasarathe: @Zishan-7  i'm pretty sure this will be needed on more places, can extract it in a utility and use it?

kushagrasarathe (approved):
left a comment, rest lgtm

PR #1102, opened 2025-08-18
fix: claim flow bugs

kushagrasarathe:
- fixes TASK-13943 : claimers name getting updated to sender's name
- fixes TASK-13936 : cross chain claiming to exchanges/wallets not working
- also fixes a issue, where token selector would reset to default chain/token after entering ens/address in input
- also fixes TASK-13977 : back btn navigations on desktop view for claim flow


Zishan-7 (approved):
lgtm

PR #1115, opened 2025-08-20
[TASK-14107] fix: don't allow claiming on xChain if route is not found

jjramirezn:
sometimes the button is clicked before the route is finished loading, it should not be enabled. Also wtf that min calculation

### DIFF
--- a/src/components/Claim/Link/Initial.view.tsx
+++ b/src/components/Claim/Link/Initial.view.tsx
@@ -560,6 +560,7 @@ export const InitialClaimLinkView = (props: IClaimScreenProps) => {
         selectedRoute,
         recipient.address,
         isValidRecipient,
+        hasFetchedRoute,
     ])
 
     const getButtonText = () => {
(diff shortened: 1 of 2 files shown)


src/components/Claim/Link/Initial.view.tsx
```diff
@@ -560,6 +560,7 @@ export const InitialClaimLinkView = (props: IClaimScreenProps) => {
         selectedRoute,
         recipient.address,
         isValidRecipient,
+        hasFetchedRoute,
```
Hugo0, line 563: makes sense

src/components/Claim/Link/Onchain/Confirm.view.tsx
```diff
@@ -58,16 +59,13 @@ export const ConfirmClaimLinkView = ({
...
     const minReceived = useMemo<string>(() => {
-        let amountNumber: number
-
-        // manual 1% slippage calculation based on the deposited token amount
-        amountNumber = Number(formatUnits(BigInt(claimLinkData.amount), claimLinkData.tokenDecimals)) * 0.99 // subtract 1%
-
-        const formattedAmount = formatTokenAmount(amountNumber)
-
-        return `$ ${formattedAmount}`
+        if (!selectedRoute || !resolvedTokenSymbol) return ''
```
Hugo0, line 63: this is indeed fucked

Hugo0: iirc @kushagrasarathe added this as a quick fix a while ago but we never revisited

Hugo0 (approved):

src/components/Claim/Link/Onchain/Confirm.view.tsx
jjramirezn: yeah I checked

PR #1120, opened 2025-08-21
[TASK-14113] fix: handle rpc outage when creating sendlinks

jjramirezn:
Identify cause of failure when sending an user op (failure of sending the op vs failure of getting the receipt). In cases where the failure is in getting the receipt, handle gracefully


src/components/Create/useCreateLink.tsx
```diff
@@ -361,9 +361,14 @@ export const useCreateLink = () => {
...
-            const contractAbi = getContractAbi(contractVersion)
-            const contractAddress: Hash = getContractAddress(chainId, contractVersion) as Hash
             const tokenAddress = PEANUT_WALLET_TOKEN as Hash
+            const contractAbi = peanut.getContractAbi(contractVersion)
+            const contractAddress: Hash = peanut.getContractAddress(
+                PEANUT_WALLET_CHAIN.id.toString(),
+                contractVersion
+            ) as Hash
+            // Get the next deposit index to store in case of failure
+            const nextIndex = await getNextDepositIndex(contractVersion)
```
Hugo0: ISSUE, blocking until response: This introduces an extra api call and thus extra latency. Can we paralelize this? Or is there another spot in the code where we read the state of the contract? If so, do that for speed/smoothness

Hugo0: Another Question: why n+1? The correct deposit idx could be n+2 or n+3

jjramirezn: We could do it before the user clicks on create link, while they are reading the data or whatever. But then we have to ensure we have it if the user clicks too fast. There is no other read in this flow.

second question: yeah, that's the whole point, thats why we check on the backend if we have an incomplete link. But the minimum index that it can possibly be is the same as the depositCount, so thats the one we are using.

It's a pain to check the index for every creation, because that number won't be used for the majority of them, only for the ones that fail after sending the user operation. And in that case, we cannot get the deposit count because we assume there is a network error

jjramirezn: I think we can group it witrh the handleSendUserOpEncoded actually

PR #1133, opened 2025-08-26
fix: send flow cta

kushagrasarathe:
- fixes TASK-14256 : request flow cta being shown in direct send
- also fixes TASK-13591 : success sound not playing on direct send


Zishan-7 (changes requested):
Send with peanut button is always disabled when paying to address

image

kushagrasarathe:
> Send with peanut button is always disabled when paying to address

@Zishan-7  works for me
image

Zishan-7 (approved):
lgtm

PR #1138, opened 2025-08-27
fix: bank claim flow runtime error

kushagrasarathe:
- fixes TASK-14287 : runtime error when claiming and BIC autocompletion fails
- also fixes TASK-14078 : similar runtime error due to iban mismatch
- also fixes race condition issue when bic completion fails and the cta keeps loading or requires multiple clicks to proceed to next view

### DIFF
--- a/public/game/peanut-game.html
+++ b/public/game/peanut-game.html
@@ -1001,7 +1001,7 @@
                  * Maximum obstacle grouping count.
                  * @const
                  */
-                ;(Obstacle.MAX_OBSTACLE_LENGTH = 3),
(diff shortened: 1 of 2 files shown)


public/game/peanut-game.html
jjramirezn, line 1: WTF is this I had the same change on a commit

jjramirezn (approved):

public/game/peanut-game.html
kushagrasarathe: @jjramirezn `pnpm format ` lol

PR #1145, opened 2025-08-28
Fix: disable add money button on default state + disable sound on IOS

Zishan-7:
TASK-13591
TASK-14289
TASK-14253

### DIFF
--- a/src/components/Global/SoundPlayer.tsx
+++ b/src/components/Global/SoundPlayer.tsx
@@ -1,5 +1,6 @@
 'use client'
 
+import { DeviceType, useDeviceType } from '@/hooks/useGetDeviceType'
 import { useEffect, useRef } from 'react'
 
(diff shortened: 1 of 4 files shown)


src/components/Global/SoundPlayer.tsx
```diff
@@ -18,6 +19,13 @@ type SoundPlayerProps = {
 export const SoundPlayer = ({ sound }: SoundPlayerProps) => {
     const audioRef = useRef<HTMLAudioElement | null>(null)
 
+    const { deviceType } = useDeviceType()
+
+    // Early return for iOS devices - completely disable sound
```
Hugo0, line 24: nit: would benefit of a @dev TODO comment explaining that we'd want to fix this in the future

Hugo0 (approved):
lgtm

PR #1151, opened 2025-08-29
HOTFIX: Invalid IBAN for UK

Zishan-7:


src/utils/withdraw.utils.ts
```diff
@@ -23,7 +23,7 @@ export const getCountryFromIban = (iban: string): string | null => {
         }
     }
 
-    return country ? country.title : null
+    return country ? country.path : null
 }
```
Zishan-7, line 27: returning path here is better because for countries like United Kingdom path is 'united-kingdom' which matches the selectedCountry variable value

Hugo0 (approved):

PR #1159, opened 2025-09-01
[TASK-14287] fix: stop cleaning error by bic field

jjramirezn:
We now always clear before starting submission and also bic field will always show, so that logic is not needed anymore.


Zishan-7 (approved):
LGTM!

PR #1160, opened 2025-09-01
[TASK-13586] fix: don't try to create account twice on setup

jjramirezn:
Dont merge if approved, we are in code freeze


Zishan-7 (approved):
LGTM!

PR #1161, opened 2025-09-02
[TASK-13531] Fix: Automatic balance refresh not working

Zishan-7:


src/hooks/wallet/useWallet.ts
```diff
@@ -103,11 +106,88 @@ export const useWallet = () => {
...
+
+            // Store cleanup function that cleans up both watchers
+            eventListenerRef.current = () => {
+                watchIncoming()
+                watchOutgoing()
+            }
+        } catch (error) {
+            console.error('Failed to setup balance monitoring:', error)
+        }
+    }, [address, fetchBalance])
```
jjramirezn, line 170: issue: this is not too efficient with credits

Some back of napkin calculations:

viem's watchContractEvent does the following:



Which in infura (ignoring alchemy for now) it has these costs:



Lets assume a session of 1 minute, that would be 30 calls to getFilterChanges (with the 2sec interval)

30 calls x 140 credits = 4200 credits
4200 credits + 80 credits (newFilter) = 4280 credits
4280 credits * 2 watchContractEvent = 8560 credits per user session
This becomes too much too quick (longer sessions, etc etc) only 100 sessions like this and we are at almost 1M credits (we have 15M per day right now)

I strongly **recommend** doing this with websocket triggered from the deposit indexer.

jjramirezn (changes requested):

Hugo0:
closed; different PR tackled this

PR #1199, opened 2025-09-10
[TASK-14544] Add Support CTA for claim and req fulfilment guest flows

Zishan-7:
image

image

### DIFF
--- a/src/app/(mobile-ui)/layout.tsx
+++ b/src/app/(mobile-ui)/layout.tsx
@@ -145,7 +145,7 @@ const Layout = ({ children }: { children: React.ReactNode }) => {
                                         'flex w-full items-center justify-center md:ml-auto md:w-[calc(100%-160px)]',
                                         alignStart && 'items-start',
                                         isSupport && 'h-full',
-                                        isUserLoggedIn ? 'min-h-[calc(100dvh-160px)]' : 'min-h-[calc(100dvh-110px)]'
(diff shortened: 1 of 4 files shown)


src/app/(mobile-ui)/layout.tsx
```diff
@@ -145,7 +145,7 @@ const Layout = ({ children }: { children: React.ReactNode }) => {
                                         'flex w-full items-center justify-center md:ml-auto md:w-[calc(100%-160px)]',
                                         alignStart && 'items-start',
                                         isSupport && 'h-full',
-                                        isUserLoggedIn ? 'min-h-[calc(100dvh-160px)]' : 'min-h-[calc(100dvh-64px)]'
+                                        isUserLoggedIn ? 'min-h-[calc(100dvh-160px)]' : 'min-h-[calc(100dvh-110px)]'
```
kushagrasarathe: I'm very much concerned this might affect other views

Zishan-7: I changed this because giving more height was adding unnecessary scroll

Hugo0 (approved):

src/app/(mobile-ui)/layout.tsx
kushagrasarathe: @Zishan-7  instead of changing root layout, cant we just individually update css in two files where we are calling `SupportCTA` atleast that would save us from other flows going off screen or shriniking

PR #1207, opened 2025-09-12
[TASK-14388] Feat/request on ramp improvements

Zishan-7:
This PR also contributes to TASK-14541

image

Rest of the changes are related to redirect flow- so cant paste SS

### DIFF
--- a/src/components/Global/PostSignupActionManager/post-signup-action.consts.ts
+++ b/src/components/Global/PostSignupActionManager/post-signup-action.consts.ts
@@ -11,4 +11,14 @@ export const POST_SIGNUP_ACTIONS = [
             icon: 'check' as IconName,
         },
     },
+    {
+        // this regex will match any path that contains the word "claim", this helps in determing if the user is coming from a claim link
+        pathPattern: /^\/[^\/]+\/[^\/\?]+\?.*chargeId=/,
+        config: {
+            title: 'Verification complete!',
+            description: `Your identity has been successfully verified. You can now send the money your friend requested!`,
+            cta: 'Send it!',
+            icon: 'check' as IconName,
+        },
+    },
 ]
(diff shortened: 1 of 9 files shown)


src/components/Global/PostSignupActionManager/post-signup-action.consts.ts
```diff
@@ -11,4 +11,14 @@ export const POST_SIGNUP_ACTIONS = [
             icon: 'check' as IconName,
         },
     },
+    {
+        // this regex will match any path that resembles the request link, this helps in determing if the user is coming from a request link
+        pathPattern: /^\/[^\/]+\/[^\/\?]+\?.*chargeId=/,
```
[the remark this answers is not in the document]
Zishan-7: the redirect url will always be relative. we never store absolute path

src/components/Global/PostSignupActionManager/post-signup-action.consts.ts
```diff
@@ -11,4 +11,14 @@ export const POST_SIGNUP_ACTIONS = [
...
+        config: {
+            title: 'Verification complete!',
+            description: `Your identity has been successfully verified. You can now send the money your friend requested!`,
```
Hugo0, line 19: note: whilst good, this seems pretty complex to have to think about so many flows

src/components/Request/views/ReqFulfillBankFlowManager.tsx
```diff
@@ -55,6 +55,7 @@ export const ReqFulfillBankFlowManager = ({ parsedPaymentData }: { parsedPayment
             setIsKycModalOpen(true)
         }
     }, [showVerificationModal])
+    console.log('first')
```
Hugo0: issue: remove this

Hugo0 (approved):
minor PR. has one unnecessary log line. I don't see anything critical

src/components/Global/PostSignupActionManager/post-signup-action.consts.ts
Zishan-7: Best way is to save these states in the database and trigger modals based on that. It will always be complicated to implement this on the FE

PR #1221, opened 2025-09-17
[TASK-14493] feat: show local currency exchange rate in claim to bank confirm view

Zishan-7:
image

### DIFF
--- a/src/components/Claim/Link/views/Confirm.bank-claim.view.tsx
+++ b/src/components/Claim/Link/views/Confirm.bank-claim.view.tsx
@@ -13,8 +13,6 @@ import { ClaimLinkData } from '@/services/sendLinks'
 import { formatUnits } from 'viem'
 import ExchangeRate from '@/components/ExchangeRate'
 import { AccountType } from '@/interfaces'
-import { useCurrency } from '@/hooks/useCurrency'
-import { getCurrencySymbol } from '@/utils/bridge.utils'
 import countryCurrencyMappings from '@/constants/countryCurrencyMapping'
 
 interface ConfirmBankClaimViewProps {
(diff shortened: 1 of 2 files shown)


src/components/ExchangeRate/index.tsx
```diff
@@ -5,12 +5,13 @@ import { useExchangeRate } from '@/hooks/useExchangeRate'
 
 interface IExchangeRateProps extends Omit<IExchangeRate, 'enabled'> {
     nonEuroCurrency?: string
+    sourceCurrency?: string
 }
 
-const ExchangeRate = ({ accountType, nonEuroCurrency }: IExchangeRateProps) => {
+const ExchangeRate = ({ accountType, nonEuroCurrency, sourceCurrency = 'USD' }: IExchangeRateProps) => {
     const { exchangeRate, isFetchingRate } = useGetExchangeRate({ accountType, enabled: !nonEuroCurrency })
     const { exchangeRate: nonEruoExchangeRate, isLoading } = useExchangeRate({
```
kushagrasarathe, line 13: @Zishan-7 can you add some comments on this component, initally it used only one hook for exchange rate but after the lp widget we are using two hooks (not sure if both required now but good if we can document why we are using two)

src/components/Claim/Link/views/Confirm.bank-claim.view.tsx
kushagrasarathe, line 1: @Zishan-7 the pr looks good but it doesnt tackle the main issue that i meant, bad on me to move the ownership without giving proper context 😭

actually in the PeanutActionDetailsCard sometimes the currency amount came up as usd value instead of the local currency, ig that's not happening anymore as your screenshot shows correct usd value, lets discuss on discord

kushagrasarathe (approved):
@Zishan-7 left two comments, one is actionable (comment one), rest lgtm

kushagrasarathe (changes requested):
actually @Zishan-7 some issues with the pr, pining on discord

src/components/ExchangeRate/index.tsx
Zishan-7: useGetExchange hook fetches the exchange rate from bridge API, but it only gives data for EURO and USD ->MXN, but we needed more currencies for exhange rate widget and thats why useExchangeRate was created, it uses combination of bridget + third part API and provides exchange rate for more currency pairs. I think useGetExchange rate can be deprecated now and should be removed in future

kushagrasarathe (approved):
lgtm

PR #1255, opened 2025-09-26
[TASK-14780] Feat/add manteca currencies to exchange rate widget

Zishan-7:


src/app/api/exchange-rate/route.ts
```diff
@@ -51,24 +134,60 @@ export async function GET(request: NextRequest) {
...
+            return fromToUsd * usdToTo
+        } else {
+            // One currency is LATAM, the other is not USD - this shouldn't happen in our flow
+            // but we'll return null to fall back to other providers
+            console.warn(`Unsupported Manteca conversion: ${from} → ${to}`)
+            return null
+        }
+    } catch (error) {
+        console.error(`Manteca API error for ${from}-${to}:`, error)
+        return null
```
jjramirezn: issue: this method is not needed

Or at least it should not call mantecaApi.getPrices. Price fetching should be from the action `getCurrencyPrice`

src/app/api/exchange-rate/route.ts
```diff
@@ -39,8 +43,87 @@ export async function GET(request: NextRequest) {
...
+            // Check if we should use Bridge for this pair or its reverse
+            const shouldUseBridge = BRIDGE_PAIRS.has(pairKey)
+            const shouldUseBridgeReverse = BRIDGE_PAIRS.has(reversePairKey)
+
+            if (shouldUseBridge || shouldUseBridgeReverse) {
+                // For Bridge pairs, we need to determine which rate to use
+                let bridgeResult
+                if (shouldUseBridge) {
+                    // Direct pair (e.g., USD→EUR): use sell_rate
+                    bridgeResult = await fetchFromBridge(fromUc, toUc, 'sell_rate', false)
```
jjramirezn: comment: this should not call bridge directly

this is not part of this PR but it should've been raised in the moment it was introduced. But if you check this is the only thing in the app that uses the BRIDGE_API that is because we moved everything related to communication with bridge to the backend. And this should be using `getExchangeRate` function (that we should rename to be bridge specific)

jjramirezn (changes requested):
Check comments

src/app/api/exchange-rate/route.ts
jjramirezn: @Hugo0 FYI this is related to a PR you reviewed. I think its understandable that you missed it. The context needed to know this was with kush or with me

jjramirezn (approved):

PR #1267, opened 2025-10-01
fix: add loading feedback in country list for smooth ux

kushagrasarathe:
- fixes TASK-14968 : added loading feedback and using transition hook from react now to make navigation feel smoother
- also added caching to useGeoLocation hook


src/hooks/useGeoLocation.ts
```diff
@@ -0,0 +1,71 @@
...
+            } finally {
+                setIsLoading(false)
+            }
+        }
+
+        fetchCountry()
+    }, [])
+
+    return { countryCode, isLoading, error }
+}
```
Zishan-7, line 71: NICE!

Zishan-7 (approved):

PR #1279, opened 2025-10-03
fix: only show token and network field when its not peanut token/chain

kushagrasarathe:
- contributes to TASK-14498 :  only show token and network field in receipt when its not peanut token/chain
- also fixes TASK-15141 : format token balance in send flow

### DIFF
--- a/src/components/Payment/PaymentForm/index.tsx
+++ b/src/components/Payment/PaymentForm/index.tsx
@@ -22,7 +22,7 @@ import { ParsedURL } from '@/lib/url-parser/types/payment'
 import { useAppDispatch, usePaymentStore } from '@/redux/hooks'
 import { paymentActions } from '@/redux/slices/payment-slice'
 import { walletActions } from '@/redux/slices/wallet-slice'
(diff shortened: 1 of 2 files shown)


src/components/Payment/PaymentForm/index.tsx
```diff
@@ -102,7 +102,7 @@ export const PaymentForm = ({
     const { initiatePayment, isProcessing, error: initiatorError } = usePaymentInitiator()
 
     const peanutWalletBalance = useMemo(() => {
-        return balance !== undefined ? formatAmount(formatUnits(balance, PEANUT_WALLET_TOKEN_DECIMALS)) : ''
```
jjramirezn, line 105: question: what is the rationale for this?

jjramirezn (dismissed):

src/components/Payment/PaymentForm/index.tsx
kushagrasarathe: @jjramirezn  answered on discord

jjramirezn (approved):

PR #1288, opened 2025-10-06
[TASK-14958] fix: reset payment state on component unmount

Zishan-7:


src/app/[...recipient]/client.tsx
```diff
@@ -410,6 +410,18 @@ export default function PaymentPage({ recipient, flow = 'request_pay' }: Props)
         }
     }, [searchParams, parsedPaymentData, chargeDetails, requestType])
 
+    // reset payment state on unmount
+    useEffect(() => {
+        return () => {
+            dispatch(paymentActions.resetPaymentState())
+            setError(null)
+            setIsUrlParsed(false)
```
Comment (kushagrasarathe) [inline src/app/[...recipient]/client.tsx]: nit: clearing local component states is unnecessary, they anyways reset on unmount

src/app/[...recipient]/client.tsx
```diff
@@ -410,6 +410,18 @@ export default function PaymentPage({ recipient, flow = 'request_pay' }: Props)
...
+            setIsRequestDetailsFetching(false)
+            setCurrencyAmount('')
+            setisLinkCancelling(false)
+        }
+    }, [dispatch])
```
Comment (kushagrasarathe) [inline src/app/[...recipient]/client.tsx]: qn: why use dispatch in dep array? empty dep array makes more sense, having disptach might result into early reset, have you tested this ?

kushagrasarathe (approved):
left some qn and a nit

src/app/[...recipient]/client.tsx
Comment (Zishan-7) [reply src/app/[...recipient]/client.tsx]: You are right, we dont need dispatch here but I dont think this interferes with when the state will reset. The return function is called only on unmount.

I will still remove it from the dependency array

PR #1293, opened 2025-10-07
Invites QA fixes

Zishan-7:
This PR - 
- Fixes login issue - sometimes user has to login twice
- Adds early user modal ( Will be shown to the user that signed up before lockup date, 11 oct 2025)
- No more jail modal

image

image


src/app/(mobile-ui)/home/page.tsx
Hugo0, line 200: nit: this comment seems outdated now

src/app/(mobile-ui)/home/page.tsx
```diff
@@ -202,7 +205,9 @@ export default function Home() {
                 !hasSeenAddMoneyPromptThisSession &&
                 !showIOSPWAInstallModal &&
                 !showBalanceWarningModal &&
-                !isPostSignupActionModalVisible
+                !isPostSignupActionModalVisible &&
+                showNoMoreJailModal !== 'true' &&
```
Hugo0, line 211: code style issue: why break convention here? The other modals have a certain. we need to standardize. This is confusing.

"hasSeen"
"Show"
"isVisible"

All three say the same thing but are unnecessarily named differently

Boy scout rule: leave code better than you found it

src/components/Global/EarlyUserModal/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { useEffect, useState } from 'react'
+import ActionModal from '../ActionModal'
+import ShareButton from '../ShareButton'
+import { generateInvitesShareText } from '@/utils'
+import { useAuth } from '@/context/authContext'
+import { updateUserById } from '@/app/actions/users'
+
+const EarlyUserModal = () => {
+    const { user } = useAuth()
+    const inviteCode = `${user?.user.username?.toUpperCase()}INVITESYOU`
```
Hugo0: issue: we have at least two other places in the app where we have invite codes. If you have 3 places where you do the same code, apply DRY

Hugo0 (approved):

src/app/(mobile-ui)/home/page.tsx
Zishan-7: The modals follow "show" convention. hasSeen is not used to control modal visibility. It is the flag to check if user has seen the money modal or not.

PR #1303, opened 2025-10-09
Fix: invite bugs

Zishan-7:


Hugo0 (approved):
approved, pending iOS confirmation

PR #1316, opened 2025-10-10
[TASK-14149] refactor: hide attachment

jjramirezn:


src/components/Global/FileUploadInput/index.tsx
```diff
@@ -1,9 +1,10 @@
+/* Uncomment to enable file upload */
```
Hugo0, line 1: nice

src/components/Global/FileUploadInput/index.tsx
```diff
@@ -1,9 +1,10 @@
...
 import BaseInput from '@/components/0_Bruddle/BaseInput'
-import { checkifImageType, fetchWithSentry } from '@/utils'
-import * as Sentry from '@sentry/nextjs'
-import { useEffect, useState } from 'react'
+//import { checkifImageType, fetchWithSentry } from '@/utils'
```
Hugo0, line 3: nit as always: `// abc` instead of `//abc`

Hugo0 (approved):

PR #1323, opened 2025-10-14
feat: add redirect logic on landing page

Zishan-7:


src/components/LandingPage/noFees.tsx
```diff
@@ -8,10 +8,27 @@ import { Star } from '@/assets'
...
     const router = useRouter()
+    const { fetchBalance, balance } = useWallet()
+    const { user } = useAuth()
+
+    const handleCtaAction = (sourceCurrency: string, destinationCurrency: string) => {
+        if (!user) {
+            router.push('/setup')
+            return
+        }
+        const formattedBalance = parseFloat(printableUsdc(balance ?? 0n))
```
kushagrasarathe, line 27: use `formatCurrency` here, jota recently updated that fn to show correct balance including decimal units

kushagrasarathe (approved):

src/components/LandingPage/noFees.tsx
Zishan-7: I think that function is useful for displaying an amount in the UI, here we dont need to do that parseFloat will work fine

PR #1351, opened 2025-10-22
attempt at more fixes

Hugo0:
attempts to fix:

1. sendlink not being fetched (db latency issue) with cachebuster
2. xchain claim stalling out  / errors in xchain claim

During review, ensure no funds can be sent to the wrong address or chain in xchain claim


Hugo0:
@coderabbitai During review, ensure no funds can be sent to the wrong address or chain in xchain claim

review in depth

Hugo0:
@coderabbitai During review, ensure no funds can be sent to the wrong address or chain in xchain claim

review in depth

PR #1389, opened 2025-10-31
[TASK-16364] feat: validate pix

jjramirezn:


Hugo0 (approved):
good - but this is perfect opportunity for tests.

I dont trust regexes that much

PR #1409, opened 2025-11-06
Hotfix: Add blur to QR code till request is created

Zishan-7:
https://github.com/user-attachments/assets/c54bd759-bcc9-4863-b11d-ec57aaf57c7c

### DIFF
--- a/src/components/Global/QRCodeWrapper/index.tsx
+++ b/src/components/Global/QRCodeWrapper/index.tsx
@@ -8,9 +8,10 @@ interface QRCodeWrapperProps {
     url: string
     isLoading?: boolean
     disabled?: boolean
+    isBlurred?: boolean
 }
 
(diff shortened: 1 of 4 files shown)


src/components/Global/QRCodeWrapper/index.tsx
```diff
@@ -21,7 +22,7 @@ const QRCodeWrapper = ({ url, isLoading = false, disabled = false }: QRCodeWrapp
                 )}
             >
                 {/* QR Code with white buffer */}
-                <div className="relative">
+                <div className={twMerge('relative', isBlurred && 'blur-[3px]')}>
```
kushagrasarathe, line 25: nit: explicit value? why not tailwind class?

kushagrasarathe (approved):
one nit, rest lgtm

src/components/Global/QRCodeWrapper/index.tsx
Zishan-7: its creating too much blur, not looking good

PR #1428, opened 2025-11-08
fix

Hugo0:
Fix QR claiming flow


Hugo0:
@coderabbitai review in depth

PR #1429, opened 2025-11-08
fix: refetch user on kyc state change

kushagrasarathe:


Hugo0:
@coderabbit review in depth. Do we ever fetch user unnecessarily or duplicate? Can this bring performance issues?

PR #1432, opened 2025-11-08
fix: set default currency in input to local for manteca countries

kushagrasarathe:
- fixes TASK-16703 : minor ui/ux issues


Hugo0:
@coderabbitai review in depth

These are the objectives:

## Description

- Default currency when withdrawing to an Argentina bank account is USD instead of ARS. It should be ARS the default
- The search icon is missing in the countries list



- QR code has the old design now
-

Hugo0 (approved):
We should probably expand this to Europe (euro) and mexico (mxn) as well

PR #1446, opened 2025-11-11
fix: wrong balance warning showing when it shouldnt be

Hugo0:
## Fix: Balance validation race condition during payments

**Problem:** "Not enough funds" warning flashed during QR payments and withdrawals after transaction completed but before API call finished.

**Root Cause:** Balance check effects only checked `hasPendingTransactions` (mutation state), missing the gap between transaction completion and API completion.

**Solution:** Added `isLoading` check to balance validation in 4 payment flows:
- QR payment
- Manteca withdraw  
- Bank withdraw
- Send link

**Result:** Balance validation now skips checks during the entire payment lifecycle (transaction + API calls), preventing premature warnings.

```typescript
// Before
if (hasPendingTransactions) return

// After  
if (hasPendingTransactions || isLoading) return
```

### DIFF
--- a/src/app/(mobile-ui)/qr-pay/page.tsx
+++ b/src/app/(mobile-ui)/qr-pay/page.tsx
@@ -761,7 +761,8 @@ export default function QRPayPage() {
         }
 
         // Skip balance check if transaction is being processed
-        if (hasPendingTransactions || isWaitingForWebSocket) {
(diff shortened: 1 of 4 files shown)


kushagrasarathe (approved):

kushagrasarathe:

Evaluate:
- code quality and DRY

PR #1450, opened 2025-11-12
fix: loading states to properly refetch user on page refresh

kushagrasarathe:
- fixes TASK-16818 : qr kyc on page refresh when paying a qr code


src/app/(mobile-ui)/qr-pay/page.tsx
```diff
@@ -146,7 +146,7 @@ export default function QRPayPage() {
         setPendingSimpleFiPaymentId(null)
         setWaitingForMerchantAmount(false)
         retryCount.current = 0
-        // reset perk states 
+        // reset perk states
```
jjramirezn, line 149: praise: very important

jjramirezn (approved):

PR #1455, opened 2025-11-13
[TASK-16583] feat: sign user operation and send it for QR payments instead of sending

jjramirezn:
the tx

PR #1461, opened 2025-11-13
Fix: use tanstack query for retries

Zishan-7:


src/app/(mobile-ui)/qr-pay/page.tsx
```diff
@@ -451,103 +450,87 @@ export default function QRPayPage() {
...
-                        }
-                    }, delayMs)
-                } else {
-                    // All retries exhausted
-                    mantecaRetryCount.current = 0
-                    setShouldRetry(false)
-                    setLoadingState('Idle')
+            // Provider-specific errors: show appropriate message
+            if (error.message.includes("provider can't decode it")) {
+                if (EQrType.PIX === qrType) {
```
Hugo0, line 510: why no MP?

Hugo0: or QR3

Hugo0 (approved):
might conflict w/ https://github.com/peanutprotocol/peanut-ui/pull/1455/files

Also check CR comments

src/app/(mobile-ui)/qr-pay/page.tsx
Zishan-7: for other providers, we show them the screen - waiting for merchant to set the amount. because manteca throws the same error when the amount is not set

PR #1504, opened 2025-11-23
fix: qr payment race condition on be api error

kushagrasarathe:
- there was a race condition issue on qr payments, when user op validation on backend fails, it returns a 400 response, we show it in the ui but the settimeout causes the loading state to trigger making the error flash and qr screen stuck on loading state forever


kushagrasarathe:

Evaluate:
- code quality and DRY

jjramirezn (approved):

PR #1519, opened 2025-12-01
fix: race conditon when setting up account using 1pwd passkey

kushagrasarathe:
## issue
after users registered with external passkey managers (1Password in this particular case, also tested with bitwarden), the account creation process would fail, the code wasn't waiting for the user data to load before redirecting to the next page

## fix
made the code wait properly at three points:
1. **In `addAccount`**: Now waits for user data to be fetched before continuing (was not waiting before)
2. **Added retry logic**: If user data fetch fails, automatically retries 3 times with delays (1s, 2s, 3s)
3. **Before redirect**: Verifies user data is ready before sending user to home page

this ensures cookies and user data are ready before moving forward, fixing the issue for external passkey manager users.

### DIFF
--- a/src/components/Setup/Views/SignTestTransaction.tsx
+++ b/src/components/Setup/Views/SignTestTransaction.tsx
@@ -127,8 +127,12 @@ const SignTestTransaction = () => {
                     dispatch(setupActions.setLoading(false))
                     return
                 }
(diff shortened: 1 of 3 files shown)


Hugo0:
@coderabbitai review in depth

PR #1520, opened 2025-12-01
Fix: Incorrect exchange price on Input field and withdraw review page

Zishan-7:
Issue: Manteca price api can have different prices from the actual withdrawal rate, because USDC/USDT is not always 1:1

Solution - Always fetch USDC -> USTC price and then multiply that by USDC -> Desired manteca currency to get the accurate exchange rate


Zishan-7:

Evaluate:
- code quality and DRY

kushagrasarathe (approved):

PR #1532, opened 2025-12-04
fix: pass through error message

Hugo0:
NextJs strips server action error logs. THis is a fix to show the actual error in FE


Hugo0:
@coderabbitai review in depth. Espcially compare to #1531

PR #1550, opened 2025-12-11
[TASK-17156] add dynamic address formatting according to width of the container

Zishan-7:
iphone SE
image

Iphone 12 pro
image

Desktop
image

### DIFF
--- a/src/hooks/useAutoTruncatedAddress.ts
+++ b/src/hooks/useAutoTruncatedAddress.ts
@@ -1,0 +1,197 @@
+import { useEffect, useState, useRef, useCallback } from 'react'
+
+/** Safety margin multiplier to account for font rendering differences between canvas and DOM */
+const SAFETY_MARGIN = 0.9
+
+/**
+ * Calculates the optimal truncation for an address to fit within a container width.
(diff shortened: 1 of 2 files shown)


src/hooks/useAutoTruncatedAddress.ts
```diff
@@ -0,0 +1,199 @@
...
+ * Returns the formatted address like "0xF3...3he7e" with dynamic character counts.
+ *
+ * Shows as many characters as fit in the container, never causing overflow.
+ *
+ * @param address - The full address to truncate
+ * @param containerWidth - Available width in pixels
+ * @param charWidth - Estimated width per character in pixels (default: 8)
+ * @returns Formatted address string
+ */
+export const formatAddressToFitWidth = (address: string, containerWidth: number, charWidth: number = 8): string => {
```
kushagrasarathe, line 17: suggestion: can have a string validation check here to see if address is string or not

kushagrasarathe, line 17: charWidth default value can be a const instead of hardcoding 8

kushagrasarathe (approved):

PR #1553, opened 2025-12-12
[TASK-16880] fix: request pot redirect on signup

Zishan-7:


src/components/Setup/Views/SignTestTransaction.tsx
```diff
@@ -23,6 +23,22 @@ const SignTestTransaction = () => {
     const [error, setError] = useState<string | null>(null)
     const [isSigning, setIsSigning] = useState(false)
     const [testTransactionCompleted, setTestTransactionCompleted] = useState(false)
+    const searchParams = useSearchParams()
+
+    const handleRedirect = () => {
```
kushagrasarathe: blocking: suggestion, use `useAccountSetup` instead of having it in this screen, that hook already handles redirects iirc

kushagrasarathe (changes requested):
check comment

kushagrasarathe (approved):

PR #1568, opened 2025-12-17
[wip] feat: send flow isolated from payment from 

kushagrasarathe:


src/features/payments/flows/send/views/SendInputView.tsx
```diff
@@ -0,0 +1,117 @@
...
+    // handle back navigation
+    const handleGoBack = () => {
+        if (window.history.length > 1) {
+            router.back()
+        } else {
+            router.push('/')
+        }
+    }
+
+    // determine button text and state
```
Hugo0, line 53: thought: these "states" should be reusable across all Balance Decreasing flows. If we don't reuse them, we have duplication, which is more maintenance and more bugs.

Is there any way to consolidate?

src/features/payments/flows/send/views/SendInputView.tsx
```diff
@@ -0,0 +1,117 @@
...
+    const isLoggedIn = !!user?.user?.userId
+
+    // handle submit - directly execute payment
+    const handleSubmit = () => {
+        if (canProceed && hasSufficientBalance && !isLoading) {
+            executePayment()
+        }
+    }
+
+    // handle back navigation
```
Hugo0, line 44: nice fn. Could be abstracted across flows

PR #1573, opened 2025-12-18
Feat: Add account recovery page

Zishan-7:
image

image

image

image


kushagrasarathe (approved):
lgtm, only thing, check #1579 added a new way to handle icons if you think this pr can be wait for merging until monday then i can sort out the icon used in this one and we can merge both

PR #1586, opened 2025-12-23
fix: qr-scan kyc state bug for new users

kushagrasarathe:
- fixes TASK-17671


Hugo0:
@coderabbitai review in depth

PR #1627, opened 2026-01-20
fix: add automatic cookie migration from sameSite strict to lax

Hugo0:
## Summary

Adds automatic cookie migration mechanism to prevent users from being stuck when deploying the `sameSite` fix to production.

**Context:** PR #1618 introduced `sameSite: 'strict'` which broke authentication. While we've fixed it to use `sameSite: 'lax'`, existing users with old cookies would be stuck unable to authenticate after deployment.

## How It Works

1. Created `refreshJWTCookieIfNeeded()` utility function
2. Added to all Next.js API routes that read `jwt-token` cookies  
3. When a user makes a successful authenticated request, their old `strict` cookie automatically gets refreshed to `lax`
4. Migration happens transparently without any user intervention

## Migration Path

Users with `sameSite=strict` cookies will be auto-migrated when they:
- Navigate directly to the site (typing URL, bookmark)
- Use same-site navigation
- Make any authenticated API call through Next.js routes

This covers the majority of user flows and ensures smooth transition to production.

## Routes Updated

- ✅ `get-user-from-cookie`
- ✅ `get-decoded-token`
- ✅ `update-user`
- ✅ `submit-profile-photo`
- ✅ `add-account`

## Security

No security impact - this simply re-sets the same JWT with correct cookie attributes. The migration from `strict` → `lax` is itself a security fix (lax is OWASP-recommended).

## Cleanup

This code can be safely removed after ~30 days post-deploy once all users have been migrated.

## Test Plan

- [x] Test with existing `sameSite=strict` cookie → gets auto-refreshed to `lax`
- [x] Test authentication flows still work
- [x] Verify no linter errors

@coderabbitai review

### DIFF
--- a/src/app/actions/offramp.ts
+++ b/src/app/actions/offramp.ts
@@ -1,8 +1,8 @@
 'use server'
 
-import { cookies } from 'next/headers'
 import { type TCreateOfframpRequest } from '../../services/services.types'
 import { fetchWithSentry } from '@/utils/sentry.utils'
+import { getJWTCookie } from '@/utils/cookie-migration.utils'
 
 const API_KEY = process.env.PEANUT_API_KEY!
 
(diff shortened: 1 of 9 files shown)


Hugo0:
@coderabbitai review

Hugo0:
@coderabbitai review latest changes

- exhaustiveness
- best practice
- edge cases and bugs
- DRY

PR #1648, opened 2026-01-27
[TASK-18201] fix: use correct timestamp for Bridge KYC history entry

jjramirezn:
Bridge KYC was using bridgeKycStartedAt for sorting, which could be recent if the user tested another KYC flow. Now uses the appropriate timestamp based on status:
- approved: bridgeKycApprovedAt
- rejected: bridgeKycRejectedAt
- pending: bridgeKycStartedAt

Applied to both HistoryPage and HomeHistory components.


jjramirezn:

Evaluate:
- code quality and DRY

PR #1674, opened 2026-02-09
🐛 fix: clear expired jwt cookie and sw cache on 401

Hugo0:
## Summary
- On 401 from `/get-user`, clear the jwt cookie and send `Clear-Site-Data: "cache"` header
- This nukes the stale cookie and service worker cache so the client redirects to `/setup` for re-authentication

## Context
Users with expired JWTs (30-day expiry) or corrupted cookies (from the Jan 20 httpOnly/sameSite changes) get stuck because the browser keeps sending the invalid cookie. This fix ensures the client can recover automatically.

## Risks
- `Clear-Site-Data: "cache"` will clear the SW cache on 401, forcing a fresh download of app assets on next load. This is intentional — stale cached code was part of the problem.

## QA
- User with expired JWT should be redirected to /setup instead of infinite spinner
- After re-authenticating, app should work normally

@coderabbitai review

---

> [!NOTE]
> **Medium Risk**
> Changes auth failure handling in a user bootstrap endpoint and adds cache-clearing headers, which can impact session behavior and force asset re-downloads, but is scoped to `401` responses only.
> 
> **Overview**
> When `/api/peanut/user/get-user-from-cookie` receives a non-`200` response, it now conditionally adds recovery headers.
> 
> On `401`, the route clears the `jwt-token` cookie via `Set-Cookie` and instructs the browser to clear cached assets via `Clear-Site-Data: "cache"`, so clients stuck with expired/invalid tokens or stale service-worker code can re-authenticate cleanly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 578071be31b0ba22741e9362109d2227320bdae2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #1675, opened 2026-02-09
🐛 fix: clear stale redux user on auth failure

Hugo0:
## Summary
- Always validate user against backend on mount (remove redux gate from `enabled`)
- Use `placeholderData` instead of `initialData` (show cached UI instantly, but don't skip fetch)
- Clear redux user on fetch failure so layout redirect to `/setup` triggers

## Problem
Redux persists user data to localStorage. On app boot, the query had `enabled: !authUser?.user.userId` — if Redux had a user, the query never fired. Expired JWTs were never detected. App rendered with stale cached data while all API calls failed.

## Risks
- One extra `/get-user-from-cookie` call on mount (skipped if data < 5min stale)
- No visual change for healthy sessions (placeholderData shows cached UI while validating)

## QA
- User with expired JWT should redirect to /setup after app boot
- User with valid JWT should see no difference (instant load, no flicker)

### DIFF
--- a/src/hooks/query/user.ts
+++ b/src/hooks/query/user.ts
@@ -3,7 +3,7 @@ import { useAppDispatch, useUserStore } from '@/redux/hooks'
 import { userActions } from '@/redux/slices/user-slice'
 import { fetchWithSentry } from '@/utils/sentry.utils'
 import { hitUserMetric } from '@/utils/metrics.utils'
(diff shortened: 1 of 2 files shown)


Hugo0:
@coderabbitai review and compare to changes in #1674 and #1672

PR #1681, opened 2026-02-17
Merge production into staging (sync 130)

Hugo0:
## Summary

Syncs `peanut-wallet` (production) → `peanut-wallet-dev` (staging). One conflict in `src/hooks/query/user.ts` — resolved by combining production's critical bugfixes with dev's improvements. Also fixes two pre-existing bugs and adds retry resilience to kernel client initialization.

### Conflict resolution: `src/hooks/query/user.ts`

Both branches independently modified `useUserQuery`:

- **Production** (commit `bf43ef7`, Feb 9): Hotfixed a stale-JWT bug
- **Dev** (commit `2a9ca85`, Jan 29): Added `BackendError` class + retry logic for Card Pioneers

| Aspect | Taken from | Why |
| --- | --- | --- |
| `BackendError` class + smart retry | **Dev** | Good improvement — retry 5xx, don't retry 4xx |
| `enabled: dependsOn` (no Redux gate) | **Production** | **Critical fix** — the old `enabled: !authUser?.user.userId` permanently disabled the query after first success, making `refetchOnMount`, `refetchOnWindowFocus`, and `staleTime` all dead code. Expired JWTs were never detected. |
| `placeholderData: authUser` (not `initialData`) | **Production** | **Critical fix** — `initialData` tells TanStack "this IS real data, start in success state". Combined with the Redux gate, the query never fetched. `placeholderData` shows cached data instantly but still validates. |
| `dispatch(setUser(null))` on 4xx | **Production** | **Critical fix** — clears stale Redux so layout redirects to `/setup` |
| Remove `keepPreviousData` import | **Production** | No longer needed |

### Bug fix: `isKernelClientReady` always set to `true`

**File:** `src/context/kernelClient.context.tsx`

Previously, `setIsKernelClientReady(true)` was dispatched regardless of whether any kernel client actually initialized. If passkey was corrupted or ZeroDev was down, all chain clients would fail but the user would see a fully authenticated UI where every transaction throws "No client found for chain" with no recovery path.

**Fix:** Only set `isKernelClientReady(true)` if the primary chain (PEANUT_WALLET_CHAIN) client was created. If it fails, retries up to 3 times with exponential backoff (1s, 2s) before forcing logout — prevents unnecessary logouts during transient ZeroDev outages.

### Bug fix: Transaction query key concatenation

**File:** `src/hooks/useTransactionHistory.ts`

`TRANSACTIONS + username` created a single string like `'transactionshugo'` as the query key. `invalidateQueries({ queryKey: [TRANSACTIONS] })` prefix-matches on `'transactions'` but string equality against `'transactionshugo'` fails — so filtered transaction queries were never properly invalidated. After sending money to a contact, the mutual transaction list could show stale data.

**Fix:** Changed to proper array key `[TRANSACTIONS, 'latest', { limit, targetUsername }]` so prefix invalidation works correctly.

### New: `retryAsync` shared utility

**File:** `src/utils/retry.utils.ts`

Added a generic async retry wrapper with exponential backoff, reusing the existing `createExponentialBackoff` helper. For imperative async code outside React Query (kernel client init, future use cases).

### Auto-merged production changes (no conflicts)

- **layout.tsx**: Remove redundant `hasToken` guard, add `isRedirecting` ref with hard-nav fallback
- **user-slice.ts**: `PayloadAction<IUserProfile | null>` for `setUser(null)` support
- **get-user-from-cookie/route.ts**: Clear JWT cookie + `Clear-Site-Data: "cache"` on 401
- **qr-pay**: PIX-specific error message for unsupported QR codes
- **ExchangeRate**: Show local currency conversion amount on withdraw
- **FAQs**: Auto-linkify URLs in FAQ answers
- **PaymentInfoRow**: Refactor tooltip to shared `<Tooltip>` component

### Known issues not addressed (TODO)

- [ ] **Rules of Hooks violation in `useTransactionHistory`** — conditional `useQuery`/`useInfiniteQuery` based on `mode` prop. If `mode` changes between renders → React crash. Currently safe because callers always pass a static `mode`, but it's a landmine for future refactors.
- [ ] **Exchange rate silently returns `'1'` on API error** (`useGetExchangeRate.tsx`) — non-USD users could see incorrect rates. Needs design decision on error UX.
- [ ] **Settings page `logoutUser()` called without `await`** + redundant `invalidateQueries()` — causes wasted failed refetches during logout. Harmless due to hard navigation that follows.
- [ ] **Side effects in `queryFn`** — `hitUserMetric('login')` fires on every refetch, inflating login metrics. Redux dispatch in queryFn creates dual source of truth. Architectural concern for later.
- [ ] **SW "two-tier caching" is not implemented** — `sw.ts` uses `NetworkOnly` for all routes. No SW caching exists. Cache cleanup on logout targets caches that never exist.

## Test plan

- [ ] `npm run build` passes (verified — only pre-existing type errors in `/dev/ds/playground/`)
- [ ] Clear cookies → visit /home → redirects to /setup (not infinite loading)
- [ ] Login → wait 6+ min → tab away/back → revalidates against backend
- [ ] Login → kill backend → reload → shows BackendErrorScreen
- [ ] Login → clear localStorage → reload → forces logout to /setup
- [ ] Public routes (/claim, /pay/*) work without auth
- [ ] Send money to contact → transaction history refreshes correctly (query key fix)
- [ ] Simulate ZeroDev outage → kernel client retries before logout (retry resilience)

### DIFF
--- a/src/app/(mobile-ui)/layout.tsx
+++ b/src/app/(mobile-ui)/layout.tsx
@@ -7,7 +7,6 @@ import WalletNavigation from '@/components/Global/WalletNavigation'
 import OfflineScreen from '@/components/Global/OfflineScreen'
 import { ThemeProvider } from '@/config'
 import { useAuth } from '@/context/authContext'
-import { hasValidJwtToken } from '@/utils/auth'
 import classNames from 'classnames'
 import { usePathname } from 'next/navigation'
 import { useCallback, useEffect, useRef, useState } from 'react'
@@ -32,7 +31,6 @@ const Layout = ({ children }: { children: React.ReactNode }) => {
 
     const { isFetchingUser, user } = useAuth()
     const [isReady, setIsReady] = useState(false)
-    const [hasToken, setHasToken] = useState(false)
     const isUserLoggedIn = !!user?.user.userId || false
     const isHome = pathName === '/home'
     const isHistory = pathName === '/history'
@@ -48,9 +46,6 @@ const Layout = ({ children }: { children: React.ReactNode }) => {
     const scrollableContentRef = useRef<Element | null>(null)
 
     useEffect(() => {
-        // check for JWT token
-        setHasToken(hasValidJwtToken())
-
         setIsReady(true)
     }, [])
 
@@ -80,9 +75,17 @@ const Layout = ({ children }: { children: React.ReactNode }) => {
     // enable pull-to-refresh for both ios and android
     usePullToRefresh({ shouldPullToRefresh })
 
+    const isRedirecting = useRef(false)
+
     useEffect(() => {
-        if (!isPublicPath && isReady && !isFetchingUser && !user) {
-            router.push('/setup')
+        if (!isPublicPath && isReady && !isFetchingUser && !user && !isRedirecting.current) {
+            isRedirecting.current = true
+            router.replace('/setup')
+            // hard navigation fallback in case soft navigation silently fails
+            const fallback = setTimeout(() => {
+                window.location.replace('/setup')
+            }, 3000)
+            return () => clearTimeout(fallback)
         }
     }, [user, isFetchingUser, isReady, isPublicPath, router])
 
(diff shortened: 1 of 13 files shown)


Hugo0:
@coderabbitai review with focus on auth changes. Any potential bugs or edge cases introduced?

Hugo0:
@coderabbitai review with focus on auth changes. Any potential bugs or edge cases introduced?

PR #1697, opened 2026-02-26
Fix PIX key validation failing when pasted with whitespace

jjramirezn:
Strip all whitespace from PIX keys at the input level in the withdraw flow so spaces from copy-paste don't cause Manteca API errors.


jjramirezn:

Evaluate:
- code quality and DRY

PR #1715, opened 2026-03-06
[TASK-18633] fix: setup flow bugs - install copy, notification stuck state, PWA loop

jjramirezn:
1. Install screen: Replace tech-speak fallback with user-friendly copy
2. Waitlist notification: Add 'Not now' skip so users without invite codes don't get stuck on 'Enable Notifications' forever
3. PWA install loop: Remove target=_blank /setup link that causes redirect loop between browser and PWA; show instruction to open from Home Screen with fallback 'Continue here instead' button


jjramirezn:

Evaluate:
- code quality and DRY

PR #1890, opened 2026-04-15
fix: apply Peanut dev fee to ExchangeRateWidget destination amount

Hugo0:
## Summary

Follow-up to #1889. The landing-page `ExchangeRateWidget` calculator was showing "Recipient Gets" amounts using the raw exchange rate without deducting Peanut's 0.5% developer fee. Users planning a deposit based on the widget would receive less than shown.

## Scope

- `src/components/Global/ExchangeRateWidget/index.tsx` only
- Cross-currency pairs (EUR/GBP/MXN ↔ other non-USD). USD pairs pass through unchanged.

## Fix

Apply `applyBridgeCrossCurrencyFee()` (shared helper from #1889) to the destination amount before display. Reuses the same constant and logic as the onramp/offramp quote fix.

Bidirectional editing handled with an inverse helper (`reverseBridgeCrossCurrencyFee`): when the user types a net "Recipient Gets" value, we gross it up before computing the source amount, so source = (net / (1 - fee)) / rate.

Swap-currencies button uses the displayed (net) amount as the new source so the post-swap value matches what the user saw.

## Test plan

- [ ] Landing page widget: USD → EUR shows net amount (rate × 0.995)
- [ ] Landing page widget: EUR → MXN shows net amount (rate × 0.995)
- [ ] USD → USDC unchanged (no fee applied)
- [ ] Bidirectional editing stays consistent (gross/net math reverses correctly)

### DIFF
--- a/src/components/Global/ExchangeRateWidget/index.tsx
+++ b/src/components/Global/ExchangeRateWidget/index.tsx
@@ -1,12 +1,29 @@
 import CurrencySelect from '@/components/LandingPage/CurrencySelect'
 import countryCurrencyMappings from '@/constants/countryCurrencyMapping'
+import { BRIDGE_DEVELOPER_FEE_RATE } from '@/constants/payment.consts'
 import { useDebounce } from '@/hooks/useDebounce'
 import { useExchangeRate } from '@/hooks/useExchangeRate'
+import { applyBridgeCrossCurrencyFee } from '@/utils/bridge.utils'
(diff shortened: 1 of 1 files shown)


Hugo0:
Consolidated into #1889. The widget commit has been cherry-picked onto `fix/bridge-fee-display-quote` so all fee-display fixes ship as one hotfix PR. Closing this stacked PR.

PR #1967, opened 2026-05-11
feat(card): push notifications for issued, spend settled, declined, rebalance

jjramirezn:
Pairs with **peanut-api-ts#<push-notifications PR>**. BE drives all four push triggers + the synthetic decline category; this FE PR teaches the receipt drawer to read the new category so the in-app copy matches the push body.

## Why

The card lifecycle has zero push notifications today (other than KYC + 3DS). Users get card-issued events, spend confirmations, declines, and auto-rebalance completions surfaced after the fact — or not at all. This PR (paired with the BE PR) closes that gap and adds nuanced copy for the most-confusing decline case.

## Decline category — the meaty bit

Rain returns \`INSUFFICIENT_FUNDS\` for **both** genuine shortfalls AND limit-too-low cases. A real user hit this: had funds in their smart wallet, but the per-tx \`cardLimit\` was below the spend, the auto-rebalancer didn't have headroom to top up, and Rain reported it as \"insufficient funds\". The misleading copy sent the user to top up funds they already had.

BE now computes a synthetic \`declineCategory\` (\`'limit_too_low' | 'insufficient_balance' | 'other'\`) by comparing the declined amount against the card's \`cardLimit\` AND keying off Rain's raw reason. The category gates both the push body and the receipt drawer copy.

## Files

- \`src/utils/cardDeclineReason.ts\` — \`friendlyDeclineReason(code, category?)\`: prefer category when present, fall back to the raw-code map.
- \`src/components/TransactionDetails/transactionTransformer.ts\` — surface \`declineCategory\` on the typed \`extraDataForDrawer.cardPayment\`.
- \`src/components/TransactionDetails/provider-rows/CardPaymentRows.tsx\` — read either field for visibility + copy.

## Test plan

- [ ] Mock a card-spend intent with \`metadata.declineCategory: 'limit_too_low'\` — receipt drawer shows \"Card limit reached — increase your limit\" regardless of raw \`declineReason\`.
- [ ] Same with \`'insufficient_balance'\` → \"Insufficient balance\".
- [ ] Legacy intent with only \`declineReason: 'INSUFFICIENT_FUNDS'\` (no category) → falls back to \"Insufficient balance\" via the raw-code map.
- [ ] Blocked merchant (\`declineReason: 'blocked_merchant'\`, no category) → \"This merchant isn't supported\" via raw fallback.
- [ ] On staging, trigger a real card decline with a low \`cardLimit\`, confirm the push notification body matches the receipt drawer copy.

## Cross-repo deploy order

BE first (sets the synthetic category + sends the push). Then FE (reads the category). Both are backwards-compatible: legacy intents without the category still render via the raw-code fallback.

### DIFF
--- a/src/components/TransactionDetails/provider-rows/CardPaymentRows.tsx
+++ b/src/components/TransactionDetails/provider-rows/CardPaymentRows.tsx
@@ -41,7 +41,7 @@ export function hasCardPaymentRowsContent(transaction: TransactionDetails): bool
     if (nonBlank(card.merchantCategory)) return true
     if (nonBlank(card.merchantCity) || nonBlank(card.merchantCountry)) return true
     if (card.settlementAdjusted && parseCents(card.authAmount) != null) return true
-    if (transaction.status === 'failed' && card.declineReason) return true
+    if (transaction.status === 'failed' && (card.declineReason || card.declineCategory)) return true
     if (card.cancellationReason === 'auto_closed') return true
     return false
 }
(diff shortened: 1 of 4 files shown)


jjramirezn:

Evaluate:
- code quality and DRY

PR #1988, opened 2026-05-12
feat(auth): consume sliding-refresh JWT from /get-user

Hugo0:
## Summary

Client half of the JWT sliding-refresh — kills the 30-day hard logout for active users.

When `/get-user` includes a refreshed `token` in its response (companion peanutprotocol/peanut-api-ts#751), swap it in via `setAuthToken` so the cookie/localStorage holds the new token going forward. Strip `token` from the payload before it reaches the user store so consumers never see auth state mixed into domain data.

## What changes

**`src/hooks/query/user.ts`** — `fetchUser`:
```ts
const payload: (IUserProfile & { token?: string }) | null = await userResponse.json()
if (payload?.token) {
    setAuthToken(payload.token)
    delete payload.token
}
// ... existing 401/404 cleanup, redux dispatch, etc.
```

That's it for behavior. 14 lines added, 5 deleted.

## Tests

`src/hooks/query/__tests__/user.test.tsx` — 5 tests, all passing:

- calls `setAuthToken` when the response includes a refreshed token
- does NOT call `setAuthToken` when the response has no `token` field
- strips `token` from the data exposed to consumers
- clears the token on 401 (dead JWT) without touching `setAuthToken`
- clears the token on 404 (user no longer exists)

## Forward compatibility

If this lands before the backend PR, nothing breaks — the API just doesn't include `token`, the `if (payload?.token)` branch is skipped, behavior is identical to today.

If the backend lands first, behavior is also unchanged until the UI updates — backend just sends an extra JSON field that the old UI ignores.

So merge order doesn't matter.

## Verifiability

- [x] `npm run typecheck` — clean
- [x] `npm test` — 1034 passed, 4 skipped, 0 failed (47 suites)
- [x] Hook test mocks `apiFetch` + `setAuthToken`/`clearAuthToken`; asserts call counts + arguments + that `token` is stripped from `result.current.data`

## Risk

- **Low.** Strictly additive — the change only fires when the API ships the new field. Existing flow unchanged.
- `setAuthToken` is the same call site `auth-token.ts` already uses on login. No new storage code paths.

### DIFF
--- a/src/hooks/query/__tests__/user.test.tsx
+++ b/src/hooks/query/__tests__/user.test.tsx
@@ -78,6 +78,21 @@ describe('useUserQuery — JWT sliding refresh', () => {
         expect(result.current.data).not.toHaveProperty('token')
     })
 
+    it.each([
+        ['empty string', ''],
+        ['null', null],
+    ])('strips falsy token (%s) without calling setAuthToken', async (_label, falsyToken) => {
+        mockApiFetch.mockResolvedValueOnce(
+            mockResponse(200, { user: { userId: 'u1', username: 'alice' }, token: falsyToken })
+        )
+
+        const { result } = renderHook(() => useUserQuery(), { wrapper: makeWrapper() })
+        await waitFor(() => expect(result.current.isSuccess).toBe(true))
+
+        expect(result.current.data).not.toHaveProperty('token')
+        expect(mockSetAuthToken).not.toHaveBeenCalled()
+    })
+
     it('clears the token on 401 (dead JWT) without touching setAuthToken', async () => {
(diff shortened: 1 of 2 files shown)


Hugo0:
@coderabbitai applied (commit 00325edac):

**Fixed** — switched to always-strip-the-key, only-call-setAuthToken-when-truthy. The previous `if (payload?.token)` could leak `token: ''` or `token: null` into the user store. Added `it.each` cases for both falsy values.

Backend currently can't send a falsy token (the response only includes `token` when truthy), so the bug was theoretical — but the code is now obviously correct without having to reason about cross-repo invariants.

PR #768, opened 2025-03-25
[TASK-9940] perf: improve claim flow

jjramirezn:
Made the following changes to improve claim flow speed:

- There was a lot of time wasted on creating a provider from the sdk, now we use a viem client directly. Biggest gain here
- Sensible caching where applicable
- Token price is a server action instead of a route (small gain, but important in slow networks)

## Summary by CodeRabbit

- **Chores**
  - Upgraded a key blockchain integration dependency to improve overall performance and stability.

- **New Features**
  - Enhanced the claim process with improved validation and secure link handling.
  - Introduced real-time token pricing and details retrieval with caching for faster, more accurate market data.
  - Added streamlined blockchain client interactions for a more reliable transaction experience.
  - Implemented a new method for retrieving comprehensive token details.

- **Refactor**
  - Revised error handling and data retrieval workflows across claim, payment, and token details to ensure smoother user interactions.

PR #2018, opened 2026-05-13
fix(history): refuse to build /claim URL when send-link IDs are missing

Hugo0:
## Summary

- **Defensive guard** against the BE SEND_LINK extraData regression on `dev` (decomplexify `mapGenericIntent` dropped `contractVersion` + `depositIdx`).
- Without this guard, raw template interpolation produced literal `?v=undefined&i=undefined` query params. Sender clicked Cancel → SDK threw `No Peanut vault for chainId=42161 version=undefined`. Recipients hit the same broken URL on claim.
- The BE fix lives in https://github.com/peanutprotocol/peanut-api-ts/pull/769. This PR is the FE side of the same incident: even after the BE is fixed, this guard ensures a future regression can't silently leak `undefined` URLs.

## Behavior change

`completeHistoryEntry` SEND_LINK arm at `src/utils/history.utils.ts:294`:

| extraData | Before | After |
|---|---|---|
| `{contractVersion: 'v4.3', depositIdx: 7}` (well-formed) | `/claim?c=42161&v=v4.3&i=7#p=...` | `/claim?c=42161&v=v4.3&i=7#p=...` (unchanged) |
| `{depositIdx: 7}` (missing version) | `/claim?c=...&v=undefined&i=7#p=...` ❌ | `link=''` → drawer falls back to "use device where created" copy |
| `{contractVersion: 'v4.3'}` (missing idx) | `/claim?c=...&v=v4.3&i=undefined#p=...` ❌ | `link=''` → drawer falls back |

## Verify gap

The existing `src/utils/__tests__/history.utils.test.ts` mocked `getFromLocalStorage → null`, which skipped the `link` builder branch entirely — so even with deliberately-broken extraData the test passed. Extended the test to mock a password and assert link well-formedness directly.

## What's in this PR

1. **Guard** in `src/utils/history.utils.ts:293` — only build the URL when both `contractVersion` and `depositIdx` are non-nullish.
2. **3 new test cases** in `src/utils/__tests__/history.utils.test.ts` covering well-formed / missing-version / missing-idx. Verified the two new "missing" cases fail without the guard.

## Cross-repo

Pair with: https://github.com/peanutprotocol/peanut-api-ts/pull/769 (BE fix that restores the dropped fields).

## Test plan

- [x] `npx jest src/utils/__tests__/history.utils.test.ts` — 11/11 pass
- [x] Verified the 2 new "missing field" tests fail when the guard is reverted (stash + run)
- [ ] After merge + BE merge + staging deploy: create a fresh send-link on staging, copy URL, confirm `?v=X.X&i=N` (not `undefined`), confirm cancel succeeds

### DIFF
--- a/src/utils/__tests__/history.utils.test.ts
+++ b/src/utils/__tests__/history.utils.test.ts
@@ -16,8 +16,9 @@ import { completeHistoryEntry } from '../history.utils'
 import type { HistoryEntry } from '../history.utils'
 
 jest.mock('@/app/actions/currency', () => ({ getCurrencyPrice: jest.fn() }))
+const mockGetFromLocalStorage = jest.fn(() => null as string | null)
 jest.mock('@/utils/general.utils', () => ({
-    getFromLocalStorage: jest.fn(() => null),
+    getFromLocalStorage: (...args: unknown[]) => mockGetFromLocalStorage(...(args as [])),
     getTokenDetails: jest.fn(() => ({ symbol: 'USDC', decimals: 6 })),
 }))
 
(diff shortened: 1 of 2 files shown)


Hugo0:
Closing as superseded.

This PR was the FE half of an incident pair — defensive guard at the **write side** (refuse to build /claim URL when send-link IDs are missing). Two related PRs landed first and together cover the same bug class:

- **peanut-ui#2015** (merged 2026-05-13) — read-side fallback: `getParamsFromLink` substitutes `getLatestContractVersion` when `v` is missing/`'undefined'`/`'null'`. Handles broken URLs already in the wild.
- **peanut-api-ts#768** (merged 2026-05-13) — BE source fix: `mapGenericIntent` now propagates `contractVersion` + `depositIdx` to `extraData`, so the FE never receives missing fields again.

Together they make this PR's guard dead code in practice — it never triggers post-#768 because the BE always sends well-formed extraData. The 47 lines of tests it adds lock behavior that doesn't fire.

If anyone wants to keep a write-side assertion that "we never build a /claim URL with `v=undefined`," it's worth a tests-only follow-up that asserts the well-formed shape against the current code path. Flagging for post-Mega-Release backlog.

Paired BE PR peanutprotocol/peanut-api-ts#769 closed for the same reason.

PR #2098, opened 2026-05-25
fix(routes): reserve /card and other app paths from the [...recipient] catch-all

Hugo0:
## What was broken
`/card/asfafsafs` redirected to `/pay/...` with "Invalid recipient". Trace:

1. Next.js matches `/card` exactly via `(mobile-ui)/card/page.tsx`. No child route matched `asfafsafs`.
2. Fell through to the root `/[...recipient]/page.tsx` catch-all with `recipient = ["card", "asfafsafs"]`.
3. `isReservedRoute("/card")` returned **false** — `card` wasn't in `DEDICATED_ROUTES` — so the catch-all happily treated `card` as a Peanut username and pushed the user into the pay/request flow.

## What changed
Added the missing app paths to `DEDICATED_ROUTES` in `src/constants/routes.ts`. With these reserved, `isReservedRoute()` returns true on the first segment, the catch-all's `Page()` calls `notFound()`, and Next.js renders our 404 (the ragdoll).

```
+ card
+ badges
+ limits
+ notifications
+ pay-request
+ recover-funds
+ jobs
```

`pay-request` and `recover-funds` only had the bug for `/<name>/anything` since their direct route handles `/<name>` exactly. Same for the others — every one of them is a real route folder that was previously susceptible to "catch-all hijack" on any unknown sub-segment.

## Side effect
These names can no longer be claimed as Peanut usernames. Worth it — none are user identities, all are dedicated routes.

## Test plan
- [ ] `/card/asfafsafs` on a staging preview → 404 (ragdoll), not the pay/invalid-recipient screen.
- [ ] `/card` still loads the card hub.
- [ ] `/card/add-to-wallet`, `/card/pin`, `/card/physical`, `/card/limit` still load.
- [ ] `/badges/foo`, `/limits/foo/bar`, `/notifications/x`, `/pay-request/x`, `/recover-funds/x` → 404 (previously: catch-all hijack).
- [ ] Normal recipient flows still work — e.g. `/alice`, `/0x...`, `/alice/10USDC` continue to hit `[...recipient]` and render the payment page.


Hugo0:
Added `src/constants/__tests__/routes.test.ts` — walks `src/app/`, unwraps route groups, skips dynamic segments, asserts every page.*/route.*-bearing folder is in `DEDICATED_ROUTES`. Fails CI on the next missing entry.

Verified the walker output matches the seven additions in this PR:
```
add-money, api, badges, card, card-payment, careers, claim, crisp-proxy,
dev, exchange, history, home, invite, jobs, kyc, limits, lp, maintenance,
notifications, pay, pay-request, points, profile, qr, qr-pay, quests,
receipt, recover-funds, request, rewards, send, settings, setup, withdraw
```
All present in `DEDICATED_ROUTES` after this PR; `missing: []`.

PR #2108, opened 2026-05-26
feat(qr-pay): live card-vs-local-rail savings for ARS + BRL

0xkkonrad:
## Summary

A single live source for "how much does a foreign card cost vs Peanut's local rail", shared by **every** card-comparison surface. Replaces the hardcoded 9.13% — the BCRA-vs-MEP spread fluctuates ~5–25% in practice, so the launch-day snapshot was wrong most days.

- New server action [`getCardMarkupRate`](src/app/actions/card-comparison.ts) — **single source of truth**:
  - **ARS live** via `dolarapi.com/v1/dolares/oficial` (free, public, the de facto AR FX source). Markup = `mantecaPrice / (bcraOfficial × 0.97) − 1`. The 3% factor is a typical issuer FX fee on top of the network rate. Cached 5 min server-side + 5 min client via React Query.
  - **BRL static** at 7% (IOF 3.5% + issuer ~3%). Statutory + slow-moving, no live source needed. Same hook surface so swapping in a live feed later is trivial.
  - Falls back to the static constant if the live fetch fails — payment flow never blocks on a third-party feed.
  - **Self-sufficient**: callers without a Manteca rate (e.g. LocalRailNudge) just pass a currency code; the action fetches Manteca itself.

- Wires **three** surfaces to the same hook so they can't drift:
  1. QR pay **confirm** screen — new `"Save vs card ~$X"` row between Exchange Rate and Peanut fee (this surface previously had no comparison at all).
  2. QR pay **success** screen — existing "saved ~\$X compared to card!" message, now live and extended to BRL (was ARS-only).
  3. **LocalRailNudge** on Rain-card-spend receipts — was `"around 9% less"` hardcoded (anchored to the old 9.13% in `calculateSavingsInCents` per its commit message); now `"around {N}% less"` where N comes from the same hook. Loose "around" copy stays — it's a post-hoc educational nudge, not a precision moment.

- `calculateSavingsInCents` now takes a markup *rate* instead of a currency code — decoupled from any static table; the table is fallback + eligibility gate only.

## Why this shape

Juan José's commit message for [LocalRailNudge](src/components/TransactionDetails/provider-rows/LocalRailNudge.tsx) (5333032, 2026-05-21) is explicit:

> The ~9% figure is the same delta `calculateSavingsInCents` already uses (9.13% vs card).

That intent — **one figure across surfaces** — was load-bearing. This PR preserves it while fixing the staleness. A first revision of this PR went live-rate on the QR-pay surfaces but left LocalRailNudge unchanged, which would have created drift between two surfaces that were meant to agree. Fixed.

## Honest framings

- Manteca's ARS price has its own spread baked in. The live `mantecaPrice / (bcraOfficial × 0.97) − 1` calc reflects "what Peanut actually delivers vs what a card delivers", **not** a pure BCRA-vs-MEP gap. That's the right thing to display (it's what the user gets), but the variable name shouldn't oversell as a pure spread.
- 3% issuer fee is a constant in [`card-comparison.ts:10`](src/app/actions/card-comparison.ts) (`ISSUER_FX_FEE`). Same drift class as the 9.13% it replaces — just moved. Product/marketing should know it lives here.
- BRL doesn't really gain from the dynamic path today (returns the static value). The infrastructure pays for itself the day IOF steps down or we wire in a live BRL source. Until then it's "free" cost (~10 LOC).

## Test plan

- [x] `npm test` — new `card-comparison.test.ts` (10 cases, covering eligibility, live ARS, static BRL, all dolarapi failure modes, the no-Manteca-price branch, and inverted-spread guards)
- [x] `npm test` — `qr-pay-states.test.tsx` (22/22 active, no regressions)
- [x] `npm test` — `LocalRailNudge.test.tsx` refreshed (live-rate path tested, static fallbacks updated to per-currency: AR 9%, BR 7%)
- [x] `tsc --noEmit` — clean on the touched files
- [ ] Manual: scan a MercadoPago/QR3 ARS QR on the preview, verify "Save vs card" amount tracks today's BCRA-vs-MEP (cross-check against `dolarapi.com/v1/dolares/oficial`)
- [ ] Manual: scan a PIX BRL QR, verify ~7% of USD value
- [ ] Manual: open a past Rain card-spend receipt in AR/BR, verify the nudge percentage now matches the QR-pay surface (no drift)

### DIFF
--- a/src/app/(mobile-ui)/qr-pay/__tests__/qr-pay-states.test.tsx
+++ b/src/app/(mobile-ui)/qr-pay/__tests__/qr-pay-states.test.tsx
@@ -153,6 +153,10 @@ jest.mock('@/app/actions/currency', () => ({
     getCurrencyPrice: jest.fn(() => Promise.resolve({ sell: 1200, buy: 1250 })),
 }))
 
+jest.mock('@/hooks/useCardMarkupRate', () => ({
+    useCardMarkupRate: jest.fn(() => ({ data: null })),
+}))
+
 jest.mock('@/app/actions/increase-limits', () => ({
     initiateIncreaseLimits: jest.fn(),
 }))
@@ -271,6 +275,7 @@ jest.mock('@/utils/perk.utils', () => ({
 
 jest.mock('@/utils/qr-payment.utils', () => ({
     calculateSavingsInCents: jest.fn(() => 0),
+    hasCardMarkupComparison: jest.fn(() => false),
     isArgentinaMantecaQrPayment: jest.fn(() => false),
     getSavingsMessage: jest.fn(() => ''),
 }))
@@ -1073,12 +1078,12 @@ describe('GROUP 4: Success States', () => {
 
     test('Argentina QR3 success shows savings message', async () => {
         const {
-            isArgentinaMantecaQrPayment,
+            hasCardMarkupComparison,
             calculateSavingsInCents,
             getSavingsMessage,
         } = require('@/utils/qr-payment.utils')
 
-        isArgentinaMantecaQrPayment.mockReturnValue(true)
+        hasCardMarkupComparison.mockReturnValue(true)
         calculateSavingsInCents.mockReturnValue(150)
         getSavingsMessage.mockReturnValue('You saved $1.50 vs card!')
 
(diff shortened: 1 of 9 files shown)


0xkkonrad:
Bundled into #2109 — same currency-adjacent UX work, single Vercel preview and review surface.

0xkkonrad:
Reopening — splitting back out of #2109 so card-markup and the landing scroll fix get reviewed independently. Scroll fix is now in #2116.

jjramirezn (approved):

PR #2109, opened 2026-05-26
fix(landing): stop currency widget from scrolling to top

0xkkonrad:
## Summary
- Clicking a currency in the landing-page exchange widget would sometimes scroll the user back to the top.
- Root cause: `router.replace('?...', { scroll: false })` is flaky in Next.js 16 when there's a `<Suspense>` boundary above — search-param changes occasionally bubble a scroll reset through it. The landing page has Suspense at `src/app/page.tsx:38`, which is why the bug surfaced there.
- Fix: switch both `Global/ExchangeRateWidget` and its MDX wrapper `Marketing/mdx/ExchangeWidget` to `nuqs` with `shallow: true, history: 'replace', scroll: false`. nuqs goes through `window.history.replaceState` and never re-enters the Next.js router. Also brings these two files in line with the "use nuqs, never URLSearchParams" rule in CLAUDE.md.

## Test plan
- [ ] On `/` (and any SEO page that embeds `<ExchangeWidget destinationCurrency="ARS" />`), scroll down to the exchange widget and open either currency picker.
- [ ] Pick a different currency — the page should stay put, not jump to the hero.
- [ ] Swap currencies via the swap button — same: no scroll.
- [ ] Type a new source amount — no scroll on debounce.
- [ ] Reload with `?from=BRL&to=USD&amount=250` and confirm the widget restores those values.
- [ ] On an SEO page (e.g. `/en/send-to/argentina`), confirm the widget still seeds `to=ARS` on first mount when no query string is present.

### DIFF
--- a/src/app/(mobile-ui)/qr-pay/__tests__/qr-pay-states.test.tsx
+++ b/src/app/(mobile-ui)/qr-pay/__tests__/qr-pay-states.test.tsx
@@ -153,6 +153,10 @@ jest.mock('@/app/actions/currency', () => ({
     getCurrencyPrice: jest.fn(() => Promise.resolve({ sell: 1200, buy: 1250 })),
 }))
 
+jest.mock('@/hooks/useCardMarkupRate', () => ({
+    useCardMarkupRate: jest.fn(() => ({ data: null })),
+}))
+
 jest.mock('@/app/actions/increase-limits', () => ({
     initiateIncreaseLimits: jest.fn(),
 }))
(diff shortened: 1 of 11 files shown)


0xkkonrad:
Bundled in the QR-pay live-card-comparison work from #2108 (now closed) — two commits added: `1c1c51c9b` (live BCRA-driven savings for ARS, dynamic plumbing for BRL across QR pay confirm + success screens) and `06ae91ec0` (single live source incl. LocalRailNudge — aligned with Juan José's intent in 5333032).

The scroll-fix and card-comparison work are independent: different files, no overlap. Vercel preview will exercise both.

0xkkonrad:
Splitting back out — review feedback flagged the bundled scope (scroll fix + card-markup feature have different rollback domains). 

- Scroll fix: #2116
- Card-markup: #2108 (reopened)

Closing this one.

PR #906, opened 2025-06-12
[TASK-11578] feat: abstract squid route fetching

jjramirezn:
Stop using the skd and use the squid API directly, this give us more control and access to all the data that returns squid (for example, we now have access to the fees and don't have to recalculate them ourselves)

PR #2132, opened 2026-05-29
kyc(gate): consume wait:bridge NextAction + SEPA v2 ToS copy

Hugo0:
## Summary

Two FE-side follow-ups to peanut-api-ts#906 (BE Bridge requirement vocabulary fix) — neither feature is usable without that BE PR.

- **New `waiting-on-provider` gate kind**. BE now emits `{kind:'wait', key:'wait:bridge'}` when Bridge is doing internal review (\`kyc_approval\`, \`post_processing\`, generic \`stripe_database_lookup_*\`). Without a FE handler the wait-only rail fell through to \`needs-enrollment\` → user saw a misleading "verify your identity" CTA and got bounced back to Sumsub for nothing. Now: \`deriveGate\` returns \`waiting-on-provider\` with the BE-supplied "we're finalizing with Bridge" message, and the 4 consumer call sites suppress the KYC modal (treated like \`loading\`).
- **BridgeTosStep copy varies for SEPA v2**. New \`reasonCode\` prop; \`bridge_tos_v2_required\` → SEPA-specific copy ("Accept SEPA Terms of Service" / "...to enable EUR (SEPA) and GBP (Faster Payments)…"). \`bridge_tos_required\` (base) keeps existing copy. Endpoint call is unchanged (Bridge's \`tos_acceptance_link\` is opaque to endorsement — Bridge picks the right ToS server-side based on customer state).
- **Manteca \`handleOnboardingError\` intentional-fallback comment**. The Manteca hosted widget is dead-by-design per KYC 2.0 (centralized on Sumsub → \`submitToManteca\` via API), but the error-recovery redirect stays as a last-resort escape hatch for users stuck in incomplete Manteca state. Comment clarifies the intent so the next reader doesn't think it's a primary flow.

## Priority order change in deriveGate

\`waiting-on-provider\` slots between \`pending\` and \`needs-identity\` — AFTER \`ready\`, so a usable rail wins (don't block a user from a rail they can use), BEFORE \`needs-identity\` so the wait copy beats a misleading "verify" CTA.

\`\`\`
1. loading
2. blocked-rejection
3. accept-tos
4. fixable-rejection
5. ready              ← user has a usable rail → use it
6. pending            ← provisioning, also a wait but more concrete
7. waiting-on-provider ← NEW
8. needs-identity
9. needs-enrollment
\`\`\`

## Test plan

- [x] \`pnpm jest --testPathPattern='capability-gate'\` → 12/12 green (new file)
- [x] \`pnpm jest\` full suite → 1187 pass, 3 pre-existing content.test.ts failures (unrelated)
- [x] \`pnpm typecheck\` → clean (asset-import errors are Next.js worktree quirk, pre-existing)
- [x] \`pnpm prettier --check\` on all touched files → clean
- [ ] Manual: in sandbox, force a Bridge \`kyc_approval\` state on a user, hit \`/add-money/EU/bank\`, verify the wait copy renders instead of the verify-identity modal
- [ ] Manual: in sandbox, EU user mid-SEPA-ToS, verify the BridgeTosStep modal shows SEPA-specific title/description

## Risks

- Behavioral change: \`waiting-on-provider\` is new, so consumers' \`gate.kind !== 'ready'\` guards now have one more branch they suppress (same as \`loading\`). Wide-coverage Jest suite passes.
- Reason-prop addition on \`accept-tos\` GateState is additive (optional field). Existing consumers unaffected.

### DIFF
--- a/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
+++ b/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
@@ -213,9 +213,11 @@ export default function OnrampBankPage() {
         if (!validateAmount(rawTokenAmount)) return
 
         if (gate.kind !== 'ready') {
-            // capabilities still loading — silently no-op instead of flashing a
-            // needs_kyc modal on top of state we don't know yet.
-            if (gate.kind === 'loading') return
+            // capabilities still loading OR provider doing internal review —
+            // silently no-op instead of flashing a misleading needs_kyc modal.
+            // `waiting-on-provider` means the user has nothing to do; opening
+            // a KYC modal would imply otherwise.
+            if (gate.kind === 'loading' || gate.kind === 'waiting-on-provider') return
             if (gate.kind === 'accept-tos') {
                 guardWithTos()
             } else {
@@ -437,6 +439,7 @@ export default function OnrampBankPage() {
                         handleWarningConfirm()
                     }}
                     onSkip={hideTos}
+                    reasonCode={gate.kind === 'accept-tos' ? gate.reason?.code : undefined}
                 />
             </div>
         )
--- a/src/app/(mobile-ui)/withdraw/[country]/bank/page.tsx
+++ b/src/app/(mobile-ui)/withdraw/[country]/bank/page.tsx
@@ -178,7 +178,9 @@ export default function WithdrawBankPage() {
 
     const handleCreateAndInitiateOfframp = async () => {
         if (gate.kind !== 'ready') {
-            if (gate.kind === 'loading') return
+            // Loading and waiting-on-provider both mean "user has no action to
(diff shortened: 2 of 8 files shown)


Hugo0:
**Merge order:** depends on https://github.com/peanutprotocol/peanut-api-ts/pull/906 — the BE resolver PR that emits the new `kind:'wait'` NextAction (`wait:bridge` key) and the `bridge_tos_v2_required` reason code this FE consumes. Without #906 merged, this FE is functionally a no-op (no rail will ever return `waiting-on-provider` from `deriveGate`, and no `BridgeTosStep` will see `reasonCode === 'bridge_tos_v2_required'`). Merging this before #906 is safe (no regressions) but produces no observable user-facing change until #906 ships.

PR #2146, opened 2026-06-01
fix(card): gate share-asset row on "went through the flow" (part 2/2 FE)

Hugo0:
**Follow-up to #2144. Pairs with peanut-api-ts #927 (the BE half).**

## What

Two things, both in `HomeHistory` / the card-unlock gate:

1. **Broadens the gate** from #2144's *issued-card-only* to Hugo's actual intent — **"went through the flow"**: saw the celebration (`cardInfo.skipCelebrationSeen`, persisted by #927) **OR** holds a card. `deriveCardUnlockEntry`'s param is now `wentThroughFlow`. Stamps the write-once server marker (`cardApi.markCelebrationSeen()`) when the user dismisses the celebration.
2. **Re-applies the eslint cleanup** that was lost when #2144 was merged before its 2nd commit (`f1bd4be29`) landed — typed the badge synthetic row (`BadgeHistoryEntry` in the `entries` union, `b.id` guard), dropping two pre-existing `any`s in `HomeHistory.tsx`. Without this, the next PR touching the file fails eslint.

## Why

#2144's issued-card gate stopped the bleeding (3,230 bogus → 9) but is stricter than intended — it hides the row from users who went through the flow without a card. The truthful signal is the celebration-seen marker (#927 wires the long-deferred column).

## Safe deploy order / back-compat

`skipCelebrationSeen` is **optional** in `CardInfoResponse` — `undefined` until #927 ships, treated as "not seen". So:
- Merge/deploy **#927 (BE) first**; then this FE broadens automatically.
- If this merges first, it just behaves like today's issued-card gate (no regression) until the BE catches up.
- `markCelebrationSeen()` failures are swallowed (console.error) — a 404 before the BE ships won't break the celebration dismiss.

## Verify

- `cardUnlock.types.test.ts` updated (5 cases incl. the gabby access-only → null regression). All pass.
- eslint clean, changed files typecheck clean, prettier clean.
- Back-merge to `dev` after merge.

### DIFF
--- a/src/app/(mobile-ui)/card/page.tsx
+++ b/src/app/(mobile-ui)/card/page.tsx
@@ -500,9 +500,20 @@ const CardPage: FC = () => {
                         badgeCode={skipCode}
                         username={user?.user?.username ?? undefined}
                         badges={allBadges}
-                        onContinue={() => {
+                        onContinue={async () => {
+                            // localStorage drives the in-flow re-celebration gate;
+                            // setState transitions the screen immediately.
                             markSkipCelebrationSeen()
                             setSkipCelebrationSeen(true)
+                            // Persist the write-once server marker so the home
+                            // activity-feed share-asset row knows the user went
+                            // through the flow. Best-effort; await before refetch so
+                            // cardInfo.skipCelebrationSeen comes back true.
+                            try {
+                                await cardApi.markCelebrationSeen()
+                            } catch (err) {
+                                console.error('[card] markCelebrationSeen failed:', err)
+                            }
                             invalidateOverview()
                             void refetchCardInfo()
                         }}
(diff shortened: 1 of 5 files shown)


Hugo0:
Closing as stale (open since 2026-06-01, ~28d; card/page.tsx + HomeHistory have churned heavily since → would conflict on rebase). This was the FE half (pairs with peanut-api-ts #927) of refining the card-unlock activity-feed gate from #2144's issued-card-only to 'went through the flow' via skipCelebrationSeen. #2144 already merged and cut the bogus rows 3,230 → 9, so this only chases the last ~9 + an eslint cleanup — not worth reviving two conflicting cross-repo PRs on launch eve. Not superseded by #2274 (that's the share-asset redesign, a different concern). Reopen if we still want the celebration-seen gate.

PR #2154, opened 2026-06-01
feat(payments): gate post-on-chain Retry on the 4 unprotected money-flows

Hugo0:
Follow-up to [peanut-ui #2147](https://github.com/peanutprotocol/peanut-ui/pull/2147) — Konrad's offramp double-pay defuse. The post-incident audit found the **exact same on-chain-then-ack-at-10s shape** in four unprotected flows; this PR ports the fix to them.

## The 4 flows

| Flow | File | sendMoney call site |
|---|---|---|
| Crypto withdraw | \`withdraw/crypto/page.tsx:319\` | \`sendMoney → recordPayment\` (or Rain webhook on collateral skip) |
| Direct send | \`useDirectSendFlow.ts:138\` | \`sendMoney → recordPayment\` |
| Semantic request pay | \`useSemanticRequestFlow.ts:298 + :492\` | \`sendMoney → recordPayment\` (2 sites) |
| Contribute pot | \`useContributePotFlow.ts:197\` | \`sendMoney → recordPayment\` |

Every one of these would have produced Konrad's double-pay had his exact incident shape hit them.

### 1. Extends \`usePaymentRecorder\` with a post-on-chain gate

The shared hook now exposes \`submittedTxHash\` — **set as the first action inside \`recordPayment\`, before the BE call**. Timing is the safety contract: the gate must engage when the on-chain hash is known, NOT after the BE ack. Otherwise a timeout surfaces as error → user retries → double-pay.

Also exposes \`markSubmitted(hash)\` for the rare skip-recordPayment paths (same-chain Rain-collateral crypto withdraw, where the Rain webhook reconciles instead).

### 2. Wires the gate into each of the 4 flow handlers

Each consumer:
- Destructures \`submittedTxHash\` from \`usePaymentRecorder()\`
- Early-returns at the top of its \`executePayment\` / \`handlePayment\` / \`handleConfirmWithdrawal\` callback when set
- \`SendInputView\` (direct-send UI) also OR's it into \`isButtonDisabled\` so the button stays disabled post-on-chain — clearer UX than 'enabled but does nothing on click'

### 3. Bumps \`recordPayment\` timeout 10s → 60s

\`services/charges.ts\` was using the 10s \`fetchWithSentry\` default. Matches \`confirmOfframp\`'s bump in #2147. Reduces how often the gate fires in the first place; the gate itself is still the defuse.

## Why the gate isn't \`useOnChainThenAck\`

The audit floated extracting a hook with a status enum + callbacks + storage. That over-engineers a same-session safety gate — the actual reusable kernel is one boolean. Naming it \`submittedTxHash\` inside the hook where the on-chain hash already lives earns its keep without inventing new abstractions. Future devs grep the field name and immediately see the safety invariant.

Cross-session recovery (page reload after on-chain leg) stays offramp-only via [#2150](https://github.com/peanutprotocol/peanut-ui/pull/2150) — charges have multi-payment semantics (pots, splits) that make a tx-hash-keyed BE recovery endpoint ambiguous. No incident has forced cross-session for these 4 flows yet.

## Regression risks

| Risk | Severity | Notes |
|---|---|---|
| Gate is wrong for one of the 4 flows | **Low** — each verified to fire an irreversible \`sendMoney\` on-chain leg before the BE ack. Retry = double-pay in all 4. | Tests + manual QA |
| 60s timeout hides genuine BE breakage | Low — same tradeoff as #2147's \`confirmOfframp\` bump | n/a |
| Hook generalization touches all 4 consumers' call paths | Medium — hot paths | Surgical edits: each call site adds ~3 lines (destructure + early-return + deps array entry) |
| User stuck on a stale gate after recoverable error | Low | \`reset()\` clears the gate; called in success paths (existing) and on flow reset |
| New any-types or eslint regressions | None — verified | pnpm prettier --check clean |

## Test plan

- [x] 6 new unit tests on \`usePaymentRecorder\` — all pass (timing test, throw-leaves-gate-set, markSubmitted, reset, happy path, null-on-mount)
- [x] \`pnpm prettier --check\` on touched files — clean
- [x] \`npm run typecheck\` — only the pre-existing SEOFooter submodule worktree warning (unrelated)
- [ ] **Manual QA** before merge: force \`recordPayment\` timeout on each of the 4 flows (block the call in DevTools), verify the Submit/Confirm button does NOT re-trigger \`sendMoney()\` on a second click.

## What's NOT in scope

- \`useClaimLink.tsx\` raw-\`fetch\` cleanup (different shape, separate concern — audit flagged but it's not Konrad-shape)
- Cross-session recovery for these 4 flows (would need BE \`GET /charges/by-tx-hash/:hash\` plus a generalized hook; deferred until incident forces it)
- \`useOnChainThenAck\` abstraction (audit recommendation; rejected as over-engineering for the actual scope)

Fixes the four flows the audit flagged as Konrad-shape-unprotected. Companion to #2147 + #2150 (offramp) + #929 (BE idempotency).


Hugo0:
Back to draft.

PR #2135, opened 2026-05-29
fix(badges): port EthFloripa + Token Nation event badges to dev (FE)

Hugo0:
Pairs with **peanut-api-ts#912** (BE).

## Why
The event-badge hotfix (#2133) shipped straight to `main` for the June São Paulo / Floripa events. `dev` never received it (FE `dev` refactored badges into a single `BADGES` record + has no `utm_campaign` handling). Without this, the next `dev`→`main` release would **regress** the live FE award path:
- no icons for `ETHFLORIPA_HUB` / `TOKEN_NATION_SP_2026` (fall back to peanutman logo)
- no `utm_campaign` → badge-code translation, so `/invite?...&utm_campaign=ethfloripa` would stop resolving to a badge

## What
- `BADGES`: add both codes (icon path + third-person copy, matching dev convention)
- `public/badges/`: add `token_nation_2026.svg` + `ethfloripa_hub.svg` (copied from main)
- `InvitesPage`: add `UTM_CAMPAIGN_TO_BADGE_MAP` + read `utm_campaign`; resolution order = explicit `campaign`/`campaignTag` → invite-code map → `utm_campaign` (utm last so an explicit tag always wins)

Mirrors merged main #2133.

## Verification
- typecheck: clean for changed files (asset-module `TS2307`s are the usual fresh-worktree `next-env.d.ts` noise, regenerated by CI)
- prettier: clean
- jest: `invite-suffix` passes; the only 2 failing suites (`content.test.ts`, `countryCurrencyMapping.test.ts`) are pre-existing and unrelated

## Note
FE main's `INVITE_CODE_TO_CAMPAIGN_MAP` has a **temporary** `squirrelinvitesyou → ARBIVERSE` mapping (also on dev). Since invite-code map is checked *before* utm, an ethfloripa link using the squirrel code with **only** `utm_campaign` (no explicit `campaign=`) would award ARBIVERSE. Event QR carries explicit `campaign=ETHFLORIPA_HUB`, so it's correct — flagging for whoever owns that temporary mapping.

PR #2155, opened 2026-06-01
chore(sync): back-merge main into dev — restore event badge catalog + guard regression

Hugo0:
Clean back-merge of `main` into `dev`, plus a regression guard so the underlying
codesmell can't bite a third time.

## Why
`dev` was missing `TOKEN_NATION_SP_2026` + `ETHFLORIPA_HUB` from the `BADGES`
catalog, so the live `/invite` award path renders those event badges with the
Peanutman fallback art + raw backend name. `main` already has the fix
(`32699f171`, the June 1 hotfix). This merge carries it to `dev`.

`git merge-tree` reports **zero conflicts**; the entire `dev↔main` delta is the
12-line `badge.utils.ts` block (plus `dev`-ahead `qr-pay` work, preserved).

## How the drop happened (narrative)
1. **May 23** — `551766bc0` collapses 3 parallel badge maps into one `BADGES`
   record on `dev`. Not yet released to `main`.
2. **May 29** — event-badge hotfixes ship to `main` (`#2133`), written against
   `main`'s **old parallel-map shape** (`CODE_TO_PATH` + `PUBLIC_DESCRIPTIONS`).
3. Those hotfixes reach `dev` via merge — but the parallel maps no longer exist
   there, so git auto-resolves the map-key additions as *deletion wins*. The
   `badge.utils.ts` entries **silently evaporate** (no conflict). The InvitesPage
   `utm_campaign` line + the SVG assets survive (refactor never touched them) —
   which is why `dev` today has the utm path + assets but not the catalog entries.
4. The same refactor later crosses into `main` on release, dropping `main`'s
   entries the same way → prod regression → hotfix `32699f171`.

Root cause: a hand-maintained FE badge registry refactored on one long-lived
branch while hotfixed on the other. Cross-branch merges don't conflict — they
drop keys whose target structure changed. Only signal was wrong art in prod.

## What's in this PR
- **Back-merge** `origin/main` → `dev` (the 12-line catalog re-add).
- **Extract** the `/invite` campaign maps out of the `InvitesPage` client
  component into pure-data `campaign-maps.ts` (import-only change in InvitesPage).
- **Guard test** `campaign-maps.test.ts`: asserts every value in
  `INVITE_CODE_TO_CAMPAIGN_MAP` + `UTM_CAMPAIGN_TO_BADGE_MAP` exists as a `BADGES`
  key. Verified it **fails** on the two dropped codes when the entries are absent,
  and passes (6/6) with the merge applied — i.e. it would have caught both the
  `dev` and `main` regressions at CI.

## Supersedes
Closes #2135 (`chore/sync-event-badges-to-dev`) — verified redundant: identical
SVG blobs already on `dev`, same `badge.utils.ts` entries, and its only
InvitesPage delta was comment re-wording (a downgrade — it drops the
"keep in sync with peanut-api-ts" note).

## Verification
- `git merge-tree origin/dev origin/main` → no conflicts
- `npx jest src/components/Invites/campaign-maps.test.ts` → 6/6 pass; proven to fail when entries removed
- typecheck clean on changed files; prettier clean
- BE side already in sync (peanut-api-ts `main ⊆ dev`; codes whitelisted+seeded on both)

### DIFF
--- a/src/components/Badges/badge.utils.ts
+++ b/src/components/Badges/badge.utils.ts
@@ -94,6 +94,18 @@ export const BADGES: Record<string, BadgeMeta> = {
         description: '$1K swiped. They put their money where their card is.',
         // TODO(card-launch): award on cumulative card spend ≥ $1K
     },
+    // Event badges — assets shipped to main via the May 29 hotfix but the catalog
+    // entries were dropped when the parallel maps collapsed into this single BADGES
+    // record, so the backend codes fell back to the Peanutman logo + raw backend name.
+    TOKEN_NATION_SP_2026: {
+        path: '/badges/token_nation_2026.svg',
+        description: 'São Paulo, baby. They came, they claimed, they tagged the wall.',
+    },
+    ETHFLORIPA_HUB: {
+        path: '/badges/ethfloripa_hub.svg',
+        description: 'Ilha da Magia, baby. Coconuts and consensus.',
+        displayName: 'Ethereum Hub Floripa',
+    },
 }
 
 /** All known badge codes — derived from BADGES so we never duplicate the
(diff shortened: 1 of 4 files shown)


Hugo0:
Closing — stale sync branch superseded by merged back-merges (#2190 etc). Certification found its **only** unique commit was `6800def7e` (campaign→BADGES guard test), which was in neither main nor dev — that work has been rescued into **#2197** so nothing is lost. The branch is retained (not deleted) for reference.

PR #964, opened 2025-07-08
[TASK-12603] reset send flow state when navigating to/from payment link creation

FacuBozzi:

PR #2173, opened 2026-06-02
fix(offramp): derive destination rail from account type, not country (PEANUT-API-5P/5M/5N)

Hugo0:
## Why

**2026-06-02 21:24 UTC. PEANUT-API-5P/5M/5N + Discord on-call page.**

A user with a UK GBP bank account tried to offramp USDC → EUR via SEPA. Bridge correctly rejected:
\`\`\`
country is not supported for SEPA
\`\`\`
SEPA only credits EUR-denominated Eurozone accounts. You can't SEPA-credit a GBP account.

## Root cause (FE)

\`BankFlowManager.handleCreateOfframpAndClaim\` picked the destination from country, not from the bank account's actual type:

\`\`\`ts
const destination = getOfframpCurrencyConfig(
    account.country ?? selectedCountry!.id
)
\`\`\`

When \`account.country\` is missing or doesn't match, the picker falls through to \`selectedCountry\`. \`getOfframpCurrencyConfig\`'s "everything unknown → EUR/SEPA" default then fires — even for a clearly GBP/UK account. Result: a GB account was paired with EUR/SEPA → Bridge 400.

The bank account's own \`type\` already carries the right answer for every Bridge destination we support (\`us\` / \`gb\` / \`clabe\` / \`iban\`). Deriving directly from type closes this class of bug at the source.

## What this PR does

New helper \`getOfframpConfigFromAccount(account)\` in \`src/utils/bridge.utils.ts\`:

| account.type | currency | paymentRail |
|---|---|---|
| \`us\` | usd | ach |
| \`gb\` | gbp | faster_payments |
| \`clabe\` | mxn | spei |
| \`iban\` | eur | sepa |
| \`manteca\` | **throws** (must use the Manteca offramp path, not Bridge) |
| missing/unknown | falls back to country-based picking (prior behavior) |

Also accepts BE Prisma-shape suffixes (\`BANK_IBAN\`, \`BANK_ACCOUNT_GB\`, …) so it works regardless of which side of the API the account row came from.

Use site: \`BankFlowManager.view.tsx:248\` — the only call site for the offramp destination picker on this code path.

## Tests

\`src/utils/__tests__/bridge.utils.test.ts\` — 7 new cases under \`getOfframpConfigFromAccount (PEANUT-API-5P/5M/5N regression)\`:

- ✅ **GB account → GBP/faster_payments** (was EUR/SEPA in the incident — locked in)
- ✅ US → USD/ach, CLABE → MXN/spei, IBAN → EUR/sepa
- ✅ BE Prisma-suffix shapes (\`BANK_IBAN\`, \`BANK_ACCOUNT_GB\`)
- ✅ Manteca throws to surface a wrong-path call early
- ✅ Missing-type fallback to country picking still works

**71 tests passing** in the file (was 64 before).

## What this PR does NOT do (deliberate)

- The companion BE PR fixes error surfacing + Discord pager filter: [peanut-api-ts #964](https://github.com/peanutprotocol/peanut-api-ts/pull/964). Even after this FE fix, *other* user-input mistakes (e.g. a Bridge customer status change mid-flow) shouldn't page on-call.
- Doesn't audit every other call site of \`getOfframpCurrencyConfig\`. The \`withdraw/[country]/bank/page.tsx\` flow uses the URL country, which by design matches the account type — out of scope here.

## Test plan

- [x] Unit tests cover the regression scenarios (above).
- [x] Typecheck clean for changed files.
- [x] Prettier clean.
- [ ] Post-merge: watch Sentry — \`country is not supported for SEPA\` should drop to zero from non-test traffic.

### DIFF
--- a/src/components/Claim/Link/views/BankFlowManager.view.tsx
+++ b/src/components/Claim/Link/views/BankFlowManager.view.tsx
@@ -15,7 +15,7 @@ import useClaimLink from '../../useClaimLink'
 import { type AddBankAccountPayload } from '@/app/actions/types/users.types'
 import { useAuth } from '@/context/authContext'
 import { type TCreateOfframpRequest, type TCreateOfframpResponse } from '@/services/services.types'
-import { getOfframpCurrencyConfig } from '@/utils/bridge.utils'
+import { getOfframpConfigFromAccount } from '@/utils/bridge.utils'
 import { getBridgeChainName, getBridgeTokenName } from '@/utils/bridge-accounts.utils'
 import { generateKeysFromString, getParamsFromLink } from '@/utils/peanut-link.utils'
 import { getContractAddress } from '@/utils/peanut-claim.utils'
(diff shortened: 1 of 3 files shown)


Hugo0:
Superseded by #2263, which folds in this commit (helper + Claim-flow fix, authorship preserved via cherry-pick) and completes it: applies the derive-from-account-type fix to the **withdraw** flow too, fixes the IBAN-country false-reject, and pairs with the BANK_GB backend typing (peanut-api-ts #1042) that makes the GB branch actually fire. Thanks for the original direction here.

PR #2189, opened 2026-06-05
fix(passkey): pin native signing ceremony to the kernel's own credential

jjramirezn:
## Why (follow-up to #2188)

#2188 made sure the kernel **client** holds the right credential. But Natalia *still* failed afterward (deployed 18h, no re-login): the recovery proved her kernel client was correct (`c4f6…`), yet her assertion was signed by her **second account's** key (`nnatalia`, `52ca1e36…`) — this time via the **native** surface.

Root cause: on Android native, signing goes through a custom `signMessageCallback` (`native-webauthn.ts`), and its `navigator.credentials.get()` was **unconstrained** — `allowCredentials` was only set if a caller supplied it, and ZeroDev doesn't for this shim. With two `peanut.me` passkeys on the device, the platform handed back the wrong account's passkey even though the client was right. The kernel then rejects it on-chain (`0xffffffff` → "invalid admin signature").

## What
Pin `allowCredentials` in the native callback to the kernel's **own** `authenticatorId`:
- `createNativeSignMessageCallback(rpId, pinnedCredentialId?)` defaults the get()'s allow-list to the kernel's credential **only when the caller didn't supply one**.
- Call site passes `storedWebAuthnKey.authenticatorId` (`kernelClient.context.tsx:285`).

## Safety (the key concern)
- **Native-only** — web signing goes through ZeroDev's own path and is untouched.
- It only **adds** the kernel's own credential to the allow-list. A working signature already uses that exact credential, so it can never be excluded → **a previously valid signature cannot become invalid.**
- **Worst case is a no-op:** if the `@capgo/capacitor-passkey` shim ignores `allowCredentials`, behavior is unchanged.

## Honest residual risk
Whether this *fixes* Natalia depends on the capacitor shim honoring `allowCredentials` for `get()` — needs a **real-device smoke test** (a 2-account Android device, or Natalia). If the shim honors it (expected), the platform is forced to her own passkey. If her own passkey isn't actually present on that device, signing will fail *cleanly* rather than substitute — which is correct, and points to recovery.

## Verification
- `npm run typecheck` ✓
- `npm test` — 1305/1306; the single failure (`add-money-states` crypto-deposit assertion) is pre-existing, reproduced on clean `origin/main`, unrelated.
- `prettier` ✓
- **Not yet device-tested** — please smoke on a 2-account Android build before relying on it.

### DIFF
--- a/src/context/kernelClient.context.tsx
+++ b/src/context/kernelClient.context.tsx
@@ -282,7 +282,13 @@ export const KernelClientProvider = ({ children }: { children: ReactNode }) => {
             // the native capacitor plugin callback so signing works after restore.
             if (isAndroidNative() && !storedWebAuthnKey.signMessageCallback) {
                 const rpId = storedWebAuthnKey.rpID || getNativeRpId()
-                storedWebAuthnKey.signMessageCallback = createNativeSignMessageCallback(rpId)
+                // Pin the native assertion to THIS account's credential so the
+                // platform can't return another Peanut account's passkey for the
+                // same RP (would sign with the wrong key → "invalid admin signature").
+                storedWebAuthnKey.signMessageCallback = createNativeSignMessageCallback(
+                    rpId,
+                    storedWebAuthnKey.authenticatorId
+                )
             }
             // Only update if the key actually changed to avoid re-triggering kernel client init
             // Note: WebAuthnKey contains BigInt fields (pubX, pubY) which JSON.stringify cannot handle,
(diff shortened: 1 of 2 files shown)


jjramirezn:
Closing — wrong on two counts:

1. It targets the **native** (`createNativeSignMessageCallback`) path, but we don't ship native — `isAndroidNative()` is always false, so that code never runs.
2. Even if it did, the pin is **redundant**: `@zerodev/passkey-validator` already passes `allowCredentials: [{ id: webAuthnKey.authenticatorId }]` to both the native callback and the web ceremony (`toPasskeyValidator.js:83-86`).

Root cause is not an unconstrained ceremony. A user video shows Google Password Manager going straight to biometric with **no account picker** → the platform *is* honoring `allowCredentials` with a single credential — it's just the **wrong** one. Her cached `WebAuthnKey` is the other account's (cached-wrong-key bug). That's what the #2188 guard fixes; it just hasn't reached her because she's on a stale service-worker-cached bundle. Fix = reload onto the new build + clean re-login, not this.

PR #2196, opened 2026-06-07
feat(shhhhh): polish card share asset for Twitter (peanut blue, hero card, clean year)

Hugo0:
## What
Visual polish of the card **share asset** (`ShareAssetD3`) per the [Waitlist Feedback](https://www.notion.so/peanutprotocol/37283811757980478b90dc3a8a12c47d) doc, optimised for how it reads on a Twitter timeline.

| Feedback | Change |
|---|---|
| "BG violates brandbook / doesn't pop" | `#efe4ff` lavender → **`#90A8ED` peanut blue** (reused from the prod LP `businessBgColor` / `--background-color`, as one DRY const) |
| "card too small / whitespace huge / should fill the space" | card is the hero: `CARD_W` 620→760 (~52%→~63% of canvas), centred |
| "the `'''` look buggy" | stamp year `'25` → **`2025`** (the leading apostrophe read as a stray tick) |
| "badges too small" | low badge-count stamps bumped for thumbnail legibility (high counts unchanged — the non-overlap invariant + jitter leaves no headroom) |

The `/shhhhh` LP hero uses the same `PixelatedCardFace`, so its scale was compensated (0.645→0.526) to keep its on-screen footprint **unchanged** by the shared `CARD_W` bump.

## Gate (local)
- `prettier --check` clean · `typecheck` exit 0 · `npm test` = 1306 passed / 76 suites
- `shareAssetLayout.test.ts` non-overlap + canvas-bounds invariants stay green (24/24)

## Notes
- **Stacks on #2190** (the main→dev back-merge) — based on `chore/sync-main-dev-0606` so the diff is just these 3 files; GitHub auto-retargets to `dev` once #2190 merges.
- **Follow-ups (not in this PR):** the `/shhhhh` desktop hero card bleeds slightly off the right (pre-existing rotation overflow) — left for a deliberate hero-framing pass; and the `SINCE OCT '25` stat keeps its apostrophe (reads as a normal date, unlike the stamp tick).


Hugo0:
**Post-review audit (Fable) — 1 new commit `8b2479e11`:**

- Bare door now checks access at click: **hasCardAccess holders go to /card** (before: 'you're on the waitlist' with no position); join failures show an inline error (before: silent console.error — and pre-launch *every* join silently 403'd against the old BE gate, both click and post-signup cookie path).
- **Removed the `?press_door=1` backdoor** on /card + `cardApi.grantFlowEarlyAccess` — orphaned by the door rework, and a shareable URL that silently stamps flowEarlyAccess contradicts 'the door is not a bypass'.
- **History row**: chronological, un-pinned, now also injected on the paginated **/history** page (same synthetic pattern as KYC/badge). Celebration-seen BE wiring reverted (see api#990 comment for why it was inverted).
- **?campaign=skip lands on /card** (signed-out signup carries `redirect_uri=/card`; the badge award is awaited inside registration, so no race).

Local gate: typecheck ✅, jest 1306 ✅, eslint not-worsened (4 pre-existing errors on touched files, 0 new), `next build` ✅. Pairs with peanut-api-ts#990 c9408139 + 72b21c02 — **BE must deploy before/with this FE** (FE now relies on un-gated join + inner-implies-outer flowEarlyAccess).

PR #2207, opened 2026-06-10
fix(card): render failure-state rails instead of looping users back to add-card

Hugo0:
## Why

`computeCardState` only understood `REJECTED` and `PENDING` (plus a dead `'IN_REVIEW'` branch that isn't a real `RailStatus`). A Rain rail in **REQUIRES_INFORMATION / REQUIRES_EXTRA_INFORMATION / FAILED / REQUIRES_SUPPORT** fell through to `add-card` → user taps Apply → backend answers "Application already submitted" → overview invalidates → `add-card` again. **Infinite loop with no remediation screen.** Real prod victims diagnosed today (2026-06-10), e.g. `barbara` (REQUIRES_INFORMATION, `INCORRECT_SOCIAL_NUMBER`).

## What

- `cardState.utils.ts`: all seven `RailStatus` values now resolve to a real state — `requires-info` (new, surfaces the capabilities `reason.userMessage`), `requires-support` (new, contact-support screen), FAILED reuses the existing rejection screen; dead `IN_REVIEW` branch removed (verified unreachable).
- `ApplicationStatusScreen`: two new variants + optional `reasonMessage`; support CTA via the existing Crisp-gated support drawer (PR #2130 pattern). No provider names in any copy.
- `card/page.tsx`: wires `useCapabilities` + the two new switch cases.

| RailStatus | Card state | Screen |
|---|---|---|
| REQUIRES_INFORMATION / REQUIRES_EXTRA_INFORMATION | `requires-info` (new) | "We need more information" + reason + support CTA |
| REQUIRES_SUPPORT | `requires-support` (new) | "Something went wrong on our side" + support CTA |
| FAILED | `rejected` (reused) | existing rejection screen |

## Notes

- Companion BE PR: peanutprotocol/peanut-api-ts#1006 (fixes the upstream causes: Rain missing-address 400, orphan applicant row, webhook crash, alert dedupe).
- Pre-existing gap left as-is: failure-state users without `flowEarlyAccess` and no card still hit the page-level `notFound()` first (also true for `rejected` before this PR).
- Follow-up opportunity (not in scope): dispatch the capabilities sumsub NextAction to re-open the WebSDK from `requires-info` instead of support-only.

## Test plan

- [x] `npm run typecheck` clean
- [x] `npm test` — 81 suites / 1395 green; new table-driven suite covers all 7 RailStatus values + a "failure states never resolve to add-card" regression guard (12 new tests)
- [x] prettier clean

### DIFF
--- a/src/app/(mobile-ui)/card/page.tsx
+++ b/src/app/(mobile-ui)/card/page.tsx
@@ -69,7 +69,7 @@ const CardPage: FC = () => {
 
     const { overview, isLoading: overviewLoading, error: overviewError } = useRainCardOverview()
     const { serializeGrant } = useGrantSessionKey()
-    const { railsForProvider } = useCapabilities()
+    const { railsForProvider, isLoading: capabilitiesLoading } = useCapabilities()
     const { setIsSupportModalOpen } = useModalsContext()
     const onBack = useSafeBack('/home')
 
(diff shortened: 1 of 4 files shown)


Hugo0:
@coderabbitai review

PR #2208, opened 2026-06-10
feat: advertise ETH on the crypto deposit screen (EVM only)

Hugo0:
## Summary

[TASK-19911](https://app.notion.com/p/peanutprotocol/3738381175798183bf56e02cb2a8e874) — ETH deposits already work end-to-end (Rhino inflow SDAs bridge ETH to Arbitrum USDC and the backend credits it), but the deposit screen only advertised USDC/USDT, so users never knew they could send ETH.

This adds ETH to the supported-token chips on every Rhino deposit surface (CryptoDeposit view, RhinoDeposit view / request-fulfilment, TokenAndNetworkConfirmationModal), **EVM only**.

**Verified against Rhino's live `depositAddresses.getSupportedConfigs()`** (read-only probe, 2026-06-10): ETH is accepted on the EVM SDA chains (Arbitrum, Base, BNB, Ethereum, Katana, Mantle, Polygon, Optimism, Unichain) but **not** on Solana or Tron. A token sent to an SDA that Rhino doesn't accept is silently lost (no webhook, no intent), so the token list moves from a flat list with a TRON special-case to an explicit per-network map, with a unit test guarding that ETH never leaks to SOL/TRON.

Side benefit: `TOKEN_LOGOS.ETH` means ETH deposit history/status rows now resolve a proper ETH icon instead of the USDT fallback (`add-money/crypto/page.tsx`, `getTokenLogo`).

## Risks / breaking changes

- Display-only change; no backend or API contract impact.
- **Known imprecision (accepted):** the EVM screen shows one flat token list for all EVM chains, but Rhino does not accept ETH on Gnosis/Celo (live config). Practically unreachable — exchanges/wallets don't offer native-ETH withdrawals on those chains — and the existing screen has the same flat-list semantics today.
- ⚠️ **Pre-existing drift flagged (NOT fixed here):** Rhino's live SDA config no longer lists **Scroll** at all, but `SUPPORTED_EVM_CHAINS` still advertises it on the deposit screen. If Rhino really dropped Scroll, a USDC/USDT deposit on Scroll today would be silently lost. Follow-up needed (confirm with Rhino, then drop the chip).

## QA

- `npm run typecheck` ✅ · `npm test` 81/81 suites (1381 passed) ✅ · `npm run build` ✅ · prettier ✅
- New test: `src/constants/__tests__/rhino.consts.test.ts` (ETH EVM-only; TRON USDT-only; SOL stablecoins-only).
- Manual: Add Money → From Crypto → EVM shows USDT/USDC/ETH chips; Solana/Tron tabs unchanged.

### DIFF
--- a/src/app/(mobile-ui)/add-money/__tests__/add-money-states.test.tsx
+++ b/src/app/(mobile-ui)/add-money/__tests__/add-money-states.test.tsx
@@ -268,8 +268,9 @@ jest.mock('@/constants/rhino.consts', () => ({
     getSupportedTokens: jest.fn(() => [
         { name: 'USDC', logoUrl: '/usdc.png' },
         { name: 'USDT', logoUrl: '/usdt.png' },
+        { name: 'ETH', logoUrl: '/eth-token.png' },
     ]),
-    TOKEN_LOGOS: { USDC: '/usdc.png', USDT: '/usdt.png' },
+    TOKEN_LOGOS: { USDC: '/usdc.png', USDT: '/usdt.png', ETH: '/eth-token.png' },
 }))
 
 jest.mock('@/constants/zerodev.consts', () => ({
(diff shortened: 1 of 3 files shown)


Hugo0:
@coderabbitai review

PR #2210, opened 2026-06-10
fix: actionable UX + telemetry for WebAuthn NotAllowedError during signing

Hugo0:
## Summary

When an iOS third-party credential provider (1Password) wedges, every passkey assertion during transaction signing throws `NotAllowedError` until the provider is unlocked or the device restarts. Users hitting this mid-send saw the generic **"There was an issue with your request. Please contact support."** — a dead end for a failure that a retry (after unlocking/rebooting) fixes. **45 users hit this in the last 30 days, 40 confirmed iOS** (TASK-20000, Crisp user `ariel`: 4 failed send attempts on 2026-06-06, QR payment succeeded on a later retry).

- `ErrorHandler` now maps the iOS Safari message ("not allowed by the user agent…") to retry/unlock-your-password-manager guidance. Only the *timeout* variant was mapped before.
- iOS `PASSKEY_TROUBLESHOOTING_STEPS.NotAllowedError` now leads with the password-manager unlock hint and ends with device restart (the two fixes that actually work in the field).
- New `passkey_sign_failed` PostHog event (`error_name`, `context`) emitted from the two low-level signing entry points (`useZeroDev.handleSendUserOpEncoded`, `useSignUserOp.signCallsUserOp`) via `capturePasskeySignFailure`, which no-ops for non-WebAuthn errors. Spend-bundle EIP-712 failures already carry `error_kind` on `card_withdraw_failed`, so they're intentionally not double-captured.

## Risks / breaking changes

None expected. Copy + analytics only; no flow logic changed. No backend involvement. The new ErrorHandler branch sits after the existing WebAuthn-timeout matcher and before the generic fallback, so existing mappings are untouched (ordering covered by tests).

## QA

- `npm test` — new tests: verbatim prod error message → new copy; ordering guard for the existing timeout variant; `capturePasskeySignFailure` captures for `NotAllowedError` and ignores non-WebAuthn errors.
- `npm run typecheck`, `npm run build`, `pnpm prettier --check .` all green locally.
- Manual repro of the wedged-provider state isn't feasible in sandbox; the mapped string is the verbatim DOMException message captured in PostHog $exception events from prod.

### DIFF
--- a/src/constants/analytics.consts.ts
+++ b/src/constants/analytics.consts.ts
@@ -201,6 +201,11 @@ export const ANALYTICS_EVENTS = {
     CARD_WITHDRAW_ATTEMPTED: 'card_withdraw_attempted',
     CARD_WITHDRAW_SUCCEEDED: 'card_withdraw_succeeded',
     CARD_WITHDRAW_FAILED: 'card_withdraw_failed',
+
+    // WebAuthn ceremony failed while signing a transaction (userOp / EIP-712).
+    // `error_name` is the DOMException name (NotAllowedError = provider
+    // refused/wedged, e.g. 1Password on iOS), `context` is the signing call site.
+    PASSKEY_SIGN_FAILED: 'passkey_sign_failed',
 } as const
 
 /**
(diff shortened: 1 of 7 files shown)


Hugo0:
@coderabbitai review

PR #2211, opened 2026-06-10
fix(rain-cooldown): mount intro modal globally + track swallowed cooldowns

Hugo0:
## Summary

The Rain cooldown explainer modal was mounted only in the `(mobile-ui)` layout, but the `rain:cooldown` event fires from **every** spend path — including public pay/send/request pages (`/send/<user>`, `/<user>/<amount>`) that live outside that route group. A user paying a request from one of those pages got a bare inline error with no modal (Sentry [PEANUT-UI-QJ1](https://peanut-c34d84c05.sentry.io/issues/PEANUT-UI-QJ1), June 4–5). The `RainCooldownProvider` + ToastProvider already live at the root via `ClientProviders`; the modal now mounts beside them so coverage can't drift per route group.

Second part: a `rain_cooldown_hit` PostHog event captured in `rainRequest`'s 425 branch (services/rain.ts) — the single choke point every spend surface's cooldown flows through. Fires for **both** shapes, including the `retryAfterSec: null` one that shows no cooldown UI (the QJ1 blind spot, previously zero telemetry: the qr-pay catch handles these gracefully with no `captureException`). Flow context comes from PostHog's auto-captured `$pathname`.

## Risks / breaking changes

- Low blast radius: mount-point move (modal portals to `document.body`, so stacking is unchanged) + one analytics event. No API contract changes.
- First `posthog` usage inside `src/services/` — deliberate: per-callsite captures would undercount the other 10+ spend surfaces and were coupled to the `rainMsg` substring match.
- Companion BE PR peanutprotocol/peanut-api-ts#1008 makes `retryAfterSec` always present so the countdown/modal engage on every cooldown shape — independent, no deploy-order requirement.

## QA

- `npm test` — 1378 passed. Typecheck, prettier, `npm run build` clean (re-run after each review fix).
- Existing `RainCooldownContext` tests cover modal-trigger behavior; mount move is structural (verified: single root layout wraps all route groups in `ClientProviders`).
- Manual: dispatch `window.dispatchEvent(new CustomEvent('rain:cooldown', { detail: { retryAfterSec: 120, message: 'x' } }))` on a public `/send/<user>` page → intro modal + countdown toast now appear.

### DIFF
--- a/src/app/(mobile-ui)/layout.tsx
+++ b/src/app/(mobile-ui)/layout.tsx
@@ -15,7 +15,6 @@ import '../../styles/globals.css'
 import QRScannerOverlay from '@/components/Global/QRScannerOverlay'
 import SecurityVerificationOverlay from '@/components/Global/SecurityVerificationOverlay'
 import SupportDrawer from '@/components/Global/SupportDrawer'
-import RainCooldownIntroModal from '@/components/Global/RainCooldown/IntroModal'
 import JoinWaitlistPage from '@/components/Invites/JoinWaitlistPage'
 import { useRouter } from 'next/navigation'
 import { Banner } from '@/components/Global/Banner'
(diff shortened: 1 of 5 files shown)


Hugo0:
@coderabbitai review

PR #2212, opened 2026-06-10
fix: auto-recover from chunk-load failures instead of dead-ending

Hugo0:
## Summary

**TASK-19993 ("Vraj 500 error")** — a user on `/withdraw/germany/bank` hit an uncaught `ChunkLoadError` after entering the withdrawal amount and dead-ended on Next's static **"Application error: a client-side exception has occurred"** screen. He could not withdraw at all. This isn't one user: PostHog `$exception` shows ChunkLoadErrors **every day** (a few users/day background from flaky networks) plus **deploy-day spikes** (2026-05-26: 37 users in ~2.5h, all pinned to one superseded deployment).

The only real fix for a failed chunk is **reloading the page** — refetching the HTML re-pins to the current deployment, whose chunks resolve. This PR adds that recovery on both surfaces where the error can land:

1. **Uncaught** → inline `beforeInteractive` script in the root layout (`CHUNK_ERROR_RECOVERY_SCRIPT`). It must be inline: the May-26 data shows the **error-boundary chunks themselves** (`global-error-*.js`, `(marketing)/error-*.js`) were among the top failures — boundary-based recovery can't load when chunk loading is what's broken. Inline-with-the-HTML code is the only code guaranteed to be in memory at error time.
2. **Caught by a boundary** (no window event fires) → `recoverFromChunkError()` from `LazyLoadErrorBoundary` (failed lazy chunks silently rendered `null` — e.g. the transaction-details drawer just never opened), `[...recipient]/error.tsx` + `(marketing)/error.tsx` ("Try again" `reset()` re-renders against the same dead deployment — can never succeed under skew), `MarketingErrorBoundary`, and `global-error.jsx`.

Safety: both paths share one sessionStorage guard — **max one auto-reload per 60s** — so they can't compound into a reload loop; when the guard blocks (e.g. hard-offline), behavior degrades to exactly today's. Standalone PWA is excluded (reload there can bounce the user out to the browser — same hazard the sw-registration script documents). Non-chunk errors are untouched.

## Risks / breaking changes

- **Low.** All additions are no-ops for non-chunk errors; the inline script only acts on uncaught chunk failures that today produce a guaranteed dead-end.
- An auto-reload mid-flow discards in-memory state — but the page was already broken, and URL-as-state means flows resume where they were. No reload happens while anything is healthy.
- No new deps; inline script is ES5, no CSP change (inline scripts already used in layout).

## QA

- `npm test` — 19 new tests cover the inline script (executed via `new Function` against shadowed globals — the exact string that ships) + `isChunkLoadError` + `recoverFromChunkError`: reload-once, guard window, standalone skip, broken-sessionStorage skip, non-chunk no-op.
- Manual repro: build + serve, then block a `_next/static/chunks/*.js` URL in DevTools and navigate — page reloads once instead of dead-ending; second failure within 60s shows the previous behavior.

Crisp ref: session_430d7370-f4ab-4c9c-9e61-e2c4d2b689f6

### DIFF
--- a/src/app/[...recipient]/error.tsx
+++ b/src/app/[...recipient]/error.tsx
@@ -5,6 +5,7 @@ import { useRouter } from 'next/navigation'
 import { useModalsContext } from '@/context/ModalsContext'
 import { Button } from '@/components/0_Bruddle/Button'
 import { Card } from '@/components/0_Bruddle/Card'
+import { recoverFromChunkError } from '@/utils/chunk-error-recovery'
 
 export default function PaymentError({ error, reset }: { error: Error & { digest?: string }; reset: () => void }) {
     const router = useRouter()
@@ -12,6 +13,9 @@ export default function PaymentError({ error, reset }: { error: Error & { digest
 
     useEffect(() => {
         console.error(error)
+        // "Try again" re-renders against the same dead deployment under skew —
+        // for chunk errors only a reload (re-pin to current deployment) works.
+        recoverFromChunkError(error)
     }, [error])
 
     return (
--- a/src/app/[locale]/(marketing)/error.tsx
+++ b/src/app/[locale]/(marketing)/error.tsx
@@ -2,10 +2,14 @@
 
 import { useEffect } from 'react'
 import Link from 'next/link'
+import { recoverFromChunkError } from '@/utils/chunk-error-recovery'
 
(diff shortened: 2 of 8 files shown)


Hugo0:
@coderabbitai review

PR #2233, opened 2026-06-16
fix(card): revert to collateral-first spend routing (incident: smart-first reverts on-chain)

jjramirezn:
## 🚨 Incident mitigation — reverts #2230

**Symptom (prod):** card users whose funds live in collateral get `Paymaster error … UserOperation reverted during simulation … ERC20: transfer amount exceeds balance` when trying to pay — the payment hard-fails.

**Cause:** #2230 reordered `computeSpendStrategy` to **smart-first**. But card funds are swept smart→collateral to back the card, so the smart account is normally ~empty. The FE's `smartBalance` (`useBalance`, 30s-cached / read before a sweep) can be `>= amount` while the on-chain smart account is empty → `smart-only` is chosen → the kernel USDC transfer reverts at the paymaster. The old **collateral-first** ordering never trusted the smart balance for these users, so it was safe.

**This PR:** restores collateral-first ordering + its test, and leaves a `NOTE` in the code documenting why smart-first was reverted.

## Trade-off
- Restores the pre-#2230 behavior → **reopens the original Rain withdrawal-signature cooldown** on rapid back-to-back payments when the user *does* hold smart-account USDC (kkonrad's report). That's a tolerable annoyance vs. broken payments.
- The correct smart-first redo (follow-up) needs: a **live/uncached** smart-balance read at routing time **AND** a **fallback to collateral** when the smart-only path can't be funded (catch the revert / pre-simulate). Not safe to land without that.

## QA
- `npx jest src/hooks/wallet` — strategy returns `collateral-only` when collateral covers (single recipient), `smart-only` only when collateral path ineligible and smart covers, `mixed` for shortfall.
- Sandbox: collateral-funded card user (smart ~0) makes a payment → routes collateral-only, no paymaster revert.

### DIFF
--- a/src/hooks/wallet/__tests__/useSpendBundle.test.ts
+++ b/src/hooks/wallet/__tests__/useSpendBundle.test.ts
@@ -4,7 +4,7 @@
  * The hook itself orchestrates kernel clients, Rain API calls, and the
  * session-key grant flow — those paths are covered by integration + manual
  * testing on sandbox. These tests lock down the deterministic pieces:
- *   - `computeSpendStrategy` routing (smart → collateral → mixed → insufficient)
+ *   - `computeSpendStrategy` routing (collateral → smart → mixed → insufficient)
  *   - `usdcUnitsToRainCents` amount conversion at the Rain API boundary
  */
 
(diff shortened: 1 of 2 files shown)


jjramirezn:
Superseded by #2234 — keeping smart-first and fixing the root cause (route on a live on-chain balance) instead of reverting. Branch retained as an emergency fallback.

PR #2260, opened 2026-06-19
feat(card): show registered cardholder name on reveal

abalinda:
## Summary
Render the Rain-registered cardholder name on the card face, only in the revealed state (alongside
PAN/CVV/expiry), `ph-no-capture` since it's PII.

## Risks / cross-repo
⚠️ **Depends on peanut-api-ts#1038**, which returns the optional `cardholderName` on
`GET /rain/cards/:cardId/details`. Deploy **backend first**; the field is optional so the FE degrades
gracefully if the BE isn't live yet.

## QA
- Unit: `CardFace.test.tsx` (name shown revealed / hidden masked / renders when name absent) +
  `useCardReveal` passthrough. Typecheck + prettier clean.
- Sandbox: reveal shows the name; masked shows none.

## Follow-up (not blocking)
- Refresh the OpenAPI snapshot (`api.openapi.json` + `api.generated.ts`) to include `cardholderName` via
  `pnpm gen:api:live` against the updated backend. Runtime works without it (FE uses the hand-written
  `RainCardDetailsResponse`); `check:api` stays green (snapshot self-consistent).

### DIFF
--- a/src/components/Card/CardFace.tsx
+++ b/src/components/Card/CardFace.tsx
@@ -11,6 +11,9 @@ export interface RevealedCardDetails {
     cvv: string
     expiryMonth: number
     expiryYear: number
+    /** Registered cardholder name from Rain. Optional — the backend resolves it
+     *  best-effort, so a Rain hiccup leaves it absent and the field is hidden. */
+    cardholderName?: string
 }
 
 interface Props {
@@ -125,6 +128,13 @@ const CardFace: FC<Props> = ({
                                     </button>
                                 )}
                             </div>
+                            {/* Registered cardholder name — PII, kept out of session
+                             * recordings like the other revealed fields. */}
+                            {revealed.cardholderName && (
+                                <span className="ph-no-capture mt-1 text-sm font-bold uppercase tracking-wide">
+                                    {revealed.cardholderName}
+                                </span>
+                            )}
                             <div className="flex items-end justify-between">
                                 <div className="text-s flex gap-6">
                                     <div>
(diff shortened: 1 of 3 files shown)


abalinda:
Shows the Rain registered name on the card, but only once you reveal it (next to PAN/CVV/expiry) and _ph-no-capture_ like the other sensitive fields  nothing on the masked card. Just threads _cardholderName_ through the reveal type → _useCardReveal_ → _CardFace_

CardFace tests cover shown-on-reveal / hidden-when-masked / still-renders-if-absent, plus the reveal-hook passthrough. typecheck + prettier clean, unit + e2e + Vercel preview green. Red eslint is pre-existing. Depends on peanut-api-ts#1038 (returns the optional field) so deploy BE first — the field's optional, FE just no-ops without it. Low risk, mergeable.

abalinda:
@coderabbitai 

any comments?

abalinda:
dropped the "Expiry"/"CVV" text labels on the revealed card (and in the loading skeleton too). They were adding a line of height that pushed the cardholder name + PAN up into the hand artwork — without the labels everything sits a touch lower and clears it. Values still render as before, and matching the skeleton means there's no height jump when the reveal animates in. Also added a test that asserts the name keeps its `ph-no-capture`class, so it can't drift out of the session-recording guard in a later refactor.

Overall this PR just surfaces the registered cardholder name on reveal.

needs visual check (I wasnt able to make it on my end)

jjramirezn (approved):

PR #2266, opened 2026-06-21
fix(request): show spendable balance incl. card collateral

abalinda:
## Summary

A user reported the "Balance: $XXX" on `/request` reading **lower** than `/home` (their example: `$2213.38` vs `$2472.03`). Both screens read the same `useWallet()` hook but picked different fields:

- `/home` (and `/send`) show `spendableBalance` = smart-account USDC **+ Rain card collateral** (landed + in-transit).
- `/request` showed `balance` = smart-account USDC **only**, silently dropping card collateral.

The gap is exactly the funds the user holds as card collateral. This is the case `useWallet.ts:257-263` explicitly warns against ("payment-input forms … should show spendableBalance … otherwise a user with funds split across smart and collateral sees a smaller balance than they actually have"). The **Send** view already follows this; the two **Request** views were the only outliers.

Fix: alias `spendableBalance: balance` in both Request views, mirroring `Initial.link.send.view.tsx`.

## Risks / breaking changes

- **Display-only.** `AmountInput`'s `walletBalance` prop is rendered as text only (`AmountInput/index.tsx:279-284`); it is **not** wired to any affordability / spend gate. No spend-routing logic changes.
- No cross-repo / API impact. `spendableBalance` is already returned by `useWallet`.

## QA

Sandbox, user with funds split across smart account **and** Rain card collateral:
1. `/home` balance, `/send` balance, and `/request` balance (both the create-link view and a direct `/<username>` request) now show the **same** number.
2. Previously `/request` read lower by the collateral amount.

---

## Update — 2nd commit (`d2cfdfeda`): unify the affordability gate (scope expanded)

The display fix above only patched a symptom. The same *"which balance field?"* bug class lives in the affordability **gates** of the legacy money-flows, so this commit fixes the root cause across them.

**The bug class.** The displayed spendable balance includes in-transit card-collateral top-ups (added in #2170 so the balance doesn't crater to $0 during the ~10–45s smart→collateral handoff). Five legacy flows hand-rolled their "insufficient balance" gate by comparing the entered amount against that **display** number — which can briefly read higher than what's actually routable. `useWallet` already exposes `hasSufficientSpendableBalance()` (gates on *available-now* = smart + **landed** collateral) for exactly this, but only the newer `features/payments/flows` used it.

**Fix.** Route the gates through the shared predicate:
- `Send/link`, `qr-pay`, `withdraw/bank`, `withdraw/manteca` → `hasSufficientSpendableBalance(amount)`
- `withdraw` page ceiling (`maxDecimalAmount`) → newly-exposed `availableSpendableBalance`
- **Displayed balances are unchanged.**

**Risk.** Outside the top-up window `available == display`, so behaviour is identical — the change only tightens the rare in-transit window (block at input vs. fail at execution). One known, by-design nuance: during a top-up the display can read higher than the gate allows for ~30s, self-healing once collateral lands.

**Tests.** Revived the qr-pay insufficient-balance test (was `.skip`ped on mock-shape drift) + added a request-view regression locking the spendable-balance display field. 6 source + 3 test files; typecheck ✅, prettier ✅, affected jest suites ✅.

### DIFF
--- a/src/app/(mobile-ui)/qr-pay/__tests__/qr-pay-states.test.tsx
+++ b/src/app/(mobile-ui)/qr-pay/__tests__/qr-pay-states.test.tsx
@@ -138,6 +138,10 @@ jest.mock('@/hooks/useRainCardOverview', () => ({
 
 jest.mock('@/utils/balance.utils', () => ({
     rainCentsToUsdcUnits: jest.fn(() => 0n),
+    SPEND_BLOCK_MESSAGE: {
+        settling: 'Your balance is updating. Try again in a few seconds.',
+        insufficient: 'Not enough balance. Add funds to continue.',
+    },
 }))
 
 const mockUseTransactionDetailsDrawer = jest.fn()
@@ -561,7 +565,7 @@ function applyDefaults() {
     mockUseWallet.mockReturnValue({
         balance: parseUnits('100', 6), // $100 USDC
         spendableBalance: parseUnits('100', 6),
-        hasSufficientSpendableBalance: () => true,
+        spendBlockReason: () => null, // affordable by default
         sendMoney: jest.fn(),
     })
 
@@ -784,13 +788,13 @@ describe('GROUP 2: Payment Form States', () => {
 
     test('Insufficient balance shows pay button disabled + error', async () => {
         // Payment needs ~$18.4 but the available-now spendable is only $5, so the
-        // affordability gate (hasSufficientSpendableBalance) blocks it. Revived from
-        // skip once the gate moved off the raw display balance onto the shared hook
-        // predicate (the original mock-shape drift this test hit).
+        // gate (spendBlockReason) returns 'insufficient'. Revived from skip once the
+        // gate moved off the raw display balance onto the shared hook classifier
+        // (the original mock-shape drift this test hit).
         mockUseWallet.mockReturnValue({
             balance: parseUnits('5', 6),
             spendableBalance: parseUnits('5', 6),
-            hasSufficientSpendableBalance: () => false,
+            spendBlockReason: () => 'insufficient',
             sendMoney: jest.fn(),
         })
 
--- a/src/app/(mobile-ui)/qr-pay/__tests__/qr-pay-states.test.tsx
+++ b/src/app/(mobile-ui)/qr-pay/__tests__/qr-pay-states.test.tsx
@@ -137,9 +137,10 @@ jest.mock('@/hooks/useRainCardOverview', () => ({
 }))
 
 jest.mock('@/utils/balance.utils', () => ({
+    // keep the real isAmountWithinBalance / messages so the gate is genuinely
+    // exercised; only stub the Rain widening helper.
+    ...jest.requireActual('@/utils/balance.utils'),
     rainCentsToUsdcUnits: jest.fn(() => 0n),
-    INSUFFICIENT_BALANCE_MESSAGE: 'Not enough balance. Add funds to continue.',
-    BALANCE_SETTLING_MESSAGE: "Your balance isn't fully available yet. Please try again in a few seconds.",
 }))
 
 const mockUseTransactionDetailsDrawer = jest.fn()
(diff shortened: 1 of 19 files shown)


abalinda:
Switches the two /request views from the smart-account-only `balance` to `spendableBalance`, so the "Balance:" line now includes Rain card collateral and matches /home + /send. Fixes the report where /request read lower than the homescreen 

image
image


Root-caused
- `useWallet` already exposes both; /send was correct, the two request views (create-link + direct) were the only ones still reading smart-only. Lines up with useWallet's own comment that payment-input forms should show spendable.
- Confirmed `AmountInput`'s balance is display-only — not wired to any affordability/spend gate, so zero risk to send routing.

CodeRabbit no actionable comments + /code-review clean,
Mergeable. Display-only 2-line change, low risk.

Hugo0:
### Heads-up: 2nd commit expands the scope of this PR

Pushed `d2cfdfeda` on top of the original display fix. I've updated the PR description with full detail — short version here so reviewers/`@abalinda` aren't surprised by the extra files.

**Why the scope grew.** The original one-line fix (show `spendableBalance` on `/request`) corrected a *symptom*. The same "reads the wrong balance field" bug also lives in the affordability **gates** of the legacy money-flows. The displayed spendable balance includes in-transit card-collateral top-ups (#2170, so it doesn't crater to $0 during the ~10–45s smart→collateral handoff), but `Send/link`, `qr-pay` and `bank`/`manteca` withdraw validate the entered amount against that **display** number rather than the available-now `hasSufficientSpendableBalance()` predicate the newer `features/payments/flows` already use. During a card top-up that can green-light a spend whose funds aren't routable yet (fails at execution instead of being blocked at input).

**What the commit does.** Routes those gates through the shared predicate and exposes `availableSpendableBalance` for the withdraw ceiling. **Displayed balances are unchanged.** Outside the top-up window `available == display`, so behaviour is identical — it only tightens the rare in-transit window.

**Tests.** Revived the qr-pay insufficient-balance test (was `.skip`ped on mock drift) + added a request-view regression. typecheck ✅, prettier ✅, affected jest suites ✅.

One by-design nuance to be aware of: during a top-up the displayed balance can read higher than the gate allows for ~30s (self-heals once collateral lands) — consistent with `useWallet`'s documented intent.

Hugo0:
### Course-correction: simplify to fail-late + refetch (dropping the input-gate tightening)

After review, the earlier "gate on available-now" approach (blocking in-transit spends at input) was the wrong call, and I'm reverting it. Reasoning:

- The FE balance **lags up to ~30s** (`useBalance` and `useRainCardOverview` both poll at `staleTime/refetchInterval: 30s`), but the **spend routing reads the live on-chain balance** at submit (#2234). So gating at input on the stale FE number can *block a spend that would actually succeed live*.
- Blocking early also defeats the natural overlap where the ~15s flow time lets the ~10–45s collateral rebalance settle — by submit, the funds have often landed.

So the right model is **fail-late**: gate only on the **displayed** balance (block when the amount exceeds the full visible balance — a true shortfall), let everything else proceed, and rely on live execution.

**What this commit does (net simpler than before):**
1. Gates back to the **displayed** balance via the single `hasSufficientSpendableBalance` predicate — **deletes** the `spendBlockReason` / `availableSpendableBalance` / `SPEND_BLOCK_MESSAGE` "settling at input" apparatus (and with it the Rain-mechanic copy).
2. **Refetch-on-failure (TanStack):** on `InsufficientSpendableError` the routing already re-read live smart balance, so we just invalidate `[RAIN_CARD_OVERVIEW_QUERY_KEY]` at the two throw sites — the FE de-stales immediately instead of waiting out the 30s poll.
3. **Informative failure copy:** the post-gate failure (in-transit not yet landed) now says *"Your balance isn't fully available yet — please try again in a few seconds,"* centralized in one constant, instead of a misleading "add funds."
4. Keeps the genuinely good parts: Aleks's display fix + the `formattedSpendableBalance` formatting consolidation.

Followed by a full `/code-review` pass over the assumptions and flows.

PR #2268, opened 2026-06-23
feat(add-money): flag BRL-via-PIX onramp as under maintenance (warn-only)

abalinda:
## Why
BRL deposits via PIX (the Manteca Brazil onramp) are currently **unstable / intermittently unavailable**. We want users to still see and use the option, but to clearly know it may be delayed or fail.

## What
A new **warn-only**, Brazil-scoped config lever — `underMaintenanceConfig.pixBrazilOnrampMaintenance` (`src/config/underMaintenance.config.ts`):

- **List tag** — a "Maintenance" badge on the **Pix** option in `/add-money/brazil`. The option stays **clickable** (warn-only).
- **In-flow banner** — a warning `InfoCard` at the top of the deposit screen (`/add-money/brazil/manteca`), shown only for Brazil (`selectedCountry.id === 'BR'`).

Why a new lever instead of the existing `disabledPaymentProviders: ['MANTECA']`: that one is too coarse — it would also kill the Argentina/ARS Manteca onramp **and** Manteca QR pay, and it hard-blocks rather than warns.

## Scope / notes
- **Frontend-only.** The API still accepts PIX deposits by design (warn-only). If a hard server-side stop is ever needed, that's a separate change.
- **To turn off** when PIX recovers: flip `pixBrazilOnrampMaintenance` to `false` in `underMaintenance.config.ts`.
- Shared copy (`PIX_BRAZIL_ONRAMP_MAINTENANCE`) keeps the badge + banner wording aligned.

## Base / history
Single commit, cut **clean onto `main`** (hotfix — the onramp is degraded now). This supersedes #2261, whose base got flipped dev→main and swelled to the whole dev→main delta (45 files); this PR is just the one intended change (6 files). main→dev back-merge tracked.

## Tests
- `InputAmountStep` renders the `maintenanceBanner` and keeps **Continue enabled** (warn-only).
- `AddWithdrawCountriesList` tags the Pix option "Maintenance" and it **still navigates** into the flow; no tag when the flag is off; non-Pix methods never tagged.
- Local gate: `prettier` ✓, `tsc` ✓ (0 errors), `jest` ✓ (add-money-states + AddWithdrawCountriesList + qr-pay-states, 73 passed). `next build` left to CI (authoritative; runs with submodules + env).

### DIFF
--- a/src/app/(mobile-ui)/add-money/__tests__/add-money-states.test.tsx
+++ b/src/app/(mobile-ui)/add-money/__tests__/add-money-states.test.tsx
@@ -1503,6 +1503,27 @@ describe('GROUP 8: InputAmountStep Component', () => {
         expect(screen.getByText('Continue')).not.toBeDisabled()
     })
 
+    test('renders maintenanceBanner and keeps Continue enabled (warn-only)', () => {
+        renderWithProviders(
+            <InputAmountStep
+                tokenAmount="100"
+                setTokenAmount={jest.fn()}
+                onSubmit={jest.fn()}
+                isLoading={false}
+                error={null}
+                setCurrencyAmount={jest.fn()}
+                limitsValidation={{ isBlocking: false, isWarning: false, currency: 'USD' }}
+                limitsCurrency="USD"
+                onBack={jest.fn()}
+                maintenanceBanner={<div data-testid="pix-maintenance">PIX deposits are under maintenance</div>}
+            />
+        )
+
+        expect(screen.getByTestId('pix-maintenance')).toBeInTheDocument()
+        // warn-only: the banner is informational and must not block submission
+        expect(screen.getByText('Continue')).not.toBeDisabled()
+    })
+
     test('Continue disabled when limits blocking', () => {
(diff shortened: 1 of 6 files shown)


abalinda:
Quick context on this one 👇

This is just the BRL-via-PIX "under maintenance" warning — a Maintenance badge on the Pix option + a heads-up banner in the deposit flow. It's warn-only, so deposits still go through; we just want people to know PIX might be flaky right now. One config flag (pixBrazilOnrampMaintenance) flips it off when PIX is healthy again.

Why a fresh PR instead of [#2261](https://github.com/peanutprotocol/peanut-ui/pull/2261): that branch was cut off dev, and when its base got switched dev → main the diff blew up to basically a whole prod release (~45 files) — none of which was actually my change, just everything dev had ahead of main showing through. The real change was always a single commit.

So I cherry-picked that one commit clean onto main here (verified identical via range-diff, 6 files). Local gate + CI are green, CodeRabbit's happy. #2261 can just be closed — nothing's lost there since all those other commits already live in dev.

#ran /peanut-pr

PR #2269, opened 2026-06-23
chore(add-money): PIX-BR onramp recovery toggle + maintenance-flag cleanup

Hugo0:
> **Draft — parked on purpose.** Merge this **when the BRL-via-PIX onramp (Manteca Brazil) is stable again.** Merging it flips PIX maintenance **off** in prod.

Follow-up to #2268.

## Why
#2268 shipped the warn-only "PIX under maintenance" tag + in-flow banner with the flag **on**. This is the matching recovery PR — ready in advance so flipping PIX back to normal is a one-merge action, not a scramble. It also cleans up two smells from the #2268 review.

## What
1. **Restore to not-under-maintenance** — `pixBrazilOnrampMaintenance: true → false` in `underMaintenance.config.ts`. The tag + banner disappear; the warn-only machinery stays in place, ready to flip back on (set the flag to `true`) if the onramp degrades again.
2. **One selector, no drift** — both surfaces (the list "Maintenance" tag and the in-flow banner) now read the maintenance state through a single `isPixBrazilOnrampUnderMaintenance()` selector instead of two separate `underMaintenanceConfig.pixBrazilOnrampMaintenance` reads. The maintenance determination lives in one place, so if the rule ever gains complexity (a time window, more regions) the two surfaces can't diverge. Each caller still keeps its own *context* check (the `pix-add` method / country `BR`).
3. **Robust test restore** — the maintenance test snapshots and restores the *shipped* flag value instead of hardcoding the `afterEach` restore to `true`. Previously, once the committed default became `false` (i.e. this PR), the old `afterEach` would have left the shared config singleton stuck at `true` for anything running after it.

## Base
Cut onto **`main`**, not `dev`: the `pixBrazilOnrampMaintenance` flag + UI only exist on `main` so far (the #2268 main→dev back-merge is still pending), so `main` is the only base where this is a clean, minimal diff. When the back-merge lands, this change rides along with it.

## Scope / notes
- **Frontend-only.** No API or schema changes.
- No behavior change while the flag is `false` beyond the maintenance UI being hidden — i.e. the pre-#2268 deposit experience.

## Tests (local gate, run in worktree)
- `prettier` ✓ (all 4 files unchanged after `--write`)
- `tsc` ✓ — 0 errors in changed files
- `jest` ✓ — `AddWithdrawCountriesList` (7) + `add-money-states` (41) pass; maintenance tests pass with the flag both on and off
- `next build` left to CI (authoritative — submodules + env)

### DIFF
--- a/src/app/(mobile-ui)/qr-pay/__tests__/qr-pay-states.test.tsx
+++ b/src/app/(mobile-ui)/qr-pay/__tests__/qr-pay-states.test.tsx
@@ -866,8 +866,8 @@ describe('GROUP 2: Payment Form States', () => {
     })
 
     test('Provider maintenance shows maintenance banner', async () => {
-        const maintenanceConfig = require('@/config/underMaintenance.config').default
-        maintenanceConfig.disabledPaymentProviders = ['MANTECA']
+        const underMaintenanceConfig = require('@/config/underMaintenance.config').default
+        underMaintenanceConfig.disabledPaymentProviders = ['MANTECA']
 
         setupMantecaPayment()
 
@@ -878,7 +878,7 @@ describe('GROUP 2: Payment Form States', () => {
         })
 
         // Clean up
-        maintenanceConfig.disabledPaymentProviders = []
+        underMaintenanceConfig.disabledPaymentProviders = []
     })
 })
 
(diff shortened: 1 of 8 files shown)


jjramirezn:
Obsolete: #2335 retires the PIX maintenance surface as part of shipping the dynamic-QR deposit flow (flag flips false there, and the snapshot/restore test fix from this PR is adopted).

PR #2297, opened 2026-06-29
feat(badges): fullscreen badge-receipt celebration (TASK-19791) [PARKED]

Hugo0:
**Status: PARKED (draft) — do NOT merge for the card launch.** Tracked so the work isn't lost; see #2292 for the direction we're shipping instead.

## What this is
TASK-19791 — a one-time fullscreen "badge unlocked!" moment (badge art + confetti + haptics + Continue), fired on the next `/users/me` refetch for any freshly-earned visible badge except `WAITLIST_SKIP`. FE-only: per-user `localStorage` seen-set + a 7-day freshness window (no backend, no migration). Done + tested — 19 unit tests, full suite green.

## Why it's parked
Fullscreen-every-badge collides with onboarding. Every badge that fires at/around launch is **incidental**: `BETA_TESTER` (signup), `SHHHHH` (everyone getting the card), `EVENT_ALUMNI` (event attendees), `NOT_SO_SHHHH` (influencers). A user going through `/shhhhh` card registration would get **2–3 fullscreen takeovers stacked** — "wtf, I wanted my card." A fullscreen *moment* only earns its interruption for badges from deliberate achievement (Card Pioneer after real spend, invite milestones) — none of which fire at launch.

## What we're shipping instead
Non-interruptive: the history item is already tappable (`BadgeStatusItem` → drawer), and **#2292** adds the improved shared `BadgeDetailModal`. That covers TASK-19791's "too subtle" intent without the collision. (A non-blocking earn toast is under consideration as a fast-follow.)

## Reviving later
For a genuine *achievement* badge that warrants a moment: rebase onto `dev`, gate the trigger to an allowlist of achievement codes (exclude all onboarding/event badges), then re-open for review.

Stacked on #2274 — base is `feat/card-share-sticker-collage` so the diff is just the badge commit; GitHub retargets to `dev` when #2274 merges.

### DIFF
--- a/src/app/ClientProviders.tsx
+++ b/src/app/ClientProviders.tsx
@@ -8,6 +8,7 @@
  */
 import { ConsoleGreeting } from '@/components/Global/ConsoleGreeting'
 import RainCooldownIntroModal from '@/components/Global/RainCooldown/IntroModal'
+import BadgeReceiptCelebration from '@/components/Badges/BadgeReceiptCelebration'
 import { ScreenOrientationLocker } from '@/components/Global/ScreenOrientationLocker'
 import { TranslationSafeWrapper } from '@/components/Global/TranslationSafeWrapper'
 import { PeanutProvider } from '@/config'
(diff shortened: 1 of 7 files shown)


Hugo0:
Update: the non-blocking badge-earn toast (the fast-follow noted above) is now implemented + tested in #2292 (commit 2c7456ca8). This PR (fullscreen) stays parked.

PR #2301, opened 2026-06-29
feat(card): 'Tweet to appeal' CTA on the waitlist cooldown screen

Hugo0:
## What
The waitlist-joined cooldown screen (`CardWaitlistJoinedScreen`) only had a "Back to home" button — a dead end. Per Konrad/Hugo it should let users **beg to jump the queue**.

- **Primary CTA: "Tweet to appeal"** — reuses the rejection screen's `pickRejectionCaption()` and fires the X intent tagging `@joinpeanut`. Text-only (no asset to capture on the friendly cooldown screen — same path the rejection screen falls back to on desktop).
- **Secondary: "Back to home"** (demoted to `variant="stroke"`).

## Notes
- Hotfix off `main` (launch is live). `dev` has an unreleased `BadgesRow` addition on this same file, so the back-merge will need to keep both BadgesRow + this CTA — flagged.
- No asset/image share here by design; if you want the full image-share (capture + native share) instead of a text tweet, say so and I'll wire it.

### DIFF
--- a/src/components/Card/CardWaitlistJoinedScreen.tsx
+++ b/src/components/Card/CardWaitlistJoinedScreen.tsx
@@ -21,6 +21,7 @@ import { PeanutWhistling } from '@/assets/mascot'
 import { useRouter } from 'next/navigation'
 import posthog from 'posthog-js'
 import { ANALYTICS_EVENTS } from '@/constants/analytics.consts'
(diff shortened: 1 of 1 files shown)


Hugo0:
Superseded: the waitlist cooldown screen is being removed (hotfix/waitlist-end-on-rejection). The rejection screen is now terminal and keeps the appeal asset + Tweet-to-appeal, so a separate cooldown CTA is moot.

PR #2316, opened 2026-07-01
fix(kyc): resume Sumsub SDK after PWA reload (TASK-20396)

kushagrasarathe:
## Problem — TASK-20396

Users report that mid-KYC, if they tab out of the PWA (to camera, gallery, or another app) and come back, they're dropped out of Sumsub and lose their progress. Reproduced on Android.

## Root cause

The app ships as a **standalone PWA** (`display: standalone`). Android aggressively evicts backgrounded standalone PWAs from memory — and opening the native camera/gallery (which the KYC flow pushes users toward) is a strong trigger. On return the page **cold-reloads**; there's no bfcache to restore.

The "SDK is open" state is `useState` in both KYC integrations, so a cold reload resets it and the SDK vanishes:
- `/card` (Rain card): gated by `sumsubToken !== null`.
- shared flow (add-money / withdraw / qr-pay / KYC drawer): gated by `showWrapper` in `useSumsubKycFlow`.

## Fix (approach A — URL-persisted resume)

Mirror the SDK-open state to the URL (`?kyc=true`) via nuqs. An app-switcher relaunch restores the last route + query, so the flag survives. On mount, if set, re-acquire a fresh token for the **same in-progress applicant** and reopen the SDK.

- **`/card`** — re-`applyForCard({ termsAccepted: false })` (the same idempotent call the token-refresh path uses).
- **shared flow** — re-`handleInitiateKyc()` in `useMultiPhaseKycFlow`, with `autoStart` so it lands straight in the SDK (user already consented) instead of the intro screen.

A sync effect keeps the URL flag in step with the SDK's open state (skipping the first run so the mount-time resume reads the persisted flag first).

## ⚠️ This is a PARTIAL fix — read before closing the ticket

**What this fixes:** the SDK reopening. After the reload, the user is taken straight back into Sumsub on the **same applicant**, with all previously-**submitted** steps preserved — instead of being dumped on the underlying page and having to navigate back and restart the flow.

**What this does NOT fix:** the data the user was **actively typing on the current step** (e.g. the personal-info / tax-ID form) is still lost. If they had filled fields but not yet advanced to the next step, those fields come back blank.

**Why we can't fix that part:**

- Sumsub's own docs confirm the SDK autosaves in-progress applicant-data / questionnaire input to the browser's **`sessionStorage`**, scoped to the verification session ([verification-steps docs](https://docs.sumsub.com/docs/verification-steps)): *"Sumsub automatically saves the information an applicant enters ... in the browser session storage ... If the session expires before the applicant finishes entering their information, they will need to re-enter their data when they return."*
- `sessionStorage` survives an in-tab reload but is **destroyed when the browsing context is torn down** — which is exactly what Android does when it kills the backgrounded PWA. That's why submitted steps survive (they're on Sumsub's server) but the mid-typed step does not.
- There is **no Sumsub config flag / init option / dashboard setting** to switch this to `localStorage` or make the draft durable (confirmed across WebSDK settings + get-started + about docs).
- We **cannot recover it ourselves** either: the Sumsub form runs in a cross-origin iframe (`static.sumsub.com`), so we can't read its `sessionStorage`, and the WebSDK only emits step-level events (`onStepCompleted`, `onApplicantSubmitted`) — there's no field-level/keystroke event to capture and re-push.
- We also can't reliably stop Android from killing a backgrounded standalone PWA, so preventing the reload isn't an option.

**The only way to fully preserve typed data** would be an architectural change: collect the personal-info / tax-ID fields in our **own** PWA form (persisted in our storage, which survives the kill), push them to Sumsub via the `fixedInfo` + `/questionnaires` API before the SDK opens, and enable prefill on the Sumsub level (`applicantDataSettings.prefill`). That's a separate, larger project touching FE + BE + Sumsub level config — tracked separately, not in this PR.

## Known limitation (summary)

Recovers the bounce-out + manual re-entry, **not** the fields on the step the user was actively typing. See the section above for why.

## Testing

- `npm run typecheck` clean, `npm run build` passes, prettier clean.
- `jest card` 137/137; `useSumsubKycFlow` / add-money-states / qr-pay-states suites pass (76).
- Manual (preview): start identity verification on Android PWA → tab to camera/gallery → return → SDK should reopen into the resumed applicant instead of the underlying page.

## Files

- `src/app/(mobile-ui)/card/page.tsx`
- `src/hooks/useMultiPhaseKycFlow.ts`
- `src/components/Kyc/SumsubKycModals.tsx`

### DIFF
--- a/src/app/(mobile-ui)/card/page.tsx
+++ b/src/app/(mobile-ui)/card/page.tsx
@@ -1,7 +1,6 @@
 'use client'
 import { type FC, useCallback, useEffect, useRef, useState } from 'react'
 import { notFound } from 'next/navigation'
-import { useQueryState, parseAsBoolean } from 'nuqs'
 import { useQuery, useQueryClient } from '@tanstack/react-query'
 import posthog from 'posthog-js'
 import { ANALYTICS_EVENTS } from '@/constants/analytics.consts'
@@ -26,6 +25,7 @@ import { useGrantSessionKey } from '@/hooks/wallet/useGrantSessionKey'
 import { useCapabilities } from '@/hooks/useCapabilities'
 import { useModalsContext } from '@/context/ModalsContext'
 import { useSafeBack } from '@/hooks/useSafeBack'
+import { useSumsubReloadResume } from '@/hooks/useSumsubReloadResume'
 
 // localStorage key for the one-time celebration gate (per-device by design:
 // re-doing the funnel re-celebrates, see the eligibility-check effect below).
@@ -76,13 +76,6 @@ const CardPage: FC = () => {
     // Sumsub card-application token — populated when POST /rain/cards reports
     // the user still needs to complete the rain-card-application level.
     const [sumsubToken, setSumsubToken] = useState<string | null>(null)
-    // Mirror "SDK is open" to the URL so it survives a PWA reload. Android
-    // evicts backgrounded standalone PWAs from memory (opening camera/gallery
-    // mid-KYC is the common trigger); on return the page cold-reloads and
-    // sumsubToken — being useState — resets to null, dropping the user out of
-    // Sumsub back to the card screen. With the flag in the URL, a reload can
-    // re-acquire a fresh token for the same applicant and reopen the SDK.
-    const [kycInProgress, setKycInProgress] = useQueryState('kyc', parseAsBoolean.withDefault(false))
     const [applyError, setApplyError] = useState<string | null>(null)
     // When backend returns status:'terms-required', we capture it here so
     // the dispatcher can render the terms screen between Sumsub and submit.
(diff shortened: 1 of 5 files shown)


kushagrasarathe:

Evaluate:
- code quality and DRY

kushagrasarathe:
Addressed the review in cdb191b:

- **#5 duplication + ordering** — extracted the shared `useSumsubReloadResume(isOpen, onResume)` hook (`src/hooks/useSumsubReloadResume.ts`). Both call sites (card page + `useMultiPhaseKycFlow`) now supply just `isOpen` + a resume callback; the hook owns the `?kyc` query param, the skip-first-run ordering, and the error/cleanup contract in one place.

- **#2 wrong-path resume + #3 stuck flag** — fixed with the same lever: `handleInitiateKyc` (and the `useSumsubKycFlow` implementation) now returns whether the SDK actually opened. The hook's `onResume` returns that boolean, and a resume that doesn't reopen clears the flag. So a reload mid self-heal / start-action / restart-identity no longer silently reopens the standard level — for those (base-approved) users `handleInitiateKyc` resolves without a token → returns `false` → flag cleared, SDK not reopened (graceful fallback to today's manual re-entry, not a regression). Same path fixes the stuck-flag-on-init-error case.

- **#4 `||` → `??`** — `SumsubKycModals` now uses `autoStart={autoStartSdk ?? flow.sdkAutoStart}` so an explicit `false` from a caller wins.

- **#1 scope** — the PR description was updated earlier; the shared-flow changes are intentional (they match the commit).

Remaining known edge (documented, accepted): `restart-identity` for a Manteca wrong-country user could resume into the standard identity level rather than the exact restart level — narrow remediation case, and re-doing identity is close to intended. Not worth per-flow arg persistence here.

Verified: typecheck clean, `npm run build` passes, prettier clean, `jest` 213 passed across the KYC/card suites.

kushagrasarathe:

Evaluate:
- code quality and DRY

kushagrasarathe:
Addressed round-2 in a82a0795a.

**`sdkAutoStart` never resets (valid, fixed).** Note the suggested one-liner (reset inside `handleInitiateKyc`) is actually racy — the resume calls `setSdkAutoStart(true)` *before* `handleInitiateKyc()`, so a reset inside it would clobber the resume's own autostart. Fixed it correctly instead:

- On a **failed** resume (`handleInitiateKyc()` returns `false`, SDK never opened) → disarm immediately in the resume callback.
- On the **falling edge** of `showWrapper` (resumed session closed) → disarm via a `prevShowWrapperRef` effect that fires only on `true → false`, so it never disarms the pending resume open (no race with the intermediate `showWrapper=false` renders while the resume's re-init is in flight).

Result: only the reload-resume open skips `StartVerificationView`; a later retry-after-rejection / self-heal / start-action / restart-identity in the same session shows it again. Covers your new QA scenario #3.

**`/card` doesn't wire `autoStart` (minor) — intentional for now.** Making card's resume seamless would mean duplicating this same arm/disarm state machine (fresh card opens must still show the consent view, only resume skips it) into the card page's bespoke integration. For a separate flow the reporters aren't hitting, that complexity isn't worth it — the extra "Start Verification" tap on card resume is harmless (it's a light consent screen). Easy follow-up if we want parity later.

Verified: typecheck clean, `npm run build` passes, prettier clean, `jest` 213 passed across the KYC/card suites.

kushagrasarathe:

Evaluate:
- code quality and DRY

kushagrasarathe:
Round-3 nit addressed in 836c3a2c9 — resume-triggered initiations now carry `resumed: true` on the `KYC_INITIATED`/`MANTECA_KYC_INITIATED` capture, so a PWA-reload resume is distinguishable from a genuine new session in the funnel. No signature change (ref-based, read at the synchronous capture, reset after the resume settles).

That was the last open item — CodeRabbit's round-3 review confirmed all prior findings resolved with no blocking concerns. PR is ready for human review.

PR #2329, opened 2026-07-02
fix: surface bridge deposit errors instead of failing silently

kushagrasarathe:
## Summary

EEA-uplift users whose Bridge KYC docs are under review were creating deposits that failed with **zero feedback** — they retried blind (user recording confirmed; ~369 prod users currently sit in the waiting-on-provider state). Three silent-failure holes on the bank deposit path:

1. `useCreateOnramp` threw a hardcoded generic message without parsing the backend's `{ error }` body (the comment claimed it parsed; it didn't). Backend reasons like *"Could not create transfer: customer is under review"* were discarded.
2. The page's catch gated the visible error on the hook's `error` state, which hasn't flushed when the synchronous catch runs — always `null` on first attempt, so nothing displayed.
3. `gate.kind === 'waiting-on-provider'` silently no-op'd the Continue button — no "under review" UX existed anywhere.

Fix: parse the error body (`error`/`message` keys, generic fallback), show the caught message directly in the catch, and show a friendly "your verification details are being reviewed" message on the waiting-on-provider gate.

## Risks / breaking changes

- FE-only; no backend or contract changes.
- Backend error strings are now shown verbatim to users — they're already user-appropriate on this route ("Bridge Terms of Service must be accepted…", "Additional KYC may be required"), and any string beats the silent failure.
- Blast radius: bank deposit page + one hook (single consumer).

## QA

- `npm test` — 107 suites green, incl. new `useCreateOnramp` hook tests + 2 page tests (error surfacing with unflushed hook state; waiting-on-provider message instead of dead button).
- The strengthened page test reproduces the incident: old code fails it (hook `error: null` → no `setError` call), new code passes.
- Manual repro path: sandbox user with Bridge rail in `REQUIRES_EXTRA_INFORMATION` + `kyc_approval` wait signal → deposit → Continue now shows the under-review message.

### DIFF
--- a/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
+++ b/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
@@ -13,7 +13,7 @@ import { useAuth } from '@/context/authContext'
 import { useCapabilities } from '@/hooks/useCapabilities'
 import { getKycModalVariant, getGateUserMessage } from '@/utils/capability-gate'
 import { useModalsContext } from '@/context/ModalsContext'
-import { useCreateOnramp } from '@/hooks/useCreateOnramp'
+import { useCreateOnramp, GENERIC_ONRAMP_ERROR } from '@/hooks/useCreateOnramp'
 import { useParams, useSearchParams } from 'next/navigation'
 import { useCallback, useEffect, useMemo, useState } from 'react'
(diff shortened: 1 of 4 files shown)


kushagrasarathe:
@coderabbitai review

kushagrasarathe:
Superseded by #2345 — consolidated with the EEA-uplift funnel change and the new Bridge re-verification pending modal into one PR targeting main (per request).

PR #2332, opened 2026-07-02
feat(analytics): fire eea_uplift funnel on modal open, cover all remediation

kushagrasarathe:
## Summary

Makes the `eea_uplift_started` / `eea_uplift_completed` funnel reliable for watching who **attempts** the EEA uplift and whether they finish — so session recordings can be filtered by the event. Two gaps fixed:

1. **Fired on the modal CTA, not on open** — anyone who opened the uplift modal and abandoned emitted nothing. Now fires at **modal-open**.
2. **Missed the urgent cohort entirely** — post-`2026-06-29`-cliff, the urgent uplift is no longer a future-dated advisory; the BE reclassifies it to **blocking** (`fixable-rejection` → KYC modal), a path that called no tracking. Now covered.

Fires across **both remediation paths** and **both channels** (deposit + withdraw):
- **urgent / blocking** → `gate.reason.code` starts with `eea_uplift` (BE sets reason code to the questionnaire cluster).
- **upcoming / advisory** → `requirementKey` ∈ {`sof_individual_primary_purpose`, `has_foreign_tax_registration`, `place_of_birth_missing`, `nationalities`} — derived from prod remediation data (172 EEA-affected users).

Event now carries `source` (advisory|blocking), `urgent` (blocking = past the cliff = urgent), and `effective_date`, so **urgent vs upcoming (dob-type)** split directly in PostHog.

**Excluded** `proof_of_address_document` / `government_id_expired`: they co-occur on the cliff but are separate remediation (cluster `null`) — keying them to eea would over-fire on ordinary doc rejections for non-EEA users.

## Risks

- Analytics-only; no user-visible behavior change. No BE change.
- Firing moved from CTA to modal-open ⇒ `started` counts rise (now includes abandoners) — that's the intent. `completed` still latches on a recorded start, so completion rate stays meaningful.
- Slight limitation: a blocking uplift whose reason resolves to `document_rejected` (freeform classifier) instead of `eea_uplift` wouldn't be tagged — best signal available FE-side without a BE flag.

## QA

- `npm test` — 107 suites green; new `eea-uplift.utils` (10) + rewritten `useEeaUpliftFunnel` (6) tests, incl. key-set match, reason-code prefix, urgent flag, and the co-occurring-non-uplift exclusion.
- Separate from the deposit silent-failure fix (#2329). Both touch the two bank pages; if #2329 merges first I'll rebase (small, different regions of `handleAmountContinue`).

### DIFF
--- a/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
+++ b/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
@@ -37,7 +37,7 @@ import { InitiateKycModal } from '@/components/Kyc/InitiateKycModal'
 import AdvisoryPreemptModal from '@/components/Kyc/AdvisoryPreemptModal'
 import { useAdvisoryPreempt } from '@/hooks/useAdvisoryPreempt'
 import { useEeaUpliftFunnel } from '@/hooks/useEeaUpliftFunnel'
-import { eeaUpliftReasonCode, isEeaUpliftAdvisory } from '@/utils/eea-uplift.utils'
+import { upliftTriggerFromGate, upliftTriggerFromAdvisory } from '@/utils/eea-uplift.utils'
 import posthog from 'posthog-js'
 import { ANALYTICS_EVENTS } from '@/constants/analytics.consts'
 import { addMoneyCountryUrl } from '@/utils/native-routes'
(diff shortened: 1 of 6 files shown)


kushagrasarathe:
@coderabbitai review

kushagrasarathe:
Superseded by #2345 — consolidated into one PR targeting main (per request).

PR #2336, opened 2026-07-02
fix(kyc): back off verification-progress poll instead of a fixed 5s interval

Hugo0:
## Summary

**Incident (2026-07-02):** while the KYC verification-progress modal is open, `useSumsubKycFlow` fired `initiateSumsubKyc` on a fixed `setInterval(..., 5000)` as a websocket fallback — for the *entire* modal-open. That endpoint is **mutating**: for approved-LATAM users in the self-recovery state each call re-runs a full provider submission (**86 re-submissions in 20 min** for one user). Even with the new BE cooldown (api #1106), each poll still costs ~3 Sumsub API calls + DB writes on the create-action fall-through.

**Why time-based, not error-based:** the poll requests **succeed (HTTP 200)** even when the backend reprocess fails, so an error-count backoff would never escalate. Backoff must be purely time-driven.

**The change (hook-only):** replace the fixed 5s `setInterval` with a self-rescheduling `setTimeout` chain on a time-escalating schedule, plus an overall cap:

| elapsed since modal opened | poll cadence |
|---|---|
| 0 – 60s | 5s (fast path for the common quick transition) |
| 60 – 120s | 10s |
| 120 – 180s | 20s |
| 180s+ | 60s |
| ≥ ~15 min | **stop polling entirely** |

After the ~15 min cap the modal stays in its existing long-running "Almost there" state (with its "Go to Home" CTA) and the websocket remains the only signal. Re-opening the modal restarts polling fresh. Cleanup on modal close / unmount cancels the pending timer (mirrors the old `clearInterval` semantics; adds an in-flight guard so a poll resolving after teardown can't re-arm).

Untouched by design: the websocket path, `handleInitiateKyc`, and the initial `fetchCurrentStatus` effect. **Zero behavior change for flows that transition within the first minute** (the common case).

## Risks / blast radius

- **KYC verification UX only.** Worst case after the first minute: a status transition that the websocket misses is noticed up to 60s late (previously up to 5s). The websocket remains the **primary** signal, so in practice transitions still surface promptly; this poll is only the fallback.
- No API/schema/contract change; no cross-repo deploy ordering. FE-only.

## QA

- **Automated:** 5 new fake-timer hook tests in `src/hooks/__tests__/useSumsubKycFlow.test.ts` — cadence escalates 5s→10s→20s on schedule, polling stops at the cap, timer cleanup on modal close, timer cleanup on unmount, poll still calls `initiateSumsubKyc` with the same `{ regionIntent, levelName, targetCountry }` args. Full suite: 106 suites / 1645 passing.
- **Manual:** open the verification-progress modal (trigger the SDK-submitted state), watch the Network tab — the initiate calls should fire at 5s for the first minute, then space out to 10s → 20s → 60s, and stop after ~15 min. Confirm the modal stays visible after the cap and that the websocket still drives the completion transition.

### DIFF
--- a/src/hooks/__tests__/useSumsubKycFlow.test.ts
+++ b/src/hooks/__tests__/useSumsubKycFlow.test.ts
@@ -340,7 +340,7 @@ describe('useSumsubKycFlow — verification-progress poll backoff', () => {
     })
 
     it('cancels the pending timer on unmount', async () => {
-        const { result, unmount } = openModal()
+        const { unmount } = openModal()
 
         await act(async () => {
             await jest.advanceTimersByTimeAsync(5_000)
--- a/src/hooks/__tests__/useSumsubKycFlow.test.ts
+++ b/src/hooks/__tests__/useSumsubKycFlow.test.ts
@@ -230,3 +230,141 @@ describe('useSumsubKycFlow — terminal-error exits clear the user-initiated gua
         await waitFor(() => expect(onKycSuccess).toHaveBeenCalledTimes(1))
     })
 })
+
+// Incident 2026-07-02: while the verification-progress modal was open, this hook
+// fired initiateSumsubKyc — a MUTATING endpoint — on a fixed 5s setInterval for
+// the entire modal-open, re-running provider submissions for approved-LATAM
+// self-recovery users (86 in 20 min for one user). The poll now backs off on a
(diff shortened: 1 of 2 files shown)


Hugo0:
@coderabbitai review

Hugo0:
@coderabbitai review

PR #2342, opened 2026-07-03
feat(kyc): provide-email sheet for the no-email verification dead-end

Hugo0:
## Summary
Companion to **peanutprotocol/peanut-api-ts#1119**. Users whose email was never captured dead-end on 'message support' — the only recovery for all 61 affected users (none have a Sumsub email either; verified against prod). The BE emits a `provide-email` NextAction for email-blocked rails; this PR renders it:

- `provide-email` added to `NextActionKind` + capability gate (`GateState` variant, checked before `restart-identity` in the blocked branch).
- New `ProvideEmailStep` (ActionModal + email field, BridgeTosStep shell): validates, POSTs `update-user` (first-time email set is allowed while KYC-locked), refetches capabilities. BE flips rails → PENDING and resubmits automatically.
- Wired into both surfaces where the block manifests: home `ActivationCTAs` (detects `reason.code === 'email_required'`, opens the sheet instead of the support drawer) and the bank add/withdraw gate (`AddWithdrawCountriesList`, mirrors accept-tos).

## Risks / breaking changes
- **Deploy order: BE (#1119) first.** Until the BE emits the action, nothing changes here (gate branch never matches). If this FE lags the BE, blocked users see the generic state — same as today, no crash.
- FE-only otherwise; no new deps.

## QA
- Unit: 1667 pass incl. 2 new gate tests (provide-email wins the blocked branch; plain blocked still dead-ends). Typecheck + build green.
- Manual path once BE is live: user with email-blocked rail → home CTA or bank flow → email form → save → rail goes PENDING → normal in-progress UI.

### DIFF
--- a/src/components/AddWithdraw/AddWithdrawCountriesList.tsx
+++ b/src/components/AddWithdraw/AddWithdrawCountriesList.tsx
@@ -34,6 +34,7 @@ import { railJurisdictionForBank } from '@/utils/bridge.utils'
 import { getRegionIntent } from '@/utils/regions.utils'
 import { useTosGuard } from '@/hooks/useTosGuard'
 import { BridgeTosStep } from '@/components/Kyc/BridgeTosStep'
+import ProvideEmailStep from '@/components/Kyc/ProvideEmailStep'
 import { useModalsContext } from '@/context/ModalsContext'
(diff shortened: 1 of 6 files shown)


Hugo0:
@coderabbitai review

PR #2346, opened 2026-07-03
fix(offramp): let the offramp campaign skip the invite-code gate at signup

kushagrasarathe:
## Summary

Offramp-campaign arrivals (`peanut.me/invite?utm_campaign=offramp`) were being asked for an **invite code** at signup — even though the campaign is supposed to *replace* the invite. Root cause is **frontend-only**:

- The backend already treats `OFFRAMP_USER` as invite-equivalent — it's in `SKIP_BADGE_CODES` **and** `FULL_BYPASS_BADGE_CODES`, so awarding it flips `hasAppAccess` (the invite jail) + card-waitlist skip.
- `useZeroDev.handleRegister` already registers a **campaign-only** user (no invite code) → `awardBadge(campaignTag)`.
- But `src/app/(setup)/setup/page.tsx` only jumped to the `signup` step when an **invite-code** cookie was present. An offramp user (has `campaignTag` cookie, no invite code) fell through to the `welcome` invite-code gate.

**Fix:** `/setup` now treats a **bare-claim campaign** cookie (offramp, skip, event_alumni + the vanity badges — the exact set `/invite` already lets you claim *without* an invite) like an invite code: straight to `signup`, no invite prompt. Uses the existing, unit-tested `classifyBareCampaign`. Makes `/setup` consistent with `/invite`.

Result: `/invite?utm_campaign=offramp` → Claim your badge → `/setup` → **signup (passkey), no invite code** → badge awarded + app access granted.

## Note on scope

An earlier commit on this branch embedded the campaign in the desktop QR + added an app-wide capture hook. Once the campaign acts as the invite code, the user goes straight to signup and never reaches the QR screen, so that was moot (and the app-wide hook was over-broad — it captured skip codes on every page). **Net change is one file** — the QR/hook approach is reverted within the branch.

## Risks / breaking changes

- FE-only, no API/contract change. No cross-repo deploy ordering.
- Behavior change: any **bare-claim campaign** cookie now skips the invite-code gate at `/setup` (previously only invite codes did). This is intentional and matches `/invite`'s existing classification — these campaigns are *designed* to be claimed without an invite. Non-bare campaigns still require an invite code.
- `skip` / `event_alumni` / `offramp` granting access from a shared link is the existing product deal for those cohorts (unchanged).

## QA

1. Desktop & mobile: `/invite?utm_campaign=offramp` → **Claim your badge** → lands on the passkey **signup** screen directly, **no invite-code prompt**.
2. After signup: `OFFRAMP_USER` badge awarded, `hasAppAccess` true (no waitlist).
3. Regression: a normal organic `/setup` (no invite, no campaign) still shows the landing → invite-code gate. An invite-code user still jumps straight to signup.
4. `campaign-maps.test.ts` (38) covers `classifyBareCampaign` incl. `offramp_user` → bare-claim.

Base **main** (hotfix) — back-merge to `dev` after merge.


kushagrasarathe:
**CodeRabbit nitpicks addressed:**
1. `57-58` (pass invite code explicitly to `classifyBareCampaign`) — ✅ applied in 6af29f703.
2. `156-159` (bounds-check `signupScreenIndex !== -1`) — declining: the `signup` step is always present in `steps`, so this guards an impossible case, and the pre-existing invite-code branch never had it. Per repo rules (no error handling for impossible scenarios / no unrelated scope creep) leaving as-is.

jjramirezn (approved):

PR #2347, opened 2026-07-03
feat(offramp): require the migrant's Offramp username/email before revealing the deposit address

jjramirezn:
## Summary

Follow-up to #2346 and companion to **peanutprotocol/peanut-api-ts#1127**. The public offramp campaign link awards the $10 OFFRAMP_USER badge to anyone who clicks it — fine for the $10, but Peanut owes the partner **$30+$25 per _actual_ migrant** and today we can't tell which badge holders are real offramp.xyz users (Konrad's ask, 2026-07-03).

This gates the migration deposit screen (`/add-money/crypto?source=offramp`) behind a **required input**: the migrant's offramp.xyz username or email.

- New `OfframpHandleGateView`: NavHeader + copy + input + Continue (disabled until ≥3 chars). Saves via the existing `updateUserById` server action (`POST /update-user { offrampHandle }`), fires `offramp_handle_submitted` PostHog event, then reveals the deposit address.
- Skipped entirely once `user.user.offrampHandle` is set (new field on `/users/me`, BE #1127) — one-time friction per user.
- Value is deliberately **unverified**: per Konrad, asking is deterrent enough; reconciliation with the partner happens later via `SELECT username, offramp_handle FROM app.users WHERE offramp_handle IS NOT NULL`.

## Risks / breaking changes

- **Cross-repo deploy ordering: peanut-api-ts#1127 must deploy first.** Fastify strips unknown body fields, so FE-first would silently drop the handle (gate would re-appear on next visit; no data stored).
- Scope: only the `source=offramp` variant of the crypto deposit page — normal crypto deposits unaffected.
- While the auth user is still loading, offramp visitors see a spinner instead of a gate flash; migrants with a stored handle go straight through.
- Base **main** (hotfix path, matching #2346) — back-merge to `dev` after merge.

## QA

- 4 new tests in `add-money-states.test.tsx` (GROUP 4b): gate blocks address for migrants without a handle; submit trims + saves + reveals; save failure keeps gate up with error copy; stored handle skips gate.
- Local gate: prettier ✅ typecheck ✅ jest 47/47 in the suite (full run has 2 pre-existing env-only failures: locale-dependent `10,000 USD` assertion + gitignored `public/flags` in fresh worktree; both pass with en_US locale / postinstall) ✅ `next build` ✅.

Manual QA path: log in as a user with the OFFRAMP_USER badge → Add money → Migrate from Offramp → gate appears → enter handle → deposit QR revealed; revisit → no gate.

### DIFF
--- a/src/app/(mobile-ui)/add-money/__tests__/add-money-states.test.tsx
+++ b/src/app/(mobile-ui)/add-money/__tests__/add-money-states.test.tsx
@@ -584,6 +584,14 @@ jest.mock('@/components/AddMoney/hooks/useCryptoDepositPolling', () => ({
     useCryptoDepositPolling: (...args: any[]) => mockUseCryptoDepositPolling(...args),
 }))
 
+// updateUserById — the offramp handle gate persists the migrant's offramp.xyz
+// username/email through this server action
+const mockUpdateUserById = jest.fn()
+jest.mock('@/app/actions/users', () => ({
+    ...jest.requireActual('@/app/actions/users'),
+    updateUserById: (...args: any[]) => mockUpdateUserById(...args),
+}))
+
 // Country list
 jest.mock('@/components/Common/CountryList', () => ({
     CountryList: (props: any) => (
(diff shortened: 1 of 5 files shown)


kushagrasarathe (approved):
lgtm

PR #1132, opened 2025-08-26
Fix Daimo bugs

Zishan-7:
Contributes to tasks - 
TASK-14251 
TASK-14252 
TASK-14247 
TASK-14248 
TASK-14250  
TASK-14253

PR #2351, opened 2026-07-03
fix(home): activation-rejection CTA respects users who can already transact

Hugo0:
## Summary
The home activation-rejection override (`ActivationCTAs`) showed "Complete your setup → Upload document" (fixable) or "Verification issue" (blocked) to **any** Sumsub-approved user with a rejected bank/qr rail — even users who already hold an active card or another enabled rail, or whom the BE has marked `isActivated`. For a card-holder that's a nag on an optional capability they don't need (and, for a terminally-rejected bank rail, can't fix).

Gate the override on `!canAlreadyTransact`, where `canAlreadyTransact = rails.some(status === 'enabled') || user.isActivated`. The card's own rail reads `enabled`, so a card-holder is covered. A genuinely-fixable bank RFI is unaffected on the home screen only — it still surfaces in context in the `/add-money/[country]/bank` flow, which runs its own capability gate.

## Risk
Low. One added boolean guard on an existing memo; the fixable/blocked branches themselves are untouched. Covered by new tests.

## Cross-repo
Pairs with the peanut-api-ts hotfix *terminal Bridge → blocked* (peanutprotocol/peanut-api-ts #1132). Independent — either can ship first. The API fix flips a terminal user from the "fixable/upload" card to the "blocked/contact-support" card; this FE fix suppresses **both** for users who can already transact.

## QA
`npm test` — new `src/components/Home/__tests__/ActivationCTAs.test.tsx`:
- card-holder (enabled card rail) + rejected bank rail → no "Complete your setup"; falls through to the normal funnel step
- BE-activated user + rejected bank rail → no nag
- user with no working rail + fixable rejection → still nagged (behavior unchanged)

### DIFF
--- a/src/components/Home/ActivationCTAs.tsx
+++ b/src/components/Home/ActivationCTAs.tsx
@@ -126,10 +126,27 @@ export default function ActivationCTAs({ activationStep, onDismissCard }: Activa
         }
     }, [activationStep])
 
+    // A user who can already transact — they hold an active card (its rail reads
+    // `enabled`), have any other enabled rail, or the BE has marked them
+    // activated — is NOT mid-activation. A rejected *bank* rail is then an
+    // optional extra capability, not a setup blocker, so the home activation CTA
+    // must stand down. A genuinely-fixable bank RFI still surfaces in context in
+    // the /add-money bank flow (which runs its own gate). Without this, a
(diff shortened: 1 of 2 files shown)


Hugo0:
@coderabbitai review

PR #2353, opened 2026-07-03
Update content submodule

Hugo0:


0xkkonrad:
Superseded by #2375, which publishes the same content (and everything after it) straight to `main`/production. These `dev`-targeted submodule bumps never reach production (`dev` is ~39 commits behind `main` and doesn't promote), so closing to avoid a stale-pointer merge. Content itself is safe in peanut-content.

PR #2362, opened 2026-07-04
content(help): fees-pricing — one all-in rate, always shown, live widget (prod)

0xkkonrad:
Publishes peanutprotocol/mono#38 to prod (same pipeline as #2355): bumps `src/content` to peanut-content `40a1a5b` (mirror of mono@2f37590).

## What ships (en / es-419 / pt-br)

- Fees page drops all "spread" language, the pinned ~0.50%/~0.80% percentages, and both provider-attribution claims ("passed through at cost" / "not from Peanut"). The pinned EUR figure also contradicted `product/pricing.md`.
- New framing: **one all-in exchange rate** that moves with the market and **is always shown** — what you see is what you get.
- `<ExchangeWidget />` now embedded on the page in every locale (ARS for en/es-419, BRL for pt-br) — the transparency proof is the live rate itself, not cost descriptions.
- PT-BR uses "cotação" for the quote sense (fixes the taxa = fee/rate collision); ES drops the "spread" anglicism and gets a locale-correct pricing link.
- Card pricing section from #2355 is untouched.

## Verify on preview

- `/en/help/fees-pricing` (+ `/es-419/`, `/pt-br/`): widget renders and fetches a rate in the Exchange Rates section; no layout break in the prose column.
- Anchor `#the-peanut-card` still lands on the card section.

Merging this deploys to prod.


0xkkonrad:
Superseded by #2375 — that bump goes to peanut-content `0a87f45`, which is later than this PR's `40a1a5b` and already includes the fees rate-always-shown change. Merging #2375 publishes this and everything after it to production in one go.

PR #2376, opened 2026-07-07
Update content submodule

Hugo0:


0xkkonrad:
Superseded by #2379, which bumps the same `src/content` pointer (peanut-content@1f82265) directly on `main` (production). These auto-PRs target `dev`, which never promotes.

PR #2380, opened 2026-07-07
Update content submodule

Hugo0:


0xkkonrad:
Dead-end: targets dev, which never promotes to production. Content ships via the publish-to-main PR instead.

PR #2383, opened 2026-07-07
feat: make transaction counterparty name open their profile

0xkkonrad:
## Summary

Tapping a peer's **name in the Activity row** now deep-links to that person's profile, so you can immediately send or request again — the exact "get back to this person" flow. Before, the row's name was inert: only tapping the whole card opened the details drawer.

This reuses the two things the codebase already had for exactly this:
- `profileUrl()` — the app-wide way to reach a profile (Invites / Rewards / Contributors / the drawer header).
- `VerifiedUserLabel`'s existing `onNameClick` prop — the clickable-name affordance the receipt header already uses. The row passes a handler; no new click surface is built.

The eligibility rule the receipt header had inline (`isAvatarClickable`) is extracted into a shared `hasUserProfile()` predicate — a *fact* about the counterparty ("a real Peanut user with a profile: non-link send/request/receive to a username, not a raw crypto address"), so the row and the drawer header can't drift. What each surface does with the fact (clickable name, clickable avatar) stays a call-site decision under its own local name.

- Name tap navigates to the profile; `VerifiedUserLabel` stops propagation internally whenever `onNameClick` is set, so a name with its own action can never also fire its container (here: the drawer). The rest of the row still opens the drawer, including blank space next to a short name — the tap target is the name text itself.
- Only eligible for a non-link **send / request / receive** to a **username**. Address detection uses `isCryptoAddress` (EVM **+ Solana + Tron**) — the same rule `VerifiedUserLabel` renders by — not EVM-only `isAddress`, so a Solana/Tron counterparty can't slip through to a garbage `/<address>` URL. This also hardens the receipt header's avatar, which shared the narrower check.

## Repro (the bug this fixes)

On a completed send to a Peanut user, opening the transaction details shows the recipient's name (e.g. **"natalia"**) — but the name was **not clickable**, so there was no one-tap way back to that person to send/request again.

_Screenshot of the repro to be attached._

## Files (production: +50 / −11)
- `transaction-predicates.ts` — new `hasUserProfile()` (single source of truth for the eligibility fact).
- `UserHeader/index.tsx` — `VerifiedUserLabel` stops propagation when `onNameClick` is set (the header, its only prior consumer, has no parent handler — behaviour there unchanged).
- `TransactionCard.tsx` — passes `onNameClick` to the existing `VerifiedUserLabel`; no wrapper click-div.
- `TransactionDetailsReceipt.tsx` — `isAvatarClickable` now calls the predicate (net −4; EVM behaviour identical, Solana/Tron now correctly ineligible).
- `__tests__/transaction-predicates.test.ts` (+49, eligibility gate incl. EVM/Solana/Tron address rows) & `__tests__/TransactionCard.test.tsx` (real-component RTL: name→profile with no drawer, non-name→drawer, ineligible-inert).

## Design note — why the existing prop, not a new abstraction
First version hand-rolled a click-div around `VerifiedUserLabel`; review caught that the label already exposes `onNameClick` (used by the receipt header, gated on the same predicate). Reusing it deleted the wrapper, tightened the tap target to the name text, and made address-shaped names inert by construction (the label renders `AddressLink` for those and ignores `onNameClick`). The remaining overlap between row and header is a single `router.push(profileUrl(username))` line — below the abstraction threshold; no shared hook.

_Intentional divergence:_ a name tap in the **row** fires haptic (consistent with the row's other taps); the **header** name/avatar tap stays haptic-free (it never buzzed). Not an oversight — do not "fix" it by reaching into the header handler.

## Risk
Low, FE-only, no API/contract change, no new deps. Blast radius = the counterparty name in the Activity feed (`TransactionCard`, shared by Home + `/history` + pending), the receipt header (avatar now also ineligible for Solana/Tron counterparties), and `VerifiedUserLabel`'s name-click propagation (only prior consumer verified unaffected).

## QA
- `npx jest src/components/TransactionDetails` — 160 green (predicate 13 address/eligibility cases + `TransactionCard` RTL 3/3 mounting the real component: name→`/natalia` with no drawer, non-name→drawer, ineligible→inert).
- Full suite: 110 suites / 1698 tests green. `npm run typecheck` — clean (exit 0).
- Manual (sandbox): open the Activity feed, tap a peer's name on a completed send → lands on their profile with Send/Request. Tap elsewhere on the row → details drawer still opens. A link/withdraw/address row's name stays inert.

> Note: local `next build` OOMs on this box (known infra heap limit, not this change — typecheck already validates imports/types); CI build is the confirmation.

### DIFF
--- a/src/components/TransactionDetails/TransactionCard.tsx
+++ b/src/components/TransactionDetails/TransactionCard.tsx
@@ -5,7 +5,11 @@ import TransactionAvatarBadge from '@/components/TransactionDetails/TransactionA
 import { getBankAccountCountryCode } from '@/constants/countryCurrencyMapping'
 import { type TransactionDirection, type TransactionType } from '@/components/TransactionDetails/transaction-types'
 import { type TransactionDetails } from '@/components/TransactionDetails/transactionTransformer'
-import { isCardPaymentEntry, isPerkReward } from '@/components/TransactionDetails/transaction-predicates'
+import {
+    canNavigateToUserProfile,
+    isCardPaymentEntry,
+    isPerkReward,
+} from '@/components/TransactionDetails/transaction-predicates'
 import { useTransactionDetailsDrawer } from '@/hooks/useTransactionDetailsDrawer'
 import {
     formatNumberForDisplay,
@@ -28,6 +32,8 @@ import { useHaptic } from 'use-haptic'
 import LazyLoadErrorBoundary from '@/components/Global/LazyLoadErrorBoundary'
 import { PEANUTMAN } from '@/assets/mascot'
 import InvitesIcon from '../Home/InvitesIcon'
+import { useRouter } from 'next/navigation'
+import { profileUrl } from '@/utils/native-routes'
 
 // Lazy load transaction details drawer (~40KB) to reduce initial bundle size
 // Only loaded when user taps a transaction to view details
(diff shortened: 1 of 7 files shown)


0xkkonrad:
@coderabbitai review

kushagrasarathe:

Evaluate:
- code quality and DRY

kushagrasarathe:

Evaluate:
- code quality and DRY

PR #2386, opened 2026-07-08
fix(kyc): cross-region 'Unlock region' loop + wrong copy + homepage self-heal dead-end

kushagrasarathe:
- fixes TASK-20564

Fixes the KYC deposit issues from Hugo's Discord report + a related homepage dead-end. Targets **main** (hotfix — live, customer-facing). FE-only.

## The bug (verified on prod: username `hugo` / `c511d99c`)
Hugo picks a bank deposit → gets **"Unlock United States — Your ID is already on file. To unlock payments here, we need a valid ID from United States"** → taps **Submit document** → sees **"You're unlocked"** → bounces back → same modal → loop.

His actual state: SUMSUB **APPROVED**, and he **already has** Bridge rails (US ACH, UK, SEPA, SPEI) — all in `REQUIRES_EXTRA_INFORMATION`, each with `bridgeRemediation.nextAction = DOCUMENT_SELF_HEAL / requirementKey: government_id_document` (Bridge wants his ID re-uploaded; it's expired/invalid).

**Root cause:** the backend resolver emits **no capability action** for a generic government-id self-heal (it expects the FE to call the resubmit endpoint directly). So in `deriveGate`, that in-scope `requires-info` rail matches none of the branches and **falls through to `needs-enrollment` → the `cross_region` "Unlock {region}" modal**. Tapping Submit runs a *cross-region enrollment* (wrong flow) that flashes "You're unlocked" and changes nothing → loop.

## The fixes (3)
1. **`capability-gate.ts`** — if an in-scope rail is `requires-info` (the rail **exists**, the user isn't missing a region), route to **`fixable-rejection`** instead of falling through to `needs-enrollment`. `fixable-rejection`'s `onVerify` already calls `handleSelfHealResubmit('BRIDGE')` → the document re-upload flow. `needs-enrollment` (cross_region) is preserved for the genuine "no rail in this scope" case.
2. **`InitiateKycModal.tsx`** — the `cross_region` copy claimed "we need a valid ID **from {region}**", implying a country-specific ID. False — a European doesn't need a US ID for US bank transfer. Replaced with honest, region-agnostic copy: *"Your identity is already verified. To turn on payments in {region}, we just need to confirm a few more details."*
3. **`ActivationCTAs.tsx` (homepage)** — the home "Complete your setup / Upload document" card, for a fixable rejection, `router.push`'d to `/profile/identity-verification` (the UnlockedRegions list) — a dead-end where the user had to hunt for the Upload-document CTA again. Now it calls **`handleSelfHealResubmit(provider)` inline** (opens the Sumsub document re-upload right there), matching the add-money bank flow + UnlockedRegions view. Provider derived from the fixable rail.

## Verified locally (prod-shaped seeds against local DB)
- ✅ requires-info rail → **"We need extra documents"** (not "Unlock US" loop)
- ✅ genuine cross-region → corrected copy (no "valid ID from {country}")
- ✅ homepage "Upload document" heals **inline** (no navigation to /profile)

## Risks
- `deriveGate` is the shared deposit/withdraw gate. The new branch only fires when an in-scope `requires-info` rail exists **and** no accept-tos/sumsub/wait action was surfaced. Genuine cross-region enrollment (no rail in scope) is unchanged; existing tests pin it.
- Copy-only change to the cross_region modal.
- Homepage card now mounts `useMultiPhaseKycFlow` + `SumsubKycModals` (same pattern the other two surfaces already use).

## QA
- `capability-gate.test.ts` — new: `requires-info rail with no surfaced action → fixable-rejection` (31/31).
- `ActivationCTAs.test.tsx` — new: fixable-rejection CTA calls `handleSelfHealResubmit`, does not navigate (4/4).
- typecheck clean. (One unrelated suite, `countryCurrencyMapping.test.ts`, fails locally on a missing flag SVG in the bare worktree — asset artifact, not this diff; green on CI.)

## Follow-up (not this PR)
- **Premature "You're unlocked"** on the *genuine* cross-region `bridge-direct` path (`useSumsubKycFlow.ts`) fires `onKycSuccess` off the response before the rail is `ENABLED`. Hugo no longer hits it (he routes to the doc-upload flow); for real new-region enrollments it should show a "we're verifying" state and confirm rail enablement before success.
- Backend: consider having the resolver surface an explicit self-heal action for the government-id `DOCUMENT_SELF_HEAL` so the gate classifies it via a real action rather than the requires-info fallback.

### DIFF
--- a/src/components/Home/ActivationCTAs.tsx
+++ b/src/components/Home/ActivationCTAs.tsx
@@ -15,6 +15,8 @@ import { useIdentityVerification } from '@/hooks/useIdentityVerification'
 import { useAuth } from '@/context/authContext'
 import { buildContactSupportMessage } from '@/utils/contact-support.utils'
 import ProvideEmailStep from '@/components/Kyc/ProvideEmailStep'
+import { useMultiPhaseKycFlow } from '@/hooks/useMultiPhaseKycFlow'
+import { SumsubKycModals } from '@/components/Kyc/SumsubKycModals'
 
 interface ActivationCTAsProps {
     activationStep: ActivationStep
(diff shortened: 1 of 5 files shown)


kushagrasarathe:

Evaluate:
- code quality and DRY

jjramirezn (approved):

PR #2388, opened 2026-07-09
fix(profile): let verified users save email without a name

kushagrasarathe:
## Summary

Verified users could get permanently stuck on **Edit Profile** when trying to set their email.

The Name/Surname fields are `disabled` once the user is identity-verified (the provider owns the legal name), but `handleSave` **hard-required** a non-empty name — sourced *only* from `user.fullName`. When a verified user's `fullName` was empty at form-load, the two rules collided: the field is locked (can't type) yet Save throws *"Please provide your name."* (can't save). Since the email field's whole purpose is an email-only update (`payload.email` is only added when `!isEmailSet`), the user could never complete it.

**Fix:** gate the name validation *and* the `fullName` payload on editability (`!isKycApproved`). Verified users now save email-only; the provider-owned name is left untouched. Non-verified users are unchanged (name still required, still sent).

## Root cause (from a real ticket)

Crisp ticket — user verified, wanted to set email, name form greyed + "Please provide your name" on Save. DB trace showed rails ENABLED before the Sumsub name backfill (`kyc/provider-submission/index.ts`) landed, and the FE was on a **stale cached user** (rails enabled, `fullName`/`email` still empty) → Name locked+empty, Email enabled → dead-end. Also reachable with fresh data whenever Sumsub can't extract a name (backfill skipped) → those users are permanently locked out.

## Risks / breaking changes

- FE-only, single component. No API/contract change.
- Behavior change is strictly a **relaxation**: verified users can now complete an email-only save; the required-name path for editable (non-verified) users is unchanged.

## QA

1. Verified user (`isKycApproved` true) with empty `fullName`, no email set → Edit Profile → type email → Save → succeeds, email persists, name untouched.
2. Non-verified user with empty name → Save → still shows "Please provide your name."
3. Non-verified user → sets name + email → Save → both persist.

### DIFF
--- a/src/components/Profile/views/ProfileEdit.view.tsx
+++ b/src/components/Profile/views/ProfileEdit.view.tsx
@@ -85,7 +85,7 @@ export const ProfileEditView = () => {
             }
 
             // prepare request payload
-            const payload: Record<string, any> = {
+            const payload: { userId?: string; fullName?: string; email?: string } = {
                 userId: user?.user.userId,
             }
 
--- a/src/components/Profile/views/ProfileEdit.view.tsx
+++ b/src/components/Profile/views/ProfileEdit.view.tsx
@@ -74,19 +74,24 @@ export const ProfileEditView = () => {
             setIsLoading(true)
             setErrorMessage('')
 
-            // validate form data
-            if (!formData.name?.trim()) {
+            // only validate/send the name when the field is editable. once the
+            // user is identity-verified the name is provider-owned and the field
+            // is disabled — requiring it would trap verified users whose fullName
+            // is empty at load (can't type, can't save) when they only want to
+            // set their email.
+            if (!isKycApproved && !formData.name?.trim()) {
                 setErrorMessage('Please provide your name.')
                 return
             }
 
-            // combine name and surname for fullName
-            const fullName = `${formData.name} ${formData.surname}`.trim()
-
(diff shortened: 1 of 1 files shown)


kushagrasarathe:
@coderabbitai review

kushagrasarathe:

Evaluate:
- code quality and DRY

kushagrasarathe:
Thanks @coderabbitai — addressed in 9a6c4e63c:

- **No-op `{ userId }` save**: added an early return in `handleSave` when the payload has no `fullName`/`email` (verified user with email already set clicking Save unchanged), so it navigates back without the pointless `/update-user` round-trip. Confirmed the backend tolerates a userId-only body gracefully (Prisma treats the undefined fields as no-op → 200), but skipping the call entirely is cleaner.
- **DRY**: consolidated the repeated `!isKycApproved` into a single `canEditName` derived boolean, reused by the name validation, the `fullName` payload gate, and the Name/Surname `disabled` props — one source of truth for the 'verified ⇒ name locked' invariant.

Intentionally not changing:
- **`user?.user.username` vs `user?.user?.username` optional-chaining inconsistency (L128/129)** — pre-existing, outside this PR's scope; leaving it to keep the diff surgical.

kushagrasarathe:

Evaluate:
- code quality and DRY

PR #2392, opened 2026-07-10
fix: migrate root validator before mixed spends sign the Rain admin sig

Hugo0:
## Summary

**HOTFIX (main): mixed spends from unmigrated pre-2025-09-18 accounts deterministically revert — 8 users / ~$1,044 cannot withdraw their card collateral at all.**

Root cause (proven by on-chain `eth_call` elimination matrix at the exact failing block): the `mixed` spend path pre-signs the Rain admin EIP-712 signature, routed to the account's current root validator (v0.0.2). The SDK then auto-wraps the userOp in `migrateWithCall`, which swaps the root validator to v0.0.3 **before** executing `withdrawAsset` — whose ERC-1271 check of that pre-signed signature now fails → the whole op reverts (`UserOperation reverted during simulation: Delegatecall failed`). Migration alone ✅, withdrawal alone ✅, combined ❌. Since these users' wallets are empty (auto-balance sweeps everything to collateral), *every* op they can send is mixed — so they can never organically migrate. Affected: mama1010 (13 failed attempts, ongoing), cosmgasm, leonelnico + 5 more at risk.

Fix:
1. **Pre-flight migration** (`ensureRootValidatorMigrated`, new util): before a mixed spend on an unmigrated account, fire the migration as its own no-op userOp (zero-value USDC self-transfer — the exact payload proven harmless by simulation), wait for the receipt, then **rebuild the kernel client** so the admin sig is signed AND verified under v0.0.3. One extra passkey tap, once ever, only for this cohort.
2. **Ref-backed client cache + `rebuildClientForChain`** (kernelClient.context): lookups now read through a ref, so closures captured before a mid-flow rebuild (e.g. the grant flow) resolve the fresh client instead of a stale v0.0.2-signing one.
3. **DRY the admin EIP-712 payload** (`buildRainWithdrawTypedData`, new util): was duplicated verbatim in the collateral-only and mixed branches; sign-vs-verify drift here is exactly this bug class.

`collateral-only` is deliberately exempt: it broadcasts server-side with the sig checked against *current* (pre-migration) state, which is valid.

## Risks / breaking changes

- Money path. Blast radius: `useSpendBundle` (withdraw/QR-pay/send funded from collateral) + kernel client caching. Post-cutoff and already-migrated accounts take the `activeClient === kernelClient` no-op path — zero behavior change for them.
- No backend changes required; no cross-repo deploy ordering.
- New analytics: `kernel_migration_attempted` / `kernel_migration_succeeded` (monitor the 8-user rollout).

## QA

- `npm run typecheck` ✅ · `npm test` 1803 ✅ (incl. new suites for the migration gate + EIP-712 payload) · `npm run build` ✅
- RCA + on-chain proof: `peanut-api-ts/logs/rca-delegatecall*.ts` (read-only eth_calls, Arbitrum block 482157518); tracker `ops/notify/zerodev-v002-stuck-card-withdrawals-2026-07-06.md`
- Prod verification plan post-deploy: mama1010 retries his BNB withdraw → expect `kernel_migration_succeeded` + on-chain rootValidator flip + successful withdrawal.

- **Dual state+ref client cache (deliberate, contained smell).** `getClientForChain` was memoized over React state, and state updates never reach already-captured closures — so after a mid-flow client rebuild (post-migration), older closures (e.g. `grant()` captured inside `spend()`) would still resolve the stale v0.0.2-signing wrapper and produce rejected signatures. Reads now go through a ref (always current, closure-age-independent); state remains solely to drive re-renders. Contained by a single write chokepoint (`storeClient`/`clearClients`) + a per-chain build-sequence guard so stale builds can't write either copy. The clean endgame is a `useSyncExternalStore`-based provider — proper follow-up, not hotfix scope.
- **Two spend engines remain.** The drift-prone preflight (routing → insufficient → migration gate → grant gate) is now ONE module (`spendPreflight.ts`) shared by both engines — that's where this bug class breeds and it can no longer be fixed on one surface only. The executor branches (broadcast-and-wait vs sign-and-return) stay separate on purpose: different return contracts across 6 call sites, legitimately different semantics. Merging them is a candidate follow-up, judged on its own risk.


Hugo0:
## Adversarial review triage (multi-agent /code-review, high effort — 15 verified findings → 10 distinct)

**Fixed (7):**
1. ⚠️ **CONFIRMED — 4337 receipt trap**: a reverted userOp still yields `receipt.status='success'` on the bundle tx, so the gate could declare a failed migration successful → straight back into the `Delegatecall failed` revert. Now verified against **on-chain root-validator state** (short propagation poll), never the receipt. Null receipt (timeout) self-heals if the op landed. Regression-tested.
2. **CONFIRMED — logout race**: a build resolving after logout repopulated the cleared cache with the previous user's client (and `assertClientOwnedByUser` skips mid-registration). Builds now carry a per-chain monotonic sequence; superseded/orphaned builds can't store.
3. **PLAUSIBLE — lagging-RPC wrapper regression**: rebuilt client asserted to be off the wrapper, one grace retry, then fail-closed `KernelMigrationPendingError`.
4. **PLAUSIBLE — in-flight clobber**: same sequence guard; `.finally` only clears the dedupe slot it owns.
5. **CONFIRMED — stuck loadingState** on receipt-timeout: `useZeroDev` now resets to `Idle` on that path.
6. **CONFIRMED — wasted passkey tap**: fail-closed overview check hoisted above the migration tap.
7. **CONFIRMED — wrong retry copy on deterministic revert**: reverted bundle → `KernelMigrationFailedError` ("contact support"), transient → `KernelMigrationPendingError` ("retry").
8. **CONFIRMED — sign-only flow missing terminal analytics**: `card_withdraw_failed` (flow: sign-only) now emitted on any preflight/signing failure.
9. **CONFIRMED — 4th copy of the Rain EIP-712 payload** in `card-recovery/page.tsx` migrated to the shared builder (4→1 → actually 5→1 counting this one).

**Declined (1, with evidence):** "permanent per-mixed-spend rebuild tax for pre-cutoff accounts" — incorrect: `createKernelMigrationAccount` returns a **plain** v0.0.3 account (no wrapper method) once the chain shows the account migrated, so `isMigrationWrapperAccount` is false after the first successful rebuild and the gate is a no-op from then on (verified in SDK source, lines 51-59).

jjramirezn (approved):

PR #2393, opened 2026-07-10
Merge mobile release (native iOS/Android) into main

innolope-dev:
## What

Lands `feat/mobile-release` into `dev` (for staging validation before the regular dev → main release) via an integration branch: 179 commits of native/mobile work (Capacitor iOS + Android platforms, TestFlight/Play CI release pipelines, native push via OneSignal adapters, demo/reviewer mode, passkey fixes, security hardening) merged with the ~370 commits main has gained since the branches diverged on June 19.

`origin/main` is already merged into this branch, so **the diff below is exactly the mobile-release changes plus conflict resolutions** — nothing from main is re-reviewed here.

## Conflict resolutions (20 files)

- **`useNotifications` / OneSignal**: kept the branch's platform-adapter architecture (module-level store, web + native adapters). Ported main's notification-click → PostHog UTM attribution into it: new `onNotificationClick` on the adapter interface, implemented in both `web.adapter.ts` (react-onesignal) and `native.adapter.ts` (Capacitor plugin), registered in the hook with main's exact capture logic.
- **Manteca deposit polling (hook + test + service)**: took main's version — main fixed the vanished-QR bug (settling gates on `stage >= 2`, not `PROCESSING` status) after the branch's copy was synced.
- **Card share-asset (4 files)**: took main's version (pill z-index fix, soft pill pull, `PEANUTMAN` asset rename) — branch copies were earlier syncs from main.
- **`campaign-maps`, dev routes regex, transaction-predicates test**: took main's newer versions for the same reason.
- **SupportDrawer**: kept both sides — the branch's lazy-mount + Capacitor guard (WKWebView memory-crash fix) and main's Crisp-failure retry UI.
- **TokenSelector**: main's Rhino-restricted popular list + the branch's never-empty USDC-on-Arbitrum fallback. Note for reviewer: if `restrictToRhino` is active and the popular list is empty, the fallback offers USDC on Arbitrum — flagging in case that interaction matters.
- **`sentry.utils`**: union — branch's `/invites/validate` 400 skip + main's `qr-payment/init` 400/422 skip.
- **setup page, ClientProviders, card page, ActivationCTAs, DynamicBankAccountForm, useSendMoney**: both-sides-added unions (imports/state).

## Verification

- `pnpm typecheck` ✅
- `pnpm test` ✅ 129 suites / 1863 tests
- `pnpm lint` has ~1048 pre-existing problems (non-blocking in CI per `tests.yml`); none introduced by the resolutions
- Production web build: running locally, will mark ready-for-review when green

## Merge instructions — please read before merging

**Base is `dev` on purpose**: this is a big change set, so it goes through staging first and reaches production through the normal dev → main release train. Both `origin/main` and `origin/dev` are merged into this branch, so the diff is exactly the mobile-release changes plus conflict resolutions, and merging it into dev also carries the pending main → dev back-merge commits.

**Merge with a merge commit, NOT squash.** Squash would break merge tracking: `feat/mobile-release` keeps living as the TestFlight release branch, and after a squash git no longer knows the histories are joined — every future sync between it and main/dev would re-conflict on the same ~139 overlapping files, over and over. With a true merge commit, git records the join point and future syncs only deal with new changes. The same applies to the dev → main release PR that later ships this: merge commit, not squash.

**After this reaches `main`:** merge `main` back into `feat/mobile-release` so the branches converge.


innolope-dev:
@coderabbitai review --dir src

innolope-dev:
@coderabbitai full review

innolope-dev:
## Note: how to get the CodeRabbit review running on this PR

CodeRabbit keeps skipping this PR ("Too many files: 172, limit 150"). For anyone wondering, a new PR won't help and excluding images won't help either — here's why and what actually fixes it:

**Why it skips.** The diff has 201 files; CodeRabbit's default filters already exclude the 29 images/lockfile (`!**/*.png` etc.), leaving 172 — still over the 150 limit. The overage is generated native scaffolding: 28 non-image files under `ios/**`, `android/**`, `patches/**` (pbxproj, plists, storyboards, Gradle). Excluding those brings the count to ~144, under the limit.

**Why the fix in this branch isn't working yet.** This branch adds a `.coderabbit.yaml` with exactly those path filters, but every run reports **"Configuration used: Organization UI"** — the org-level dashboard config takes precedence, so the repo YAML is ignored.

**What to do (needs CodeRabbit dashboard access):**
1. Open the CodeRabbit dashboard → `peanutprotocol/peanut-ui` settings.
2. Either **enable the repository config file** (so `.coderabbit.yaml` from the branch is honored), **or** add these path filters directly in the UI:
   - `!ios/**`
   - `!android/**`
   - `!patches/**`
3. Comment `@coderabbitai full review` on this PR to re-trigger.

Fallback if repo YAML only loads from the default branch: merge the `.coderabbit.yaml` to `main` in a tiny separate PR first, then re-trigger here.

This also permanently keeps future mobile PRs (which always carry Xcode/Gradle churn) under the review limit.

kushagrasarathe (changes requested):
## Review summary: holding this before merge to `dev`

I did a deep review across the auth/signing, demo-mode, native-infra/OTA/CI, and web-conflict surfaces, looking at security, regressions, and code quality.

Good news first. No web regressions. All 12 documented conflict resolutions are correct, main's fixes all survived, the `api-fetch.ts` web path is byte-identical to `dev`, the signing core is sound, and demo mode is correctly native-only and sandboxed. This is solid work.

That said, I'm not merging as-is. The OTA pipeline can push arbitrary JS into the wallet WebView, and right now its gating is weaker than the store-release path sitting next to it. Blockers are below.

### Blockers (must fix before merge)

**1. OTA bundles have no provenance signing.** `capacitor.config.ts:27-36`

`CapacitorUpdater` has no `publicKey`. Capgo's `checksum` and `sessionKey` come from the same server response as the bundle URL, so they only verify transport integrity, not authorship. Anyone with `CAPGO_API_KEY` or access to the Capgo account can push JS that runs inside the wallet, which means full control of the passkey/signing UI, recipient addresses, and amounts, all without going through store review. That's RCE-equivalent for a money app.

Fix: enable Capgo E2E signing (`npx @capgo/cli key create`, add `publicKey`, sign bundles on upload).

**2. OTA auto-deploys to `production` on merge to `main` with no approval gate.** `.github/workflows/capgo-deploy.yml:2-4`

`on: push: branches: [main, dev]`, and `main` maps to the `production` channel with no `environment:` reviewer gate, unlike `android-release.yml` and `ios-release.yml`. One bad or malicious merge ships straight to devices. Combined with #1, one merge equals arbitrary code on every install.

Fix: add `environment: Production` (with required reviewers) to the deploy job. Ideally make the `production` channel `workflow_dispatch`-only and leave `staging` on auto.

**3. `RECORD_AUDIO` permission with no justification.** `android/app/src/main/AndroidManifest.xml:95`

QR scanning is camera-only, and `FOREGROUND_SERVICE_MICROPHONE` is explicitly removed right beside it. This looks like a leftover from the Capacitor camera template. It triggers Play data-safety scrutiny and widens the attack surface for no reason.

Fix: delete the line.

**4. `enableDemoMode()` on a web-reachable path is missing the `isCapacitor()` guard.** `src/services/invites.ts:50-51`

This fires from web pages (`InvitesPage`, `JoinWaitlistPage`), unlike the native path which gates on `isCapacitor()`. It's inert today since `isDemoMode()` re-checks Capacitor, but it still writes the demo localStorage key on web, and it's the one spot where the web-inert guarantee falls apart if Capacitor detection ever regresses.

Fix: `if (res.data?.success && isCapacitor() && isDemoInviteCode(inviteCode))`

### Non-blocking, worth folding in while you're here

- Sentry native capture at 100% plus console capture, scrub depth 10 to 15 (`sentry.utils.ts:256`, `instrumentation-client.ts`). Please confirm `isSensitiveKey` covers `authToken`, `jwt`, `sessionKey`, `privateKey`, `mnemonic`, and `Authorization` so no logged object leaks a secret.
- Dead code: `resetDemoBalance()` (`demo-balance.ts:45`, no callers) and the unused `challenge` var (`native-webauthn.ts:159`).

### Please confirm (off-repo, can't verify from the diff)

- [ ] Capgo `production` channel is device-locked and not yet serving real installs
- [ ] Whether account-level bundle signing is already enabled (would soften #1)
- [ ] `CAPGO_API_KEY` scope. Prefer an upload-only key, not full-account.
- [ ] Prod web Vercel project does not set `NEXT_PUBLIC_CAPACITOR_BUILD=true` (the one flag that makes demo mode reachable on web)
- [ ] `pnpm typecheck && npm test && npm run build` green on the merged tree (the PR claims 1863 tests pass, just want it confirmed in CI)

Once #1 through #4 are addressed and the checklist is cleared, I'm happy to re-review and land it on `dev` for staging. #2394 (native-webauthn tests) looks good and i have merged it.

innolope-dev:
Thanks for the thorough review — agree on all four blockers. Here's what I've changed.

Quick context on the two OTA items (#1, #2): that pipeline came in with the original Capgo integration rather than this mobile-release branch, so these are hardening the existing setup rather than anything new introduced here — good catches regardless, and addressed below.

### Blockers

**1. OTA bundle signing** ✅ Enabled Capgo E2E signing. Generated the RSA keypair and added the `publicKey` to `CapacitorUpdater` in `capacitor.config.ts`; the private key stays out of git (`.capgo_key_v2` is gitignored) and is passed to the CI upload via `--key-data-v2` from a new `CAPGO_PRIVATE_KEY` secret.

**2. OTA auto-deploy gate** ✅ Went with the stronger option: dropped `main` from the push trigger so production is `workflow_dispatch`-only, and added `environment: Production` to the deploy job (mirrors `android-release.yml` / `ios-release.yml`). `dev` still auto-ships to `staging`. A merge to `main` can no longer reach devices without an explicit, approved run.

**3. `RECORD_AUDIO`** ✅ Removed — leftover camera-template noise, QR scanning is camera-only.

**4. `enableDemoMode()` web guard** ✅ Added the `isCapacitor()` guard to the `validateInviteCode` path in `src/services/invites.ts`, matching the native entry path.

### Non-blocking

**5. `resetDemoBalance()`** — I kept this one rather than removing it. It reads as unused today, but it's actually the only thing that can refill the demo wallet, and without it the balance would stay stuck at 0 once spent down (`debitDemoBalance` floors at 0 and persists). Demo mode is fully on-device, so the balance and its reset are per-device. I've now wired it up as a self-healing 7-day per-device TTL: `getDemoBalanceUnits()` refills on first read when the stored timestamp is older than the TTL (or absent). I also dropped the misleading "call on demo entry" docstring, and added `demo-balance.test.ts` covering the floor-at-0 debit, the TTL refill, and fresh-install defaults.

**6. Unused `challenge` var** (`native-webauthn.ts`) ✅ Removed, along with the now-orphaned `bufferToBase64URL` helper.

**7. Sentry** — leaving as-is for now; we'll tune the capture/scrub settings in a follow-up.

`tsc`, eslint, and the touched test suites are green locally.

kushagrasarathe (approved):
approved

one qn, on the capgo workflow before it goes further: `environment: Production` is set at the job level, so it applies to every run including the `dev` to staging auto-deploy, not just production. if the Production environment has required reviewers, doesn't that mean staging pushes also pause for manual approval? that seems to contradict the "only dev auto-ships to staging" intent in the comment above it

was that intentional, or should the gate only apply to the production workflow_dispatch run? if the latter, we could make the environment conditional, something like:

```yaml
environment: ${{ (github.event_name == 'workflow_dispatch' && github.event.inputs.channel == 'production') && 'Production' || '' }}
```

innolope-dev:
@kushagrasarathe updated according to your comment

src/app/(mobile-ui)/dev/full-graph/page.tsx
```diff
@@ -141,7 +141,7 @@ export default function FullGraphPage() {
                 }) => (
                     <>
                         {/* Controls Panel - Top Right */}
-                        <div className="absolute right-4 top-4 max-h-[calc(100vh-140px)] w-[200px] overflow-y-auto rounded-xl bg-white/95 p-3 shadow-lg backdrop-blur-sm">
+                        <div className="absolute right-4 top-4 max-h-[calc(100vh_-_140px)] w-[200px] overflow-y-auto rounded-xl bg-white/95 p-3 shadow-lg backdrop-blur-sm">
```
kushagrasarathe, line 144: praise: damn, this was new to me 🙌

kushagrasarathe (approved):
good changes, lgtm

PR #2395, opened 2026-07-10
fix: remove unauthenticated public Discord relay endpoint

jjramirezn:
## Summary

**Security hotfix — actively abused in prod.** `POST /api/send-discord-notification` forwarded arbitrary `body.message` verbatim to `DISCORD_WEBHOOK_URL` with no auth, no rate limit, and no mention filtering. An attacker used it today to post an `@everyone` message into our Discord.

The route is dead code: zero callers anywhere in the mono (peanut-ui, peanut-api-ts, chip, crispy, scripts). All real Discord alerting is server-side in peanut-api-ts (`src/log/discord.ts`). `/api/health` also posts to the same webhook but only with server-composed content — untouched and safe.

## Risks / breaking changes

None. Pure deletion of an un-referenced route. The webhook URL itself was never exposed by the route (it only ever returned a serialized `Response` object), so rotation is optional defense-in-depth, not required.

## QA

- `grep -rn send-discord-notification` across the whole mono → only the deleted file.
- typecheck ✅, jest: same 2 pre-existing suite failures on pristine `main` (countryCurrencyMapping, add-money-states) — unrelated to this diff.

## Post-merge

- Hotfix to `main` → needs back-merge `main` → `dev`.
- Optional: rotate `DISCORD_WEBHOOK_URL` in Vercel env for defense-in-depth.


Hugo0:
Thanks whitehat! awaiting report.

PR #2408, opened 2026-07-13
feat: enable SEPA bank rail for AL/MD/ME/MK/RS + treat Bulgaria as Eurozone

abalinda:
## Summary
image

from: https://apidocs.bridge.xyz/platform/orchestration/more/rail-specific#eur-sepa-standard-credit-sepa-instant

Bridge now supports SEPA for **Albania, Moldova, Montenegro, North Macedonia, Serbia** ([TASK-20580](https://app.notion.com/p/39883811757981b3a646da49914c862d)), but the FE still greys the bank rail out as "Soon" — availability is a static FE map. This PR adds the 5 entries to `BRIDGE_ALPHA3_TO_ALPHA2`, enabling the rail for **both deposit and withdraw** (SEPA support is bidirectional; both enabled-sets derive from this one map).

Separately, **Bulgaria adopted the euro in Jan 2026**: the deposit/withdraw list badge now reads **EUR** instead of BGN, and the stale "EUR accounts only" warning (driven by the `countryCurrencyMapping.ts` BGN row) is removed — Bulgarian bank accounts *are* EUR now.

The 4 new countries that keep a domestic currency (ALL/MDL/MKD/RSD) get the standard non-EUR SEPA warning rows, consistent with Czechia/Poland/Denmark/etc., so users know to send EUR.

Data-only change — no logic touched. The bank deposit page already renders EUR/SEPA for all of these via `getCurrencyConfig()`.

## Risks / breaking changes

- ⚠️ **Cross-repo dependency: merge/deploy peanutprotocol/peanut-api-ts#1175 first.** The BE IBAN allowlist (`src/utils/iban.ts` → `sepaCountries`) contained `ME` but not `AL`/`MD`/`MK`/`RS`; FE-only would show withdraw as available while `validate-iban` rejects those IBANs. #1175 adds the four codes (Bridge support verified: sandbox external accounts accepted for RS/AL; Bridge docs list the countries). Deposit has no such dependency (users send to a Bridge-issued EUR IBAN).
- Removing the BGN mapping row also stops FX-rate widgets treating Bulgaria as non-EUR — intended (euro adoption). Legacy exposure checked: the BE has zero `BGN` references (Bulgarian activity was always denominated EUR via SEPA), so no stored records lose flag/country resolution; a stale `?from=BGN` deep-link degrades gracefully to the generic add-money route.
- The BE capability gate is unchanged and coarse (`railJurisdictionForBank` collapses these countries to `EU`), so it cannot express per-country SEPA support — the FE map is the only per-country switch, which is why the BE allowlist above must move in lockstep.

- Enabling via `BRIDGE_ALPHA3_TO_ALPHA2` flips deposit **and** withdraw together — intentional (SEPA admission is bidirectional), but it means withdraw availability rides on the BE allowlist noted above.
- The map's comment was retitled: it was labeled "bridge EAA country codes" while already holding US/GB/CH — the old label invited a "clean up non-EEA entries" edit that would disable live rails.
- Follow-up (not this PR): `nonEurSepaCurrencies` inside `isNonEuroSepaCountry()` duplicates the non-Eurozone rows of `countryCurrencyMappings` in the same file (identical 14-code set, must be edited in pairs). Could be derived via a discriminator field.

## QA

- `npm test` (124 suites) + `npm run typecheck` green; flag-asset test auto-covers the 4 new mapping rows (al/md/mk/rs.svg exist).
- Manual: add-money country list → Albania / Moldova / Montenegro / North Macedonia / Serbia bank rail not greyed "Soon" (deposit + withdraw); Bulgaria badge reads EUR.
- Bulgaria → bank deposit page: no "EUR accounts only" banner. Serbia → bank page: banner **does** show.


abalinda:
@coderabbitai review

abalinda:
Bulgaria to have adopted EUR reported by the user roletna666 (email: roletna666@gmail.com)

PR #2414, opened 2026-07-14
feat(card): repair enable-bricked kernels inline in the grant preflight

jjramirezn:
## Summary

Follow-up to peanut-ui#2412 / peanut-api-ts#1179. An on-chain sweep of the 2025-09-18 root-validator migration wave found **91 kernel accounts with `validNonceFrom > currentNonce`** — a revocation floor above the counter that makes **every enable-mode validation revert `InvalidNonce` forever** (Kernel.sol's floor check; root-passkey ops are exempt, so the wallets look healthy). 3 of them already surfaced as card auto-balance victims; the rest are latent and would hit this the day they activate a card. A smaller cohort of undeployed pre-cutoff accounts mints approvals whose baked v0.0.3 initCode CREATE2-derives the wrong address (`AA14 initCode must return sender`).

This folds the repair into `useGrantSessionKey` so affected users never see a failure:

- The grant preflight now reads `validNonceFrom` alongside the existing `getCode`/`currentNonce`/`accountMetadata` reads (same parallel batch — no extra latency for healthy accounts).
- Floored account → root userOp `invalidateNonce(floor + 1)` (kernel syncs the counter up), then the grant binds the fresh nonce. Undeployed migration-wrapper account → the migration no-op deploys it (the SDK wrapper swaps the root validator in the same op), then the grant proceeds. Either way it's **one extra passkey tap during card setup** instead of a support ticket + `/fix-card-signature` round-trip.
- Repair confirmation is by re-read chain state, never the bundle receipt (a reverted userOp still yields a successful bundle) — same rationale as `ensureRootValidatorMigrated`. A retry after a confirm timeout re-diagnoses live state first, so it can never re-send an already-consumed invalidation.
- A floor beyond Kernel v3.1's invalidation cap (`> currentNonce + 10`; no valid target exists) fails loudly with contact-support copy instead of sending a deterministically-reverting op. No known account is in that state.
- New analytics: `card_session_key_preflight_repair` (`mode: invalidate | deploy`) to watch the latent cohort drain.

**Privacy note:** detection is purely on-chain state of the *connected* wallet — no account list, no addresses, nothing user-identifying ships in this open-source repo.

## Risks / breaking changes

- Healthy accounts: one additional parallel `eth_call` in the grant preflight; behavior otherwise unchanged (all existing nonce-binding tests pass unmodified).
- Affected accounts get a second passkey prompt inside `grant()` — consumers (`EnableAutoBalanceBanner`, `ReEnableModal`, `spendPreflight`) already treat `grant()` as an opaque async step with a spinner, so no consumer changes needed.
- If the floor read fails on a deployed account the grant now fails loudly (previously it would sign anyway) — deliberate: skipping the check can mint a dead-on-arrival approval.

## QA

- `kernelNonceRepair.utils` unit tests: repair-call selection per mode, retry-skips-consumed-invalidation, cap guard, confirm-by-state incl. never-confirms (7 tests).
- `useGrantSessionKey` suites: all existing nonce-binding/validator-binding tests pass unchanged; 2 new tests cover floored → repair → bind-fresh-nonce and undeployed-wrapper → deploy-first (15 total across the three suites).
- Typecheck + full jest green (1 pre-existing add-money failure, exists on dev/main).

## Design notes / accepted trade-offs (from review passes)

- **Deploy path reuses `ensureRootValidatorMigrated`** (review finding): a hand-rolled "is it deployed" confirm is satisfiable by a reverted migration inside a successful bundle (account lands on v0.0.2, approval silently dead) — the gate verifies the root swap against on-chain ground truth and rebuilds the client.
- **Flaky `validNonceFrom` read does not block healthy grants**: falls back to pre-floor-check behavior + `card_session_key_preflight_repair{mode:floor-read-failed}` analytics. A floored account slipping through is still caught by the sweep's permanent-failure path and `/fix-card-signature`.
- **Accepted: repair userOps route through `handleSendUserOpEncoded`**, which force-logs-out on the stale-webAuthnKey failure. Identical exposure already exists in prod for the migration no-op in `runCollateralSpendPreflight`; the logout is deliberate account-safety behavior, not new risk surface.
- **Accepted: the extra passkey tap has no bespoke framing overlay** (spendPreflight wraps its migration tap in `setSecurityOverlay`). Wiring the overlay through grant consumers is scope creep for a one-time tap affecting a bounded legacy cohort — follow-up if support signals confusion.
- **Accepted: undeployed pre-cutoff grants gain a bundler round-trip + confirm window.** The old "instant" path minted AA14-dead approvals for exactly these accounts — instant-but-broken isn't a baseline worth preserving.

### DIFF
--- a/src/constants/analytics.consts.ts
+++ b/src/constants/analytics.consts.ts
@@ -162,7 +162,8 @@ export const ANALYTICS_EVENTS = {
     CARD_SESSION_KEY_FAILED: 'card_session_key_failed',
     // Grant preflight found a repair-needing kernel (nonce floor from the
     // 2025-09-18 migration wave, or an undeployed pre-cutoff account) and
-    // sent the repair userOp before signing. `mode`: invalidate | deploy.
+    // completed the repair before signing (captured on success), or fell back
+    // on a flaky floor read. `mode`: invalidate | deploy | floor-read-failed.
     CARD_SESSION_KEY_PREFLIGHT_REPAIR: 'card_session_key_preflight_repair',
     // Withdraw refused with 409 STALE_CARD_APPROVAL — stored session-key
     // approval is bound to a deprecated validator; user must re-enable the card.
--- a/src/constants/analytics.consts.ts
+++ b/src/constants/analytics.consts.ts
@@ -160,6 +160,10 @@ export const ANALYTICS_EVENTS = {
     CARD_SESSION_KEY_PROMPTED: 'card_session_key_prompted',
     CARD_SESSION_KEY_GRANTED: 'card_session_key_granted',
     CARD_SESSION_KEY_FAILED: 'card_session_key_failed',
+    // Grant preflight found a repair-needing kernel (nonce floor from the
+    // 2025-09-18 migration wave, or an undeployed pre-cutoff account) and
+    // sent the repair userOp before signing. `mode`: invalidate | deploy.
+    CARD_SESSION_KEY_PREFLIGHT_REPAIR: 'card_session_key_preflight_repair',
     // Withdraw refused with 409 STALE_CARD_APPROVAL — stored session-key
     // approval is bound to a deprecated validator; user must re-enable the card.
     CARD_STALE_APPROVAL_HIT: 'card_stale_approval_hit',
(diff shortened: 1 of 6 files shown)


jjramirezn:
@coderabbitai review

PR #2415, opened 2026-07-14
fix(add-money): stop truncating Bridge deposit reference to 10 chars

abalinda:
## Summary

On the bank-transfer Add Money screen, the Bridge deposit reference was sliced to its first 10 characters in **display, the copy button, the copy-all-details blob, and the confirmation summary** (`AddMoneyBankDetails.tsx`, 4 sites). Users wired funds with the truncated reference, Bridge couldn't reconcile the incoming transfer, and the deposit sat in `AWAITING_FUNDS` until cancelled/returned — a happy-path money-flow break confirmed via a real user case and a PostHog session recording.

The truncation came in via `44c5beb63` ("limit deposit msg to 10 chars") — a display-overflow patch that also truncated `textToCopy`. This PR restores the full `depositMessage` everywhere and solves the original overflow with `break-all`, matching the correct sibling implementation in `BridgeDepositInstructions.tsx`.

Follow-up fixes from `/code-review high`:
- All four surfaces now read one hoisted `depositReference` const — the per-surface copy-paste is what let the original bug land on some surfaces only.
- **Share Details is now gated on the reference existing** — previously the literal `'Loading...'` fallback could be shared as the deposit reference if tapped before onramp data resolved (same unreconcilable-wire outcome as the truncation).

### Previous
image
image

### Now

image

- The on-screen `'Loading...'` display fallbacks stay (legit loading UI); only the machine-consumed paths (copy/share) are gated on real data.
- No wrap-guard added to the "double check" InfoCard line: Bridge references are a fixed 20-char shape, which fits at body-text size; `break-all` is only needed on the 4xl display.
- Pre-existing smell, out of scope: `AddMoneyBankDetails` and `BridgeDepositInstructions` render the same Bridge deposit instructions independently — that divergence is how this bug shipped. Consolidation is a follow-up.

## Risks / breaking changes

- None cross-repo — backend already returns the full `depositMessage`; this is display/copy only.
- Visual: the reference now wraps across lines on narrow screens (previously truncated). `break-all` keeps it inside the card.
- Hotfix to `main` → back-merge debt main→dev after merge.

## QA

- New colocated regression suite `__tests__/AddMoneyBankDetails.test.tsx` (4 tests): full reference in display, `textToCopy`, double-check summary, and shareable text; no copy/share while the reference is loading. Red-checked: reintroducing the slice fails the suite.
- Note: `add-money-states.test.tsx` stubs this component, so the regression tests live colocated with it instead.
- Full local gate: prettier ✅ · typecheck ✅ · jest 125/125 suites ✅ · changed files lint clean
- Visual QA on the local harness (`UI_REPO=worktree qa up`): full 20-char sandbox reference renders, wraps, copies.

### DIFF
--- a/src/components/AddMoney/components/AddMoneyBankDetails.tsx
+++ b/src/components/AddMoney/components/AddMoneyBankDetails.tsx
@@ -236,7 +236,7 @@ ${routingLabel}: ${routingValue}`
         }
 
         bankDetails += `
-Deposit Reference: ${onrampData?.depositInstructions?.depositMessage?.slice(0, 10) || 'Loading...'}
+Deposit Reference: ${onrampData?.depositInstructions?.depositMessage || 'Loading...'}
 
 Please use these details to complete your bank transfer.`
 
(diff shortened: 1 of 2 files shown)


abalinda:
@coderabbitai review

abalinda:
Closing — diagnosis superseded. The 10-char truncation is intentional (some banks, e.g. Wise, cap reference fields at 10 chars; Bridge matches deposits on the partial reference), so removing it was the wrong direction. The real customer-impacting bugs are (1) the transaction drawer showing the FULL reference while Add Money shows the truncated one (two-codes confusion) and (2) Cancel deposit firing instantly with no confirmation, right next to the support link. Replacement PR follows: uniform 10-char reference everywhere + confirmation step before cancelling.

PR #2427, opened 2026-07-15
fix: EIP-681 QR scans 404 — recipient guard rejects address@chainId

abalinda:
## Why

Scanning any real-world EIP-681 payment QR (e.g. `ethereum:USDC@42161/transfer?address=0x…&uint256=3.44e7`) dead-ends on the 404 page. The scanner parses it correctly and builds `/0x…@42161/34.4USDC`, and the url parser downstream fully supports `recipient@chain` — but `couldBeRecipient` (the Feb-21 `/es/argentina` SEO-regression guard) tests the EVM-address regex against the full segment including the `@42161` suffix, fails, and `notFound()`s.

Repro: Konrad's scan 07-15 13:54 UTC (PostHog `qr_scanned`, qr_type EIP_681, Crisp session_50bb92f0).

## What

- `couldBeRecipient`: strip the `@chain` suffix before the address/ENS checks (it was already stripped for the username check). Chain validation still happens downstream. Also collapsed the dead `includes('@')` ternary (`split('@')[0]` is equivalent).
- `generateMetadata` ([...recipient]/page.tsx): wrapped `parseAmountAndToken` in try/catch — it throws on malformed amount segments and the widened guard makes `/0x…@chain/<garbage>` URLs reach it (was pre-existing for `/username/<garbage>`); now falls back to default metadata instead of 500ing.

- `/0xADDR@<invalid-chain>` now renders a 200 with a clean "Invalid chain" ErrorAlert instead of a hard 404 (bogus-profile rendering is impossible — @-suffixed inputs never reach PublicProfile). These URLs were never indexable pre-fix, so crawl exposure is negligible. Tracked with other follow-ups in #2428.
- The guard remains a deliberate synchronous superset of the async parser; contract-test follow-up also in #2428.

## Testing

- New assertions: `0xADDR@42161`, percent-encoded `%4042161`, `vitalik.eth@arbitrum` → accepted; all locale-regression guards still pass.
- Full local suite: 127/127 suites, 1834 tests green. Prettier clean. eslint: the 1 pre-existing error in routes.ts:143 is untouched (advisory backlog, not worsened).


abalinda:
@coderabbitai review

abalinda:
Retargeted to `dev` per Hugo's request. Verification notes:

**Is this already fixed on dev? No — web is still broken.** `couldBeRecipient` on dev is byte-identical to main's broken version (zero dev-only commits touch `routes.ts`). What c3d8163d7 fixed is the **native** routing only: `recipientPayUrl()` returns `/send?recipient=…` on Capacitor, but on web it returns `/${path}` — the same `/0x…@42161/…` URL that hits the unchanged guard and 404s. This PR is still needed; it composes fine with the native helper (same path string, now valid on web too).

**Conflicts: none** — verified with `git merge-tree` against `origin/dev` (clean tree, 0 conflicts).

**Heads-up before merging:** this branch is based on main's tip, and main is currently **18 commits ahead of dev** (content publishes + #2416). Merging this PR into dev therefore also performs that main→dev back-merge. All 18 are legitimately destined for dev, but if you'd rather keep this PR's diff to the 2 fix commits, say the word and I'll rebase it onto dev (needs a force-push).

abalinda:
closing this one, I will cherrypick and point my change to main

PR #2432, opened 2026-07-16
fix(withdraw): stop method selection bouncing back to saved accounts

kushagrasarathe:
## Problem

On /withdraw with saved bank accounts: tapping "Select new method" → "Crypto" landed the user back on the saved-accounts list instead of the amount input. Separately, an open country list would snap back to saved accounts on its own after a few seconds (sometimes immediately).

## Fix

- `AddWithdrawRouterView.onCryptoClick` (withdraw flow) now only sets the crypto method in `WithdrawFlowContext` — same contract as bank-country selection — instead of also pushing `/withdraw/crypto` pre-amount. The withdraw page flips to the amount step and navigates to `/withdraw/crypto` after Continue.
- The default-view effect in `AddWithdrawRouterView` now latches once per mount (ref guard). User refetches still sync the saved-accounts list but no longer force the view back to saved accounts.
- New regression tests in `src/components/AddWithdraw/__tests__/AddWithdrawRouterView.test.tsx` — both fail on the unfixed code, pass with the fix.

## Risks / breaking changes

- FE-only, scoped to the withdraw method-selection screen. The `add` flow shares the component; its crypto path (confirmation modal) and country navigation are untouched.
- No cross-repo impact.

## QA

1. On staging/preview with a user that has saved bank accounts, go to /withdraw.
2. Tap "Select new method" → country list stays put (previously reverted within ~4s when a rail poll / focus refetch fired).
3. Tap "Crypto" → amount input appears; enter an amount ≥ $1 → Continue → lands on /withdraw/crypto initial view.
4. Complete a crypto withdrawal → "Back to home" still goes home (the race `da943d68d` originally fixed remains fixed — reset still happens on unmount, which now only ever fires post-amount or on genuine exit).

## Postmortem

**Regression window:** landed on `dev` 2026-07-14 via PR #2393 (mobile-release merge). Breaking commit: `da943d68d` "fix(withdraw): 'Back to home' no longer bounces to the withdraw screen" (authored 2026-07-02 on the mobile-release branch).

**What broke (symptom 1 — Crypto click):**
`da943d68d` moved `resetWithdrawFlow()` from the success view's `onComplete` to the crypto page's unmount cleanup, to fix a real race where completing a withdrawal bounced "Back to home" to /withdraw. But `/withdraw/crypto` has a render guard: no `amountToWithdraw` → `router.push('/withdraw')`. The method-selection screen had always navigated to `/withdraw/crypto` on the Crypto click *before* an amount exists, relying on that guard bouncing back with `selectedMethod` still intact in context (→ amount input step). After `da943d68d`, the bounce also unmounts the crypto page, whose new cleanup wipes the entire flow context (`selectedMethod`, `showAllWithdrawMethods`) — so the user lands on the saved-accounts list.

Chain: Crypto click → set method + push `/withdraw/crypto` → no-amount guard pushes back → unmount cleanup resets flow → no method selected + showAll=false → saved accounts.

**What broke (symptom 2 — country list reverting on its own):**
Long-latent defect, not from `da943d68d`. The default-view effect in `AddWithdrawRouterView` has `user` in its deps and unconditionally called `setShowAllWithdrawMethods(false)` whenever saved accounts exist. Every `/users/me` fetch dispatches a *fresh* user object into redux (`userActions.setUser(payload)` — no identity preservation), and `useUserQuery` refetches on window focus, on mount, and on a 4s interval while any rail is pending (`useUserAutoRefresh`). Each refetch re-fired the effect and yanked the open country list back to saved accounts — "after a few seconds" = the poll tick; "sometimes immediately" = an in-flight refetch resolving right after the click.

**Why it wasn't caught:** the pre-amount `/withdraw/crypto` entry path (method selection, as opposed to send-flow `?method=crypto` which skips the screen) had no test; `withdraw-states.test.tsx` mocks out `AddWithdrawRouterView` entirely. `da943d68d` was verified against the success-flow race it fixed, not against the pre-amount entry that shares the same guard + cleanup.

**Prevention:** the new tests pin both behaviors: crypto selection must not navigate pre-amount, and user-object identity churn must not override an explicit view choice.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented the add/withdraw screen from unexpectedly switching views when account information refreshes.
  * Improved the withdrawal crypto selection flow so users continue to the next step without unnecessary navigation.

* **Tests**
  * Added coverage for crypto method selection and view stability during account refreshes.

### DIFF
--- a/src/components/AddWithdraw/AddWithdrawRouterView.tsx
+++ b/src/components/AddWithdraw/AddWithdrawRouterView.tsx
@@ -10,7 +10,7 @@ import {
 } from '@/utils/general.utils'
 import { useRouter, useSearchParams } from 'next/navigation'
 import { addMoneyCountryUrl, withdrawCountryUrl, rewriteMethodPath } from '@/utils/native-routes'
-import { type FC, useEffect, useState, useTransition, useCallback } from 'react'
+import { type FC, useEffect, useRef, useState, useTransition, useCallback } from 'react'
 import { useUserStore } from '@/redux/hooks'
 import { AccountType, type Account } from '@/interfaces/interfaces'
 import { useWithdrawFlow } from '@/context/WithdrawFlowContext'
@@ -90,7 +90,11 @@ export const AddWithdrawRouterView: FC<AddWithdrawRouterViewProps> = ({
         shouldShowAllMethods = true
     }
 
-    const baseRoute = flow === 'add' ? '/add-money' : '/withdraw'
+    // apply the default view (saved accounts vs all methods) only once per mount.
+    // the user query re-dispatches a fresh `user` object on every refetch (window
+    // focus, 4s pending-rail poll), and re-running the default unconditionally
+    // yanked an open country list back to the saved-accounts view.
+    const hasAppliedDefaultView = useRef(false)
 
     useEffect(() => {
         setIsLoadingPreferences(true)
@@ -107,7 +111,7 @@ export const AddWithdrawRouterView: FC<AddWithdrawRouterViewProps> = ({
 
             if (bankAccounts.length > 0) {
                 setSavedAccounts(bankAccounts as unknown as Account[])
-                setShouldShowAllMethods(false)
+                if (!hasAppliedDefaultView.current) setShouldShowAllMethods(false)
             } else {
                 setSavedAccounts([])
             }
@@ -117,11 +121,14 @@ export const AddWithdrawRouterView: FC<AddWithdrawRouterViewProps> = ({
             const currentRecentMethods = prefs?.recentAddMethods ?? []
             if (currentRecentMethods.length > 0) {
                 setRecentMethodsState(currentRecentMethods)
-                setShouldShowAllMethods(false)
-            } else {
+                if (!hasAppliedDefaultView.current) setShouldShowAllMethods(false)
+            } else if (!hasAppliedDefaultView.current) {
                 setShouldShowAllMethods(true)
             }
         }
+        // latch only once the user has loaded, so the first real resolution
+        // (not the pre-auth null render) decides the default view
+        if (user) hasAppliedDefaultView.current = true
         setIsLoadingPreferences(false)
     }, [flow, user, setShouldShowAllMethods])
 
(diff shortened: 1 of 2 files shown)


kushagrasarathe:
@coderabbitai review

kushagrasarathe:
**Deep review pass (multi-agent /code-review high) — outcomes**

Applied in b8e9bc46e:
- crypto click now delegates to `handleMethodSelected` — one code path for "withdraw method selected", so the analytics payload / method shape can't silently drift.
- add flow skips the localStorage re-read + state churn once the default view is latched (was running every 4s during pending-rail polling).
- regression tests now run against the real `WithdrawFlowContextProvider` instead of a hand-rolled context copy.

Investigated and NOT applied (with reasons):
- **Stale `selectedMethod` on abandoned session** (re-entering /withdraw shows the amount step for a previously chosen method; can leak across into the send flow): real, but **pre-existing** — the bank path on dev behaves identically today, and this matches pre-#2393 behavior. Every quick fix shape breaks a legit flow: the native app (`AddWithdrawCountriesList`) intentionally does `setSelectedMethod(...) → router.push('/withdraw')` with no amount, i.e. "method set + no amount at mount" is a *valid* state, not a reliable staleness signal. Proper fix is structural (scope the provider to the withdraw segment, or move method/step to URL state per the repo's URL-as-state rule) → follow-up.
- **`useSavedAccounts` dedup** (the saved-account type filter here is a hand copy of the hook's canonical list — this drift class already bit Manteca once): worth doing, but the hook sources `user` from authContext while this component uses the redux store — swapping sources belongs in its own reviewed change → follow-up.
- **View mode should be URL state (`?view=all` via nuqs)**: agreed per repo convention; would make the bounce class impossible by construction and remove the ref latch. Structural → same follow-up as above.
- **Crypto page's no-amount redirect guard + reset-on-unmount** remains a trap for any future pre-amount navigation into /withdraw/crypto; contained for now (deep links/refreshes start with fresh context) → note for the withdraw-flow refactor.

kushagrasarathe:
@coderabbitai review

PR #2433, opened 2026-07-16
fix: exempt crypto withdrawals from the fiat $1 minimum

abalinda:
## Summary
Crypto withdrawals were blocked below $1 by the shared withdraw amount step: `minUsdAmount` derives a country-based bank minimum via `getMinimumAmount()`, and crypto (no country) fell into the $1 default. The minimum was introduced for bank offramps (295c8ec11, June 2025 — Bridge's $1 floor) and crypto only inherited it by sharing the page. Inconsistent with send-via-link, which allows sub-$1 amounts with the same sponsored gas.

**Fix:** `minUsdAmount` is 0 when the withdrawal is crypto — any amount > 0 passes. Bank rails keep their country minimums (US/EU $1, GB £3, MX 50 MXN) untouched.

Review hardening (multi-agent review on the first cut):
- **Exemption follows routing.** The crypto predicate follows `selectedMethod` when set (`selectedMethod ? selectedMethod.type === 'crypto' : isCryptoFromSend`): a stale bank method surviving in the app-wide `WithdrawFlowContext` (entered via `/withdraw?method=crypto`) still routes Continue to the bank flow, so it keeps the bank minimum. Regression test included.
- **Cross-chain floor restored where it's real.** Removing the blanket $1 exposed the Rhino SDA path to sub-minimum deposits — accepted on-chain, never bridged (stranded, uncredited). The confirm CTA now blocks when `payAmount < minDepositLimitUsd` (value the page previously discarded), with unit-tested helper `isBelowRhinoMinDeposit`. Direct same-chain USDC transfers keep no minimum.
- The two other inline `type !== 'crypto'` checks (limits card, Continue limits gate) consolidated onto the same predicate; exchange-rate polling now disabled for crypto.

- **Cross-chain dust fees can exceed principal** (e.g. ~$1.50 flat fee on a $0.10 mainnet withdrawal). The existing preview + non-blocking high-fee warning disclose this honestly; adding a proportionality block is a product decision deliberately left out.
- **Follow-ups (pre-existing, not this PR):** `getMinimumAmount('')` silently returns the $1 default for an unknown country — the next countryless method re-arms the trap; the `minUsdAmount` computation (country derivation + FX conversion + exemption) is a candidate for an extracted, tested hook — add-money already re-implements it with a different FX hook.

## Risks / breaking changes
- FE-only; no backend or cross-repo impact (BE enforces no minimum on this path).
- Dust withdrawals cost sponsored paymaster gas — same exposure sub-$1 links already have.
- Targets `main` per request → **back-merge debt main→dev after merge.**

## QA
- `withdraw-states`: sub-$1 crypto enables Continue and routes to `/withdraw/crypto`; sub-$1 bank stays blocked with the $1 message; stale-bank-method + `?method=crypto` stays blocked.
- `withdraw.utils`: `isBelowRhinoMinDeposit` unit tests (below/at/above/unknown values).
- Full jest suite green locally (127 suites / 1842 tests).

### DIFF
--- a/src/app/(mobile-ui)/withdraw/__tests__/withdraw-states.test.tsx
+++ b/src/app/(mobile-ui)/withdraw/__tests__/withdraw-states.test.tsx
@@ -425,7 +425,19 @@ describe('GROUP 3: Amount Validation', () => {
                 errorMessage: 'Minimum withdrawal is $1.',
             })
         )
-        expect(mockRouterPush).not.toHaveBeenCalled()
+    })
+
+    test('Stale bank method entering via ?method=crypto keeps the bank minimum', () => {
+        // Regression: the crypto exemption must follow selectedMethod (the
+        // routing source of truth), not the URL param. A leftover bank method
+        // from an abandoned withdraw survives in the app-wide context and
+        // still routes Continue to the bank flow — so sub-$1 must stay blocked.
+        mockWithdrawFlow.selectedMethod = { type: 'bridge', countryPath: 'us' }
+        mockWithdrawFlow.amountToWithdraw = '0.5'
+
+        renderWithdraw({ method: 'crypto' })
+
+        expect(screen.getByText('Continue')).toBeDisabled()
     })
 })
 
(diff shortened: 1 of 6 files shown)


abalinda:
@coderabbitai review

abalinda:
@coderabbitai review

jjramirezn (approved):

abalinda:
@coderabbitai full review

PR #2436, opened 2026-07-16
fix(native): don't flag P0_TRANSFORMS pages as uncovered server routes

innolope-dev:
## The bug — native builds fail on `main` today

`detectUncoveredServerRoutes()` in `scripts/native-build.js` fails the native build when it finds App Router files with server-only exports. But the script *also* maintains `P0_TRANSFORMS` — files it replaces with static-export-safe client stubs before `next build` runs. Their server-only exports never reach the export, yet the scan flags them anyway, because it only consults `ITEMS_TO_DISABLE`:

```js
if (isCoveredByDisableList(rel)) continue
```

This is not hypothetical. On `main` right now, `src/app/(mobile-ui)/claim/page.tsx` has `export const dynamic = 'force-dynamic'` (line 11) **and** a matching `P0_TRANSFORMS` entry (`native-build.js:86`). Someone did the right thing — added the directive *and* the stub — and the anti-rot guard rejects it anyway. The native build cannot get past this scan.

## The fix

Teach the scan that transformed files are already handled:

```js
if (isCoveredByDisableList(rel) || isHandledByTransform(rel)) continue
```

## Why this isn't a silent no-op

The obvious failure mode is a predicate that never matches — the scan keeps passing, the "fix" looks fine, and the guard still rejects real routes. Both halves were checked:

- **`rel` and `P0_TRANSFORMS[].path` are the same shape.** `rel` is `path.relative(APP_DIR, full)`; the transform loop consumes `item.path` as `path.join(APP_DIR, item.path)`. Both are relative to `src/app`.
- **Negative control against the real scan.** Running the actual (unmodified) scan functions over the real app tree: unpatched → reports `(mobile-ui)/claim/page.tsx (export const dynamic = 'force-dynamic')`; patched → zero offenders.
- **Mutation check.** Breaking the predicate to the plausible-but-wrong form (`'src/app/' + normalized`) fails 3 of the 4 tests, so they aren't vacuous.
- Unrelated routes (`api/foo/route.ts`, `(mobile-ui)/claim/layout.tsx`, `''`) correctly do **not** match, guarding against an always-true predicate.

## Tests

`scripts/__tests__/native-build-scan.test.js` — 4 tests, green under the existing jest config. Plain CJS to match the plain-`.js` script under test. `pnpm-lock.yaml` untouched.

## Not verified

A full `pnpm native:build` was **not** run — it's a 10–30 min native toolchain build and was deliberately skipped. What's proven is the guard logic itself, not that `next build` subsequently completes with the stubs in place. Note that this is already unknown on `main`, where the build stops at this scan; the change can only get it further, not less far.

### DIFF
--- a/scripts/__tests__/native-build-scan.test.js
+++ b/scripts/__tests__/native-build-scan.test.js
@@ -1,0 +1,57 @@
+const fs = require('fs')
+const path = require('path')
+const Module = require('module')
+
+const SCRIPT_PATH = path.join(__dirname, '..', 'native-build.js')
+
+// native-build.js is a script, not a module: it calls main() at import time and
+// exports nothing. Load the real source with that call stripped so the scan
+// helpers can be asserted against the actual app tree.
+function loadScriptInternals() {
+    const source = fs.readFileSync(SCRIPT_PATH, 'utf-8')
+    const withoutEntrypoint = source.replace(/\nmain\(\)\s*$/, '\n')
+    expect(withoutEntrypoint).not.toBe(source)
+
+    const exposed =
+        withoutEntrypoint +
+        '\nmodule.exports = { isHandledByTransform, detectUncoveredServerRoutes, P0_TRANSFORMS, APP_DIR }\n'
+
+    const mod = new Module(SCRIPT_PATH, null)
+    mod.filename = SCRIPT_PATH
+    mod.paths = Module._nodeModulePaths(path.dirname(SCRIPT_PATH))
+    mod._compile(exposed, SCRIPT_PATH)
+    return mod.exports
+}
+
+const { isHandledByTransform, detectUncoveredServerRoutes, P0_TRANSFORMS, APP_DIR } = loadScriptInternals()
+
+const toPosix = (p) => p.split(path.sep).join('/')
+
+describe('native build server-route scan', () => {
+    it('treats every P0_TRANSFORMS entry as handled', () => {
+        for (const transform of P0_TRANSFORMS) {
+            expect(isHandledByTransform(transform.path)).toBe(true)
+        }
+    })
+
+    // The scan passes `path.relative(APP_DIR, full)` into the predicate. If the
+    // predicate compared a different path shape it would silently never match.
+    it('matches the relative path shape the scan actually computes', () => {
+        for (const transform of P0_TRANSFORMS) {
+            const absolute = path.join(APP_DIR, transform.path)
+            expect(isHandledByTransform(path.relative(APP_DIR, absolute))).toBe(true)
+        }
+    })
+
+    it('does not suppress routes that are not transformed', () => {
+        expect(isHandledByTransform('some/other/page.tsx')).toBe(false)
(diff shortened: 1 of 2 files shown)


kushagrasarathe (approved):
Approved. Root-cause fix — P0_TRANSFORMS files were falsely flagged as uncovered server routes, blocking native build on main. Verified rel/`t.path` share the same shape; negative-control test guards against an always-true predicate. Non-blocking: full `pnpm native:build` not run (author flagged), acceptable since main is already stuck at this scan.

kushagrasarathe:
## Review — ship (conflict flag) ✅

`isHandledByTransform` normalizes `path.sep`→`/` then exact-matches `P0_TRANSFORMS[].path`. Verified against `origin/main`: the scan computes `rel = path.relative(APP_DIR, full)`, same relative-to-`src/app` shape as `P0_TRANSFORMS[].path` — so the predicate matches, and being exact-equality it can only suppress the specific stubbed files, never a sibling route. Strong test incl. a real negative control (runs the *unmodified* scan over the real tree, asserts zero transformed offenders + non-transformed paths don't match). Honest gap: full `pnpm native:build` not run (guard logic proven, not that `next build` completes) — acceptable since the guard currently blocks the build on `main`.

**Conflict flag:** #2447 also edits `scripts/native-build.js` (it de-barrels one `P0_TRANSFORMS` stub string) on base `dev`. You're on `main`. **Land this before #2447 and re-run `native-build-scan.test.js` on whichever lands second** — the conflict otherwise surfaces at the next `dev↔main` back-merge, not at PR time.

kushagrasarathe (changes requested):
## Requesting changes — reconcile with #2447

Predicate + tests are correct (see review comment). Required before approval:

- [ ] **Sequence with #2447** (which also edits `scripts/native-build.js`, de-barreling a `P0_TRANSFORMS` stub string, on base `dev`). Land this on `main` first, then **re-run `native-build-scan.test.js`** on whichever lands second — otherwise the collision surfaces at the next `dev↔main` back-merge, not at PR time.

kushagrasarathe:
### Still open on this PR

No code change needed — this is a sequencing decision:

- [ ] **Decide merge order vs #2447** (both edit `scripts/native-build.js`; #2447 de-barrels a `P0_TRANSFORMS` stub string, this one refactors the scan around `P0_TRANSFORMS`). Land this on `main` first, then **re-run `scripts/__tests__/native-build-scan.test.js`** on whichever lands second. The collision otherwise surfaces at the next `dev↔main` back-merge, not at PR time.

innolope-dev:
Confirming the sequencing: this lands on `main` **first**, then #2447. I'll re-run `scripts/__tests__/native-build-scan.test.js` on whichever lands second and report the result here.

No code change needed on this PR — the predicate and its negative-control test are unchanged.

kushagrasarathe (approved):
Sequencing confirmed (lands on main before #2447, scan test re-run on whichever is second) — clearing my earlier change request. No code change needed; predicate + negative-control test unchanged.

PR #2437, opened 2026-07-17
fix(card): don't render "#null" for a waitlist entry without a position

innolope-dev:
## The bug

The physical-card waitlist screen can tell a user they are **"You are #null on the list."**

`getPhysicalWaitlist` types the position as nullable (`src/services/rain.ts`):

```ts
export interface PhysicalWaitlistState {
    joinedAt: string | null
    position: number | null   // <-- nullable
}
```

…but `PhysicalCardScreen` renders the joined branch gated on `joinedAt` only, interpolating `position` regardless:

```tsx
{`You are #${data.position} on the list. We'll let you know when cards are ready to be shipped.`}
```

A JS template literal stringifies `null` to `"null"`, so a user who is on the list with no assigned queue position gets `#null`. The same component already treats the field as nullable a few lines up — `position: data?.position ?? null` for its PostHog event — which confirms null is a real, expected state rather than a type-only edge case.

This is **not** related to the earlier `#16on` → `#16 on` spacing fix (#2311); that one repaired the spacing on this same line and left the null case untouched. The bug is live on `main` today.

## The fix

Branch on the null case with its own sentence instead of interpolating a number that isn't there. Since the interpolation was being touched anyway, real positions now go through `toLocaleString()`, so a queue position of 1234 reads `#1,234` rather than `#1234`.

## Tests

New `PhysicalCardScreen.test.tsx` (the component had none) covering three cases: a normal position, the null case asserting no `#` reaches the body, and thousands-separator formatting.

**The test was verified to actually catch the bug** — against the original code the null and formatting cases fail (with exactly the reported `#null`) while the normal-position case still passes. A regression test that passes either way is worthless, so this was checked rather than assumed.

## Verification

`pnpm typecheck` ✓ · `jest PhysicalCardScreen` 3/3 ✓ · `eslint` no new errors · `prettier` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved waitlist messaging when a position is unavailable.
  * Formatted available waitlist positions using locale-aware number formatting (e.g., thousands separators).
  * Prevented confusing “#null” text from appearing.

* **Tests**
  * Added coverage for both numeric waitlist positions and unavailable positions.

### DIFF
--- a/src/components/Card/PhysicalCardScreen.tsx
+++ b/src/components/Card/PhysicalCardScreen.tsx
@@ -76,7 +76,9 @@ const PhysicalCardScreen: FC<Props> = ({ cardId, last4, onPrev }) => {
                     <Image src={PeanutWalking} unoptimized alt="" aria-hidden className="h-32 w-auto" />
                     <h1 className="text-xl font-extrabold">You are on the list!</h1>
                     <p className="text-sm text-grey-1">
-                        {`You are #${data.position} on the list. We'll let you know when cards are ready to be shipped.`}
+                        {data.position === null
+                            ? `You are on the list. We'll let you know when cards are ready to be shipped.`
+                            : `You are #${data.position.toLocaleString()} on the list. We'll let you know when cards are ready to be shipped.`}
                     </p>
                 </div>
             ) : (
--- a/src/components/Card/__tests__/PhysicalCardScreen.test.tsx
+++ b/src/components/Card/__tests__/PhysicalCardScreen.test.tsx
@@ -1,0 +1,60 @@
+/**
+ * PhysicalCardScreen — waitlist "on the list" copy contract.
+ *
+ * The joined branch is gated on joinedAt, but the API types position as
+ * nullable, so a joined user can have no queue position. Interpolating it
+ * blindly renders "You are #null on the list."
+ */
+import React from 'react'
+import { render, screen } from '@testing-library/react'
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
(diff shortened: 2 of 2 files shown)


kushagrasarathe (approved):
Approved. Live bug — nullable `position` interpolated to "#null". `=== null` branch is correct (keeps `#0` valid, unlike a falsy check). `toLocaleString()` is a justified in-line extra and is tested. Regression test verified to fail pre-fix.

kushagrasarathe:
## Review — ship ✅

`position: number | null` is a real runtime state, and the joined branch gated on `joinedAt` alone reached `#${null}`. `=== null` is precise — keeps position `0` on the numeric branch (a `??`/`||` guard would've wrongly hidden it). `.toLocaleString()` is type-safe (narrowed to `number`) and a reasonable in-passing improvement. Test covers numeric / null / thousands-separator, separator derived not hardcoded, fails pre-fix.

kushagrasarathe (approved):

innolope-dev:
Both CodeRabbit nits here are already handled — no change needed:

- **`== null` vs `=== null`**: `data.position` is normalized to `number | null` at the source (`position: data?.position ?? null`), so `undefined` can never reach the `=== null` check. The strict check is correct and — as noted in review — keeps position `0` valid, which a falsy/`??` guard would wrongly hide.
- **Hardcoded `1,234`**: the test already asserts `\`#${(1234).toLocaleString()}\``, deriving the separator from the runtime locale (with a comment stating the contract is delegation to `toLocaleString`, not the literal "1,234"). No locale flakiness.

CI green except the pre-existing `eslint` baseline.

PR #1197, opened 2025-09-09
[TASK-14633] feat(confirmation-modal): add token and network icons

jjramirezn:

PR #2445, opened 2026-07-17
chore(types): add FE types for GET /notifications/admin/recent

innolope-dev:
## What

Regenerates the committed OpenAPI snapshot (`src/types/api.openapi.json`) and TypeScript types (`src/types/api.generated.ts`) to include the new backend endpoint `GET /notifications/admin/recent`, added in [peanut-api-ts#1197](https://github.com/peanutprotocol/peanut-api-ts/pull/1197).

## Notes

- **Additive only** — the diff adds exactly one path to the snapshot (165 lines) and its generated types (75 lines); no existing path changes.
- Both files are `.prettierignore`d (snapshot is pulled verbatim from the BE spec), so the block is spliced in the BE's tab format rather than reformatted. `pnpm gen:api` output is idempotent, so `check:api` passes.
- The endpoint is admin-only (`x-admin-token`), so the FE won't call it — this keeps the generated types a faithful mirror of the backend contract, nothing more.

Depends on peanut-api-ts#1197 landing (the endpoint must exist on the BE for the types to be meaningful).

## Summary by CodeRabbit

* **New Features**
  * Added an administrative API endpoint for retrieving recent notifications.
  * Supports filtering by user and result limit, with administrator authentication.
  * Returns notification details, delivery statuses, timestamps, provider identifiers, and error information.
  * Documents validation and authentication error responses.

### DIFF
--- a/src/types/api.generated.ts
+++ b/src/types/api.generated.ts
@@ -4,6 +4,81 @@
  */
 
 export interface paths {
+    "/notifications/admin/recent": {
+        parameters: {
+            query?: never;
+            header?: never;
+            path?: never;
+            cookie?: never;
+        };
+        /** Recent notification rows for a user (all channels + statuses) */
+        get: {
+            parameters: {
+                query: {
+                    userId: string;
+                    limit: number;
+                };
+                header: {
+                    "x-admin-token": string;
+                };
+                path?: never;
+                cookie?: never;
+            };
+            requestBody?: never;
+            responses: {
+                /** @description Default Response */
+                200: {
+                    headers: {
+                        [name: string]: unknown;
+                    };
+                    content: {
+                        "application/json": {
+                            items: {
+                                id: string;
+                                eventType: string;
+                                channel: string;
+                                status: "PENDING" | "SENT" | "FAILED" | "SKIPPED";
+                                skipReason: string | null;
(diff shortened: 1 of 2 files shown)


kushagrasarathe:
## Review — ship ✅

Purely additive generated diff (242+/**0−**): adds exactly the `/notifications/admin/recent` path to both `api.generated.ts` and `api.openapi.json`. No existing path altered/dropped; types match the openapi source (query params, `x-admin-token`, response fields, 400/401). No half-regeneration.

**Merge order:** logically mirrors BE peanut-api-ts#1197 — harmless to merge before it (FE never calls this route), but the snapshot is only *truthful* once #1197 lands. Land **after #1197**.

kushagrasarathe (approved):

kushagrasarathe:
## Security note (non-blocking)

These generated types are inert — no runtime code, no secret committed. Keeping the approval. One consumption caveat tied to BE #1197: the `/notifications/admin/recent` response can carry provider PII (recipient email/phone in the `error` column) and is gated by a **server-only** support token. Any FE that uses these types must call the endpoint from a server route/action with the token injected server-side — **never from a `'use client'` component**, or both the token and the PII land in the browser bundle.

PR #2448, opened 2026-07-18
fix(sentry): stop double-counting fetch failures, add missing DrawerTitle

innolope-dev:
## Problem

Three of our top-volume Sentry issues are amplification, not new signal. `captureConsoleIntegration({levels: ['error','warn']})` turns every `console.error`/`console.warn` into a Sentry event, and `fetchWithSentry` both captures the underlying failure at the call site **and** re-throws a `ServiceUnavailableError` wrapper that global handlers capture again. A single API timeout could produce up to three events:

1. the explicit timeout `captureException` (kept — this is the real signal, e.g. [PEANUT-UI-QHG](https://peanut-c34d84c05.sentry.io/issues/PEANUT-UI-QHG))
2. the re-thrown wrapper bubbling up — [PEANUT-UI-QDJ](https://peanut-c34d84c05.sentry.io/issues/PEANUT-UI-QDJ), 1.6k events
3. a consumer's `console.error` about the same failure — [PEANUT-UI-MH5](https://peanut-c34d84c05.sentry.io/issues/PEANUT-UI-MH5), 2.5k events

Separately, [PEANUT-UI-MJS](https://peanut-c34d84c05.sentry.io/issues/PEANUT-UI-MJS) (1.7k events, 854 users) is Radix's `DialogContent requires a DialogTitle` a11y warning: `BadgeStatusDrawer` is the only drawer without a `DrawerTitle`.

## Fix

- **sentry.utils.ts**: add `ServiceUnavailableError` to the ignore list — it is always a wrapper created by `fetchWithSentry` whose cause was already captured with full request context at the fetch site.
- **fetchWithSentry**: log the raw fetch rejection at `console.info` (not `error`) — the explicit `captureException` calls right below carry the full context; the error-level log only produced a duplicate event via captureConsole.
- **useTokenPrice**: the "falling back to tokenDenomination" log becomes `console.info` — the fallback is graceful and the underlying failure is already reported. The `captureException` stays for genuinely new (non-fetch) errors; beforeSend now drops the wrapper case.
- **BadgeStatusDrawer**: the badge name heading becomes the `DrawerTitle` (real a11y fix, not a suppression). Note: `DrawerTitle` adds `leading-none tracking-tight` from the shared component — minor typographic delta on this one-line title.

## Deliberately NOT suppressed

- The timeout captures themselves (`/users/me`, `/rain/cards`, `/tokens/price`) — 10s timeouts hitting 1.7k users is genuine API-latency signal.
- Capacitor "plugin is not implemented" — a top-frame occurrence means a genuinely broken binary; [PEANUT-UI-QV3](https://peanut-c34d84c05.sentry.io/issues/PEANUT-UI-QV3)'s iframe source was already fixed in 972f35458 and residual events are stale binaries (resolve in Sentry, don't filter in code).

## Testing

- `pnpm typecheck` clean, related jest tests green, prettier applied.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Sentry error filtering to ignore re-thrown service-unavailability errors.
  * Reduced duplicate Sentry events from handled network failures.
* **UI Improvements**
  * Updated the badge status drawer header to use a consistent drawer title component.
* **Observability**
  * Adjusted fallback logging to be informational while still capturing exceptions.
* **Tests**
  * Added coverage for the new “already reported” error-filtering behavior.

### DIFF
--- a/sentry.utils.test.ts
+++ b/sentry.utils.test.ts
@@ -1,0 +1,19 @@
+import type { ErrorEvent } from '@sentry/nextjs'
+import { shouldIgnoreError } from './sentry.utils'
+
+function eventWith(partial: { message?: string; type?: string; value?: string }): ErrorEvent {
+    return {
+        message: partial.message,
+        exception: { values: [{ type: partial.type, value: partial.value }] },
+    } as unknown as ErrorEvent
+}
+
+describe('shouldIgnoreError — alreadyReported (fetchWithSentry wrapper)', () => {
+    it('ignores a re-thrown ServiceUnavailableError (already captured at the fetch site)', () => {
+        expect(shouldIgnoreError(eventWith({ type: 'ServiceUnavailableError', value: 'upstream 503' }))).toBe(true)
+    })
+
+    it('does not ignore an unrelated application error', () => {
+        expect(shouldIgnoreError(eventWith({ type: 'TypeError', value: 'x is not a function' }))).toBe(false)
+    })
+})
--- a/sentry.utils.ts
+++ b/sentry.utils.ts
@@ -38,6 +38,12 @@ const IGNORED_ERRORS = {
     // Third-party scripts we don't control
     thirdParty: ['googletagmanager', 'gtag', 'analytics', 'hotjar', 'clarity', 'intercom', 'crisp'],
 
+    // fetchWithSentry wrapper errors: the underlying timeout/network/HTTP
+    // failure is already captured at the fetch site with full context, so the
+    // re-thrown ServiceUnavailableError bubbling to global handlers (or being
+    // console.error'd by a consumer) would only double-count it (PEANUT-UI-QDJ).
+    alreadyReported: ['ServiceUnavailableError'],
+
     // Third-party SDK internal errors (not actionable)
     thirdPartySdkErrors: [
         'IndexedDB:Set:InternalError', // Vercel Analytics storage - fails in private browsing, not actionable
--- a/sentry.utils.ts
+++ b/sentry.utils.ts
@@ -42,6 +42,9 @@ const IGNORED_ERRORS = {
     // failure is already captured at the fetch site with full context, so the
     // re-thrown ServiceUnavailableError bubbling to global handlers (or being
     // console.error'd by a consumer) would only double-count it (PEANUT-UI-QDJ).
+    // Substring-matching this pattern is safe only because ServiceUnavailableError
+    // is our own internal fetchWithSentry wrapper name, not a generic string that
+    // could appear in an unrelated third-party error message.
     alreadyReported: ['ServiceUnavailableError'],
 
     // Third-party SDK internal errors (not actionable)
--- a/src/components/Badges/BadgeStatusDrawer.tsx
+++ b/src/components/Badges/BadgeStatusDrawer.tsx
@@ -1,4 +1,4 @@
-import { Drawer, DrawerContent } from '@/components/Global/Drawer'
+import { Drawer, DrawerContent, DrawerTitle } from '@/components/Global/Drawer'
 import Image from 'next/image'
 import { useState } from 'react'
 import { formatDate } from '@/utils/general.utils'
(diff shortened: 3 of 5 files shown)


innolope-dev:
Reconciliation note — this PR overlapped with two others opened later the same day; they've been rescoped so nothing conflicts:

- **#2451** no longer touches `BadgeStatusDrawer` — the MJS instance fix (visible `DrawerTitle` = badge name) ships here. #2451 now only adds the shared `accessibleTitle` sr-only fallback in `DrawerContent` plus regression tests guaranteeing every drawer renders a `DialogTitle` — a guardrail against the bug class. No file overlap; merge in any order.
- **#2450** dropped its plugin-not-implemented message filter in favor of this PR's rationale (iframe source fixed in 972f35458; residuals are stale binaries; a code filter would mask a genuinely broken binary). PEANUT-UI-QV3 will be resolved in Sentry instead. #2450 is retitled to what remains: a real cross-field false-positive fix in `shouldIgnoreError` + first-ever test coverage for the ignore list.

kushagrasarathe:
## Review — ship (nits) ✅

Correct double-count fix: `ServiceUnavailableError` is always the `fetchWithSentry` re-thrown wrapper (cause already captured at the fetch site), so ignoring it in `beforeSendHandler` + downgrading the two `console.error`→`console.info` removes the duplicate `captureConsoleIntegration` mints. Explicit `captureException` signal preserved. `DrawerTitle` fix is real (already exported).

**Nits:**
- Edits **both** `sentry.utils.ts` (root, SDK-config filter) and `src/utils/sentry.utils.ts` (`fetchWithSentry`) — these are two *legitimately distinct* files, not a symlink/dup. But the name collision is a genuine footgun (a reviewer can't tell them apart); worth a future rename (root → e.g. `sentry.filter.ts`).
- `'ServiceUnavailableError'` is a broad substring match — safe today (internal wrapper name), and #2450's per-field matching tightens it. Add a comment.
- No test for the new ignore entry / console-level change.

**Merge order:** land **after #2450** (auto-merges; per-field hardening in place first). Merge together with #2451 to clear the Sentry DrawerTitle issue.

kushagrasarathe (changes requested):
## Requesting changes — sequence + hardening

Double-count fix is correct. Required before approval:

- [ ] **Land after #2450** (per-field matching) so your new `'ServiceUnavailableError'` substring pattern is matched per-field, not against the loose concatenated blob. Hunks auto-merge; this is ordering.
- [ ] **Add a comment** on the broad `'ServiceUnavailableError'` substring match noting it's safe only because it's an internal `fetchWithSentry` wrapper name.
- [ ] **Add a test** for the new ignore entry + the `console.error`→`console.info` change (currently untested).

Non-blocking follow-up: rename root `sentry.utils.ts` (SDK-config filter) to disambiguate from `src/utils/sentry.utils.ts` (`fetchWithSentry`) — the name collision is a footgun.

innolope-dev:
## Nits addressed (`ac181ff0f`)

- **Substring-safety comment**: documented that matching `ServiceUnavailableError` by substring is safe only because it's our own internal `fetchWithSentry` wrapper name, not a generic string.
- **Test coverage**: added root-level `sentry.utils.test.ts` covering the new `alreadyReported` entry (a re-thrown `ServiceUnavailableError` is ignored; an unrelated `TypeError` is not). This is distinct from `src/utils/__tests__/sentry.utils.test.ts`, which tests the *other* same-named file (the `fetchWithSentry` wrapper) — the naming collision you flagged.

Skipped the `console.error→console.info` unit assertion (brittle to test at the hook level, and it's covered indirectly by the ignore path). **CI**: unit, typecheck, format, ci-success all green; only the pre-existing `eslint` baseline is red. Still sequenced to land after #2450 (per-field matching).

kushagrasarathe (approved):
Concerns addressed (substring-safety comment + ServiceUnavailableError ignore-entry test added) — clearing my earlier change request. Merge AFTER #2450 (non-overlapping hunks, auto-merges; per-field logic in place first).

PR #2451, opened 2026-07-18
fix(a11y): always give DrawerContent an accessible DialogTitle

innolope-dev:
## Problem

Sentry issue [PEANUT-UI-MJS](https://peanut-c34d84c05.sentry.io/issues/?query=PEANUT-UI-MJS): `DialogContent requires a DialogTitle for the component to be accessible for screen reader users.` — 1,703 events in 14 days across 854 users, culprit page `/home`.

This is Radix Dialog's dev-mode accessibility warning (our `Drawer` wrapper is built on vaul, which is built on Radix Dialog). Any `DrawerContent` rendered without a `DrawerTitle` logs it to the console on every open, and the `captureConsole` Sentry integration amplifies it into an event per occurrence. It is also a real accessibility defect: screen readers announce the dialog with no name.

The one instance currently missing a title (`BadgeStatusDrawer`, opened from the badge strip on `/home`) is fixed in **#2448** with a *visible* `DrawerTitle` (the badge name), which is the better a11y outcome there. This PR ships the reusable mechanism and the guardrail so the class of bug can't quietly come back; it no longer touches `BadgeStatusDrawer` and shares no files with #2448 — the two PRs can merge in any order.

## Fix

- `DrawerContent` (`src/components/Global/Drawer`) now accepts an `accessibleTitle` prop that renders a visually hidden (`sr-only`) `DrawerTitle`, for drawers whose design has no visible heading — same idiom already used manually in `KycStatusDrawer`, `TransactionDetailsDrawer`, and `TokenSelector`, now available as a one-prop path.
- Regression tests pin the labelling contract: `accessibleTitle` produces a DialogTitle wired to the dialog's `aria-labelledby`, an explicit `DrawerTitle` child still labels the dialog, and the Radix missing-DialogTitle warning does not fire when the prop is used.
- No warning suppression anywhere.

Audited all `DrawerContent` usages (ContributorsDrawer, CardUnlockDrawer, KycStatusDrawer, TransactionDetailsDrawer, ChooseNetworkDrawer, TokenSelector, QRBottomDrawer, dev DS pages): with #2448, all render a `DrawerTitle`. The headlessui-based `Modal` is unaffected (different library, no Radix warning).

## Impact

- Prevents recurrence of the PEANUT-UI-MJS class of bug: future drawers without a visible heading have a one-prop path to a proper accessible name, and the new tests guard the DialogTitle contract on the shared wrapper. (The instance fix that removes the current ~1.7k events / 14 days ships in #2448.)
- Screen-reader labelling behaviour of the shared drawer is now under test.

## Testing

- New `src/components/Global/Drawer/__tests__/Drawer.test.tsx`: asserts `accessibleTitle` renders an `sr-only` title wired to the dialog's `aria-labelledby`, that an explicit `DrawerTitle` child still labels the dialog, and that the Radix missing-DialogTitle console warning does not fire — 3/3 pass (re-verified after scoping out the BadgeStatusDrawer change; the suite is Drawer-generic).
- `pnpm tsc --noEmit` clean (0 errors); prettier clean on touched files.

### DIFF
--- a/src/components/Badges/BadgeStatusDrawer.tsx
+++ b/src/components/Badges/BadgeStatusDrawer.tsx
@@ -37,7 +37,7 @@ export const BadgeStatusDrawer = ({ isOpen, onClose, badge }: BadgeStatusDrawerP
     return (
         <>
             <Drawer open={isOpen} onOpenChange={onClose}>
-                <DrawerContent className="py-5" accessibleTitle={`Badge unlocked: ${displayName}`}>
+                <DrawerContent className="py-5">
                     <div className="space-y-5 p-5">
                         <Card
                             className="relative cursor-pointer p-4 md:p-6"
--- a/src/components/Badges/BadgeStatusDrawer.tsx
+++ b/src/components/Badges/BadgeStatusDrawer.tsx
@@ -37,7 +37,7 @@ export const BadgeStatusDrawer = ({ isOpen, onClose, badge }: BadgeStatusDrawerP
     return (
         <>
             <Drawer open={isOpen} onOpenChange={onClose}>
-                <DrawerContent className="py-5">
+                <DrawerContent className="py-5" accessibleTitle={`Badge unlocked: ${displayName}`}>
                     <div className="space-y-5 p-5">
                         <Card
                             className="relative cursor-pointer p-4 md:p-6"
--- a/src/components/Global/Drawer/__tests__/Drawer.test.tsx
+++ b/src/components/Global/Drawer/__tests__/Drawer.test.tsx
@@ -1,0 +1,65 @@
+import { render, screen } from '@testing-library/react'
+import { Drawer, DrawerContent, DrawerTitle } from '..'
+
+beforeAll(() => {
+    window.matchMedia =
+        window.matchMedia ||
+        ((query: string) =>
+            ({
+                matches: false,
+                media: query,
+                addEventListener: () => {},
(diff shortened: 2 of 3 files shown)


kushagrasarathe:
## Review — ship ✅

Adds one optional `accessibleTitle?` prop; when set renders `<DrawerTitle className="sr-only">` (reusing the already-exported primitive that wires Radix `aria-labelledby`) — root-cause a11y fix, not a warning suppression. Prop defaults `undefined`, so `accessibleTitle && <DrawerTitle/>` renders nothing when unset → **existing callers byte-for-byte unaffected**, no double-title risk. The rest of the diff is indentation churn from the signature change. Tests pin the contract. Scope disciplined.

**Pairs with #2448:** #2451 ships the reusable mechanism; #2448 applies a visible title to `BadgeStatusDrawer` (the drawer that generated the Sentry events). Different files, any order — but **merge both**; #2451 alone doesn't clear the Sentry issue.

kushagrasarathe (approved):

innolope-dev:
The console-spy restore nit is already addressed — the `accessibleTitle` test wraps its `render`/assertions in `try { … } finally { errorSpy.mockRestore() }`, so the spy is restored even on failure. CI green apart from the pre-existing `eslint` baseline. Ready — pairs with #2448 to clear the Sentry DrawerTitle issue.

PR #2463, opened 2026-07-21
feat(security): prove a fresh passkey assertion on sensitive actions

innolope-dev:
## Why

Card PAN/CVV reveal, PIN read/set, withdrawals and bank-account add are authorized today by the session cookie alone. Sessions persist for weeks, so a borrowed unlocked phone or an exfiltrated web cookie reaches all of it with no further checks.

This is the client half of item **A** — the last of the session-security series and the one that closes the real gap. API half: peanut-api-ts#1217. Companions: app lock (#2461), report-only CSP (#2462).

## What

`src/services/step-up.ts` runs a WebAuthn assertion against the user's own passkey and exchanges it for a 5-minute proof token, sent as `x-step-up-token`.

Gated calls:
- Rain: card details, PIN get/set, `withdraw/session-approve`, `withdraw/prepare` — opted in with `stepUp: true` at the single `rainRequest` choke point rather than threading a header through every call site.
- `POST /users/accounts` (add bank account) via `withStepUpHeader`.

## Prompt economy

The proof is cached for its lifetime with a 30s safety margin. A withdrawal (`session-approve` → `prepare`) costs **one** Face ID prompt, not two. The cache is dropped on logout — a proof outliving its session would let the next person on the device skip verification.

## Safe to merge alone

The API accepts the header today but doesn't require it: enforcement sits behind `STEP_UP_ENFORCED`, default off. So this can ship in either order, and the flag only flips once a build carrying it is out. Until then the extra prompt appears but nothing depends on it.

## Testing

8 unit tests, passing — ceremony, cache reuse, expiry-margin re-prompt, clear-on-logout, the no-passkey message, no caching on rejection, and that a cancelled prompt propagates rather than silently proceeding unverified. Full `src/services` suite green (18). Typecheck clean.

**Not verified on a device.** The web path is exercised by tests with `@simplewebauthn/browser` mocked; the native path goes through the `@capgo/capacitor-passkey` shim, which mocks can't stand in for. Worth a native pass on card reveal and a withdrawal before `STEP_UP_ENFORCED` is flipped.

## One thing worth a reviewer's eye

`rainRequest` reads the JWT with `Cookies.get('jwt-token')` and throws `Authentication required` when it's absent. On native, `getAuthToken()` returns null by design — the token lives in the native cookie jar — so this path looks like it should already fail there. I left it alone since it's pre-existing and unrelated, but if Rain card flows do work on native today, I'm misreading something and the step-up wiring may need the same treatment.

## Summary by CodeRabbit

- **New Features**
  - Added step-up authentication for sensitive banking and card actions using passkey verification.
  - Step-up verification tokens are securely refreshed, reused briefly, and added automatically to protected requests.
  - Added clear handling for missing passkeys, cancelled verification, and expired authentication proofs.

- **Bug Fixes**
  - Cleared cached step-up credentials during logout to prevent reuse across sessions.

### DIFF
--- a/src/app/actions/users.ts
+++ b/src/app/actions/users.ts
@@ -3,6 +3,7 @@ import { type AddBankAccountPayload, BridgeEndorsementType, type InitiateKycResp
 import { type CounterpartyUser } from '@/interfaces'
 import { type ContactsResponse } from '@/interfaces'
 import { serverFetch } from '@/utils/api-fetch'
+import { withStepUpHeader } from '@/services/step-up'
 
 export const updateUserById = async (payload: Record<string, any>): Promise<{ data?: ApiUser; error?: string }> => {
     try {
@@ -51,6 +52,7 @@ export const addBankAccount = async (payload: AddBankAccountPayload): Promise<{ 
         const response = await serverFetch('/users/accounts', {
             method: 'POST',
             body: JSON.stringify(payload),
+            headers: await withStepUpHeader({}),
         })
 
         const responseJson = await response.json()
--- a/src/context/authContext.tsx
+++ b/src/context/authContext.tsx
@@ -25,6 +25,7 @@ import { createContext, type ReactNode, useContext, useState, useEffect, useMemo
 import { captureException, setUser as setSentryUser } from '@sentry/nextjs'
 // import { PUBLIC_ROUTES_REGEX } from '@/constants/routes'
 import { USER_DATA_CACHE_PATTERNS } from '@/constants/cache.consts'
+import { clearStepUpToken } from '@/services/step-up'
 
 interface AuthContextType {
     user: IUserProfile | null
@@ -200,6 +201,10 @@ export const AuthProvider = ({ children }: { children: ReactNode }) => {
 
         // clear redirect url
         clearRedirectUrl()
+
+        // A cached step-up proof outliving the session would let the next user
+        // of this device skip verification on card and withdrawal screens.
+        clearStepUpToken()
 
         // cancel + remove all queries to prevent refetches with cleared jwt
         try {
--- a/src/services/__tests__/step-up.test.ts
+++ b/src/services/__tests__/step-up.test.ts
@@ -1,0 +1,103 @@
+/**
+ * The cache is the risky part: too eager and one Face ID prompt unlocks
+ * sensitive routes indefinitely, too shy and a withdrawal prompts twice.
+ */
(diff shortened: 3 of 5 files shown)


kushagrasarathe (commented):
## Review — FE half of step-up (with peanut-api-ts#1217) — SAFE-with-conditions

Sound: routes step-up through the single `rainRequest` choke point (`stepUp: true`), sends a fresh passkey assertion, and the BE independently verifies it. No client-only gating, no backdoor. Reviewed together with the BE half.

Notes / conditions (mostly shared with #1217):
- **Enforcement is contingent on #1217's `STEP_UP_ENFORCED` flag** — until flipped, this ships the prompt UX with no server enforcement.
- **Native path untested** (`@capgo/capacitor-passkey`) — needs an on-device iOS+Android pass.
- **Confirm native carries the session:** `rainRequest` reads `Cookies.get('jwt-token')`, which may not resolve on native where the token lives in the native cookie jar / Preferences. Verify native Rain/card flows authenticate before relying on step-up there.

The challenge-TTL fix (the one real security gap) is BE-side — see #1217.

innolope-dev:
Both conditions in the review are now handled on this side; the rest live on the BE PR.

**Native session — fixed in b52b2bc6b.** You were right to flag `Cookies.get('jwt-token')`: on native the webview's `document.cookie` doesn't hold the API session (it lives in the native cookie jar), so `rainRequest` would have thrown `Authentication required` before step-up ever ran. It now goes through `apiFetch` like every other service — Authorization header on web, native cookie jar on Capacitor — and the hard "no token" guard only applies on web. Typecheck and the step-up/rain unit suites are green.

**Challenge TTL** — landed BE-side (peanut-api-ts#1217, `EXTRACT(EPOCH FROM (now() - created_at))` check in verify, integration-tested both sides of the boundary).

**Still outstanding, tracked in #1217's pre-flip checklist:** the on-device iOS+Android pass through `@capgo/capacitor-passkey`, which nothing in CI exercises. Enforcement remains off until that and the `STEP_UP_ENFORCED` flip, so merging this ships prompt UX only — as the review noted.

kushagrasarathe (approved):
Re-reviewed — native-auth condition RESOLVED. `rainRequest` no longer reads `Cookies.get(jwt-token)`; it routes through `apiFetch → getAuthHeaders → getAuthToken` (localStorage on Capacitor, cookie on web). Residual guard is correctly `!isCapacitor() && !token`. Step-up options/verify go through the same native-safe path; headers preserved; cache cleared on logout. SAFE. (Follow-up, non-gating: `card.ts` still reads the cookie directly — separate ticket.)

PR #2473, opened 2026-07-22
fix(notifications): surface native OneSignal failures in Sentry

innolope-dev:
## Problem

Native OneSignal failures never reach Sentry — there are zero OneSignal issues tagged `environment:native`, while the same failures on web are plentiful (captured via `captureConsole`). The native failure paths all swallow errors with `console.warn`, so a device that silently unlinks from its user (all pushes target 0 recipients) or an app built without an OneSignal app id is invisible.

## Changes

All captures fire once per session per failure class, at `warning` level, tagged `feature:onesignal` so they're queryable with one search:

- **`useNotifications.ts` — `handleLoginError()`**: when an identity/verify error permanently disables `external_id` login (silently unlinking the device from the user), emit `captureMessage` tagged `onesignal:login-disabled` with the error message. Guarded by the existing `disableExternalIdLogin` flag so it fires once.
- **`useNotifications.ts` — `ensureInitialized()`**: the existing `captureException` now carries `feature:onesignal` and `level: warning` (init runs once per page, so already once-per-session).
- **`useNativePlugins.ts`**: the "failed to init app listeners" catch (back button + push-tap deep-link routing) now captures, gated by a module-level flag. (The "notification click listener" catch mentioned in the report no longer exists on `main` — click listening moved into the OneSignal adapter, whose failures surface via the `ensureInitialized` capture.)
- **`native.adapter.ts` — `init()`**: missing `NEXT_PUBLIC_ONESIGNAL_APP_ID` emits `captureMessage` tagged `onesignal:missing-app-id` before throwing, so a broken build config is visible even if a caller swallows the rejection. Captured once because the rejected `initPromise` is cached.

## Verification

- `tsc --noEmit` passes
- `prettier` run on all touched files

### DIFF
--- a/src/hooks/useNativePlugins.ts
+++ b/src/hooks/useNativePlugins.ts
@@ -2,6 +2,7 @@
 
 import { useEffect } from 'react'
 import { useRouter } from 'next/navigation'
+import { captureMessage } from '@sentry/nextjs'
 import { isCapacitor, getPlatform } from '@/utils/capacitor'
 import { deepLinkToNativePath } from '@/utils/native-routes'
 import { sanitizeRedirectURL } from '@/utils/general.utils'
@@ -12,6 +13,8 @@ import { sanitizeRedirectURL } from '@/utils/general.utils'
  * plugins are loaded via dynamic import with webpackIgnore since they only
  * exist in native builds (not on vercel/web ci).
  */
+let appListenersFailureCaptured = false
+
 export function useNativePlugins() {
     const router = useRouter()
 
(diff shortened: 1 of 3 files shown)


kushagrasarathe (approved):
Reviewed — SAFE. Observability-only: adds Sentry `captureMessage` inside existing swallowed catch/guard branches; no change to send/link/init flow. No PII (error strings are Capacitor/OneSignal SDK messages, no tokens/emails/player-ids); all `warning` level; native-gated + deduped so no spam. Non-blocking nits: base is `main` not `dev` (confirm intentional), and consider a comment pinning `extra.error` to SDK text only.

PR #2479, opened 2026-07-22
fix(csp): close report-only allowlist gaps for OneSignal, RPC and analytics

innolope-dev:
Follow-up to #2462. In the first hours after that merged, the report-only policy produced **~800 violations** (Sentry `logger:csp`). Nothing is broken today — disposition is `report` — but promoting this policy to enforcing as-is would break real functionality.

I audited every reported host against the code that actually originates the request.

## OneSignal was missing entirely

`cdn.onesignal.com` (24 hits, `script-src-elem`) is the web SDK loader, and `api.onesignal.com` was blocked under both `script-src-elem` (12) and `connect-src` (6). Enforcing would have stopped the SDK from loading at all and taken web push with it. Added the two hosts explicitly to `script-src` (kept tight, since that's the XSS-relevant directive) and `https://*.onesignal.com` to `connect-src`.

## Chain RPC endpoints — both sources, not just the reported ones

The reported RPC hosts (~780 of the 800 violations) are only half the story. There are two independent sources and both need allowlisting:

1. `rpcUrls` in `src/constants/general.consts.ts` — the curated per-chain list our own viem clients use.
2. viem's per-chain defaults — wagmi's transports are bare `http()` with no URL, so external-wallet flows fall back to viem's built-in RPC per chain.

Source 2 hadn't shown up in the reports yet (those flows are rarer), so allowlisting only what was reported would have meant a second round of violations. I enumerated both, which is why the list includes hosts like `eth.merkle.io` and `rpc.gnosischain.com` that aren't in the Sentry data.

Provider-level wildcards (`*.core.chainstack.com`, `*.publicnode.com`, `*.arbitrum.io`, `*.drpc.org`) because each provider serves one subdomain per network; single-host providers are listed exactly. Noted in the "known-loose parts" comment as something to revisit before promotion.

The list is a hand-kept mirror — `next.config.js` is CommonJS and `general.consts.ts` is TS, so neither source is importable here. There's a comment on the constant saying to re-check it when either source gains a chain.

## Analytics

`analytics.google.com` (50), `region1.analytics.google.com` (29), `stats.g.doubleclick.net` (16), `www.google.com` (15). These are GA4's region-specific beacon and GTM's audience/conversion pings — the marketing analytics stack depends on them, so they're allowlisted to keep reporting intact once the policy is enforcing.

## Remaining app traffic

`ipapi.co` (`useGeoLocation`), `api.justaname.id` (JustaName ENS provider), and `storage.crisp.chat` — the last via widening `client.crisp.chat` to `*.crisp.chat`, since Crisp serves assets from several subdomains.

## Verification

- Rendered the resulting policy and asserted all 16 reported hosts are covered (accounting for wildcard matching)
- `next.config.js` loads, `eslint` clean, `prettier` reports unchanged
- **Still report-only.** Promotion to enforcing should wait until the violation stream is quiet, and is its own PR.

## Summary by CodeRabbit

* **Security & Compatibility**
  * Updated content security policies to support additional blockchain RPC providers and external service endpoints.
  * Expanded analytics, location, naming, support chat, and notification service connectivity.
  * Added support for required OneSignal scripts and broader provider domains.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -31,6 +31,31 @@ function sentryCspReportUri() {
 }
 
 /**
+ * Chain RPC endpoints, which have two independent sources that must both be
+ * allowed: `rpcUrls` in src/constants/general.consts.ts (our own viem clients)
+ * and viem's per-chain defaults, which wagmi's bare `http()` transports fall
+ * back to for external-wallet flows. Neither is importable here — next.config
+ * is CommonJS and general.consts.ts is TS — so this list is a hand-kept mirror.
+ * Re-check it when either source gains a chain.
+ */
+const chainRpcHosts = [
+    'https://*.core.chainstack.com',
+    'https://*.publicnode.com',
+    'https://*.arbitrum.io',
+    'https://*.drpc.org',
+    'https://mainnet.optimism.io',
+    'https://mainnet.base.org',
+    'https://rpc.scroll.io',
+    'https://bsc-dataseed.bnbchain.org',
+    'https://eth.merkle.io',
+    'https://polygon-rpc.com',
+    'https://rpc.gnosischain.com',
+    'https://rpc.linea.build',
+    'https://forno.celo.org',
+    'https://*.rpc.thirdweb.com',
+]
+
+/**
  * First-draft CSP, shipped REPORT-ONLY.
  *
  * Nothing here is enforced yet: the app currently has no script-src at all, so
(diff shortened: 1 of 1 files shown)


kushagrasarathe (approved):

CSP allowlist gaps closed for legitimate OneSignal / Google Analytics origins, still in **report-only** mode — no enforcement change, no `unsafe-*` widening. Low risk.

No blocking issues.

PR #2480, opened 2026-07-22
fix(notifications): capture notification-click listener init failures

innolope-dev:
Small follow-up to #2473, which made native OneSignal failures visible in Sentry.

#2473 and #2470 were in flight at the same time and merged a minute apart. #2473 covered the two swallowing catches that existed in `useNativePlugins.ts` at the time; #2470 then added a new block to the same file — the notification-click listener — whose catch is `console.warn` only. So one native failure path is silent again.

It matters because launch URLs are suppressed in both the web and native SDKs (`suppressLaunchURLs` / `OneSignal_suppress_launch_urls`), which makes this listener the only thing that routes a push tap. If `getOneSignalAdapter()` or the listener registration throws, every notification tap silently does nothing — and today there'd be no Sentry event to explain why.

Same shape as the capture in the neighbouring app-listener catch: once per session via a module-level flag, `warning` level, tagged `feature:onesignal` and `source:native_click_listener`.

## Verification

- `tsc --noEmit` passes, `eslint` clean, `prettier` reports unchanged
- Native-only path, so it needs an OTA bundle or binary before it can produce events

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error tracking when notification click handling fails to initialize.
  * Prevented duplicate reports for the same initialization failure.

### DIFF
--- a/src/hooks/useNativePlugins.ts
+++ b/src/hooks/useNativePlugins.ts
@@ -15,6 +15,7 @@ import { getOneSignalAdapter } from '@/services/onesignal'
  * exist in native builds (not on vercel/web ci).
  */
 let appListenersFailureCaptured = false
+let clickListenerFailureCaptured = false
 
 export function useNativePlugins() {
     const router = useRouter()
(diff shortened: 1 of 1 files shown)


kushagrasarathe (approved):

Observability-only: captures failures in the notification-click listener to Sentry with a dedup guard. No behavior change to the click handling, no new data surface.

No blocking issues.

PR #2481, opened 2026-07-22
fix(balance): keep last-known spendable total instead of rendering $0

innolope-dev:
## The bug

The home screen's headline number is `smartAccountBalance + rainSpendingPower + rainInTransit`. Both Rain terms come from a single `/rain/cards` call, and `rainCentsToUsdcUnits(undefined)` folds them to `0n` — so a failed fetch was **indistinguishable from "this user has no collateral"**.

For a card user whose funds the auto-balancer swept into collateral, the smart account is legitimately 0, so the total cratered to a confident, wrong **$0** — no error, no retry affordance. Reported as mobile apps showing zero balance while activity loaded correctly (activity is a different endpoint and server-computed).

Pairs with peanutprotocol/peanut-api-ts#1226.

## Changes

- **`lastKnownSpendable.ts`** (new) — per-user persisted last-known-good total, built on the existing `{userId}:user-preferences` store rather than a new mechanism. A cold start paints the previous number immediately and corrects it in the background.
- **`useWallet`** — introduces `isRainReady`; display resolves as `stableSpendable ?? (isRainReady ? liveSum : cachedValue)`. Also consumes the API's new `balanceUnavailable`: a response that arrived but couldn't read Rain is *not* ready, while a null balance **without** the flag stays a real zero (a user with no card must still see their plain smart-account total). A served-but-stale balance is used as-is — it's fresher than our own cache.
- **`useRainCardOverview`** — stop the 30s poll for users with no card application. It ran for every logged-in user, making this the most-called endpoint and the top source of client 10s timeouts, for a payload that reads `hasApplication: false` every time. Card users keep the 30s cadence; the `user_rail_status_changed` WebSocket invalidation resumes polling the moment someone applies.
- **`home/page.tsx`** — a cached number renders at 50% opacity, so it reads as refreshing rather than asserted.

## Safety

The cache is **display-only**. Every affordability gate still runs on the live `availableSpendableBalance`, so a stale number can't green-light a spend into an orphan charge — there's a test pinning exactly that. The cache is only ever written from a fully-settled read where Rain actually answered, so a bad value can't become sticky.

## Two bugs caught in self-review

Both would have shipped:

1. **Demo mode makes no `/rain/cards` call**, so `isRainReady` was permanently false — the demo home screen would have sat on a loader forever. Verified the regression test genuinely catches it (reverting just that clause fails that test and nothing else).
2. `writeLastKnownSpendable` was **inside a `setState` updater**, which StrictMode double-invokes. Moved to a ref-guarded write, which also stops a redundant disk write on every 30s poll.

## Testing

New `useWalletSpendable.test.tsx` — 9 tests covering all the `balanceUnavailable` states, the cache seed/write rules, demo mode, and the gate-stays-live guarantee.

Full suite green: **148 suites, 2018 passed, 3 skipped**. Typecheck and prettier clean. The 2 eslint errors on `home/page.tsx` (`underMaintenanceConfig`, `isPostSignupSession`) are pre-existing — confirmed byte-identical on a stashed tree.


innolope-dev:
## Follow-up: the other instances of this bug class

Pushed `869c04230`, covering the remaining places where an **unknown** value is coerced to **zero** and then treated as fact.

### Correction to the spend-path severity

I earlier described this as users being "told 'not enough balance' and blocked from sending money they have." **That was wrong**, and one grep would have caught it.

`InsufficientSpendableError` already maps to `BALANCE_SETTLING_MESSAGE` — *"Your balance isn't fully available yet. Please try again in a few seconds."* — via `friendly-error.utils.tsx:59`, and every handler I checked (`qr-pay`, `withdraw/manteca`, `LockCardModal`, `CancelCardModal`, `useSendMoney`) uses that copy. Nobody is accused of having insufficient funds.

And blocking is the **correct call**: without `spendingPower` we can't size a collateral withdrawal, so the spend genuinely can't be routed. `useSpendBundle` already invalidates the Rain overview so a retry self-heals.

So no behavioural change was warranted. What *was* wrong is narrower: the failure reported `error_kind: 'insufficient'`, hiding a Rain outage inside the card-withdraw failure rate. It now reports `rain-balance-unavailable`.

### Changes

- **`isRainBalanceKnown`** extracted into `balance.utils` — the display path and the spend path now answer "can this Rain figure be trusted?" through one predicate instead of two copies that could drift.
- **`spendPreflight`** takes `rainBalanceKnown` (defaulting to `true`, so existing callers keep current reporting) and uses it for the analytics tag only.
- **`features/limits/utils.ts`** — `formatExtendedNumber(remainingLimit ?? 0)` rendered "You can add up to $0" when the limit was merely unknown. Now omits the sentence rather than inventing a figure. Note this needed guarding at message *construction*, not just at the `items.push` — otherwise the template interpolates the literal `"null"`.

### Testing

6 new cases on `isRainBalanceKnown` pinning the distinction that started all this: a user with no card is a **real zero**, a missing overview is **unknown**, and a stale-but-served balance is still usable.

Full suite: **148 suites, 2023 passed, 3 skipped**. Typecheck and prettier clean.

innolope-dev:
Superseded by #2482, which carries the identical 3 commits rebased onto `main` (this branch was cut from `dev`, and force-push is blocked by repo rules so the history had to move to a new branch).

Retargeting this PR to `main` in place would have dragged in 24 unrelated dev-only commits — SimpleFi removal, UK-resident block, card art, Sumsub watchdog, SEPA name-match — since `main` and `dev` have diverged (main 53 ahead, dev 26 ahead).

PR #2482, opened 2026-07-22
fix(balance): keep last-known spendable total instead of rendering $0

innolope-dev:
Rebased onto `main` from #2481 (which was branched off `dev`; force-push is blocked by repo rules, hence a new branch). **Supersedes #2481** — identical 3 commits, no dev-only work dragged along.

Targets `main` because this needs to ship in the next mobile build, and `main`/`dev` have diverged (main is 53 ahead of dev, dev 26 ahead of main).

## The bug

The home screen's headline number is `smartAccountBalance + rainSpendingPower + rainInTransit`. Both Rain terms come from a single `/rain/cards` call, and `rainCentsToUsdcUnits(undefined)` folds them to `0n` — so a failed fetch was **indistinguishable from "this user has no collateral"**.

For a card user whose funds the auto-balancer swept into collateral, the smart account is legitimately 0, so the total cratered to a confident, wrong **$0** — no error, no retry affordance. Reported as mobile apps showing zero balance while activity loaded correctly (activity is a different endpoint and server-computed).

Pairs with peanutprotocol/peanut-api-ts#1226, but ships independently: `balanceUnavailable` is an optional field and absent is treated as "available".

## Changes

- **`lastKnownSpendable.ts`** (new) — per-user persisted last-known-good total on the existing `{userId}:user-preferences` store. A cold start paints the previous number immediately and corrects it in the background.
- **`useWallet`** — `isRainReady` gate; display resolves as `stableSpendable ?? (isRainReady ? liveSum : cachedValue)`.
- **`isRainBalanceKnown`** in `balance.utils` — shared by the display path and the spend path so they can't drift on "can this Rain figure be trusted".
- **`useRainCardOverview`** — stop the 30s poll for users with no card application. It ran for every logged-in user, making this the most-called endpoint and the top source of client 10s timeouts, for a payload that reads `hasApplication: false` every time.
- **`spendPreflight`** — a Rain outage was reported as `error_kind: 'insufficient'`, hiding it inside the card-withdraw failure rate. Now `rain-balance-unavailable`. Analytics only; behaviour unchanged (blocking is correct — you can't size a collateral withdrawal you can't see — and users already get "try again in a few seconds", not "not enough funds").
- **`features/limits/utils.ts`** — "You can add up to $0" when the remaining limit was merely unknown. Omits the sentence instead.
- **`home/page.tsx`** — a cached number renders at 50% opacity so it reads as refreshing, not asserted.

## Safety

The cache is **display-only**. Every affordability gate still runs on the live `availableSpendableBalance`, so a stale number can't green-light a spend into an orphan charge — there's a test pinning that. The cache is only written from a fully-settled read where Rain actually answered.

## Two bugs caught in self-review

1. **Demo mode makes no `/rain/cards` call**, so `isRainReady` was permanently false — the demo home screen would have sat on a loader forever. Verified the regression test catches it (reverting just that clause fails that test and nothing else).
2. `writeLastKnownSpendable` was **inside a `setState` updater**, which StrictMode double-invokes. Moved to a ref-guarded write.

## Testing

Full suite on the `main` base: **155 suites, 2073 passed, 3 skipped**. Typecheck and prettier clean. 15 new tests across `useWalletSpendable.test.tsx` and `balance.utils.test.ts`.

## Summary by CodeRabbit

* **New Features**
  * Added a “stale spendable balance” state: when showing cached values, the displayed balance is now visually dimmed until live data is ready.
  * Wallet spendable display now seeds from a persisted last-known value on cold start for smoother rendering.
* **Bug Fixes**
  * Rain “balance unavailable/unknown” is no longer treated as an authoritative zero for availability decisions.
  * Limit warnings avoid blank or misleading messaging when remaining limits are unknown.
  * Reduced unnecessary Rain polling when there is no application.
* **Tests**
  * Added regression coverage for cached/stale spendable behavior, demo mode, user switching, and updated balance/limit edge cases.


kushagrasarathe (dismissed):
**Reviewed — SAFE (money-safe). Approving with one nit to tighten.**

Correctly separates a real `$0` balance from a failed read: the last-known spendable total is shown (fixing the `$0` flash) while a bad read no longer masquerades as zero. Verified **no stale number reaches an actual spend** — the live on-chain gate in `resolveSpendStrategy` is intact; this cache is display-only.

Nit (non-blocking): the last-known cache isn't time-bounded — the `at` timestamp is written but never read, so a stale value can be shown indefinitely if reads keep failing. Either bound the display on `at` (e.g. show stale-flagged only within N minutes) or drop the unused field. Fine to ship as-is and follow up.

innolope-dev:
Pushed `fcf5854` addressing the review:

**Time-bounding nit (your review):** `readLastKnownSpendable` now expires entries on the `at` timestamp after `LAST_KNOWN_MAX_AGE_MS` (7d), so a stale value can no longer be painted indefinitely when `/rain/cards` reads keep failing. Still display-only and stale-flagged; the affordability gate remains on live data. Added a test.

**Cross-user scoping (CodeRabbit's Major finding on `useWallet.ts`):** already handled in `f726aa46d` — `stableSpendable`, `persistedSpendableRef` and `lastKnownSpendable` are keyed by `owner: userId`, and `stableForUser`/`lastKnownForUser` only surface a value when `owner === userId`, so an account switch can't show or suppress with the previous user's balance. Covered by the two account-switch tests in the suite.

Spendable suite green (12 passing), typecheck clean on the touched files.

kushagrasarathe:
**Re-reviewed `fcf585425` — nit resolved, clean.** ✅

`LAST_KNOWN_MAX_AGE_MS` (7d) now expires the persisted total on its `at` timestamp in `readLastKnownSpendable`, with a test that advances `Date.now()` past the bound and asserts `undefined`. Backward-compat is handled (entries predating `at` are treated as fresh — a small self-healing population). The doc comment is honest that this is a display-only bound, not a correctness guard — the money path still uses the live gate. CI is green (`ci-success` passing; only the pre-existing repo-wide eslint red).

No remaining concerns — good to merge.

kushagrasarathe:
@innolope-dev merge conflicts

innolope-dev:
Resolved the merge conflict with `main` — one conflicting file (`balance.utils.test.ts`, two independent test blocks added at the same spot). Kept both. All 938 tests pass, prettier clean. Pushed to this branch; GitHub now reports it as mergeable.

kushagrasarathe:

Full diff read; contract dep and behaviour verified against `main`.

```
TIER: T2 (money wallet, balance display — but display-only cache; affordability gates provably stay on live data, pinning test present)
PAIRED PR: peanut-api-ts#1226 CLOSED — its `balanceUnavailable` field is ALREADY on BE main (rain/cards.ts:81, helpers.ts:77). FE treats it optional (rain.ts:621) → deploy-safe either direction. No dependency.
A. BREAKAGE:  pass — pref merge via spread (no clobber); poll-stop only for hasApplication===false with WS-invalidation resume; 15 new tests, unit CI green.
B. PERF:      pass (net win) — stops the 30s poll for no-application users; lastKnownSpendable write is ref-guarded to settled-total changes only.
C. QUALITY:   flag (minor, non-blocking) — see below.
D. SECURITY:  pass — client-side display-only cache, per-user keyed (owner-tag prevents cross-account bleed); no routes/secrets/deps.
CI: green on all real gates (typecheck/unit/e2e/format/Vercel); eslint red is the pre-existing repo-wide baseline, none in touched files.
VERDICT: approve-advisory
```

**Non-blocking follow-ups (not gating this merge):**
1. **Analytics gap** — `useSignSpendBundle.ts:162` doesn't pass `rainBalanceKnown` (only `useSpendBundle.ts:161` does), so the sign-only engine's Rain outages still report as `insufficient`. One-line fix to fully deliver the PR's stated "outage visible in analytics" goal.
2. **Architecture smell** — `useWallet.ts` (10/20 recent commits are balance fixes) and `balance.utils.ts` (9/20) are bug-fix magnets. This PR is the right fix, but the display-vs-gate spendable split keeps re-breaking — worth a consolidation follow-up.

<sub>🤖 AI advisory review (pr-review rulebook). Task-link gate waived per owner — external-dev series.</sub>

kushagrasarathe (approved):
Approving — no blocking findings; BE contract field already on main, cache is display-only with affordability gates provably on live data. Two non-blocking follow-ups in the comment (sign-engine analytics gap, consolidation smell).

innolope-dev:
Closing as already shipped — this work is live on `dev`.

The same fix was implemented on both lines. The `mobile-release` copies landed and reached `dev` when mobile-release was merged in (#2702), so all five commits here have equivalents on `dev`:

| this PR | on `dev` |
|---|---|
| keep last-known spendable total instead of rendering $0 | `e53aedde7` |
| consume `balanceUnavailable`, stop polling /rain/cards for non-card users | `8ac72aef3` |
| share the rain-known predicate, stop rendering unknown limits as $0 | `1d9cf30bf` |
| scope in-memory spendable holds to the active user | `43bae8c28` |
| time-bound the last-known spendable cache | `b12402822` |

`src/hooks/wallet/lastKnownSpendable.ts` is byte-identical between this branch and `dev`, and `isRainBalanceKnown` is present in `dev`'s `balance.utils.ts`. The remaining file differences are `dev` having evolved further, not the fix being absent.

Note for anyone auditing: branch *ancestry* reports this as unmerged, because the commits that landed are different SHAs with the same content. Match by commit subject instead.

PR #2488, opened 2026-07-23
fix(withdraw): show non-EVM (Tron/Solana) token selection in the selector button

abalinda:
## Summary

Picking **USDT-Tron / USDT+USDC-Solana** in the withdraw token drawer looked like it did nothing: the tap *did* select the token (context updates unconditionally, the drawer highlight even shows it), but the collapsed selector button resolves its label via `areEvmAddressesEqual` — always `false` for base58 addresses — so it kept reading **"Select a token"**. Users read that as "the coin can't be selected".

One-line fallback to a case-insensitive literal match. `areEvmAddressesEqual` stays first for its native-proxy aliasing on EVM tokens (`0xeee…` ↔ `0x000…`).

Pairs with peanutprotocol/peanut-api-ts#1229, which unblocks the actual request/charge creation for these destinations (EVM-only schema pattern + base58 case-corruption on storage). Without the BE fix, proceeding past the (now visible) selection still 400s at setup review.

## Risks

Display-only. The literal fallback cannot resolve a *different* token than before: distinct registry tokens never differ only by address case, and native aliasing is still handled by the EVM comparator first.

## QA

- typecheck + jest green vs dev-tip baseline (pre-existing `countryCurrencyMapping` suite failure is unrelated).
- Manual: withdraw → search or network-first → Tron/USDT → button shows the selection; same for Solana USDC/USDT. (Note: non-EVM tokens intentionally don't appear in the default popular list — search or network-first only.)

## Design notes / adversarial review

- Adversarially verified the full click chain (tap → context write → drawer close → `selectedTokenData` derivation → CTA → `onReview`): selection state always worked (handler unconditional; Sentry payloads prove base58 reached the BE); the collapsed-button label was the only broken link. No auto-reset effect exists that could clear the selection.
- The literal fallback is provably safe for EVM: `areEvmAddressesEqual` is a strict superset of case-insensitive equality for valid EVM addresses (adds native-proxy aliasing), so the fallback can never resolve a different token than the comparator would.


abalinda:
Superseded by #2490 — re-targeted to `main` as a hotfix (repo rules block force-push, so the rebased branch ships under `hotfix/token-selector-non-evm-label`).

PR #2491, opened 2026-07-23
fix(withdraw): complete the user-facing charge on collateral-routed crypto withdraws

abalinda:
## Summary

Crypto withdraws funded from Rain card collateral (wallet at $0 after auto-sweep — the norm for card users) succeed on-chain but stay **stuck PENDING and never appear in Activity**, while the balance visibly drops. Users read this as lost funds — 98 stuck charges / 55 users in the last 14 days, all `CRYPTO_WITHDRAW` (Notion P2 task).

Root cause: the withdraw page was the **only** charge-backed spend flow that neither passed `chargeId` into `sendMoney` nor called `recordPayment` for collateral-routed same-chain spends. The user-facing charge intent therefore never completed. History's read side was already built expecting the charge to be the canonical row: it hides never-paid charges as "abandoned drafts" and hides the rain-prepare intent as a "phantom" — so a successful withdrawal was completely invisible.

The fix mirrors the proven direct-send / contribute-pot / semantic-request contract:

- **Pass `chargeDetails.uuid` as `chargeId`** — on `collateral-only` the backend uses the charge as the preparation and settles it server-side in `/submit` (trusted completion; live on prod since `08004d52`).
- **Always call `recordPayment`** — `mixed` spends rely on it (the kernel's plain `usdc.transfer(recipient)` is validator-matchable); for `collateral-only` the backend routes it through the same trusted-completion path idempotently — the designed recovery net when `/submit`'s post-mining bookkeeping fails (the backend comment explicitly names FE `recordPayment` as that recovery).
- **Refresh 3 stale comments** (`useSpendBundle`, `useSendMoney`, `services/rain`) that instructed callers to SKIP `recordPayment` on collateral-only — the guidance that seeded this bug.

- `recordPayment` failures on collateral-routed **same-chain** paths are tolerated (Sentry warning + success view without payment details): funds have already moved, and for `collateral-only` the backend has already settled the charge. A mixed-path double failure leaves the charge PENDING exactly as pre-fix — never worse than today.
- **Mixed same-chain spends only record with a mined receipt hash.** If the receipt wait timed out, the kernel path returns only the userOp hash — the validator can never match that, and retry-exhaustion would flip a successful withdrawal to a false FAILED. We skip the record (pre-fix behavior, PENDING) with a Sentry breadcrumb instead. `collateral-only` is unaffected: its hash is a backend-broadcast EVM tx and its charge settles via the trusted path, receipt or not.
- Cross-chain behavior is byte-identical to `origin/dev`: `recordPayment` always fired there and still hard-fails on error.
- Bonus safety: passing `chargeId` closes a pre-fix double-broadcast hazard — retrying after "submit succeeded but FE errored" now gets a 409 (`Charge already completed`) from `/prepare` instead of silently preparing + broadcasting a **second** collateral withdrawal.
- Backfill of the 98 historical stuck charges is a separate ops task, not this PR.

## Risks / breaking changes

- FE-only. No BE change required — the charge-backed prepare + trusted completion + idempotent `recordPayment` re-entry contract is live on prod `main`.
- Blast radius: crypto-withdraw confirm flow only. `smart-only` and cross-chain paths verified unchanged (independent review sweep of every `sendMoney`/`spend` caller found no other flow with this bug shape).

## QA

- `npm test`: 148 suites green. New suite `crypto-withdraw-confirm.test.tsx` (6 tests): chargeId reaches `sendMoney`; mixed same-chain records the mined hash; collateral soft-fail still lands on the success view; mixed-without-receipt skips the record; cross-chain rethrow preserved; smart-only hard-fail preserved. **The bug-pinning tests verified to fail against pre-fix `origin/dev`.**
- `typecheck` + `prettier` clean.
- To verify manually (sandbox): with wallet USDC at $0 and funded card collateral, withdraw same-chain USDC to an external address → the withdrawal must appear as COMPLETED in Activity and the charge must not stay PENDING.

> Supersedes #2484 — same 5 commits, cherry-picked onto `main` so the PR carries only the fix (the old branch was cut from `dev` and dragged ~37 unrelated dev files into the diff). Diff verified byte-identical to the fix-only range; new test suite re-run green on the `main` base.

### DIFF
--- a/src/app/(mobile-ui)/withdraw/crypto/__tests__/crypto-withdraw-confirm.test.tsx
+++ b/src/app/(mobile-ui)/withdraw/crypto/__tests__/crypto-withdraw-confirm.test.tsx
@@ -330,6 +330,24 @@ describe('crypto withdraw confirm — charge completion', () => {
         expect(mockSetWithdrawError).not.toHaveBeenCalledWith(expect.objectContaining({ showError: true }))
     })
 
+    it('still shows success when recordPayment fails after a mixed same-chain spend (tolerance covers mixed too)', async () => {
+        mockSendMoney.mockResolvedValue({
+            txHash: undefined,
+            userOpHash: '0xuserop',
+            receipt: { transactionHash: '0xmined', status: 'success' },
+            strategy: 'mixed',
+            intentId: 'prep-intent-1',
+        })
+        mockRecordPayment.mockRejectedValue(new Error('network blip'))
+
+        await confirm()
+
+        await waitFor(() => expect(mockSetCurrentView).toHaveBeenCalledWith('STATUS'))
+        expect(mockRecordPayment).toHaveBeenCalled()
+        expect(mockCaptureMessage).toHaveBeenCalled()
+        expect(mockPosthogCapture).not.toHaveBeenCalledWith('withdraw_failed', expect.anything())
+    })
+
     it('skips recordPayment when a mixed spend has no mined receipt (a userOp hash would poison the validator)', async () => {
         mockSendMoney.mockResolvedValue({
             txHash: undefined,
@@ -368,7 +386,11 @@ describe('crypto withdraw confirm — charge completion', () => {
     })
 
     it('keeps rethrowing recordPayment failures on cross-chain withdrawals (mixed funding included)', async () => {
-        Object.assign(mockCrossChainTransfer, { isXChain: true })
+        // isDiffToken-only on purpose: same-chain USDC→ETH historically got
+        // silently downgraded to a plain transfer (see isCrossChainWithdrawal)
+        // — pin that token-boundary-only also routes through sendTransactions
+        // and keeps the strict rethrow.
+        Object.assign(mockCrossChainTransfer, { isDiffToken: true })
         mockSendTransactions.mockResolvedValue({
             userOpHash: '0xuserop',
             receipt: { transactionHash: '0xmined', status: 'success' },
--- a/src/app/(mobile-ui)/withdraw/crypto/__tests__/crypto-withdraw-confirm.test.tsx
+++ b/src/app/(mobile-ui)/withdraw/crypto/__tests__/crypto-withdraw-confirm.test.tsx
@@ -1,0 +1,345 @@
+/**
+ * Crypto Withdraw — Confirm Flow Tests
+ *
+ * Regression net for the "successful withdrawal stuck PENDING / missing from
+ * Activity" bug: same-chain collateral-routed withdraws must pass the charge
+ * uuid into sendMoney (so the backend settles the charge server-side) and must
+ * ALWAYS call recordPayment (mixed spends rely on it; collateral-only uses it
+ * as the idempotent recovery net). A recordPayment failure after funds moved
+ * must degrade to the success view on collateral-routed paths, while
+ * smart-only keeps its historical hard-fail behavior.
+ *
+ * Strategy (same as withdraw-states.test.tsx): mock every hook/service at the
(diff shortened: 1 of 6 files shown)


kushagrasarathe:
**Review — code quality / DRY / bug risk / security**

Correct fix at the right altitude (the withdraw page was the outlier that never passed `chargeId` / called `recordPayment`). The `mixedWithoutMinedHash` guard is a sharp catch — feeding a userOp hash to the on-chain validator would flip a *successful* withdraw to FAILED, worse than the stuck-PENDING it replaces. Test coverage across every strategy×outcome cell is the right rigor for a money path.

Two things to flag:

1. **DRY — biggest smell here.** This is now the 4th flow (direct-send, contribute-pot, semantic-request, withdraw) hand-rolling the same `strategy → chargeId / recordPayment / tolerate-failure` decision tree. The comments literally say "mirrors the proven contract" — that's copy-paste of branching logic, not reuse. The strategy-dependent completion belongs in a shared helper (e.g. `completeChargeAfterSpend(result, { isCrossChain })`). Not blocking — fixing the bug first is right — but worth a tracked follow-up, since independent drift across these flows is exactly how the withdraw path fell behind to begin with.

2. **Residual bug for the mixed path.** The failure tolerance fully closes the bug only for `collateral-only` (settled server-side, idempotent re-entry). A **mixed same-chain** spend whose `recordPayment` fails still stays PENDING with only a Sentry warning — "exactly like pre-fix," as the comment says. Fine and documented; just don't read this as "PENDING is now impossible" — it's impossible for the reported collateral-only cohort, narrowed for mixed.

Security: nothing new — server-side idempotency is what makes always-`recordPayment` safe.

Note: touches `withdraw/crypto/page.tsx` alongside #2490 (different regions); merge order may require a rebase.

abalinda:
@kushagrasarathe agreed. The DRY extraction is now tracked in mono TODO.md (shared completeChargeAfterSpend helper) but outside of the scope of this PR (that one is pretty big, needs a proper PR)

kushagrasarathe (approved):
**APPROVE** — with two non-blocking follow-ups called out below.

Correct fix at the right altitude: the withdraw page was the only charge-backed spend flow not passing `chargeId` / calling `recordPayment`, which is exactly why collateral-routed withdraws rotted PENDING and vanished from Activity (98 charges / 55 users). The fix mirrors the proven direct-send / contribute-pot / semantic-request contract.

What makes this approvable despite being a large money path:
- `mixedWithoutMinedHash` guard is a sharp catch — feeding a userOp hash to the on-chain validator would flip a *successful* withdraw to FAILED (worse than stuck-PENDING). Correct.
- The try/catch degradation is right: collateral-routed same-chain tolerates a `recordPayment` hiccup (funds already moved; collateral-only already settled server-side), other paths keep rethrowing.
- 426-line test file covers every strategy × outcome cell, including the smart-only strict-rethrow and cross-chain-keeps-recording branches. This is the rigor a money path needs.

Two follow-ups (neither blocks this bug fix):
1. **DRY:** this is now the 4th flow hand-rolling the same `strategy → chargeId / recordPayment / tolerate-failure` tree. Extract a shared `completeChargeAfterSpend(result, { isCrossChain })` — independent drift across these flows is how the withdraw path fell behind to begin with. Worth a tracked ticket.
2. **Residual:** the tolerance fully closes the bug only for `collateral-only` (server-settled, idempotent). A **mixed same-chain** spend whose `recordPayment` fails still stays PENDING with only a Sentry warning — documented as "exactly like pre-fix," so it's narrowed, not eliminated, for that path. Accepted.

CI green apart from the pre-existing repo-wide `no-explicit-any` eslint failure (new code is eslint-clean).

PR #2492, opened 2026-07-23
fix(native): refetch stale queries on app resume

innolope-dev:
## Problem

The Android app's home-screen **Activity** widget can show weeks-old transactions while web shows the fresh list for the same account.

TanStack Query's `refetchOnWindowFocus` relies on `visibilitychange`, which Android WebViews don't reliably fire when the app resumes from background. An app process kept alive/frozen by the OS resumes with its pre-background in-memory query data and never refetches — there is currently no native lifecycle → query-refetch wiring anywhere in the app.

## Fix

- `useNativePlugins`: listen to Capacitor `appStateChange` and drive TanStack's `focusManager`, so stale queries refetch on every app resume (the canonical TanStack pattern for non-browser hosts). Registered first among the App listeners so a deep-link init failure can't skip it.
- `useTransactionHistory`: send the history request with `cache: 'no-store'` — defense in depth alongside the API-side `Cache-Control: no-store` (companion peanut-api-ts PR).
- Drop stale "SW disk cache (1 week)" comments: the service worker no longer intercepts API responses (verified against the built `public/sw.js`).

## Verification

- `tsc --noEmit` clean; prettier run on touched files.
- On-device check after release: background the app, make a transaction from web, resume the app → Activity should refresh within the 30s staleTime window.

## Notes

Ships in the next binary/OTA release; no API contract changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data freshness for transaction history by preventing stale HTTP responses from being reused.
  * Improved query updates when the app moves between active and inactive states.
  * Preserved offline access to previously loaded data through in-memory caching.

### DIFF
--- a/src/config/wagmi.config.tsx
+++ b/src/config/wagmi.config.tsx
@@ -14,7 +14,6 @@ const queryClient = new QueryClient({
             refetchOnWindowFocus: true, // Refetch stale data when user returns
             refetchOnReconnect: true, // Refetch when connectivity restored
             // Allow queries when offline to read from TanStack Query in-memory cache
-            // Service Worker provides additional HTTP API response caching (user data, history, prices)
             networkMode: 'always', // Run queries even when offline (reads from cache)
         },
         mutations: {
--- a/src/hooks/useNativePlugins.ts
+++ b/src/hooks/useNativePlugins.ts
@@ -24,6 +24,13 @@ export function useNativePlugins() {
         if (!isCapacitor()) return
 
         const cleanups: Array<() => void> = []
+        let disposed = false
+        // Registrations resolve async; if the effect tore down first, run the
+        // cleanup now instead of leaking the handle.
+        const track = (cleanup: () => void) => {
+            if (disposed) cleanup()
+            else cleanups.push(cleanup)
+        }
 
         const openDeepLink = (url?: string | null) => {
             if (!url) return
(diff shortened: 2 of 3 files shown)


kushagrasarathe (approved):
**Reviewed adversarially (external contributor, native lifecycle) — SAFE, approving with conditions.**

Uses the canonical TanStack `focusManager.setFocused()` API to re-run stale mounted queries on app resume — correct choice over hand-rolled invalidation. Verified:
- **Not a refetch storm** — only mounted + stale queries refire, `staleTime` is 30s everywhere, so a resume within 30s is a no-op; the mounted set on a route is bounded. No money mutation is focus-driven.
- **Listener leak fixed** — the `disposed`/`track()` guard correctly handles the async `App.addListener` teardown race.
- **Web-safe** — whole effect is gated behind `if (!isCapacitor()) return`; the browser `visibilitychange` path is untouched.

**Conditions (non-blocking, please confirm):**
1. `setFocused(false)` on background pins `isFocused()` false app-wide, which **pauses any `refetchInterval` polling** (balance, deposit-polling) while backgrounded. For a frozen WebView that's arguably desired — but it's an undocumented side effect; add a one-line comment or confirm intent.
2. Re-run `@coderabbitai review` — the automated review was rate-limited to zero line comments here, and this is external code to a native prod path.
3. Verify sequencing once #2489 (app-lock) merges — both fire on `appStateChange`; confirm no double-trigger / stale-balance flash.

Nit: the teardown `setFocused(undefined)` can fire a spurious app-wide refetch if the last value was `true` — harmless (layout rarely unmounts) but pointless.

PR #2501, opened 2026-07-24
feat: chain-aware ENS resolution + visible resolved address in withdraw

0xkkonrad:
## Summary

Companion to [peanut-api-ts#1236](https://github.com/peanutprotocol/peanut-api-ts/pull/1236). External tester feedback on the withdraw flow's ENS handling:

1. **Wrong-chain resolution (funds-loss risk):** ENS names can hold a distinct address per chain (ENSIP-11). We resolved every name to its mainnet record and sent on the selected chain — `test.ses.eth` on Arbitrum went to its **Ethereum** address. The flow now passes the destination `chainId` through `resolveEns` → `/ens/:name?chainId=`, and switching chains after typing a name re-resolves it for the new chain (failure invalidates the input instead of keeping the stale address).
2. **Resolved address invisible until after the warning:** the resolved `0x…` now renders under the input the moment the name validates ("name resolves to 0x1234…abcd"), and the "Is this address compatible?" modal shows the concrete destination address.

## Risks / breaking changes

- **Deploy order: API first.** Until #1236 is live, the extra `chainId` query param is ignored by the old route schema → behavior identical to today (mainnet record). No hard dependency, no breakage either way.
- `GeneralRecipientInput` gains an optional `chainId` prop — all other consumers unchanged (prop omitted = today's behavior).
- Non-EVM (Solana/Tron) paths untouched: chain-change re-resolution only runs for the EVM family.

## QA

- `src/lib/validation/recipient.test.ts`: chainId forwarded to `resolveEns`; legacy callers resolve without it.
- Full suite: 149 suites / 2038 tests green; typecheck clean.
- Screenshots below (sandbox harness: worktree UI + local API + real JustaName resolution; on-chain balance read stubbed — harness EOA is unfunded, orthogonal to this change).

## Screenshots

| Resolved address under input (new) | Compatibility modal with destination (new) |
|---|---|
| resolved | modal |

Assets branch `pr-assets-2501` is deleted post-merge.


0xkkonrad:
@coderabbitai review

jjramirezn:
Reviewed (Jota, via Claude Code). Two things pushed to the branch:

1. **Merged dev** — conflict with the chain-switch error-clearing effect from #2503's branch line in `Initial.withdraw.view.tsx`. Kept both effects separate (separate prev-chain refs) rather than combining: the error-clearer's `error.showError` dep would abort an in-flight re-resolution through the cleanup.
2. **Race fix (679c359):** between switching chains and the re-resolution landing, `recipient.address` still held the old chain's address and Review stayed clickable — the exact wrong-chain send this PR exists to prevent, in a small window. The effect now invalidates the recipient before resolving and revalidates on success.

Also verified: client-side `unstable_cache` keys on stringified args so chainId is part of the cache key (no cross-chain stale hits), non-EVM families short-circuit before the ENS path, and the repo-wide eslint failure was the same flake as on the API PR (re-run passed).

API #1236 is merged, so the `?chainId=` param is live on dev. Merging once CI is green on the new head.

PR #2502, opened 2026-07-24
fix(add-money): bounce Manteca countries off the Bridge bank page (no EUR)

kushagrasarathe:
## Summary

Brazil & Argentina are Manteca-only deposit markets — they add money through their own PIX / Mercado Pago flow (`/add-money/[country]/manteca`). The **Bridge SEPA bank page** (`/add-money/[country]/bank`) derives its displayed currency from `getCurrencyConfig`, which only branches on US/MX/GB and **falls through to `EUR/SEPA` for everything else** — including BR/AR. So any Manteca country that lands on the bank page renders the amount in **euros** with no way to proceed.

This is exactly what a user hit (TASK-20225 / Crisp): verified with non-Brazilian docs, picked Brazil → PIX, and the amount screen showed EUR and wouldn't advance. The in-app Manteca amount screen was already fixed to default to BRL (Jul 2), but the bank page remained a live EUR entry point — reachable via the `onKycSuccess` redirect (`AddWithdrawCountriesList` pushes `/add-money/${slug}/bank` for `flow === 'add'`) and via deep/shared links.

**Fix:** guard at the route entry. The default export is now a thin wrapper that resolves the country and, if it's a Manteca country, `router.replace`s to the `/manteca` route and renders loading — so the Bridge page component (`BridgeBankOnrampPage`) **never mounts** for BR/AR. One chokepoint covers every entry path (KYC redirect, deep link, direct URL), and none of the Bridge page's data hooks or URL-state effects run during the redirect.

## Risk

Low, FE-only. Blast radius is the one route; non-Manteca countries (US/MX/GB/EU) render `BridgeBankOnrampPage` exactly as before. The Manteca predicate (`isMantecaSupportedCountryCode`) is the same one the root dispatcher (`add-money/page.tsx`) already routes with, so the guard and dispatcher read one source of truth (`MANTECA_SUPPORTED_EXCHANGES`) and can't disagree.

## QA

- `/add-money/brazil/bank` and `/add-money/argentina/bank` → redirect to `/add-money/<country>/manteca`, no EUR shown.
- `/add-money/germany/bank`, `/add-money/usa/bank`, `/add-money/mexico/bank` → unchanged (correct currency, Bridge flow).
- Unit: `add-money-states` suite (53) green — Brazil + Argentina redirect cases, a Mexico non-Manteca control (stays on Bridge UI, no redirect); Capacitor rewrite covered by `native-routes.test.ts`.

## Design notes

- **Route-wrapper altitude (not per-caller):** one guard on the shared route entry rather than patching each thing that can navigate to `/bank`. Also means the Bridge page's effects never fire for a Manteca deep link (raised in review).
- **Predicate unified with the dispatcher** — adding a future Manteca country updates one list and both routing sites follow.
- Not fixed here (follow-up, out of scope): `getCurrencyConfig` still has no BR/AR branch (correct — BR/AR are Manteca rails, not Bridge, so they should never reach it); the `/add-money/<path>/manteca` route string is built in a few places and could fold into a `mantecaMethodUrl()` helper.

### DIFF
--- a/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
+++ b/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
@@ -14,7 +14,7 @@ import { useCapabilities } from '@/hooks/useCapabilities'
 import { resolveKycModalVariant, getGateUserMessage } from '@/utils/capability-gate'
 import { useModalsContext } from '@/context/ModalsContext'
 import { useCreateOnramp, GENERIC_ONRAMP_ERROR } from '@/hooks/useCreateOnramp'
-import { useParams, useSearchParams } from 'next/navigation'
+import { useParams, useRouter, useSearchParams } from 'next/navigation'
 import { useCallback, useEffect, useMemo, useState } from 'react'
 import countryCurrencyMappings, { isNonEuroSepaCountry, isUKCountry } from '@/constants/countryCurrencyMapping'
 import { formatUnits } from 'viem'
@@ -42,7 +42,8 @@ import { useEeaUpliftFunnel } from '@/hooks/useEeaUpliftFunnel'
 import { upliftTriggerFromGate, upliftTriggerFromAdvisory } from '@/utils/eea-uplift.utils'
 import posthog from 'posthog-js'
 import { ANALYTICS_EVENTS } from '@/constants/analytics.consts'
-import { addMoneyCountryUrl } from '@/utils/native-routes'
+import { addMoneyCountryUrl, rewriteMethodPath } from '@/utils/native-routes'
+import { isMantecaCountry } from '@/constants/manteca.consts'
 import { useSafeBack } from '@/hooks/useSafeBack'
 import { getRegionIntent } from '@/utils/regions.utils'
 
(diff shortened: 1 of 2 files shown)


kushagrasarathe:
@coderabbitai review

kushagrasarathe:
Good catch — parameterized the regression test to cover both `brazil` and `argentina`. Both Manteca routes are now asserted.

kushagrasarathe:
@coderabbitai full review

kushagrasarathe:
Added a Mexico control case asserting a non-Manteca country stays on the Bridge amount UI with no route replacement — locks the contract that the guard doesn't over-fire. Suite now 53 green.

kushagrasarathe:
@coderabbitai full review

kushagrasarathe:
Addressed both:

- **Bridge effects running for Manteca deep links (Minor):** moved the BR/AR redirect into the route wrapper (default export), so `BridgeBankOnrampPage` never mounts for a Manteca country — none of its data hooks or URL-state effects run during the redirect. This also resolves the same class flagged by /code-review.
- **Capacitor rewrite path coverage (nitpick):** already covered by an existing unit test — `src/utils/__tests__/native-routes.test.ts:104` asserts `rewriteMethodPath('/add-money/argentina/manteca')` → `/add-money?country=argentina&view=manteca`; Brazil takes the identical regex path. Re-testing the rewrite through the component would be redundant, so leaving the component tests asserting the web target.

kushagrasarathe:
@coderabbitai full review

PR #2505, opened 2026-07-24
fix(history): resolve ENS names server-side so mobile shows them too

innolope-dev:
## Problem
On the mobile (native) app, the activity feed shows raw addresses (e.g. `0x1bbb…ea0349`) where the web app shows the resolved ENS name. Root cause: the feed reverse-resolves addresses via a **client-side** mainnet RPC call (`@justaname.id/react` `usePrimaryName`, `priority: 'onChain'`). With `CapacitorHttp` disabled, that third-party RPC call is blocked by CORS on the native WebView origin, so it silently fails and the row falls back to the address. Web (report-only CSP + `peanut.me` origin) resolves fine.

## Fix
Route the reverse-lookup through the backend instead of the client RPC:
- New `usePrimaryNameServer` hook → `GET /ens/reverse/:address` via same-origin `serverFetch` (already CORS-allowed + authed for native).
- Swaps the three `usePrimaryName` call sites: `TransactionCard`, `AddressLink`, `useRecipientDisplay`.
- **Fail-safe:** any non-OK response → `undefined` → row degrades to the address exactly as today. No regression even if the backend route isn't live yet.

## Depends on
peanut-api-ts #1237 (`GET /ens/reverse`) — deploy that first for names to actually appear on native.

## Tests
- New `usePrimaryNameServer` test (resolve / no-name / non-OK fail-safe / skips non-address & undefined).
- Updated `TransactionCard.test.tsx` mock to the new hook; existing suites green; typecheck clean.

## Notes
- Same code path is what's broken on `mobile-release`; once merged, this should be brought into that branch too.
- `JustaNameProvider` / `justaname.config` are now unused by app code and could be removed in a follow-up.

### DIFF
--- a/src/components/Global/AddressLink/index.tsx
+++ b/src/components/Global/AddressLink/index.tsx
@@ -1,6 +1,6 @@
 import { printableAddress, isCryptoAddress } from '@/utils/general.utils'
 import { normalizeEnsName } from '@/utils/ens.utils'
-import { usePrimaryName } from '@justaname.id/react'
+import { usePrimaryNameServer } from '@/hooks/usePrimaryNameServer'
 import Link from 'next/link'
 import { useEffect, useState } from 'react'
 import { twMerge } from 'tailwind-merge'
@@ -20,11 +20,7 @@ const AddressLink = ({ address, className = '', isLink = true }: AddressLinkProp
     const [urlAddress, setUrlAddress] = useState<string>(address)
 
     // Look up ENS name only for Ethereum addresses (ENS doesn't apply to Solana/Tron)
-    const { primaryName: ensName } = usePrimaryName({
-        address: isAddress(address) ? (address as `0x${string}`) : undefined,
-        chainId: 1, // Mainnet for ENS lookups
-        priority: 'onChain',
-    })
+    const { primaryName: ensName } = usePrimaryNameServer(isAddress(address) ? address : undefined)
 
     useEffect(() => {
         const normalizedEnsName = isAddress(address) ? normalizeEnsName(ensName) : null
--- a/src/components/TransactionDetails/TransactionCard.tsx
+++ b/src/components/TransactionDetails/TransactionCard.tsx
@@ -23,7 +23,7 @@ import React, { lazy, Suspense } from 'react'
 import { twMerge } from 'tailwind-merge'
 import Image from 'next/image'
 import { isAddress } from 'viem'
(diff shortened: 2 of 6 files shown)


kushagrasarathe (approved):
**Reviewed adversarially (external contributor) — SAFE. Approving with a deploy-order condition.**

Swaps ENS reverse-resolution from a client-side mainnet RPC (`usePrimaryName`) to the paired backend endpoint `GET /ens/reverse/:address` (#1237). Small, well-scoped (103/-20). `ci-success` is green.

- **XSS — safe.** The resolved name (attacker-controllable — anyone can set a reverse record to arbitrary text) is rendered only as React JSX text (`{displayName}` in `TransactionCard`, `{displayAddress}` in `AddressLink`) — auto-escaped, no `dangerouslySetInnerHTML`. The request URL interpolates the **address** (viem `isAddress()`-gated on FE, regex-re-validated on BE), never the ENS name — no path injection.
- **Fail-safe correct.** Non-OK → `null` → callers degrade to the raw address exactly as before (`usePrimaryNameServer.ts:35-40`). All three call sites migrated; leftover `usePrimaryName` refs are only a doc comment + jest mock.
- **No N+1 storm.** react-query key `['ens-primary-name', address.toLowerCase()]` dedupes the same address across the whole feed; `staleTime`/`gcTime` 24h, `retry:false`, `enabled` gated on `isAddress`. One request per distinct address, cached a day. Same-origin authed `serverFetch`.
- Good DRY (reuses `serverFetch` + `normalizeEnsName`), matching hook test covering resolve / no-name / non-OK fail-safe / non-address.

**Condition:**
1. **Deploy BE #1237 first.** Until `/ens/reverse` is live every lookup fail-safes to the raw address — no breakage, but no names either. Enforce the ordering.

**Non-blocking:**
- Latent (pre-existing, **not introduced here**): `AddressLink` builds `href={\`/${normalizedEnsName}\`}` and `normalizeEnsName` does no character sanitization — an ENS name with `/`/`?`/`:` could manipulate the route. This value was already attacker-controllable via the old client RPC (byte-identical to base), so out of scope for this diff, but worth a separate hardening PR to whitelist/sanitize the name before it becomes a URL path.
- Author's own follow-up: remove now-unused `JustaNameProvider` / `justaname.config`, and port to `mobile-release`.

CodeRabbit was rate-limited and didn't actually review here — don't read its silence as approval.

PR #2507, opened 2026-07-24
fix(receipt): copy/display crypto withdrawal destinations verbatim (Tron/Solana)

abalinda:
## Summary

QA of the Tron/Solana withdraw hotfix (#2490) found the withdrawal **receipt corrupts the destination address**. The account-details row runs the identifier through `formatIban`, which treats any string whose first two chars are letters as an IBAN and uppercases + space-chunks it. Every Tron address starts with `T…` and base58 is **case-sensitive**, so:

- **Copy:** the clipboard got an uppercased, space-chunked (invalid) address for every Tron/Solana withdrawal — pasting it into a wallet fails base58/checksum validation.
- **Display:** `maskAccountIdentifier`'s `plain` fallback runs the same `formatIban` on any identifier matching `/^[a-zA-Z]{2}\d/`, so Tron addresses whose 3rd char is a digit (~13%, e.g. `TN9R…`) were also mangled on screen. (The commit that shipped #2490 claimed "display was already safe" — it wasn't.)

## Fix

Keyed to the **actual wire vocabulary** the BE serializes for a `CRYPTO_WITHDRAW`: `recipientAccount.type` is `'address'` (history `mapGenericIntent`, sender viewer), `'evm-address'`, or `'peanut-wallet'` — **never** the Prisma `WALLET_EXTERNAL` enum, which `accountTypeToApi` maps to `'evm-address'` before it reaches the FE.

A closed 3-value `isCryptoAddressType` allowlist (every other type stays a bank rail → keeps `formatIban`) now gates **both** the copy helper (`getAccountCopyValue`) and the display mask (`maskAccountIdentifier` short-circuit → verbatim). The row labels crypto destinations **"Address"** instead of "Account Number".

## Risks / QA

- **Display-and-copy only.** Bank rails (`iban`/`us`/`clabe`/`cbu`/`cvu`/`pix`/`BANK_*`/raw `payment_rail`) are outside the closed allowlist, so their `formatIban` copy + per-rail masking are untouched.
- Tests feed the **real wire values** (`'address'`/`'evm-address'`) and cover both copy and display — each **fails on the pre-fix code** (verified: `formatIban` corrupts both sample Tron addresses; the digit-3rd-char one also on display). Full jest suite green.
- **Cross-repo:** pairs with peanut-api-ts `hotfix/non-evm-withdraw-false-failed` (the BE was flipping these same withdrawals to Failed). Independent deploys.

> Adversarial multi-agent review caught that the first pass keyed off `'WALLET_*'` — a type the wire never carries — making the fix a no-op whose tests were green on a fabricated value. This is the corrected version.

## Summary by CodeRabbit

- **Bug Fixes**
  - Crypto wallet addresses are now displayed and copied without unwanted formatting or masking.
  - Bank account identifiers continue to use appropriate IBAN formatting and masking.
  - Transaction details now label crypto destinations as “Address” for clearer identification.
  - Supports address handling across Tron, Solana, EVM, and other supported crypto address types.

### DIFF
--- a/src/components/TransactionDetails/TransactionDetailsReceipt.tsx
+++ b/src/components/TransactionDetails/TransactionDetailsReceipt.tsx
@@ -20,7 +20,7 @@ import useClaimLink from '@/components/Claim/useClaimLink'
 import { formatAmount, formatDate, isStableCoin, formatCurrency } from '@/utils/general.utils'
 import { formatPoints } from '@/utils/format.utils'
 import { getAvatarUrl } from '@/utils/history.utils'
-import { formatIban, printableAddress, shortenAddress, shortenStringLong, slugify } from '@/utils/general.utils'
+import { printableAddress, shortenAddress, shortenStringLong, slugify } from '@/utils/general.utils'
 import { maskAccountIdentifier } from '@/utils/account-mask.utils'
 import { cancelOnramp } from '@/app/actions/onramp'
 import { captureException } from '@sentry/nextjs'
@@ -40,7 +40,7 @@ import CopyToClipboard from '../Global/CopyToClipboard'
 import CancelSendLinkModal from '../Global/CancelSendLinkModal'
 import { twMerge } from 'tailwind-merge'
 import { isAddress } from 'viem'
-import { getBankAccountLabel } from './transaction-details.utils'
+import { getAccountCopyValue, getBankAccountLabel } from './transaction-details.utils'
 import { useModalsContext } from '@/context/ModalsContext'
 import { useRouter } from 'next/navigation'
 import { getBankAccountCountryCode } from '@/constants/countryCurrencyMapping'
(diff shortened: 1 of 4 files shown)


kushagrasarathe (approved):
**Reviewed for code quality, DRY, and bug-risk. Approving — ship as-is (LOW risk).**

Traced both the copy path (`getAccountCopyValue`) and the display path (`maskAccountIdentifier`) end-to-end against `formatIban`.

**Quality/DRY — clean.** `isCryptoAddressType` is defined once in `account-mask.utils` and reused by both the copy helper and the mask short-circuit — no duplication. This is a root-cause fix in the right place: it gates *both* corruption paths, not just the copy symptom the ticket named. Tests feed the real wire values (`'address'`/`'evm-address'`) and each fails on pre-fix code.

**Two small notes (non-blocking):**

1. **Closed allowlist is coupled to the BE vocabulary.** `CRYPTO_ADDRESS_TYPES = {address, evm-address, peanut-wallet}`. Any *future* crypto wire `type` the BE introduces (e.g. `'tron-address'`) falls through to `formatIban` → this exact corruption class silently returns. The comment frames this as the deliberate safe-default (new rails keep IBAN formatting), which is the right call — just flagging the coupling.
2. **`peanut-wallet` now renders verbatim + labels as "Address"** where it previously hit the mask/label default. Behavior change for the user's-own-wallet row; low risk since it's an EVM address and verbatim is correct.

PR #2509, opened 2026-07-24
Update content submodule

Hugo0:


Hugo0:
Superseded by the production publish PR #2523, which targets `main`. These auto-PRs target `dev`, which never promotes to production.

PR #2514, opened 2026-07-24
fix(step-up): stop prompting where the flow already proves the passkey

jjramirezn:
## Summary
FE half of peanutprotocol/peanut-api-ts#1241 — removes `stepUp: true` from `prepareWithdrawal` and `submitWithdrawSessionApproval`. These flows are self-proving (admin EIP-712 / enable signature follow immediately from the same passkey), so the step-up assertion added by #2463 only added a third fingerprint sheet to every collateral-funded send (the triple-prompt reports from Hugo and Jota, confirmed on-chain: the send tx carries exactly 2 WebAuthn signatures).

Card PAN/CVV/PIN and bank-account add keep step-up.

## Risks / breaking changes
- **Deploy order: backend PR first.** If `STEP_UP_ENFORCED` is on and this ships before the BE drops the gate, withdraw prepare would 401.

## QA
- prettier clean, step-up service tests pass (8/8), 0 typecheck errors in changed file. 3 failing suites are the known missing-submodule worktree gotcha (content/countryCurrency/add-money), green in CI.
- After both deploy: a collateral-funded send link = 2 fingerprint prompts (admin sig + userop); card-detail reveal still prompts.

Screenshots: N/A (no UI change — prompt-count behavior only)

### DIFF
--- a/src/services/rain.ts
+++ b/src/services/rain.ts
@@ -429,12 +429,15 @@ export const rainApi = {
     /**
      * Persist the serialized ZeroDev permission on the user's card so the
      * backend can submit session-key UserOps for collateral withdrawals.
+     *
+     * No step-up: the payload is itself a fresh passkey signature (the grant
+     * ceremony's enable sig) — an extra assertion here just doubles the
+     * fingerprint prompts on the grant flow.
      */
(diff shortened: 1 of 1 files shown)


jjramirezn:
Superseded by hotfix #2515 (targets main). dev gets it via back-merge.

PR #2516, opened 2026-07-25
fix(card): render the share asset's hand crisp, not pixelated

0xkkonrad:
## Summary

The card share asset rendered its hand through a pre-pixelated PNG, so the one image users actually post looked low-fidelity. Nothing on that asset is hidden — the card number is already `????` and the Visa mark is dropped via `hideVisa` — so the pixelation had nothing left to obscure. It was only costing fidelity.

Pixelation belongs to the closed-beta tease, which is exactly what the existing `blurAll` prop already marks. The hand now follows that switch instead of being unconditional:

- **Share asset** (`ShareAssetD3`, no `blurAll`) → crisp `peanut-card-hand.svg`, `image-rendering: pixelated` dropped.
- **Tease surfaces** (all pass `blurAll`) → pre-pixelated PNG, unchanged: `AddCardEntryScreen`, `CardEligibilityCheckScreen`, `BadgeSkipCelebration`, `/shhhhh` LP hero.

Two files, +23/−17.

## Risks / breaking changes

**None cross-repo.** Frontend-only, no API surface touched.

The one risk worth naming is the capture path. `PixelatedCardFace` carries a comment about the launch-day "blank share asset" bug, so the swap was made to preserve that fix rather than reopen it:

- The hand stays a plain ``. It is **not** returned to a runtime `<canvas>` — that was the actual cause of the blank card (`canvas.toDataURL()` returns empty on iOS Safari for an SVG-sourced canvas), and nothing here reintroduces it.
- The bug was never about an SVG-sourced ``. The peanut logo on this same card face is already `` and survives `html-to-image` today.
- Verified directly anyway: the crisp face pushed through `html-to-image` at `pixelRatio: 2` produces a 1560×1000 PNG with the hand fully present (screenshot 3 below).
- ⚠️ **That verification was desktop Chromium only.** The original blank-card bug was iOS Safari–specific, and I have not reproduced the capture there. The precedent above (peanut logo, same face, same inline path, shipping today) is the reason I believe it holds — not a test on the affected browser. Worth one manual Save-image on an iPhone before merge.

`onReady` still fires off the hand's `onLoad` / `complete`, so the Share/Save capture gate is unchanged.

## QA

1. `/dev/share-builder` — the card's hand renders crisp; badges, burst and `@username` pill are untouched.
2. Hit **Save image** — the downloaded PNG shows the crisp hand, not a blank card. This is the regression that matters.
3. `/card` eligibility check, add-card entry, badge-skip celebration, `/shhhhh` — hand still pixelated, whole card still reads as one consistent tease.

Local gate: `prettier --check .` clean · `typecheck` clean · `npm test` 162/162 suites, 2143 tests · `next build` compiles.

Recorded so the next reader doesn't have to re-derive them.

- **Hand placement left as-is (deliberate).** `CardFace.tsx` — the in-app card — renders this same SVG bottom-right at `h-[90%]`, commented as matching the finalised Rain card art. The share asset renders it top-right at 74% of card width, overflowing the top edge, rotated −15°. While pixelated that read as an abstract blur; crisp, it's a recognisable hand in a placement that disagrees with the canonical art. Reviewed and kept — re-anchoring is a design change, not a cleanup, and belongs in its own PR.
- **Component keeps the name `PixelatedCardFace`** even though pixelation is now the non-default. A rename touches 5 consumers plus tests; not worth widening a 2-file diff.
- **Pre-existing smell this PR reveals, NOT fixed here:** `CardFace.tsx` and `PixelatedCardFace.tsx` are two independent implementations of the same card artwork. This change makes them diverge visually rather than converge. Out of blast radius — follow-up.
- **No test pins the `blurAll` → source mapping.** Nothing stops a future refactor from routing the hand back through a `<canvas>` and silently reviving the blank-share-asset bug, which is the expensive failure mode here. This code doesn't move money or mutate shared state, so it isn't a hard gate — but it is the real coverage gap.

## Screenshots

| | |
|---|---|
| **1. Real `ShareAssetD3`, after** — native 1200×900 at 80%, username `edcriptofi`, 6 badges | after |
| **2. Card face, before vs after** — identical hand box (`top:-40 right:-20 560×471 rotate(-15deg)`); only fidelity changes, the hand does not move | before-after |
| **3. Through `html-to-image`** — what a user's shared PNG actually is | captured |

⚠️ **How these were produced, precisely** — this is not the standard phase-2b real-app-with-backend pipeline, so don't read it as one:

- #1 is the real `<ShareAssetD3 />` component with fixed props, rendered via a throwaway route outside the `(mobile-ui)` auth gate (deleted; not in this diff). The API harness was not running, so the component was driven directly rather than by a seeded user.
- #2 reproduces `PixelatedCardFace`'s exact markup and geometry in a standalone page, in order to show both hand sources side by side — the pixelated variant cannot otherwise be rendered next to the crisp one.
- #3 is the genuine `html-to-image` capture path from `captureShareAsset`, run against #2's crisp face.
- No 375×667 mobile shots: this asset is a fixed 1200×900 share image, not a responsive screen.

Assets branch `pr-assets-2516` is deleted post-merge.

## Summary by CodeRabbit

* **New Features**
  * Shared card images now show a crisp hand illustration by default, switching to the pixelated version when full blur is enabled.
* **Bug Fixes**
  * Share capture readiness is now correctly aligned with when the card face’s hand image has loaded.
* **Documentation**
  * Updated inline guidance and prop descriptions to reflect the new readiness behavior.

### DIFF
--- a/local/scratch/pr-peanut-ui-feat-share-asset-crisp-hand.md
+++ b/local/scratch/pr-peanut-ui-feat-share-asset-crisp-hand.md
@@ -1,0 +1,15 @@
+PR: https://github.com/peanutprotocol/peanut-ui/pull/2516   base: dev   worktree: mono/worktrees/peanut-ui-feat/share-asset-crisp-hand
+phase: 3 (CI)            last-tick: 873c98aff
+checks: prettier ✅ typecheck ✅ unit ✅ (162 suites) build ✅ (needs --max-old-space-size=5120 locally; default heap OOMs on this box)
+coderabbit: not yet pulled
+code-review: not yet run
+smells: TBD
+screenshots: attached via orphan branch pr-assets-2516 (DELETE post-merge)
+reporters: Konrad (this session) — no Crisp/Notion ticket
+notes:
+  - worktree has its OWN node_modules (not symlinked): primary peanut-ui tree is on a stale dev
+    without the @capacitor/* deps, which broke next dev + 3 test suites. pnpm via `corepack pnpm`
+    (global /usr/bin/pnpm is too old for the packages-less pnpm-workspace.yaml).
+  - dev server on this box dies mid-compile under memory pressure; screenshots were taken through
+    a throwaway route (deleted) rather than the real-backend phase-2b pipeline. Documented in the PR body.
+todo: watch CI → CodeRabbit → /code-review → smell pass → readiness report
--- a/src/app/(mobile-ui)/dev/share-builder/page.tsx
+++ b/src/app/(mobile-ui)/dev/share-builder/page.tsx
@@ -47,7 +47,7 @@ export default function ShareBuilderPage() {
     // ─── Capture (Save image) ────────────────────────────────────────────
     // Ref points at the native-size asset node (the pre-scale div) so the
     // capture renders at full 1200×900 fidelity. `assetReady` flips once the
-    // card face's async hand <canvas> mounts (ShareAssetD3.onReady) — Save is
+    // card face's hand <img> loads (ShareAssetD3.onReady) — Save is
     // disabled until then so a capture can never snapshot a blank card.
     const assetRef = useRef<HTMLDivElement>(null)
     const [assetReady, setAssetReady] = useState(false)
--- a/src/components/Card/share-asset/PixelatedCardFace.tsx
+++ b/src/components/Card/share-asset/PixelatedCardFace.tsx
@@ -1,6 +1,6 @@
 /**
- * <PixelatedCardFace /> — the pink peanut card with a pre-pixelated
- * peanut-card-hand overlay.
+ * <PixelatedCardFace /> — the pink peanut card with a peanut-card-hand
+ * overlay: crisp by default, pixelated under `blurAll`.
  *
  * Renders at native 620×391 (CARD_W/CARD_H). Wrap in `transform: scale(...)`
  * if you need a smaller preview — the inner layout uses absolute pixel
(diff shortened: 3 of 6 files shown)


0xkkonrad:
@coderabbitai review

0xkkonrad:
> `302-308` — **Make `onReady` idempotent**

Valid observation, declining for this PR. Reasons, in order:

1. **Pre-existing and outside this diff.** The `onLoad` + `img.complete` pair predates the branch; this PR only changes which `src` the `` gets. CodeRabbit correctly labelled it "outside diff range". Repo convention here is to *mention* adjacent issues rather than fold them into an unrelated diff.
2. **Inert with every current consumer.** All four call sites do exactly `setAssetReady(true)` — `BadgeSkipCelebration.tsx:171`, `CardUnlockDrawer.tsx:74`, `dev/share-builder/page.tsx:396`, and `ShareAssetD3` forwarding to those. Same-value `setState`, which React bails out on, so the re-fire converges instead of looping. Checked each one rather than assuming.
3. **The latent risk is real but hypothetical today**: a future consumer doing something non-idempotent in `onReady` (a PostHog event, say) would double-count. That's worth fixing — on its own, where the change is reviewable as what it is.

The docblock wording is the actual source of the ambiguity: "fires once the hand `` has loaded" means *as soon as*, not *exactly once*. Not tightening it here so the diff stays about the pixelation, but noted for the follow-up.

No other findings from this review; nothing else open.

0xkkonrad:
@coderabbitai review

kushagrasarathe (approved):
lgtm

PR #2407, opened 2026-07-13
Prod Release Sprint 151 — KYC verdict rendering · Solana/Tron/Base withdrawals · profile fixes (2026-07-16)

Hugo0:
# Prod Release Sprint 151 — Rhino chain expansion + profile fixes (FE)

dev → main. **Deploy AFTER** peanut-api-ts #1172 (needs the Rhino delivery-verification BE + its migration).

### Rhino chain expansion (#2398, #2401)
- **New withdraw destinations:** Solana, Tron, Base, Avalanche, HyperEVM, Ink, Katana, Linea, Mantle, Plasma, Stable, Tempo (SCROLL removed — Rhino disabled it). Per-family recipient validation (EVM 0x / Solana + Tron base58).
- **New deposit chains:** Tempo, Kaia, Plasma (USDT-only chains annotated).
- **CHAIN_REGISTRY** — one source of truth for every FE chain fact (7 hand-maintained maps consolidated into derived views, equality-proven).
- **Per-chain PostHog rollout flags** (`chain-rollout-*`) — launch chains one at a time with a toggle, no deploy. Staging/preview bypass; prod fails closed. Doctrine: `engineering/patterns/feature-gates.md`.
- Polish: Linea logo fix, deposit fee note, receipt explorer links.

### Profile-edit fixes (#2388)
Verified users can save email without a name; backend save errors surface instead of false-success; typed update payload.

## ⚠️ Rollout-flag state at cutover
On prod, a chain shows only if its `chain-rollout-*` flag is ON. **All 13 `chain-rollout-*` flags are ON** (Hugo, 2026-07-14) — so every chain incl. Solana/Tron goes live with the FE deploy. Re-flip any in PostHog before deploy if a staged launch is wanted.

## QA
1,831 tests; live-verified on staging. Testing guide in the Notion release page.

## Summary by CodeRabbit

* **New Features**
  * Added native demo mode with simulated balances, transactions, onboarding, and payments.
  * Added smart-paste extraction (incl. PIX keys and recipients) powered by clipboard detection, plus suggestion UI.
  * Added a connectivity banner and improved QR scanning using native clipboard detection and camera readiness.

* **Improvements**
  * Made chain availability rollout-aware across deposit/withdraw selectors and flows.
  * Updated withdraw UX text (“You’re withdrawing”) and added a small bridging-fee note for non-offramp flows.
  * Enhanced non‑EVM recipient handling (Solana/Tron) with address-family validation and automatic input reset when switching chains.

[...]

PR #2520, opened 2026-07-27
fix(qr-scanner): stop clipboard paste-prompt racing the camera; platform-gate the copied-address chip

abalinda:
## Summary

**The bug (TASK-20906):** since 32454bb83 (2026-07-02) the QR scanner read the clipboard the moment it opened — no user gesture — to offer a copied EVM address as a one-tap target. On iOS that raises the system "Allow Paste" prompt at open: it lands before `getUserMedia` (camera start deliberately sleeps 200ms on iOS), so users got a paste prompt instead of a camera, and the camera only appeared on the second attempt. On the iOS web/PWA path the blocked permission dialog also let the 5s watchdog fire a false `NotAllowedError` ("Camera start timed out").

**This PR (three commits):**

1. `cfd9797cb` — byte-exact revert of the pre-read (@abalinda). Kills the prompt-at-open on every surface; the explicit "Click to paste" button (gesture-driven) is untouched.
2. `9c78dc2bd` — platform-gated reintroduction of the shortcut (@innolope-dev, folded in from #2521): **Android native** keeps the read-at-open + address-preview chip exactly as before (only side effect is the system paste toast). **iOS native** probes `UIPasteboard.hasStrings` via a new app-local `ClipboardDetect` plugin — metadata-only, provably prompt-free — and shows a generic "Use copied address" chip whose real `Clipboard.read()` runs on tap (user gesture, camera already live). **Web/PWA** has no pre-read at all. Older iOS binaries running OTA'd JS resolve `hasStrings` → false: no chip, no error.
3. `3af6b6dc3` — review fixes: `onScan` failures on the chip path no longer misreport as "Could not access clipboard"; empty-clipboard copy unified with "Click to paste"; new component test pinning the platform split (read never fires at open off Android-native).

**Fail-safe property:** if the native plugin is missing, broken, or unregistered, `hasStrings` rejects → `false` → no chip, no error, camera unaffected. The Swift code gates only whether the *feature appears*, never whether the *bugfix works*.

## Impact

- Fixes the reported symptom directly (2 users confirmed via support; Crisp link in TASK-20906). The pre-read shipped in native binaries since 07-02 and prod web since 07-16 (#2407).
- **Sentry attribution, corrected during adversarial review:** the PEANUT-UI-PYW spike (~4.3→17/day from late June) is **not** this bug — release ancestry shows the spike ran on prod web builds that did not yet contain the pre-read. It tracks the QR-perk campaign wave: first-time users answering the real camera permission prompt slower than the pre-existing 5s watchdog. This PR removes the clipboard-caused share of current volume (post-07-16 residual) and the iOS prompt-at-open; the watchdog baseline needs its own fix (below).

## Follow-ups (separate PRs, not bundled)

1. **Web camera-start watchdog false-positive** — `useQRScanner.ts:270-278` races `scanner.start()` against 5s while the user may still be answering the real permission prompt; already fixed on the native path only (`useQRScanner.ts:260-265`). Dominant cause of PEANUT-UI-PYW.
2. **`useClipboardSuggestion` gesture-gating** — same gesture-less `Clipboard.read()` pattern (1s after mount) on claim/withdraw/request screens; still fires iOS paste prompts there.

## Risks / release notes

- **iOS chip activates only with the next iOS binary release** (new native plugin). Until then iOS behaves as pure-revert: camera immediately, no chip — which is the bugfix.
- ⚠️ **CI cannot compile Swift.** Before the next iOS binary release, one Xcode/device pass must confirm: project compiles with the hand-edited pbxproj, storyboard boots `AppViewController` (a bad customClass would crash at launch), chip appears with no Allow Paste alert at open, tap-allow pays / tap-deny leaves the camera live, and `npx cap sync ios` is run before archiving (the committed `CapApp-SPM/Package.swift` is stale — release CI syncs automatically, a raw local Xcode build does not).
- Android behavior is byte-equivalent to the pre-revert feature; web loses nothing (never had the chip working — only the prompt bug).
- No cross-repo impact, no API change.

## QA

- `pnpm typecheck` 0 errors · 2151 tests green (5 new: platform-split pin + error-attribution + copy) · `npm run build` green (Node 22) · pbxproj passes plist lint.
- 4-agent adversarial pass: revert verified byte-exact and regression-free; rework verified across the 5-surface matrix (no system UI can fire at open anywhere; OTA fallback confirmed against @capacitor/core 8.2.0 internals; plugin registration timing correct — registered in `capacitorDidLoad` before any JS, including Capgo-swapped bundles, loads).
- Post-deploy expectation: paste-prompt-at-open reports stop; PEANUT-UI-PYW drops partially, not to baseline (watchdog follow-up owns the rest).

Screenshots: ⚠️ NONE — the fix is OS permission-prompt behavior (not capturable headless); visual delta is the removed/regated chip, fully visible in the diff. Device screenshots belong to the pre-binary-release checklist above.

### DIFF
--- a/ios/App/App.xcodeproj/project.pbxproj
+++ b/ios/App/App.xcodeproj/project.pbxproj
@@ -11,6 +11,8 @@
 		4D22ABE92AF431CB00220026 /* CapApp-SPM in Frameworks */ = {isa = PBXBuildFile; productRef = 4D22ABE82AF431CB00220026 /* CapApp-SPM */; };
 		50379B232058CBB4000EE86E /* capacitor.config.json in Resources */ = {isa = PBXBuildFile; fileRef = 50379B222058CBB4000EE86E /* capacitor.config.json */; };
 		504EC3081FED79650016851F /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 504EC3071FED79650016851F /* AppDelegate.swift */; };
+		7A1C0DE7C11B0A4D2E5F3901 /* AppViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A1C0DE7C11B0A4D2E5F3902 /* AppViewController.swift */; };
+		7A1C0DE7C11B0A4D2E5F3903 /* ClipboardDetectPlugin.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A1C0DE7C11B0A4D2E5F3904 /* ClipboardDetectPlugin.swift */; };
 		504EC30D1FED79650016851F /* Main.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 504EC30B1FED79650016851F /* Main.storyboard */; };
 		504EC30F1FED79650016851F /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 504EC30E1FED79650016851F /* Assets.xcassets */; };
 		504EC3121FED79650016851F /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 504EC3101FED79650016851F /* LaunchScreen.storyboard */; };
@@ -22,6 +24,8 @@
 		50379B222058CBB4000EE86E /* capacitor.config.json */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.json; path = capacitor.config.json; sourceTree = "<group>"; };
 		504EC3041FED79650016851F /* App.app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = App.app; sourceTree = BUILT_PRODUCTS_DIR; };
 		504EC3071FED79650016851F /* AppDelegate.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AppDelegate.swift; sourceTree = "<group>"; };
+		7A1C0DE7C11B0A4D2E5F3902 /* AppViewController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AppViewController.swift; sourceTree = "<group>"; };
+		7A1C0DE7C11B0A4D2E5F3904 /* ClipboardDetectPlugin.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ClipboardDetectPlugin.swift; sourceTree = "<group>"; };
 		504EC30C1FED79650016851F /* Base */ = {isa = PBXFileReference; lastKnownFileType = file.storyboard; name = Base; path = Base.lproj/Main.storyboard; sourceTree = "<group>"; };
 		504EC30E1FED79650016851F /* Assets.xcassets */ = {isa = PBXFileReference; lastKnownFileType = folder.assetcatalog; path = Assets.xcassets; sourceTree = "<group>"; };
 		504EC3111FED79650016851F /* Base */ = {isa = PBXFileReference; lastKnownFileType = file.storyboard; name = Base; path = Base.lproj/LaunchScreen.storyboard; sourceTree = "<group>"; };
@@ -64,6 +68,8 @@
 			children = (
 				50379B222058CBB4000EE86E /* capacitor.config.json */,
 				504EC3071FED79650016851F /* AppDelegate.swift */,
+				7A1C0DE7C11B0A4D2E5F3902 /* AppViewController.swift */,
+				7A1C0DE7C11B0A4D2E5F3904 /* ClipboardDetectPlugin.swift */,
 				504EC30B1FED79650016851F /* Main.storyboard */,
 				504EC30E1FED79650016851F /* Assets.xcassets */,
 				504EC3101FED79650016851F /* LaunchScreen.storyboard */,
@@ -156,6 +162,8 @@
 			buildActionMask = 2147483647;
 			files = (
 				504EC3081FED79650016851F /* AppDelegate.swift in Sources */,
+				7A1C0DE7C11B0A4D2E5F3901 /* AppViewController.swift in Sources */,
+				7A1C0DE7C11B0A4D2E5F3903 /* ClipboardDetectPlugin.swift in Sources */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
--- a/ios/App/App/AppViewController.swift
+++ b/ios/App/App/AppViewController.swift
@@ -1,0 +1,9 @@
+import Capacitor
+import UIKit
+
+// App-local plugins aren't auto-discovered; register them once the bridge is up.
+class AppViewController: CAPBridgeViewController {
+    override open func capacitorDidLoad() {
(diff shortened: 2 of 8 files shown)


abalinda:
@coderabbitai review

innolope-dev:
@abalinda this is me who worked on the clipboard feature. During testing I didn't experience the issue you describe. Let me rework the feature. On Android it works, I'll do some tweaks on iOS

abalinda:
> @abalinda this is me who worked on the clipboard feature. During testing I didn't experience the issue you describe. Let me rework the feature. On Android it works, I'll do some tweaks on iOS

@innolope-dev hi! overcommunicating here, you will own this PR and work on it/open a new one that will fix both PWA and native apps?

innolope-dev:
Pushed 9c78dc2bd on top of the revert (folding in what was briefly #2521): reintroduces the clipboard shortcut with a design that can't race the camera prompt — Android keeps the pre-read + address-preview chip exactly as before this PR; iOS gets a prompt-free `UIPasteboard.hasStrings` check via a new app-local `ClipboardDetect` plugin and a generic 'Use copied address' chip whose actual `Clipboard.read()` runs on tap (user gesture, camera already live); web/PWA has no pre-read (the surface where PEANUT-UI-PYW spiked). Older iOS binaries resolve `hasStrings` to false → no chip, no error; the iOS chip activates with the next binary release. ⚠️ CI doesn't compile Swift — needs one native build/device pass to confirm plugin registration.

innolope-dev:
@abalinda I added an update on top of your commit

abalinda:
reviewed `9c78dc2bd` with an adversarial pass (full matrix: iOS/Android native, iOS PWA, Android/desktop web, old-binary OTA). The design holds — `hasStrings` is metadata-only so nothing can prompt at open, registration in `capacitorDidLoad` lands before any JS (incl. Capgo-swapped bundles), and the whole feature fails safe (plugin missing/broken → `hasStrings` rejects → no chip, no error, camera untouched).

Pushed `3af6b6dc3` on top with three small review fixes rather than round-tripping them to you:

1. `handleUsePasteChip` had `await onScan(address)` inside the clipboard `try` — a payment-routing failure toasted "Could not access clipboard" and hid the chip. Split it out; processing failures now use the existing "Error processing QR code" copy.
2. Empty clipboard on chip tap now maps to "Clipboard is empty", same as `handlePaste` for the same rejection.
3. Added `QRScanner/__tests__/index.test.tsx` pinning the load-bearing claim: `Clipboard.read()` never fires at scanner open off Android-native; iOS reads only on tap; Android keeps read-at-open.

**Before the next iOS binary release** (CI can't compile Swift — your ⚠️), the device checklist:

- [ ] Run `npx cap sync ios` before building — the committed `CapApp-SPM/Package.swift` predates the Clipboard/Camera deps; release CI syncs automatically but a raw local Xcode build ships a chip whose tap always fails UNIMPLEMENTED.
- [ ] Project compiles; app boots (storyboard now instantiates `AppViewController` — a class-resolution failure would crash at launch).
- [ ] iOS 16–18: open scanner with text copied → chip appears, **no** Allow Paste alert, first-run camera permission dialog unimpeded.
- [ ] Chip tap → Allow: pays. Deny: toast, chip hides, camera stream still live and scanning.
- [ ] Old store binary + Capgo OTA of this JS: no chip, no error.
- [ ] Android 12+: chip at open with an address copied; only the system clipboard toast.

One paper trail note: the Sentry step-change on PEANUT-UI-PYW predates your feature reaching prod web (details in the PR body) — the spike is the pre-existing 5s camera watchdog racing real permission prompts, which is now a named follow-up. Your instinct that the feature "worked in testing" was right for the surfaces you tested; the prompt collision was iOS-at-open specifically.

@kushagrasarathe

kushagrasarathe:
### Review notes — 1 should-fix + 2 minor

Overall solid: platform split is justified, tests pin the right invariant, risks disclosed honestly. Two changes worth folding in before merge (they're really one small change), plus two nits.

**1. Should-fix — the misreport bug you just fixed on the chip path still lives in `handlePaste`**

Commit `3af6b6dc3` fixed `handleUsePasteChip` so an `onScan` rejection no longer surfaces as "Could not access clipboard". But the sibling **"Click to paste"** handler still has the identical pattern:

```js
// handlePaste
try {
    const { value } = await Clipboard.read()
    const text = (value ?? '').trim()
    if (text) {
        await onScan(text)          // <-- onScan is INSIDE the clipboard try
    } ...
} catch (err) {
    toast.error('Could not access clipboard')   // any onScan/routing failure misreports here
}
```

Root-cause fix: pull `onScan` out of the clipboard `try` in `handlePaste` too, mirroring `handleUsePasteChip`. Otherwise the same class of bug you fixed is still live on the paste button.

**2. DRY — clipboard error-mapping duplicated**

The `no data on the clipboard → 'Clipboard is empty' : 'Could not access clipboard'` block now exists in both `handlePaste` and `handleUsePasteChip`. Extract one helper (e.g. `clipboardErrorToast(err)`) and call from both — folds naturally into #1.

**3. Minor — `onUseDetected` leaves `onScan` unguarded**

```js
onUseDetected={() => onScan(detectedAddress!)}   // Android path — no try/catch
```

Every other tap path now wraps `onScan`. If it rejects here → unhandled rejection, no user feedback. Pre-existing, but it's the one remaining asymmetry in a PR that's normalizing exactly this.

**4. Minor UX (fine to ship)** — `hasStrings` is content-blind, so the iOS "Use copied address" chip can appear for non-address clipboard text, then the tap resolves to "Copied text is not a wallet address". Graceful, just over-promises. Noting only.

Bug risk on the whole is low — `registerPlugin` at load is SSR/web-safe, the two chip states are mutually exclusive by platform, and the OTA fallback is correct + tested. The pbxproj/Swift device pass in the PR body remains the real release gate.

abalinda:
@kushagrasarathe all three folded into `a7a466b86`:

1. `handlePaste` no longer wraps `onScan` in the clipboard `try` — routing failures now toast "Error processing QR code" there too.
2. The read + error mapping lives once in `readClipboardText()`; the guarded scan lives once in `scanValue()` — all three tap paths (paste button, iOS chip, Android chip) funnel through them.
3. `onUseDetected` now calls `scanValue`, so the Android chip gets the same guard instead of an unhandled rejection.

Two new tests pin 1 and 3. On 4 — agreed it's inherent to a content-blind `hasStrings`; leaving as shipped.

kushagrasarathe (approved):
Approving. All three review items from the earlier pass are closed in `a7a466b86`:

- `handlePaste` onScan-misreport fixed (onScan now outside the clipboard try, via `scanValue`)
- error-mapping DRY'd into one `readClipboardText` helper
- Android detected-address chip now guarded (no more unhandled rejection)

Traced the new three-state return contract (`null` = read failed + already toasted / `''` = empty, not toasted / text) — both callers handle it correctly, no double-toast, paste-button raw-text behavior preserved. Two new tests pin the newly guarded paths.

One non-code gate remains before the next iOS binary release (not blocking this merge to dev): the manual Xcode/device pass for the hand-edited pbxproj + `AppViewController` storyboard swap + new Swift plugin — CI can't compile Swift. Already documented in the PR body.

PR #2523, opened 2026-07-27
content: publish latest to production (src/content → peanut-content@05fffeb)

Hugo0:
Bumps `src/content` on **production (`main`)** from `055f60a` → `05fffeb` (peanut-content latest = `mono@fb32c83`). Single-file submodule pointer change publishing content already merged to mono. The automated `update-content` PRs reach this same pointer but target `dev`, which never promotes.

## Summary by CodeRabbit

* **Content Updates**
  * Updated the bundled content version used by the application.


Hugo0:
Superseded — the content was amended after this PR opened (removed three service-level promises: guaranteed reply, "usually the same day", and "we will tell you when a fix ships"). Reopening at the corrected content SHA.

PR #2527, opened 2026-07-27
fix(receipt): shorten crypto destination address so it can't overflow the card

abalinda:
## Summary
A withdrawal to a Solana address renders the full 44-char base58 address in the receipt drawer's **Address** row. It's one unbreakable token, so the whole card blows out of the viewport on mobile (user report: Crisp session_779df6b2 / Notion TASK-20889). EVM (42-char) and Tron (34-char) destinations overflow the same way.

**Root cause:** the 2026-07-23 Tron fix made `maskAccountIdentifier` return crypto addresses **verbatim** (correctly bypassing the `formatIban` base58 corruption) — but nothing shortens them for display, unlike every other long value in the drawer (header, TX ID, Transfer ID all shorten). Structurally, `PaymentInfoRow` also had no flex containment: `min-width: auto` resolves to the token's full width, so any unbreakable value escapes the card.

**Fix (3 pieces):**
1. `maskAccountIdentifier` crypto branch shortens for display (`BfbXuD...NKNb19` — same shape the drawer header shows). Copy untouched: `getAccountCopyValue` still yields the verbatim address.
2. `PaymentInfoRow` value wrapper gets `min-w-0 break-words` — any future unbreakable token wraps inside the card instead of stretching the row. `break-word` only activates when a word can't fit, so normal values render unchanged.
3. The **To** row drops its viem `isAddress` pre-guard (EVM-only — Solana/Tron counterparties rendered full-length); `printableAddress` already passes usernames through untouched.

**Why not `printableAddress` in (1):** it re-validates shape via viem `isAddress`, which rejects wrong-checksum mixed-case strings and would silently fall back to the full overflowing address. The wire `type` (`address` / `evm-address` / `peanut-wallet`) already tells us it's a crypto address — shorten unconditionally, with a length guard for degenerate short strings.

**Base = `main`:** `account-mask.utils.ts` / `isCryptoAddressType` only exist on `main` (the Tron hotfix hasn't back-merged to `dev`). Merging adds to the existing main→dev back-merge debt.

- Adversarially reviewed by two independent agents (correctness attack + codebase-wide overflow sweep). Findings: fix covers the reported row; deposit-instruction surfaces (where the full address must stay readable) are on a separate render path and untouched.
- Known remaining sibling (out of scope, contained by (2) but not shortened): the `isGuestBankClaim` branch renders identifiers verbatim — a full IBAN now wraps instead of overflowing. Follow-up candidate, not bundled.
- `CRYPTO_ADDRESS_TYPES` is a closed 3-value set; if the BE ever emits a new crypto wire type it falls to the `plain` branch (pre-existing property, unchanged here).

## Risks / breaking changes
- Display-only. Bank rails (IBAN/CLABE/PIX/ACH/alias) untouched — pinned by existing tests. `break-words` affects all 22 `PaymentInfoRow` consumers but only when a single word exceeds the line — no layout change for fitting values.
- No cross-repo impact.

## QA
- `npm test`: 2154 passed (updated the 2 verbatim-display assertions to shortened; added EVM case).
- Full typecheck clean; eslint on touched files: 0 new problems (5 pre-existing in TransactionDetailsReceipt, none on changed lines).

## Screenshots
Captured on the local sandbox (mobile 375×667, seeded user, mocked Solana `CRYPTO_WITHDRAW` via route interception — same entry shape the BE serializes). Before = `origin/main`, After = this branch.

| Before (main) | After (this PR) |
|---|---|
| before | after |

The before shot reproduces the user's report exactly: the 44-char address escapes the card's right edge and the copy icon is pushed out of view. After: shortened display, copy icon inline, copy still yields the full address. (Assets live on the `pr-assets-2527` branch — delete after merge.)

## Summary by CodeRabbit

* **New Features**
  * Crypto wallet addresses are now displayed in a shortened, readable format in transaction receipts.
  * Transaction details better support displaying usernames and addresses across supported networks.

* **Bug Fixes**
  * Long wallet addresses and transaction identifiers now wrap within payment information rows instead of breaking the layout.
  * Improved receipt formatting for Tron, Solana, and EVM addresses.

### DIFF
--- a/src/components/Payment/PaymentInfoRow.tsx
+++ b/src/components/Payment/PaymentInfoRow.tsx
@@ -49,8 +49,13 @@ export const PaymentInfoRow = ({
                 <Loading />
             ) : (
                 <div className="flex items-center justify-between">
-                    <div className={twMerge('flex w-fit justify-end text-sm font-bold')}>
-                        <span>{value}</span>
+                    {/* min-w-0 + break-words: a single unbreakable token (wallet
+                        address, tx hash) must wrap inside the card, not stretch
+                        the row to the token's full width and escape the layout.
+                        break-word only activates when a word can't fit, so
+                        normal values render unchanged. */}
+                    <div className={twMerge('flex w-fit min-w-0 justify-end break-words text-sm font-bold')}>
+                        <span className="min-w-0">{value}</span>
                     </div>
                     {allowCopy && typeof value === 'string' && (
                         <CopyToClipboard textToCopy={copyValue ?? value} fill="black" iconSize="4" />
--- a/src/components/TransactionDetails/TransactionDetailsReceipt.tsx
+++ b/src/components/TransactionDetails/TransactionDetailsReceipt.tsx
@@ -39,7 +39,6 @@ import { TransactionDetailsHeaderCard } from './TransactionDetailsHeaderCard'
 import CopyToClipboard from '../Global/CopyToClipboard'
 import CancelSendLinkModal from '../Global/CancelSendLinkModal'
 import { twMerge } from 'tailwind-merge'
-import { isAddress } from 'viem'
 import { getAccountCopyValue, getBankAccountLabel } from './transaction-details.utils'
 import { useModalsContext } from '@/context/ModalsContext'
 import { useRouter } from 'next/navigation'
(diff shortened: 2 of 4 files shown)


abalinda:
@coderabbitai review

kushagrasarathe (approved):
lgtm

PR #2528, opened 2026-07-27
fix(ws): don't render minimal charge-completion pings as $0.00 history rows

abalinda:
## Summary

Fixes the "withdrawal briefly renders as **Sent to Transaction $0.00 · Completed**" flash (TASK-20808, Sentry [PEANUT-UI-QCW](https://peanut-c34d84c05.sentry.io/issues/PEANUT-UI-QCW) — 104 users / 144 events in 30d).

**Root cause:** on charge completion the BE (`charges-ws`) pushes a deliberately *minimal* `history_entry` payload — `{uuid, type, status: 'COMPLETED', isRequestLink}` — with the stated contract "clients refetch full entry from /users/history". The FE instead rendered the ping as a full entry: `useWebSocket` buffers it, `HomeHistory` / the history page merge it into the feed, and with no `extraData.kind` the transformer routes to `intentFallback` → name "Transaction" (initials TR), direction `send`, no `usdAmount` → `$0.00`, plus a green Completed badge. Worse, HomeHistory's uuid-keyed merge *replaces* the real fetched row with the ping, so the broken row can outlive the next refetch until remount.

**Fix:** honor the existing contract at the shared funnel. In `useWebSocket.handleHistoryEntry`, a kindless entry is never rendered or forwarded — it invalidates the `[TRANSACTIONS]` and `['balance']` queries instead (with `cancelRefetch: false` so the several mounted hook instances join one in-flight refetch), and the real row appears via refetch. This covers **both** kindless ping sources by design: the charge-completion ping (`charges-ws`) and the sendlink-claim ping (`routes/claim.ts` — same minimal shape, same "clients refetch" contract). Full WS entries (crypto deposits, Bridge offramps) carry `kind` and keep today's real-time rendering path. The `['balance']` invalidation preserves the /history page's previous behavior of refreshing balance on pings (its per-page callback no longer sees them).

## Risks / breaking changes

- FE-only; no API change. The BE ping (`charges-ws`) is untouched.
- Behavior delta: the completed charge now appears after a fast refetch instead of an instant-but-broken row. `fetchBalance` on home never keyed off the ping (it gates on `kind`), so no balance-refresh regression.
- Consumers without `onHistoryEntry` (card overview, KYC hooks) each invalidate on ping — TanStack Query coalesces concurrent refetches of the same key; negligible cost.
- Base is `main` (prod hotfix): needs the usual main→dev back-merge after merge.

## Adversarial review (2 independent agents)

- **Diagnosis-refutation agent: CONFIRMED** every link at file:line, ruled out alternative culprits (REST serializers all stamp `kind`; SW never caches `/users/history`; PaymentSuccessView is flow-prop-fed). Refinements folded in: the emitter also fires on charge creation/cancel/FAILED (payload hardcodes COMPLETED — BE follow-up), and pre-fix the buffered ping *re-clobbered* the refetched real row on every merge pass until remount, so the bug was persistent-while-mounted, worse than "briefly".
- **Regression-hunt agent: SHIP.** No consumer relies on kindless pings (requester's pending→paid flip never came from the ping — different uuids; it always came from the refetch, which the invalidate still triggers). QueryClient is a module-level singleton → no provider/dep-array churn risk. Its two concerns (balance invalidation, `cancelRefetch`) are addressed in commits `abff09d41` and `58d9ef654`.

## Follow-ups (filed, not in this PR)

- BE: stop hardcoding `status:'COMPLETED'` in the charge ping / don't emit it for non-terminal transitions (creation, cancel, FAILED currently ping "COMPLETED"; also 3 refetches per withdrawal).
- BE: move both minimal pings (`charges-ws`, `claim.ts`) off the render-shaped `history_entry` message type so future FE code can't regress into rendering them.
- FE: /history page never *updates* an existing cached row from a kinded WS entry (pre-existing); HomeHistory's uuid-merge can replace richer fetched rows with kinded WS entries (pre-existing).

## QA

- `src/hooks/__tests__/useWebSocket.test.tsx`: ping → not rendered + `invalidateQueries([TRANSACTIONS])`; kinded entry → rendered + forwarded; pending-request skip preserved.
- Full suite 2156 ✅, typecheck ✅, prettier ✅, prod build ✅.
- Prod evidence trail: Sentry PEANUT-UI-QCW events carry `entryUuid` = charge uuid of the just-completed intent (verified against prod DB for the reporting user's withdrawal, 2026-07-21 16:40 UTC).

Screenshots: ⚠️ NONE — the bug is a transient WS race (broken row exists only between the ping and the next refetch); not capturable in a static state. The jest suite pins the behavior instead.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of incoming transaction history updates that don’t include full charge details.
  * Automatically refreshes transaction and balance information when charge completion events are received without complete metadata.
  * Prevents incomplete or pending requests from appearing in history or triggering callbacks.
  * Complete history entries continue to display and trigger the expected notifications.
* **Tests**
  * Added coverage for WebSocket `history_entry` scenarios to ensure the above behaviors remain consistent.

### DIFF
--- a/src/hooks/useWebSocket.ts
+++ b/src/hooks/useWebSocket.ts
@@ -169,11 +169,12 @@ export const useWebSocket = (options: UseWebSocketOptions = {}) => {
                 // hits the transformer's fallback strategy and shows "Sent to
                 // Transaction $0.00 · Completed" (PEANUT-UI-QCW). Balance moves
                 // with these events too, so refresh it alongside the feed.
-                // cancelRefetch: false — several mounted hook instances see the
-                // same ping; let them join the in-flight refetch instead of
-                // abort-restarting it.
-                queryClient.invalidateQueries({ queryKey: [TRANSACTIONS] }, { cancelRefetch: false })
-                queryClient.invalidateQueries({ queryKey: ['balance'] }, { cancelRefetch: false })
+                // Default cancelRefetch (true) on purpose: a fetch already in
+                // flight when the ping arrives started pre-commit and may lack
+                // the new row — joining it would clear the invalidation with
+                // stale data. Abort-restart guarantees a post-event response.
+                queryClient.invalidateQueries({ queryKey: [TRANSACTIONS] })
+                queryClient.invalidateQueries({ queryKey: ['balance'] })
                 return
             }
             if (
--- a/src/hooks/useWebSocket.ts
+++ b/src/hooks/useWebSocket.ts
@@ -163,11 +163,14 @@ export const useWebSocket = (options: UseWebSocketOptions = {}) => {
         const handleHistoryEntry = (entry: HistoryEntry) => {
             const kind = entry.extraData?.kind
             if (!kind) {
-                // Charge-completion pings are minimal {uuid, status} payloads —
-                // the BE (charges-ws) expects clients to refetch, not render.
-                // Rendering one hits the transformer's fallback strategy and
-                // shows "Sent to Transaction $0.00 · Completed" (PEANUT-UI-QCW).
+                // Kindless entries are minimal {uuid, status} pings (BE:
+                // charges-ws charge completions, claim.ts sendlink claims) —
+                // the BE expects clients to refetch, not render. Rendering one
+                // hits the transformer's fallback strategy and shows "Sent to
+                // Transaction $0.00 · Completed" (PEANUT-UI-QCW). Balance moves
+                // with these events too, so refresh it alongside the feed.
                 queryClient.invalidateQueries({ queryKey: [TRANSACTIONS] })
+                queryClient.invalidateQueries({ queryKey: ['balance'] })
                 return
             }
             if (
--- a/src/hooks/useWebSocket.ts
+++ b/src/hooks/useWebSocket.ts
@@ -1,4 +1,5 @@
 import { useEffect, useState, useCallback, useRef } from 'react'
+import { useQueryClient } from '@tanstack/react-query'
 import {
     PeanutWebSocket,
     getWebSocketInstance,
@@ -6,6 +7,7 @@ import {
     type RailStatusUpdate,
     type RainCardBalanceChangedData,
 } from '@/services/websocket'
+import { TRANSACTIONS } from '@/constants/query.consts'
 import { type HistoryEntry } from './useTransactionHistory'
 
 type WebSocketStatus = 'connecting' | 'connected' | 'disconnected' | 'error'
@@ -50,6 +52,7 @@ export const useWebSocket = (options: UseWebSocketOptions = {}) => {
     const [status, setStatus] = useState<WebSocketStatus>('disconnected')
     const [historyEntries, setHistoryEntries] = useState<HistoryEntry[]>([])
     const wsRef = useRef<PeanutWebSocket | null>(null)
+    const queryClient = useQueryClient()
 
     const callbacksRef = useRef({
         onHistoryEntry,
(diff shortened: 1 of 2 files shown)


abalinda:
@coderabbitai review

src/hooks/useWebSocket.ts
```diff
@@ -159,6 +162,20 @@ export const useWebSocket = (options: UseWebSocketOptions = {}) => {
...
+                // Kindless entries are minimal {uuid, status} pings (BE:
+                // charges-ws charge completions, claim.ts sendlink claims) —
+                // the BE expects clients to refetch, not render. Rendering one
+                // hits the transformer's fallback strategy and shows "Sent to
+                // Transaction $0.00 · Completed" (PEANUT-UI-QCW). Balance moves
+                // with these events too, so refresh it alongside the feed.
+                // cancelRefetch: false — several mounted hook instances see the
+                // same ping; let them join the in-flight refetch instead of
+                // abort-restarting it.
+                queryClient.invalidateQueries({ queryKey: [TRANSACTIONS] }, { cancelRefetch: false })
```
jjramirezn: **`cancelRefetch: false` can serve this invalidation from a *pre-event* in-flight fetch, silently dropping the real-time update.**

The scenario: the transactions query is already refetching for an unrelated reason (30s stale-timer poll, navigation-triggered fetch) when the WS ping arrives. With `cancelRefetch: false`, `invalidateQueries` joins that in-flight request instead of cancelling and re-issuing it — but that request's response was produced **before** the charge/claim was committed. It resolves without the new entry and clears the invalidated state, so the row the ping exists to deliver doesn't appear until the next 30s poll or a navigation. Same race applies to the `['balance']` invalidation below. TanStack's default `cancelRefetch: true` exists precisely to close this window: any fetch started before the invalidation is by definition potentially stale.

The tradeoff as written is inverted: `false` optimizes the N-mounted-instances case (one shared request instead of N-1 abort/restarts) but pays for it with a correctness hole in the single flow this handler exists for. The default gives the opposite — a few redundant aborts per ping, but the surviving request is always post-event.

Suggestion: take the default on both calls and, if the abort churn is worth eliminating, dedupe at the source instead — handle the kindless ping once per event (e.g. at the PeanutWebSocket singleton layer, or a per-uuid guard) rather than once per mounted hook instance. That gets one fresh request per ping *and* closes the race. (Related, separate concern: the bare `[TRANSACTIONS]` prefix also matches the /history infinite query and refetches every loaded page per ping — worth scoping or accepting explicitly.)

abalinda: You're right — accepted and fixed in 396cbfe8f. A pre-event in-flight fetch satisfying the invalidation is exactly the flow this handler exists for (mount fetch racing the completion ping), so `cancelRefetch: false` inverted the tradeoff: it optimized redundant aborts at the cost of the update itself. Both calls now take the default; the comment in the code states why the abort-restart is load-bearing.

On the two side points:
- **Singleton-layer / per-uuid dedupe:** declined for now as speculative — the churn is a few aborted fetches per ping and the default is correct; if it ever shows up in network profiles, the dedupe belongs at `PeanutWebSocket` as you say. Noted in the PR follow-ups.
- **Bare `[TRANSACTIONS]` prefix refetching every loaded /history page:** accepted explicitly. It's the pre-existing idiom (HomeHistory invalidated the same prefix on every WS entry before this PR) and correctness-preserving; scoping it is orthogonal to this fix.

jjramirezn (approved):

PR #2532, opened 2026-07-27
fix(fetch): context-aware fetch timeouts — stop applying a stale Vercel limit to browsers

Hugo0:
## Summary

`fetchWithSentry` capped **every** `api.peanut.me` call at 10s, on a comment that said *"vercel function timout is 15s"*. That number was hand-copied from the platform years ago and never followed it:

- our own `vercel.json` declares `maxDuration: 300` for the `app/api` glob;
- Vercel's Fluid compute (default-on) [defaults to 300s](https://vercel.com/docs/functions/configuring-functions/duration) on every plan, max 800s, and [up to 1800s](https://vercel.com/changelog/vercel-functions-can-now-run-up-to-30-minutes) for Pro/Enterprise since 2026-06-15.

So a *server-side* safety margin ended up governing *browser* requests, which never run inside a function and have no platform ceiling at all.

**That budget is smaller than a page load in high-latency markets.** Measured over 90d (PostHog + Sentry):

| | Nigeria | Global |
|---|---|---|
| p90 LCP | 11,326 ms | 6,140 ms |
| `$exception` rate | 40.6% | 16.8% |
| `…timed out after 10000ms` | 15.1% | 5.1% (**3.0×**) |
| `Load failed` / `Failed to fetch` | 25.8% | 8.6% (**3.0×**) |
| service-worker `no-response` | 3.0% | 0.14% (**21×**) |

Top offenders: `/users/me` (33 persons), `/tokens/price` (23), `/rain/cards` (16). Our API origin is `gcp-us-west1` (Oregon); TTFB from Lagos is 0.86–0.93s vs 0.13s from Spain (~7×). This is every high-latency market, not just NG.

## What changed

One magic number → two named budgets picked by the constraint that actually applies, in one place (`src/utils/sentry.utils.ts`):

| Constant | Value | Why |
|---|---|---|
| `VERCEL_FUNCTION_MAX_DURATION_MS` | 300,000 | Mirrors `vercel.json`. A test parses `vercel.json` and fails if they drift — the whole bug was a platform constant copied by hand and then abandoned. |
| `SERVER_FETCH_TIMEOUT_MS` | 30,000 | Must abort **before** the platform kills the function so we emit our own error instead of an opaque 504. Vercel→`api.peanut.me` is a datacenter hop, so 30s is already generous; the 300s ceiling is the safety net, not the target. |
| `CLIENT_FETCH_TIMEOUT_MS` | 20,000 | No platform ceiling in a browser — only real mobile networks. 20s clears NG's 11.3s p90 LCP with margin while bounding the retried worst case (below). |

`NEXT_PUBLIC_FETCH_TIMEOUT_MS` still overrides **both** (`.env.example` updated: it was pinned to `10000`, now empty so the context-aware defaults apply). The per-call third-arg `timeoutMs` on `fetchWithSentry` still wins over everything — unchanged, and now covered by a test.

## Risk: the retry multiplier

**This is the main thing to review.** The timeout is per *attempt*, and React Query retries on top — `RETRY_STRATEGIES.FAST` (2 retries, 1s/2s backoff) is the global default in `src/config/wagmi.config.tsx`.

```
worst case = 3 attempts × timeout + 1s + 2s backoff
  before:  3×10s + 3s = 33s
  after:   3×20s + 3s = 63s     ← accepted
  (at 30s: 3×30s + 3s = 93s     ← rejected, real UX regression)
```

**Accepted trade-off.** 63s is the pathological tail where all three attempts hang to full budget — i.e. the network is genuinely dead, and the user was going to fail anyway. The common case *improves*: an NG request that really takes 12s now succeeds on attempt 1 instead of erroring at 10s. That tail is exactly the 15.1% of NG users currently generating timeout events.

**Considered and rejected:** dropping `FAST` from 2 retries to 1 (total 41s). It's the global wagmi default, so it would also halve retry resilience for *fast* failures (transient 5xx, which fail in <1s and cost ~3s for all 3 attempts today) — a resilience regression in a case unrelated to this bug. Not worth bundling. Happy to revisit if you'd rather cap the total.

**Financial mutations are unaffected** — `RETRY_STRATEGIES.FINANCIAL` is `retry: false` and stays that way.

**Bonus, no action needed:** a longer budget *reduces* double-submit risk — fewer POSTs that actually succeeded get aborted client-side and look failed.

## Tests

`src/utils/__tests__/sentry.utils.test.ts`:
- client vs server default resolution (and that the two contexts genuinely differ)
- `NEXT_PUBLIC_FETCH_TIMEOUT_MS` overrides both contexts; unset/unparseable falls back
- **drift guard**: parses `vercel.json` and asserts `maxDuration × 1000 === VERCEL_FUNCTION_MAX_DURATION_MS`, and that the server budget stays strictly under it
- regression pin: neither budget may return to 10s
- `fetchWithSentry` defaults to the client budget, and a per-call `timeoutMs` still wins

## Out of scope (flagged, not touched)

- `src/app/(mobile-ui)/card/page.tsx` polling (`pollUntilReady` 30000, `pollUntilApplyAdvances` 5000) — separate mechanism, unaffected.
- 🐛 **`src/hooks/useGeoLocation.ts:48` fetches `ipapi.co/country` with no timeout, no AbortController and no retry**, and it gates a skeleton in `CountryList.tsx:145`. On the same slow networks this hangs the country picker indefinitely — a real bug, but a genuinely separate PR.

## QA

Unit tests cover the resolution logic. Behavioural verification is inherently network-dependent; the drift guard is the durable protection against this recurring.

## Summary by CodeRabbit

- **New Features**
  - Added context-aware fetch timeout defaults for client and server environments.
  - Added an optional configuration override for fetch timeouts, with automatic defaults when left unset.
  - Improved timeout reporting for failed requests.

- **Tests**
  - Added coverage for default, override, fallback, and environment-specific timeout behavior.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -73,7 +73,9 @@ export PROMO_LIST={}
 export NEXT_PUBLIC_VAPID_PUBLIC_KEY=
 export VAPID_PRIVATE_KEY=
 export NEXT_PUBLIC_VAPID_SUBJECT="mailto:hello@peanut.me"
-export NEXT_PUBLIC_FETCH_TIMEOUT_MS=10000
+# Optional override of BOTH fetch budgets (client 20s / server 30s — see
+# src/utils/sentry.utils.ts). Leave empty to keep the context-aware defaults.
+export NEXT_PUBLIC_FETCH_TIMEOUT_MS=
 
 # one signal 
 export NEXT_PUBLIC_ONESIGNAL_APP_ID=
--- a/src/utils/__tests__/sentry.utils.test.ts
+++ b/src/utils/__tests__/sentry.utils.test.ts
@@ -1,5 +1,16 @@
+import { readFileSync } from 'node:fs'
+import { join } from 'node:path'
 import * as Sentry from '@sentry/nextjs'
-import { fetchWithSentry, sanitizeRequestBody, sanitizeResponseBody, scrubObject } from '../sentry.utils'
+import {
+    CLIENT_FETCH_TIMEOUT_MS,
+    SERVER_FETCH_TIMEOUT_MS,
+    VERCEL_FUNCTION_MAX_DURATION_MS,
+    fetchWithSentry,
+    resolveDefaultTimeoutMs,
+    sanitizeRequestBody,
+    sanitizeResponseBody,
+    scrubObject,
+} from '../sentry.utils'
 
 jest.mock('@sentry/nextjs', () => ({
     withScope: jest.fn((cb: (scope: unknown) => void) => cb({ setFingerprint: jest.fn(), setTag: jest.fn() })),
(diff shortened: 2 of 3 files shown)


Hugo0:
@coderabbitai review

Hugo0:
Superseded by #2533, which targets `main` (production) directly per Hugo — same branch content, re-cut from `origin/main` so the PR contains only this fix rather than the 35 unreleased commits `dev` is ahead by. CodeRabbit's override-validation finding from this PR was applied there. Closing.

PR #2533, opened 2026-07-27
fix(fetch): context-aware fetch timeouts — stop applying a stale Vercel limit to browsers

Hugo0:
> **Targets `main` (production) directly**, per Hugo: one constant becoming context-aware — no schema, no money path, trivially revertible — and staging cannot validate it anyway, since the whole point is behaviour on real high-latency mobile networks. **Back-merge to `dev` is required immediately after merge** (see bottom).

## Summary

`fetchWithSentry` capped **every** `api.peanut.me` call at 10s, on a comment that said *"vercel function timout is 15s"*. That number was hand-copied from the platform years ago and never followed it. So a *server-side* safety margin ended up governing *browser* requests, which never run inside a function and have no platform ceiling at all.

**That budget is smaller than a page load in high-latency markets.** Measured over 90d (PostHog + Sentry):

| | Nigeria | Global |
|---|---|---|
| p90 LCP | 11,326 ms | 6,140 ms |
| `$exception` rate | 40.6% | 16.8% |
| `…timed out after 10000ms` | 15.1% | 5.1% (**3.0×**) |
| `Load failed` / `Failed to fetch` | 25.8% | 8.6% (**3.0×**) |
| service-worker `no-response` | 3.0% | 0.14% (**21×**) |

Top offenders: `/users/me` (33 persons), `/tokens/price` (23), `/rain/cards` (16). Our API origin is `gcp-us-west1` (Oregon); TTFB from Lagos is 0.86–0.93s vs 0.13s from Spain (~7×). This is every high-latency market, not just NG.

## What changed

One magic number → two named budgets chosen by the constraint that actually applies (`src/utils/sentry.utils.ts`):

| Constant | Value | Why |
|---|---|---|
| `CLIENT_FETCH_TIMEOUT_MS` | **20,000** (was 10,000) | **The fix.** No platform ceiling exists in a browser — only real mobile networks. 20s clears NG's 11.3s p90 LCP with margin. |
| `SERVER_FETCH_TIMEOUT_MS` | **10,000** (unchanged) | Deliberately left alone — see below. |

`NEXT_PUBLIC_FETCH_TIMEOUT_MS` still overrides both, and is now **validated as a positive integer**: `parseInt` previously read `"30s"` as 30 ms and `"0"` as 0, either of which aborts every request instantly. Malformed values fall back to the defaults instead of bricking fetches *(CodeRabbit, Major)*. The per-call third-arg `timeoutMs` still wins over everything — unchanged, and covered by a test.

### Why the server budget did NOT move

The original plan was to derive the server budget from `vercel.json`'s `maxDuration: 300`. **That entry is dead config.** It declares the glob `app/api/**/*`, but this project keeps its routes under `src/app/api/`, and [Vercel requires the `src/` prefix for src-directory projects](https://vercel.com/docs/functions/configuring-functions/duration) — so the entry currently matches nothing.

The real ceiling is therefore the *project default*, which can't be read from the repo and differs by an order of magnitude depending on whether Fluid compute is enabled (300s with, far lower without). Setting a 30s server budget on top of an unverified ceiling risks sitting **above** it — reintroducing exactly the opaque 504s the original 10s existed to prevent. And it buys little: a Vercel→`api.peanut.me` call is a datacenter hop, not a mobile network.

So the server budget stays at 10s. The context split is still the real fix — when someone confirms the ceiling, they change **one named constant with a documented rationale**, not a global magic number.

## Risk: the retry multiplier

The timeout is per *attempt*, and React Query retries on top — `RETRY_STRATEGIES.FAST` (2 retries, 1s/2s backoff) is the global default in `src/config/wagmi.config.tsx`.

```
worst case = 3 attempts × timeout + 1s + 2s backoff
  before:  3×10s + 3s = 33s
  after:   3×20s + 3s = 63s     ← accepted, and pinned by a test
  (at 30s: 3×30s + 3s = 93s     ← rejected, real UX regression)
```

**Accepted trade-off.** 63s is the pathological tail where all three attempts hang to full budget — i.e. the network is genuinely dead and the user was going to fail anyway. The common case *improves*: an NG request that really takes 12s now succeeds on attempt 1 instead of erroring at 10s. That tail is exactly the 15.1% of NG users currently generating timeout events.

**Considered and rejected:** dropping `FAST` from 2 retries to 1 (total 41s). It's the global wagmi default, so it would also halve retry resilience for *fast* failures (transient 5xx, which fail in <1s) — a regression in a case unrelated to this bug.

⚠️ **The 63s bound covers the default strategy only.** `src/components/Claim/Claim.tsx:108` sets its own `retry: 4` with ~15s of backoff → **5 × 20s + 15s = 115s** worst case (up from 65s) on the send-link claim screen. Not changed here, and flagged as a follow-up below.

**Financial mutations are unaffected** — `RETRY_STRATEGIES.FINANCIAL` is `retry: false` and stays that way.

**Bonus, no action needed:** a longer budget *reduces* double-submit risk — fewer POSTs that actually succeeded get aborted client-side and look failed.

## ❗ Verify before merging

**`NEXT_PUBLIC_FETCH_TIMEOUT_MS` must not be set in Vercel.** `.env.example` shipped `=10000` before this PR. If that value was ever copied into the Vercel project env for production, **the override wins and this PR is a no-op**. I can't read Vercel's env store from here — please confirm it's unset (or delete it) in Production/Preview/Development, and in the Capgo/native build env.

## Tests

`src/utils/__tests__/sentry.utils.test.ts`:
- client vs server default resolution — **both branches explicitly**, since an inverted check is the failure mode that actually bites
- `NEXT_PUBLIC_FETCH_TIMEOUT_MS` overrides both contexts; per-call `timeoutMs` still wins over both
- malformed-override table: `undefined`, `''`, `not-a-number`, `30s`, `0.5`, `0`, `-1`, `Infinity` all fall back
- worst-case-total guard: derives `attempts × CLIENT_FETCH_TIMEOUT_MS + backoff` from `RETRY_STRATEGIES.FAST` itself and pins it at 63s / <70s, so changing either the budget or the retry policy fails the build instead of silently shipping a long spinner
- regression pin: the client budget may not return to 10s
- **hermetic** — verified green under `NEXT_PUBLIC_FETCH_TIMEOUT_MS=10000` and `=30s`

## Follow-ups (flagged, deliberately NOT in this PR)

1. **`vercel.json` glob is dead** — `app/api/**/*` should be `src/app/api/**/*`. Pre-existing; a platform-config change doesn't belong in a production hotfix.
2. **Confirm Fluid compute** is enabled, then the server budget can be raised with a provable ceiling.
3. **`src/hooks/useGeoLocation.ts:48`** fetches `ipapi.co/country` with **no timeout, no AbortController, no retry**, gating a skeleton in `CountryList.tsx:145` — on these same slow networks it hangs the country picker indefinitely. Real bug, separate PR.
4. **`src/components/Claim/Link/Onchain/Success.view.tsx:94`** — `setInterval(async …, 250)` with no in-flight guard stacks ~80 concurrent fetches under a stall at 20s (was ~40). Pre-existing, amplified 2×. The correct pattern already exists at `src/hooks/useUserAutoRefresh.ts:50`.
5. **`health/mobula` + `health/justaname`** are the only health routes calling `fetchWithSentry` without an explicit timeout; siblings self-cap at 5–8s. Worth passing `5000`.
6. **Playwright `expect.timeout`** (15s / 10s) is now below the 20s client budget, so a hung request expires the assertion instead of surfacing the app's error state.
7. **`src/services/rain.ts:326,466,560`** JSDoc says "10s default" — now stale for the client path.
8. **Mono docs** (markdown, separate from code per the repo rule): `ops/nigeria-launch-readiness.md` L89/L123-124/L182 cites the 10s value and lists raising it as an open pre-launch item this PR ships; `product/feedback/problems/payments-double-charged.md` L12 states 10s as live fact.

## Back-merge to `dev` — open and linked: #2535

peanut-ui releases flow `dev → main`, so a fix that lands only on `main` gets clobbered by the next release. **#2535** back-merges this same branch into `dev` and is already open.

**Merge order: this PR first, then #2535.** If this one is reworked or rejected, close #2535 too — they must not diverge. The three touched files are byte-identical on `main` and `dev` today, so it is conflict-free.

## Summary by CodeRabbit

* **Improvements**
  * Added separate default timeout budgets for browser and server requests.
  * Added support for a valid shared timeout override through environment configuration.
  * Invalid or empty timeout values now safely fall back to the appropriate default.
  * Timeout details are accurately captured for request error reporting.

* **Documentation**
  * Clarified the timeout configuration and fallback behavior in the environment template.

* **Tests**
  * Added coverage for timeout selection, overrides, invalid values, retry budgets, and per-request timeout handling.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -73,8 +73,8 @@ export PROMO_LIST={}
 export NEXT_PUBLIC_VAPID_PUBLIC_KEY=
 export VAPID_PRIVATE_KEY=
 export NEXT_PUBLIC_VAPID_SUBJECT="mailto:hello@peanut.me"
-# Overrides BOTH fetch budgets (client 20s / server 30s, src/utils/sentry.utils.ts).
-# Positive integer of milliseconds; empty or malformed = keep the defaults.
+# Overrides BOTH fetch budgets (client 20s / server 10s, src/utils/sentry.utils.ts).
+# Positive integer of milliseconds, max 2147483647; empty or malformed = keep the defaults.
 export NEXT_PUBLIC_FETCH_TIMEOUT_MS=
 
 # one signal 
--- a/.env.example
+++ b/.env.example
@@ -73,7 +73,9 @@ export PROMO_LIST={}
 export NEXT_PUBLIC_VAPID_PUBLIC_KEY=
 export VAPID_PRIVATE_KEY=
 export NEXT_PUBLIC_VAPID_SUBJECT="mailto:hello@peanut.me"
-export NEXT_PUBLIC_FETCH_TIMEOUT_MS=10000
+# Overrides BOTH fetch budgets (client 20s / server 30s, src/utils/sentry.utils.ts).
+# Positive integer of milliseconds; empty or malformed = keep the defaults.
+export NEXT_PUBLIC_FETCH_TIMEOUT_MS=
 
 # one signal 
 export NEXT_PUBLIC_ONESIGNAL_APP_ID=
--- a/src/utils/__tests__/sentry.utils.test.ts
+++ b/src/utils/__tests__/sentry.utils.test.ts
@@ -1,5 +1,17 @@
+import { readFileSync } from 'node:fs'
+import { join } from 'node:path'
 import * as Sentry from '@sentry/nextjs'
-import { fetchWithSentry, sanitizeRequestBody, sanitizeResponseBody, scrubObject } from '../sentry.utils'
+import { RETRY_STRATEGIES } from '../retry.utils'
+import {
+    CLIENT_FETCH_TIMEOUT_MS,
+    SERVER_FETCH_TIMEOUT_MS,
+    VERCEL_FUNCTION_MAX_DURATION_MS,
+    fetchWithSentry,
+    resolveDefaultTimeoutMs,
+    sanitizeRequestBody,
+    sanitizeResponseBody,
+    scrubObject,
+} from '../sentry.utils'
 
 jest.mock('@sentry/nextjs', () => ({
     withScope: jest.fn((cb: (scope: unknown) => void) => cb({ setFingerprint: jest.fn(), setTag: jest.fn() })),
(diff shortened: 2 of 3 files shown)


Hugo0:
**Re: CodeRabbit's Major on `NEXT_PUBLIC_FETCH_TIMEOUT_MS` validation** (raised on the predecessor PR #2532, carried over here).

**Applied.** The override is now validated as a positive integer of milliseconds — `Number(...)` + `Number.isInteger` + `> 0` — so `"30s"`, `"0.5"`, `"0"`, `"-1"`, `"Infinity"` and empty all fall back to the context default instead of collapsing the timeout and aborting every request instantly. `.env.example` documents the contract, and there is an `it.each` table covering exactly those cases.

**Declined** the second half ("reject server values at or above the Vercel function budget"): the override is a deliberate operator escape hatch, and this PR establishes that the effective ceiling is *not* readable from the repo — `vercel.json`'s `app/api/**` glob does not match this project's `src/app/api/` routes, so it is dead config. There is no trustworthy number to validate against. Tracked as follow-ups 1 and 2 in the PR description.

Hugo0:
Both findings applied in `db89ba8af`:

1. **`.env.example` server budget** — correct catch, the comment still said 30s after the server side was deliberately left at 10s. Now reads `client 20s / server 10s`.
2. **32-bit timer cap** — accepted, and the empirical probe was the convincing part. `setTimeout` clamps anything above `2_147_483_647` to ~1ms, so an over-large override aborts every request instantly, which is the same failure the `"0"` / `"30s"` guards already cover from the other end of the range. Added `parsed <= MAX_TIMER_MS`, plus `2147483648` to the malformed table and `2147483647` as an accepted boundary case.

Suite is green (55 tests across the two touched files) and verified hermetic under `NEXT_PUBLIC_FETCH_TIMEOUT_MS=10000` and `=30s`.

Note the earlier `vercel.json` drift test is **gone** by design, not oversight: while addressing your first review I found that entry declares the glob `app/api/**` while this project keeps routes under `src/app/api/`, and Vercel requires the `src/` prefix for src-directory projects — so it matches nothing. Asserting against it would have enshrined dead config. The server budget stays at its historical 10s instead of being raised on top of an unverifiable ceiling. Details in the PR description.

PR #2538, opened 2026-07-28
ci(content): publish content to production automatically

Hugo0:
Makes content publish itself to production, safely, and say so loudly when it doesn't.

## The problem

Content merged to mono never reached peanut.me on its own — every publish needed a manual signed PR to `main`.

`update-content.yml` opened its `src/content` bump PRs against **`dev`**. But content only ever flows `main` → `dev` (via the back-merge), so `dev` was never *ahead* of `main` and a release could never carry a bump up:

- **29 `auto/update-content` PRs opened against `dev`. Zero ever merged** — each closed by hand as superseded.
- The last 5 `dev` → `main` releases (#2506, #2407, #2391, #2340, #2298) all left `src/content` untouched.
- At time of writing `main` = `peanut-content@c268e5d`, `dev` = `@1078c28` — dev's is a strict *ancestor*.

## What changes

**1. Retarget the auto-PR to `main`** and hand the gate to `content-publish-automerge.yml`.

**2. Add a `merge-on-green` fallback.** `update-content.yml` authors as `CONTENT_BOT_TOKEN`'s user — the same user the automerge workflow approves with — and GitHub forbids self-approval, so the existing approve → `--auto --merge` path can never complete. The fallback merges via the org-admin bypass instead.

That bypass skips the **required status check** as well as the review, so the fallback re-establishes green CI *itself*. It merges only when all of: ≥1 `ci-success` check run on the exact head commit and **all** of them green; diff **exactly** `src/content`; PR open, non-draft, base `main`, same-repo head, head SHA still equal to the tested commit. It keys off the `ci-success` check run rather than `workflow_run.conclusion`, because `ci-success` is what "Protect Prod" actually gates on.

**3. Stop publishes racing and piling up.** `concurrency: update-content` — a burst of pushes otherwise branches several PRs off the same `main`, and once the first merges the rest conflict on the gitlink and can *never* merge. Plus close superseded auto-PRs on open, automating a chore done by hand 29 times.

**4. Pin the "content can't execute" invariant.** Raw MDX compiles expressions to executable JS, which would run during `next build` in CI and on Vercel.

**Correction, and it matters:** we are **not** exposed to this today, and this guard is not what prevents it. `next-mdx-remote@6` already strips the same node set by default (`removeJavaScriptExpressions` via `blockJS`, `removeImportsExportsPlugin` via `useDynamicImport: false`). My original proof compiled through raw `@mdx-js/mdx`, which is not what `renderContent` calls. There was no live hole.

It is still worth keeping, for two smaller and truer reasons:
1. That protection is a *dependency default*, on a path that now publishes to production unreviewed. A major upgrade, or someone setting `blockJS: false` for an unrelated reason, silently re-opens it. This asserts the invariant in our repo, visibly and under test.
2. next-mdx-remote *silently strips*; we run first in `remarkPlugins`, so we **throw**. An author writing `{price}` expecting interpolation currently gets a page that quietly renders nothing and auto-publishes. They should get a failed build naming the file and line.

It checks *executable*, not *braced* — content legitimately uses two inert forms that must keep working:

| Still works | Now a build error |
|---|---|
| markdown, GFM tables | `{require('fs')...}` |
| `<ExchangeWidget currency="ARS" />` | `<Foo bar={process.env.X} />` |
| `<Step number={1} />` | `<Foo {...process.env} />` |
| `{/* editor comments */}` | `import fs from 'node:fs'` |

An expression passes only if its parsed program is empty (comment) or a single literal. Fail-closed: an expression that can't be parsed is rejected.

**Reviewer: this is the one item I'd understand you dropping.** It is defence-in-depth, not a fix. Items 1-3 and 5 are the PR.

**5. Fail loud.** `content-pipeline-watchdog.yml`, every 15 min, asserts one end-to-end invariant — *is production serving what's on mono `main`?* One question covering every failure mode: dead mirror, missed dispatch, failed update-content, red CI, conflicted PR, and whatever we haven't thought of. It checks hop 1 too (parsing the `mirror: sync from mono@<sha>` stamp), because comparing only peanut-content against peanut-ui looks perfectly consistent while the mirror is dead and mono races ahead. It lives here rather than in mono because peanut-ui already holds tokens for all three repos; mono has no secrets.

A stall **always** fails the run (red in Actions, re-failing every 15 min so it can't scroll away) **and** posts to Discord. A missing webhook degrades the alert, never silences it. Red `ci-success` skips the 45-min grace period since it will never self-resolve.

## Risks

- **Content now reaches production with no human in the loop.** Approved on the condition that it is *content-bump only* and *tests pass*; both are enforced, fail-closed (any extra file, any `gh` error, any missing/red check → no merge).
- Code cannot ride this path — verified live on this very PR: `approve-and-merge` ran, saw a non-content diff, and logged *"Not content-only — normal review gate stays in place."*
- Worst case is wrong copy on peanut.me, revertible in one pointer bump. Content cannot execute (next-mdx-remote strips JS by default; item 4 pins that).

- **Depends on `CONTENT_BOT_TOKEN` belonging to an OrganizationAdmin.** API-made commits are unsigned and `required_signatures` is on both rulesets; the admin `bypass_mode: always` is the only reason this works. Documented inline — move that token to a non-admin account and the bump must become a locally signed commit. (Considered granting a bot App a ruleset bypass and rejected it: bypass is per-*ruleset*, not per-rule, so it would also hand out branch-deletion and force-push.)
- Both merge jobs run from the base branch (`pull_request_target` / `workflow_run`) — a PR cannot tamper with the guard judging it. Neither checks out or executes PR code.
- Corrected two materially wrong comments: the Git Data API commit was labelled "signed ... (verified signature)" (it is unsigned), and the Discord alert needs an explicit `User-Agent` (Cloudflare 403s python-urllib's default — caught by firing the webhook rather than trusting it, which would otherwise have been a silent failure in the thing built to prevent silent failures).

## QA

- Both workflows: `yaml.safe_load` parses, every `run:` block passes `bash -n`.
- `merge-on-green` jq filters dry-run against the real merged publish #2537 (`029e5566`): `TOTAL ci-success: 2, GREEN: 2` — the two-run (`push` + `pull_request`) case the all-green rule exists for. PR-selection correctly returns empty for an already-merged PR.
- Watchdog logic dry-run against live data: `mono content@0b0740a | mirrored from @0b0740a | peanut-content c268e5d | live c268e5d` → CURRENT, stays quiet.
- Discord webhook fired for real: HTTP 204 delivered.
- MDX guard against the real compiler: 7 attack payloads rejected (fs read, env exfil, spread, import, IIFE, template literal, bare identifier); **all 758 files under `src/content` compile unchanged**. Full inventory of every `{...}` in content: 15 literals, 2 comments, **0 executable**.
- 13 unit tests for the guard. Full suite: 4 pre-existing worktree failures (`@capacitor/clipboard` missing, `public/flags` ungenerated) — environment, not this diff.
- Prettier clean; typecheck clean for these files.

End-to-end can only be verified after merge: push content to mono → expect an auto-PR to `main` that self-merges on green, live in ~15–20 min.

## Docs

Updated on mono `main` (539b81f4, 15457469): `content/_system/ARCHITECTURE.md`, `skills/publish-content/SKILL.md` (+ frontmatter, now the *fallback* skill), `skills/publish-content/references/slack-draft.md`, `skills/update-content/SKILL.md`, `skills/README.md`. Each carries a note that the new behaviour activates when this PR merges — **delete those notes on merge**.

## Follow-up (not blocking)

`skills/publish-content/scripts/pipeline-status.sh` still prints `dev` for hop ③; flagged in the skill, worth a tidy-up pass later.

## Summary by CodeRabbit

- **New Features**
  - Added automated monitoring for content publishing, including alerts when production content becomes outdated or publishing is blocked.
  - Improved automatic publishing so validated content updates can merge reliably once checks pass.

- **Bug Fixes**
  - Prevented executable or unsafe MDX content from being published, including scripts, imports, exports, and unsafe JSX expressions.
  - Added validation to protect content rendered in production while allowing supported static content formats.

- **Tests**
  - Added coverage for safe and unsafe MDX content scenarios.

### DIFF
--- a/.github/workflows/content-pipeline-watchdog.yml
+++ b/.github/workflows/content-pipeline-watchdog.yml
@@ -1,0 +1,121 @@
+name: Content pipeline watchdog
+
+# Content publishes to production with no human in the loop. The failure mode
+# that creates is silence: a broken internal link fails `validate-links`, the
+# publish PR never merges, and nothing anywhere goes red — the content simply
+# never appears and nobody finds out until someone notices the site is stale.
+#
+# This asserts the end-to-end invariant instead of instrumenting each failure:
+#
+#     is production's content the same as the content on mono `main`?
+#
+# One question, every failure mode — mirror died, dispatch never fired,
+# update-content errored, CI red, PR conflicted, and whatever we haven't
+# thought of yet.
+#
+# It lives here rather than in mono because peanut-ui already holds tokens to
+# read all three repos (MONO_READ_TOKEN, SUBMODULE_TOKEN); mono has no secrets.
+#
+# Loud by construction: a stuck pipeline ALWAYS fails the run (red in Actions),
+# and additionally posts to Discord when the webhook is configured. A missing
+# webhook degrades the alert; it never silences it.
+
+on:
+    schedule:
+        - cron: '*/15 * * * *'
+    workflow_dispatch:
+
+permissions:
+    contents: read
+    pull-requests: read
+
+# How long the pipeline may legitimately be mid-flight. Mirror ~1min + CI ~5min
+# + Vercel ~8min lands around 15-20min in practice; 45 leaves room for a queue
+# without letting a real stall hide for long.
+env:
+    GRACE_MINUTES: 45
+
+jobs:
+    check:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Is production's content current?
+              env:
+                  UI_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                  CONTENT_TOKEN: ${{ secrets.SUBMODULE_TOKEN }}
+                  MONO_TOKEN: ${{ secrets.MONO_READ_TOKEN }}
+                  DISCORD_WEBHOOK: ${{ secrets.CONTENT_PIPELINE_DISCORD_WEBHOOK }}
+                  REPO: ${{ github.repository }}
+                  RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+              run: |
+                  set -uo pipefail
+
+                  # ---- hop 3: what production is serving -----------------------
+                  LIVE=$(GH_TOKEN=$UI_TOKEN gh api "repos/$REPO/contents/src/content?ref=main" --jq '.sha')
+
+                  # ---- hop 2: what the mirror has published --------------------
+                  TIP=$(GH_TOKEN=$CONTENT_TOKEN gh api repos/peanutprotocol/peanut-content/commits/main --jq '.sha')
+                  TIP_MSG=$(GH_TOKEN=$CONTENT_TOKEN gh api repos/peanutprotocol/peanut-content/commits/main --jq '.commit.message' | head -1)
+                  TIP_AT=$(GH_TOKEN=$CONTENT_TOKEN gh api repos/peanutprotocol/peanut-content/commits/main --jq '.commit.committer.date')
+
+                  # ---- hop 1: what mono says the content should be -------------
+                  # The mirror stamps "mirror: sync from mono@<sha>", so a mismatch
+                  # here means the mirror itself is stuck — invisible from hops 2/3
+                  # alone, which would look perfectly consistent with each other.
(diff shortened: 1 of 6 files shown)


Hugo0:
@coderabbitai review

PR #2548, opened 2026-07-28
fix(add-money): PIX deposit Done exits to home instead of starting a new deposit

abalinda:
## Summary

After a completed BRL deposit via PIX, tapping **Done** started a *new* deposit instead of returning home — the successful deposit read as unfinished and invited an accidental repeat payment. ([Discord report](https://discord.com/channels/972435984954302464/1484181812715585576/1531632169594650664), TASK-20961)

**Root cause:** the PIX screen wired its *terminal* exits to `onBack`, which the parent defines as `setUrlState({ step: 'inputAmount' })`. Once the fiat has been sent there is nothing to go "back" to, so "back" meant "restart the flow".

**Fix:** `MantecaPixQrDeposit` gets a distinct `onDone` exit; `MantecaAddMoney` wires it to `router.replace('/home')`.

- `replace`, not `push` — device/browser back must not pop into the finished flow, whose `step=showQR` with `depositDetails` gone redirects to `inputAmount`, still carrying `?amount=…&currency=BRL` in the URL, i.e. the same trap one back-press later. Matches the `replaceOnDone` convention in `PaymentSuccessView` (used by the crypto-deposit success path). `useSafeBack` would be wrong here: its in-app-history branch calls `router.back()`, which lands on the country list, not home.
- **Also fixed: the `processing` state**, found by the adversarial review pass. `processing` means the poller saw `stage >= 2` — [documented](../blob/main/src/components/AddMoney/hooks/useMantecaDepositPolling.ts) as "the fiat actually arrived" — and the spinner's only affordance was a back arrow to a **pre-filled** amount input, one tap from paying twice while the first deposit settles. It now exits like the settled state. Leaving is safe: the credit is posted server-side by the webhook/poller, not by this screen's polling.
- The expired-QR **Go back** keeps `onBack` — that one is a genuine go-back (re-quote the deposit).

## Risk

Low, narrowly scoped: two components in the BRL/PIX add-money flow, no API or state-shape change. Blast radius is `step === 'showQR'` in `MantecaAddMoney` only. The ARS/other-currency path (`MantecaDepositShareDetails`) has no completion screen and is untouched; the crypto-deposit path already routes home via `PaymentSuccessView`. No cross-repo impact.

## QA

1. Brazil → Add money → enter a BRL amount → PIX QR screen.
2. Pay the QR (Manteca sandbox) and wait for polling to flip to **Deposit received!**
3. Tap **Done** → lands on `/home`, **no** new deposit flow, balance/history refreshed.
4. Device back from `/home` → does **not** return to the completed deposit.
5. While the spinner is up (payment detected, settling), tap the header back arrow → also lands on `/home`, not a pre-filled amount input.
6. Regression: let a QR expire → **Go back** still returns to the amount input.

Automated — `MantecaPixQrDeposit.test.tsx` + new `MantecaAddMoney.test.tsx`: the settled and processing screens exit via `onDone` (never `onBack`), the expired screen still uses `onBack`, and the parent sends the user to `/home` rather than back into the flow. Each assertion was mutation-checked: reverting the corresponding fix line makes exactly that test fail.

- **Not reused: `PaymentSuccessView`.** It already implements terminal-exit semantics and would delete this screen's bespoke success block. Deliberately deferred — it pulls in Redux, the receipt drawer, confetti and `useTokenChainIcons`, changes the layout (`justify-between` vs this screen's `my-auto`) and the header icon, and duplicates the parent's `TRANSACTIONS` invalidation. That's a UI change, not a bug fix. Worth a follow-up.
- **Not fixed here: `['balance']` is never invalidated** after a deposit — only `[TRANSACTIONS]` is. It self-heals (no observer during the flow, `staleTime` 30s, `refetchOnMount`), and the omission is codebase-wide (`PaymentSuccessView`, `withdraw/manteca`), so fixing it only here would be inconsistent. Follow-up, not a regression from this PR.
- **`onDone={() => router.replace('/home')}` is not covered by the `no-restricted-syntax` back-button guard**, which only matches `onPrev`/`onBack` attributes. The guard's hazard (a `push` growing in-app history so `useSafeBack` cycles) doesn't apply to a `replace` on a terminal screen — but a future `replace` → `push` edit on that line would be lint-silent. The inline comment holds the invariant.

## Screenshots

⚠️ **NONE** — no pixel change. Both screens render identically; only the exit *destination* changes. Reaching them for a capture needs a settled Manteca-sandbox PIX payment, which the local harness can't drive. Behavior is covered by the tests above.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved navigation after PIX deposits complete or remain in processing.
  - The deposit flow now correctly returns to the home screen when users finish or exit terminal deposit states.
  - Preserved the option to return to the amount-entry step when backing out before completion.
  - Retained the existing behavior for expired deposits.

- **Tests**
  - Added coverage for deposit completion, processing, back navigation, and expiration scenarios.

### DIFF
--- a/src/components/AddMoney/components/MantecaAddMoney.tsx
+++ b/src/components/AddMoney/components/MantecaAddMoney.tsx
@@ -4,7 +4,7 @@ import MantecaDepositShareDetails from '@/components/AddMoney/components/Manteca
 import MantecaPixQrDeposit from '@/components/AddMoney/components/MantecaPixQrDeposit'
 import CyclingLoading from '@/components/Global/PeanutLoading/CyclingLoading'
 import InputAmountStep from '@/components/AddMoney/components/InputAmountStep'
-import { useParams, useSearchParams } from 'next/navigation'
+import { useParams, useRouter, useSearchParams } from 'next/navigation'
 import { addMoneyCountryUrl } from '@/utils/native-routes'
 import { useSafeBack } from '@/hooks/useSafeBack'
 import { type CountryData, countryData } from '@/components/AddMoney/consts'
@@ -38,6 +38,7 @@ const MantecaAddMoney: FC = () => {
     const params = useParams()
     const searchParams = useSearchParams()
     const queryClient = useQueryClient()
+    const router = useRouter()
 
     // URL state - persisted in query params
     // Example: /add-money/argentina/manteca?step=inputAmount&amount=100&currency=ARS
@@ -334,6 +335,9 @@ const MantecaAddMoney: FC = () => {
                 depositDetails={depositDetails}
                 currencyAmount={localCurrencyAmount}
                 onBack={() => setUrlState({ step: 'inputAmount' })}
+                // Terminal exit — `replace` so device/browser back can't pop into the
+                // finished deposit (whose step=showQR would redirect to a new one).
+                onDone={() => router.replace('/home')}
                 onComplete={() => queryClient.invalidateQueries({ queryKey: [TRANSACTIONS] })}
             />
         )
--- a/src/components/AddMoney/components/MantecaPixQrDeposit.tsx
+++ b/src/components/AddMoney/components/MantecaPixQrDeposit.tsx
@@ -68,10 +68,14 @@ const MantecaPixQrDeposit: FC<{
     }
 
     // Payment detected, settling — show the branded processing screen (same as PIX payments).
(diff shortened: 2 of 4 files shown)


abalinda:
@coderabbitai review

jjramirezn (approved):

PR #2550, opened 2026-07-28
content: publish latest to production (src/content → peanut-content@847ecbe)

Hugo0:
Auto-generated: bumps `src/content` to latest peanut-content `main`.

Based on `main`, so the diff is exactly the submodule pointer.
`content-publish-automerge.yml` approves and merges this once
`ci-success` is green. Anything other than a lone `src/content`
change falls back to the normal 1-review gate.

## Summary by CodeRabbit

* **Chores**
  * Updated the bundled content to a newer revision.

PR #2546, opened 2026-07-28
content: publish latest to production (src/content → peanut-content@63abb43)

Hugo0:
Auto-generated: bumps `src/content` to latest peanut-content `main`.

Based on `main`, so the diff is exactly the submodule pointer.
`content-publish-automerge.yml` approves and merges this once
`ci-success` is green. Anything other than a lone `src/content`
change falls back to the normal 1-review gate.

## Summary by CodeRabbit

* **Chores**
  * Updated the bundled content package to its latest version.
  * No user-facing feature or behavior changes.

PR #2551, opened 2026-07-28
fix(automerge): pass --admin, and stop swallowing a failed merge

Hugo0:
**The content pipeline is currently broken — nothing is reaching production.** The end-to-end test found the real defect, on the last hop.

## What the test showed

Everything worked until the merge itself:

| Hop | Result | Elapsed |
|---|---|---|
| push → mono | `7d255265` | — |
| mirror → peanut-content | ✅ `63abb439` | +20s |
| auto-PR → **`main`** | ✅ #2546 | +48s |
| guard + handoff | ✅ content-only → merge-on-green | +67s |
| supersede on collision | ✅ #2546 auto-closed when #2547 opened | +3.5m |
| **merge** | ❌ | — |

```
Merging #2547 — diff is exactly src/content, ci-success green (2/2)
merge failed for #2547 (already merged, or bypass unavailable)
```

## Cause

**Being an OrganizationAdmin does not make `gh pr merge` bypass "Protect Prod".** gh enforces the branch policy and requires an explicit `--admin`:

```
X Pull request #2544 is not mergeable: the base branch policy prohibits the merge.
  To use administrator privileges to immediately merge the pull request, add the `--admin` flag.
```

I hit that exact error merging #2544 by hand an hour earlier, added the flag, and never carried it back into the workflow.

`--admin` also waives the required status check — note what it is **not** waiving. This job has already established, itself, that the diff is exactly `src/content` and that every `ci-success` run on the commit is green. The gate is enforced in code above; `--admin` only drops the human review, which is the entire purpose of this workflow.

## And the reason nobody noticed for 40 minutes

`|| echo "::warning::merge failed"` — the run went **green** while content sat unpublished. A merge we decided to make and then failed to make now **fails the run loudly**. That silent-warning pattern is precisely what this project exists to eliminate, and it was sitting in the middle of it.

## Two more, found by running the step under `bash -e`

- **`set +e`** with explicit checks. GitHub runs steps as `bash -e`; an unguarded non-zero kills the loop mid-flight with no output. (Same bug class as #2544.)
- **Numeric validation** of `TOTAL`/`GREEN`/`PRS`. `gh` prints its JSON error body to **stdout**, so on an API failure the loop was iterating over the words of an error message as if they were PR numbers.

## QA

Executed the step under `bash -e` against real APIs:
```
bogus SHA (422):        ci-success not green (-1/0) — nothing to merge.   EXIT=0
real commit, no PR:     No open content-publish candidates.               EXIT=0
```

## After merge

#2550 is open with Konrad's real content waiting. Once this lands, the next `Tests` completion on it should merge it automatically — which is the end-to-end proof finishing itself.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of automatic content updates after successful tests.
  * Prevented transient service or data errors from being overlooked during processing.
  * Added clearer failure reporting when an automatic merge cannot be completed.
  * Blocked merges when validation results are incomplete or invalid.

### DIFF
--- a/.github/workflows/content-publish-automerge.yml
+++ b/.github/workflows/content-publish-automerge.yml
@@ -102,7 +102,13 @@ jobs:
                   SHA: ${{ github.event.workflow_run.head_sha }}
                   REPO: ${{ github.repository }}
               run: |
-                  set -euo pipefail
+                  set -uo pipefail
+                  # `set +e` deliberately. GitHub runs this as `bash -e {0}`, so any
+                  # unguarded non-zero — a transient `gh` 5xx, a PR that closed between
+                  # the list and the diff — kills the step mid-loop with no output and
+                  # a bare red run. Every branch below checks its own inputs and says
+                  # what it decided; dying silently is strictly worse than continuing.
+                  set +e
 
                   # Key off the `ci-success` check run, NOT `workflow_run.conclusion`.
                   # `ci-success` is the single check the "Protect Prod" ruleset gates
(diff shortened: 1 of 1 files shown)


0xkkonrad (approved):
Approving per Konrad to unblock the content publish pipeline. The gate that matters is still enforced in-code above the merge — diff must be exactly src/content and every ci-success run on the commit must be green — and --admin only waives the human review, which is this workflow's whole purpose. Failure handling is fail-closed throughout: GREEN=-1 on an API error makes the equality check impossible to pass, non-numeric PR ids are skipped, and an unreadable diff becomes UNKNOWN which never equals src/content.

PR #2447, opened 2026-07-17
Localization (en / es-419 / pt-BR) + zero ESLint errors

innolope-dev:
## What this is

Two tightly-coupled efforts on one branch (the lint guard is what enforces the localization, so they ship together):

1. **App localization** — all product-UI copy extracted to `next-intl`: **1,982 message keys** per locale in `src/i18n/app/messages/` for **English, Latin-American Spanish (es-419), and Brazilian Portuguese (pt-BR)**, an `AppIntlProvider` mounted above the context tree, and an ESLint guard (`react/jsx-no-literals`, scoped to translated surfaces) that fails CI when someone hard-codes new UI copy.
2. **ESLint cleanup** — the `eslint` CI job goes from **832 errors on `main` to 0** on this branch.

Synced with `main` as of 2026-07-17 (all 55 commits merged in, 9 conflicts resolved in favor of main's newer copy/logic with the i18n calls preserved).

## Lint stats: 832 → 0 errors

Baseline on `main` (232 files) vs this branch:

| Rule | main | branch | How it was resolved |
|---|---:|---:|---|
| `@typescript-eslint/no-explicit-any` | 307 | 0 | 112 production sites properly typed (see below); tests/mocks and dev-only tooling exempted via commented config overrides |
| `@next/next/no-html-link-for-pages` | 198 | 0 | `<a>` → `next/link` across marketing/app pages |
| `@typescript-eslint/no-unused-vars` | 120 | 0 | dead bindings removed; intentional ones `_`-prefixed |
| `no-restricted-imports` (barrel ban) | 119 | 0 | deep imports of `@/interfaces`, `@/assets`, `@/context`, `@/components`, `@/config`; dead barrels deleted |
| `@typescript-eslint/no-require-imports` | 72 | 0 | converted to ES imports; the Jest `require`-after-`jest.mock()` idiom exempted for test files |
| `react/no-unknown-property` | 6 | 0 | fixed; styled-jsx's `jsx`/`global` attrs whitelisted |
| `import/first` | 5 | 0 | imports hoisted |
| misc (`display-name`, `alt-text`, `no-empty-object-type`, stale disable directives) | 5 | 0 | fixed individually |
| **Total errors** | **832** | **0** | |

Warnings: 183 → 61, almost all `react-hooks/exhaustive-deps`. The provably-safe deps were fixed on this branch; the remainder genuinely change behavior when "fixed" and are left for a dedicated pass rather than blind suppression.

### The `no-explicit-any` sweep (307 → 0)

[...]

PR #2554, opened 2026-07-28
fix(i18n): localize backend errors, drop locale-unsafe comparisons

innolope-dev:
## Why

Follow-up to #2447 (next-intl, en / es-419 / pt-BR). Two gaps remained on the error path:

1. Backend errors reached non-English users **in English** — `friendlyError` passed them through untranslated because it had no stable discriminant to key off.
2. One branch compared a **translated** string against an English literal, so it was dead in every locale.

Backend companion: peanutprotocol/peanut-api-ts#1249. **This PR does not depend on it** — see below.

## Wire codes over prose

`friendlyError` now checks a backend `code` **before** any message matcher and maps it through an allow-list.

**The allow-list is load-bearing.** ethers sets `.code` to `NETWORK_ERROR` / `CALL_EXCEPTION`, and EIP-1193 wallets use numeric codes like `4001`. Mapping an arbitrary `.code` straight to a translation key would turn every one of those into a missing-message error. Anything unlisted falls through to today's matchers — which is also what makes this safe to merge **before** the API deploys the contract, since pre-contract errors simply carry no code. Both cases have regression tests.

`rain.ts` was already reading `code: 'STALE_CARD_APPROVAL'` off the 409 and then throwing it away in favour of the English string. The three typed Rain errors now carry their discriminant, so **the Rain paths localize without waiting on the backend at all** — the FE constructs those errors itself.

## The cooldown needed a type change

`errors.rainCooldownRetry` takes an ICU plural on `retryAfterSec` (rounded up, floored at 1 — a 20s wait must not read "0 minutes"), plus a countdown-less variant for when the backend sends no number, mirroring a distinction `rain.ts` already made.

This required a **third `FriendlyError` variant** rather than another member of `FriendlyErrorCode`: next-intl types `t()`'s `values` as the *intersection* of every union member's ICU args, so a parameterized key inside the param-less union makes `{minutes}` required at every call site and breaks `t(result.code)`. There's a comment on the type explaining it for whoever adds the next one.

## Also fixed

- **`ServiceUnavailableError`** — `fetchWithSentry` rethrows it with the real timeout on `.cause`, which the classifier doesn't walk, so *every* server fetch timeout was collapsing to the generic support fallback instead of the retryable code. Matched on the name we set ourselves rather than walking `.cause` generically, which would have silently reclassified every wrapped error in the app.
- **`withdraw/[country]/bank`** compared the mapper's *output* to `'Something failed. Please try again.'` — a literal in no catalog, so dead in every locale including English. **Deleting the branch would have regressed**: two pre-`throw` `setError` calls set already-correct copy (backend-authored, and the confirm-pending notice) that the `else` would then clobber. Replaced with a local `errorAlreadyDisplayed` flag.
- The Rain **cooldown pill** rendered a hardcoded `Card cool-down`; `src/context/**` is outside the `jsx-no-literals` globs, which is how it shipped.
- The SDK's `transactionHash` fetch failure is static English, so it gets a key instead of passing through.
- `regions.utils` compares an internal region name — locale-safe today, hoisted to a named constant with a note.

## Second commit is revertible on its own

`qr-pay` and `withdraw/manteca` re-implemented the classifier's precedence by hand, so every code above would have silently failed to reach them. Routing them through `friendlyError` is **not** a straight swap — three behaviours are deliberately preserved: the per-flow `SessionKeyGrantRequiredError` copy (the two screens intentionally differ), the looser `.includes('not allowed')` match, and the flow-specific `signFailed` fallback with its `captureException` (falling through to `genericSupport` would have been a copy regression on what is always a signing failure). Split out so it can be reverted if QA on the Rain paths disagrees.

## What stays a passthrough

The liquidity text — it does not originate in peanut-api-ts at all (it's SDK/routing-layer), so **no backend code can address it**. Verified by grep across the API repo.

## Verification

`pnpm typecheck` clean · 174 suites / 2311 tests pass · `eslint` 0 errors · `prettier --check .` clean

7 new keys × 3 locales; catalog parity + ICU gates pass. es-419 / pt-BR copy is machine-assisted, same caveat #2447 flagged.

**Not covered:** no manual multi-locale walkthrough — the "try again in N minutes" rendering is unit-tested but hasn't been seen on a device.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -141,6 +141,21 @@ module.exports = [
                     message:
                         "Don't pass { history: 'push' } to nuqs useQueryState(s) — a history entry per URL write breaks the back button (useSafeBack steps through same-screen states instead of leaving). Use the default 'replace'; the URL stays shareable. If a flow genuinely needs push-per-step, add a scoped file exemption with a comment (see useNativePlugins).",
                 },
+                {
+                    // Toast copy must come from next-intl. `react/jsx-no-literals` below
+                    // only inspects JSX children, so toasts fired from hooks and contexts
+                    // (authContext, useLogin, useSendMoney, QRScanner) shipped English to
+                    // every locale unnoticed.
+                    //
+                    // Deliberately NOT extended to `throw new Error('…')`: those messages
+                    // are developer/Sentry breadcrumbs that the friendly-error mapper
+                    // collapses to `errors.genericSupport` before any user sees them, so
+                    // translating them would only fragment Sentry issue grouping.
+                    selector:
+                        "CallExpression[callee.object.name='toast'][callee.property.name=/^(error|success|info|warning|loading)$/] > :matches(Literal, TemplateLiteral):first-child",
+                    message:
+                        "Don't pass a string literal to toast.* — copy must come from next-intl. Import the right namespace with useTranslations and pass t('…'). If the value genuinely isn't copy (an id, a URL), assign it to a named const first.",
+                },
             ],
         },
     },
--- a/src/app/(mobile-ui)/add-money/__tests__/add-money-states.test.tsx
+++ b/src/app/(mobile-ui)/add-money/__tests__/add-money-states.test.tsx
@@ -16,8 +16,8 @@
 /* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unused-vars, react/display-name */
 import React from 'react'
 import { render, screen, fireEvent, waitFor, act } from '@testing-library/react'
+import { IntlWrapper } from '@/test-utils/intl'
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
-import { NextIntlClientProvider } from 'next-intl'
 import en from '@/i18n/app/messages/en.json'
 
 // ---------- module-level mocks (must be before imports that depend on them) ----------
@@ -897,9 +897,9 @@ function createQueryClient() {
 function renderWithProviders(component: React.ReactElement) {
     const queryClient = createQueryClient()
     return render(
-        <NextIntlClientProvider locale="en" messages={en} timeZone="UTC">
+        <IntlWrapper>
             <QueryClientProvider client={queryClient}>{component}</QueryClientProvider>
-        </NextIntlClientProvider>
+        </IntlWrapper>
     )
 }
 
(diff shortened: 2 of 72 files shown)


kushagrasarathe:

Read the full diff. Verified the load-bearing claims: allow-list gating, no circular imports (`api-error` imports nothing back), and pre-contract safety.

**Verified:**
- `WIRE_CODE_MAP` / `COOLDOWN_WIRE_CODES` are a true allow-list — `wireErrorCode` reads `.code` off any error, but only listed codes map; ethers (`NETWORK_ERROR`/`CALL_EXCEPTION`) and numeric EIP-1193 codes (dropped by `typeof === 'string'`) fall through to the existing matchers. Behaviour before the contract is unchanged.
- The `kind:'params'` variant as a separate union member (not a `FriendlyErrorCode`) is a real next-intl typing constraint — `t(key, values)` types `values` as the intersection of all members' ICU args. Correctly reasoned and documented.
- Dead-branch deletions (`withdraw/[country]/bank`, `qr-pay`, `manteca`) are justified: the `errorAlreadyDisplayed` flag preserves the two pre-throw `setError` calls the old `else` would have clobbered.

```
TIER: T2 (money-movement error paths — withdraw, rain cooldown, signing)
PAIRED PR: peanut-api-ts#1249 — OPEN, not required. This PR is independent by design: pre-contract errors carry no `.code` and fall through to matchers (tested). No deploy-order constraint; the allow-list IS the permanent compat layer, not a shim.
A. BREAKAGE:  pass — behaviour-preserving; Rain paths localize FE-side without the backend (FE constructs those errors).
B. PERF:      pass — pure classification, no new I/O.
C. QUALITY:   pass — strong reuse and reasoning; the ServiceUnavailableError name-match (not a .cause walk) correctly avoids reclassifying every wrapped error.
D. SECURITY:  pass — wire codes are display-only, allow-list prevents arbitrary translation-key injection.
```

**Non-blocking:** es-419/pt-BR copy is machine-assisted (flagged, same caveat as #2447); no manual multi-locale device walkthrough of the "try again in N minutes" render.

kushagrasarathe (approved):
Approving — no blocking findings. Allow-list gating verified (ethers/EIP-1193 codes fall through), no circular imports, safe to ship independently of peanut-api-ts#1249. Non-blocking: machine-assisted es-419/pt-BR copy, no device multi-locale walkthrough.

PR #2555, opened 2026-07-28
chore(i18n): string sweep, shared intl test helper, toast lint guard

innolope-dev:
> **Stacked on #2554** — review that first; this PR's diff is against it, and the base retargets to `dev` once #2554 merges.

## Why

Follow-up to #2447. `react/jsx-no-literals` only inspects JSX **children**, so hardcoded copy kept reaching users through toasts, placeholders, aria-labels and alt text.

## The ticket was right about QRScanner

It named QRScanner as a known offender with "3–4 toasts". It had **four**, on `dev`, and three of them had keys that already existed and simply weren't being used. (An earlier pass of mine looked at a divergent branch where they'd been extracted and wrongly concluded the ticket was stale.)

## The lint rule

A `no-restricted-syntax` selector banning string literals as the first argument to `toast.error|success|info|warning|loading`.

**Appended to the existing array, not a new config block** — ESLint flat config *replaces* rule options when a later block names the same rule, so a second `no-restricted-syntax` entry would have silently disabled the four `router.back` selectors. That array already sits under `files: ['src/**/*.{ts,tsx}']`, which is exactly the coverage needed, since every offender lives in `src/hooks` or `src/context` — outside the `jsx-no-literals` globs.

Verified it fires by planting a synthetic violation, not just by observing a green run.

**Scoped to `toast.*`, not `throw new Error('…')`.** Every throw site flagged in the ticket was traced to its catch handler: they all funnel through `toFriendlyError` to `errors.genericSupport`, or into an already-localized toast. Those strings are Sentry breadcrumbs and issue-grouping keys — translating them would fragment triage for zero user-facing gain.

Note the rule can't actually block a merge: the `eslint` job is still `continue-on-error` and absent from `ci-success.needs`. Left as-is deliberately; promoting it is a separate team decision.

## Shared test helper

`src/test-utils/intl.tsx` replaces **39** hand-rolled `NextIntlClientProvider` wrappers (the ticket estimated ~15). Exports the three shapes call sites actually need: `IntlWrapper` (composable, for suites nesting `QueryClientProvider`), `renderWithIntl`, and `renderHookWithIntl`.

Two suites had omitted `timeZone`, so next-intl fell back to the runner's system zone and their date formatting was machine-dependent — the shared wrapper pins UTC. `renderHookWithIntl` exists so `useSumsubKycFlow.test.ts` (a `.ts` file, no JSX) can drop its `React.createElement` workaround and the `eslint-disable react/no-children-prop` that came with it.

`ClientProviders.test.tsx` deliberately keeps its own reference — it asserts on provider *ordering*, so the provider is the subject under test.

## Extractions

Four QRScanner toasts, both logout/login toasts, two card-approval toasts, the waitlist email placeholder, two `"Peanut Logo"` aria-labels (reusing the existing `navigation.peanutLogoAlt`), two descriptive alt strings. The QRScanner payment-method row keeps its brand names verbatim — proper nouns — and only the EVM row, whose label is prose, moves to a key.

`NavHeader` gains an optional `titleKey`: `/receipt/[entryId]` is an async **server** component and this app has no server-side next-intl setup (locale is resolved entirely client-side by `AppIntlProvider`), so `useTranslations` is unavailable there.

## Two structural fixes

**Clipboard.** The match on `"There is no data on the clipboard"` was **not** a locale bug — that string is a hardcoded English literal inside `@capacitor/clipboard`'s Java (`ClipboardPlugin.java:53`), not an OS message. It *was* a dependency coupling: a plugin patch bump silently flips the branch and an empty clipboard starts reporting as a failure. `readClipboard()` now classifies the rejection and returns a reason.

**`'Enjoy Peanut!'`** is consolidated behind one `isTestTransaction()` predicate. It compares backend data, so it's locale-safe, but it was repeated in four places and one of them rendered the raw literal instead of the `transaction.enjoyPeanut` key that already existed *and was already translated*. The backend should own this flag — worth a ticket.

## Verification

`pnpm typecheck` clean · 174 suites / 2311 tests pass · `eslint` 0 errors · `prettier --check .` clean

Note: peanut-ui's `pull_request` CI trigger only fires for `main`/`dev`/`develop` bases, so this PR's checks come from the `push` trigger.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -141,6 +141,21 @@ module.exports = [
                     message:
                         "Don't pass { history: 'push' } to nuqs useQueryState(s) — a history entry per URL write breaks the back button (useSafeBack steps through same-screen states instead of leaving). Use the default 'replace'; the URL stays shareable. If a flow genuinely needs push-per-step, add a scoped file exemption with a comment (see useNativePlugins).",
                 },
+                {
+                    // Toast copy must come from next-intl. `react/jsx-no-literals` below
+                    // only inspects JSX children, so toasts fired from hooks and contexts
+                    // (authContext, useLogin, useSendMoney, QRScanner) shipped English to
+                    // every locale unnoticed.
+                    //
+                    // Deliberately NOT extended to `throw new Error('…')`: those messages
+                    // are developer/Sentry breadcrumbs that the friendly-error mapper
+                    // collapses to `errors.genericSupport` before any user sees them, so
+                    // translating them would only fragment Sentry issue grouping.
+                    selector:
+                        "CallExpression[callee.object.name='toast'][callee.property.name=/^(error|success|info|warning|loading)$/] > :matches(Literal, TemplateLiteral):first-child",
+                    message:
+                        "Don't pass a string literal to toast.* — copy must come from next-intl. Import the right namespace with useTranslations and pass t('…'). If the value genuinely isn't copy (an id, a URL), assign it to a named const first.",
+                },
             ],
         },
     },
--- a/src/app/(mobile-ui)/add-money/__tests__/add-money-states.test.tsx
+++ b/src/app/(mobile-ui)/add-money/__tests__/add-money-states.test.tsx
@@ -16,8 +16,8 @@
 /* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unused-vars, react/display-name */
 import React from 'react'
 import { render, screen, fireEvent, waitFor, act } from '@testing-library/react'
+import { IntlWrapper } from '@/test-utils/intl'
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
-import { NextIntlClientProvider } from 'next-intl'
 import en from '@/i18n/app/messages/en.json'
 
 // ---------- module-level mocks (must be before imports that depend on them) ----------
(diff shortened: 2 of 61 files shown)


kushagrasarathe:

Read the full diff (61 files, mostly mechanical test-wrapper dedup). Verified the structural changes.

**Verified:**
- ESLint rule appended to the existing `no-restricted-syntax` array — correct: flat config *replaces* rule options when a later block re-declares the same rule, so a second block would have silently killed the four `router.back` selectors. Real footgun avoided.
- `readClipboard()` refactor is behaviour-preserving: empty→`{reason:'empty'}`, `NotAllowedError`/`SecurityError`→`unavailable`, else→`empty`. Callers' `text===''` branch goes dead but harmless (no double-toast).
- `useSendMoney` `tErrors` is declared (line 54) — compiles.
- `NavHeader.titleKey` uses a type-only `en.json` import (erased at build, no catalog bundled).

```
TIER: T2/T3 (i18n sweep + shared test helper + lint guard; touches QRScanner paste and auth/login toasts)
STACKED: base is fix/locale-safe-errors (#2554). MERGE #2554 FIRST — base auto-retargets to dev on its merge, per the PR body.
A. BREAKAGE:  pass — clipboard/QRScanner refactors preserve UX; 39 hand-rolled intl wrappers collapse to one helper.
B. PERF:      pass.
C. QUALITY:   pass — good dedup; two suites' missing timeZone now pinned to UTC (removes machine-dependent date formatting).
D. SECURITY:  pass.
```

**Non-blocking (author-flagged):** the new `toast.*` lint rule can't block merge — the `eslint` job is still `continue-on-error` and absent from `ci-success.needs`. Promoting it is a separate team call.

kushagrasarathe (approved):
Approving — no blocking findings. Clean i18n sweep; eslint-array-append reasoning is correct, clipboard refactor behaviour-preserving. Merge order: this is stacked on #2554 — merge #2554 first, then this (base auto-retargets to dev).

PR #2559, opened 2026-07-28
feat(native): deferred deep linking through the store install (TASK-20772)

kushagrasarathe:
## Summary

TASK-20772 — deferred deep linking v1 (DIY). When a mobile-web user is bounced to the Play Store / App Store, their context (locale, invite code, campaign tag, destination path) now survives the install into the native app:

- **Android** — the web side appends `&referrer=<payload>` to the Play URL; a new app-local Capacitor plugin (`InstallReferrerPlugin.java`, Play Install Referrer library) reads it back on first launch. Deterministic, first-party.
- **iOS** — clipboard hand-off: the store-bounce tap copies `https://peanut.me/?pnutdl=1&…` to the clipboard; first launch reads it via `@capacitor/clipboard`, gated by the existing prompt-free `clipboardHasStrings()` so organic installs with an empty clipboard never see the iOS paste prompt.
- **Restore** (one-shot, `deferredLinkConsumed` flag): sets the existing `inviteCode`/`campaignTag` cookies (setup flow already reads them), applies the locale live via `useAppLocale().setLocale` (normalized through `resolveLocale` — unsupported tags are dropped so a garbage payload can never override the device language), and navigates to the destination only when the launch wasn't already a real deep link.
- Web side ships **helpers + `/dev/deferred` test page only** — the user-facing download modal (TASK-20769) will consume `buildDeferredPayload` / `playStoreUrlWithReferrer` / `copyIOSHandoff` when it lands.

- Consumed flag lives in `localStorage` (survives Capgo OTA; resets on reinstall — which is exactly fresh-install semantics).
- iOS clipboard is cleared after a successful consume, after the flag is set, so an interrupted clear can't cause a re-read.
- The iOS paste prompt is double-gated by prompt-free metadata checks (`hasStrings` + new `detectPatterns` probable-web-URL in `ClipboardDetectPlugin.swift`): organic installs with unrelated clipboard text (a password, a draft) are never prompted — only a clipboard that plausibly holds the hand-off URL is worth the alert.
- Restore is deliberately not awaited in native init: a pending paste prompt or a slow referrer service can't block the push-tap listener or splash hide. The Java plugin additionally resolves exactly once with a 5s timeout so its promise can never hang.
- A deep link that **actually navigated** wins the landing over the deferred dest (a rejected/off-host launch URL doesn't burn it); cookies/locale are restored regardless. The restored locale applies after the startup locale paints, so the provider's initial swap can't clobber it.
- Cookies are written with 30-day expiry and normalized (trim/lowercase/strip-@) to match the existing invite/campaign writers — a session cookie died before signup completed, and an uppercase campaign tag would break the case-sensitive badge award.
- Locale-prefixed marketing paths (`/es-419/…`) are stripped from dests on both sides — those routes don't exist in the native static export.
- Organic Play referrer (`utm_source=google-play…`) is rejected by the `pnutdl=1` marker.

## Risks / breaking changes

- **Requires a new Android binary** (new plugin + Gradle dep) — cannot ship via Capgo OTA. Older binaries no-op safely (plugin bridge throws → caught → null).
- **iOS**: adds one prompt-free method to `ClipboardDetectPlugin.swift` — rides the next IPA (iOS is mid-submission; no binaries in the field). On a binary without the method the JS skips the clipboard read entirely: no prompt, no restore, status quo.
- No backend changes. No user-facing web surface changes.

## QA

- 15 unit tests (`deferred-link.test.ts`): build→parse round-trip, Play URL encoding hop, iOS hand-off form, organic-referrer rejection, one-shot consume, off-origin dest rejection, locale normalization.
- `:app:compileDebugJavaWithJavac` green with the new plugin.
- Device plan: iOS e2e via dev build (copy hand-off in Safari → fresh install → paste prompt → `/dev/deferred` shows restored state); Android via Play internal testing track with a `&referrer=` install link; sideloaded builds verified to no-op.

Screenshots: N/A (no visible change — dev-only page + native plumbing)

## Summary by CodeRabbit

- **New Features**
  - Added deferred deep linking across Android, iOS, and web.
  - Preserves supported language, invitation, campaign, and destination details when opening the app after installation.
  - Restores deferred navigation without overriding an active deep link.
  - Added Google Play Store integration for deferred hand-off links.
  - Added safeguards to prevent repeated restoration and safely handle unavailable native capabilities.
  - Added an internal development page for inspecting and testing deferred-link hand-offs.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -92,6 +92,9 @@ dependencies {
     implementation "androidx.credentials:credentials:1.5.0"
     implementation "androidx.credentials:credentials-play-services-auth:1.5.0"
 
+    // play install referrer — deferred deep linking (InstallReferrerPlugin.java)
+    implementation "com.android.installreferrer:installreferrer:2.2"
+
     testImplementation "junit:junit:$junitVersion"
     androidTestImplementation "androidx.test.ext:junit:$androidxJunitVersion"
     androidTestImplementation "androidx.test.espresso:espresso-core:$androidxEspressoCoreVersion"
--- a/android/app/src/main/java/me/peanut/wallet/InstallReferrerPlugin.java
+++ b/android/app/src/main/java/me/peanut/wallet/InstallReferrerPlugin.java
@@ -1,5 +1,8 @@
 package me.peanut.wallet;
 
+import android.os.Handler;
+import android.os.Looper;
+
 import com.android.installreferrer.api.InstallReferrerClient;
 import com.android.installreferrer.api.InstallReferrerStateListener;
 import com.getcapacitor.JSObject;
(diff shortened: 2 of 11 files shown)


kushagrasarathe:
@coderabbitai review

kushagrasarathe:
Superseded by #2560 — this needs to base on `main`: the native release pipeline (AAB for Play internal testing, release tags, Capgo production) builds from main, and the feature must be in a store-delivered binary to be device-tested. #2560 is the same feature adapted for main (which lacks dev's in-app i18n — the restored locale is persisted under the `app-locale` key that system reads, instead of applied live). Closing.

PR #2562, opened 2026-07-29
fix(kyc): route every native entry point through the Sumsub Cordova SDK

innolope-dev:
## Problem

Only `handleInitiateKyc` branched to the native Sumsub SDK. `SumsubKycFlow` was the single call site that hid the KYC modals on Capacitor — the other ~12 (card application, POA re-upload, self-heal resubmit, start-action, restart-identity, the Manteca and Claim flows) rendered `SumsubKycWrapper` unconditionally, so on native those ran the **browser WebSDK inside the WebView**.

When Sumsub's own init fails there, it paints their "Initialization error" screen inside a cross-origin iframe: no exception, no `onError`, nothing in Sentry or PostHog. Native KYC could fail completely silently, and did — there are no native-origin Sumsub errors in Sentry at all, which is the symptom, not the reassurance.

## Approach

Move the platform decision into `SumsubKycWrapper` — the choke point every entry point already funnels through — so all of them are fixed at once and there is **one** native path instead of two.

## Changes

| File | |
|---|---|
| `SumsubNativeSdk.tsx` | new — drives the Cordova plugin, reports every exit path |
| `SumsubSdkErrorView.tsx` | new — failure UI shared by both drivers |
| `sumsubSdk.types.ts` | new — shared props contract |
| `SumsubKycWrapper.tsx` | platform switch; web impl renamed `SumsubWebSdkModal`, otherwise untouched |
| `useSumsubKycFlow.ts` | −49 lines: the ad-hoc native branch is gone |
| `SumsubKycFlow.tsx` | dropped the `!isCapacitor()` guard |
| `types/sumsub-cordova.d.ts` | real plugin API, typed off its `dist` bundle + Java source |

Three bugs fixed beyond the routing:

- **Submissions were dropped.** The old handler only fired on `newStatus === 'Approved'`. The plugin forwards no `onApplicantSubmitted` equivalent, so a native user who submitted documents for async review (`Pending`) got neither the in-progress modal nor `onKycSuccess`. Submission is now inferred from the `SNSSDKState` set.
- **A token refresh would have killed the native screen.** `refreshToken()` writes a new token into the same state; the launch effect keys on token *presence*, not identity, so it can't relaunch mid-verification. Covered by a regression test.
- **`dismiss()` on close.** The plugin holds a module-level single-instance lock — without releasing it, every later launch rejects with `"Aborted since another instance is in use!"` for the rest of the app's lifetime.

Every native failure path — missing plugin, throwing init, rejected launch, `success: false` result — now reaches Sentry + `kyc_sdk_init_failed` and surfaces an actionable screen instead of a dead end.

## Testing

- `tsc --noEmit` clean; `eslint` and `prettier` clean on all touched files.
- Full unit suite: **175 suites / 2313 tests passing**, including 9 new tests for the native driver and one asserting the wrapper picks it under Capacitor and never loads the WebSDK.
- **Web behaviour is unchanged** — the existing `SumsubKycWrapper` suite passes untouched.

## Release note

This cannot ship as an OTA — the fix is about which SDK gets driven, so it needs a new binary. Please QA the card application and a LATAM multi-level flow on both platforms before release.

### DIFF
--- a/src/components/Kyc/SumsubKycFlow.tsx
+++ b/src/components/Kyc/SumsubKycFlow.tsx
@@ -2,7 +2,6 @@ import { Button, type ButtonProps } from '@/components/0_Bruddle/Button'
 import { SumsubKycModals } from '@/components/Kyc/SumsubKycModals'
 import { useMultiPhaseKycFlow } from '@/hooks/useMultiPhaseKycFlow'
 import { type KYCRegionIntent } from '@/app/actions/types/sumsub.types'
-import { isCapacitor } from '@/utils/capacitor'
 
 interface SumsubKycFlowProps extends ButtonProps {
     onKycSuccess?: () => void
@@ -10,8 +9,8 @@ interface SumsubKycFlowProps extends ButtonProps {
     regionIntent?: KYCRegionIntent
 }
 
-// entry point for kyc. delegates to useMultiPhaseKycFlow which handles
-// both web (sumsub web sdk) and native (sumsub cordova plugin) paths.
+// entry point for kyc. delegates to useMultiPhaseKycFlow for the logic;
+// SumsubKycModals -> SumsubKycWrapper picks the web or native SDK by platform.
 export const SumsubKycFlow = ({ onKycSuccess, onManualClose, regionIntent, ...buttonProps }: SumsubKycFlowProps) => {
     const flow = useMultiPhaseKycFlow({ onKycSuccess, onManualClose, regionIntent })
 
@@ -23,7 +22,7 @@ export const SumsubKycFlow = ({ onKycSuccess, onManualClose, regionIntent, ...bu
 
             {flow.error && <p className="text-red-500 mt-2 text-sm">{flow.error}</p>}
 
-            {!isCapacitor() && <SumsubKycModals flow={flow} />}
+            <SumsubKycModals flow={flow} />
         </>
     )
 }
--- a/src/components/Kyc/SumsubKycWrapper.tsx
+++ b/src/components/Kyc/SumsubKycWrapper.tsx
@@ -5,12 +5,16 @@ import { useTranslations } from 'next-intl'
 import Modal from '@/components/Global/Modal'
 import ActionModal from '@/components/Global/ActionModal'
 import { Icon, type IconName } from '@/components/Global/Icons/Icon'
-import { Button, type ButtonVariant } from '@/components/0_Bruddle/Button'
+import { type ButtonVariant } from '@/components/0_Bruddle/Button'
 import Loading from '@/components/Global/Loading'
 import posthog from 'posthog-js'
 import { useModalsContext } from '@/context/ModalsContext'
 import { ANALYTICS_EVENTS } from '@/constants/analytics.consts'
+import { isCapacitor } from '@/utils/capacitor'
 import { evaluateSumsubStatusEvent, type SumsubStatusEventPayload } from './sumsubStatusEvent.utils'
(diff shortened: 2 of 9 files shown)


innolope-dev:
### Declining one code-analysis finding

> **medium** `react-effect-derives-state` — `src/components/Kyc/SumsubNativeSdk.tsx:140` — small useEffect that only sets state from deps

```tsx
useEffect(() => {
    if (!visible) setFailure(null)
}, [visible])
```

Keeping this as-is. The obvious way to satisfy the rule is folding the reset into the launch effect, but that effect runs *after* the reopened render — so a user who hit an error, closed, and reopened would get one frame of the stale error modal before it cleared. Resetting on close avoids that. The alternatives that also avoid it (the `useState`-during-render adjustment pattern, or an open-generation counter) are more state and more code, not less.

Fair heuristic in general; it points the wrong way here.

### On the rest of the report

The `+27 new / -20 resolved` headline is mostly line-number churn: nearly every "new" complexity finding in `useSumsubKycFlow.ts` and `SumsubKycWrapper.tsx` has a matching resolved entry at its old line (e.g. `useSumsubKycFlow: MDD 193.2` at :64 → `182.4` at :52). Both files got measurably simpler — `useSumsubKycFlow` is CC 85 → 78, SLOC 395 → 354, since the ad-hoc native branch came out. The real `+12.61` is the two new files existing.

kushagrasarathe:

Read the full diff. This reroutes ~12 KYC entry points to one native SDK path — the highest-risk of the series — so I checked the two things that would regress it.

**Verified:**
- **Token-refresh parity.** The old inline native path threaded `levelName`+`targetCountry` into its refresh handler (a mismatch mints a token for the wrong applicant action). The shared `flow.refreshToken` (`useSumsubKycFlow.ts:351-364`) threads the same two — no regression when native now uses the shared prop.
- **Modals are native-safe.** With the `!isCapacitor()` guard gone, `SumsubKycModals` renders `SumsubKycWrapper` (→ native SDK, renders only failure UI), the in-progress modal, and the Bridge ToS iframe — all standard WebView-safe React. `isMultiLevel` being unused by the native driver is correct: the native SDK owns level transitions internally.
- **Single-instance lock** released via `dismiss()` in cleanup — matches the plugin's documented "Aborted since another instance is in use!" failure mode.
- Every native failure path (missing plugin, throwing init, rejected launch, `success:false`) now reaches Sentry + `kyc_sdk_init_failed` + an actionable screen. That silent-cross-origin-iframe dead end was the real bug.

```
TIER: T2 (native KYC — access-gating, ~12 entry points rerouted)
PAIRED PR: none — self-contained FE.
A. BREAKAGE:  pass in code — refresh parity, modal safety, and instance-lock all check out.
B. PERF:      pass.
C. QUALITY:   pass — collapses two native paths into one at the choke point; shared error view + props contract.
D. SECURITY:  pass — no new authz surface; token handling unchanged.
```

**Non-blocking but REQUIRED before release (author-flagged):** cannot ship as OTA — needs a new binary. Real-device QA of (a) card application, (b) a LATAM multi-level flow on both platforms, (c) the dismiss-lock across a close→relaunch. The submitted-state inference (`SUBMITTED_STATES`) and lock release are reasoned, not yet observed on hardware.

kushagrasarathe (approved):
Approving — no blocking code findings. Token-refresh parity, modal native-safety, and the instance-lock dismiss all check out. REQUIRED before release (not a code block): real-device QA of card application + a LATAM multi-level flow on both platforms, and the dismiss-lock across close/relaunch — this needs a new binary, not an OTA.

Issue #2543, opened 2026-07-28
csp-report collector: edge rate-limit + route-level test coverage


Hugo0:
Two follow-ups deliberately left out of #2519 (shipped as a prod noise hotfix, kept tight).

### 1. Edge rate-limit on `/api/csp-report`

CodeRabbit flagged the endpoint as unauthenticated ([thread](https://github.com/peanutprotocol/peanut-ui/pull/2519#discussion_r3664863417)). Declined in-PR because an in-memory per-IP limiter would be close to theater — the route is `force-dynamic`, so state is per-instance and resets on cold start, and a distributed source sidesteps it entirely.

Worth being precise about what is and is not exposed:

- **Sentry quota: unchanged by #2519.** The policy previously pointed `report-uri` straight at Sentry, and that URL embeds the public DSN key that ships in the client bundle. Anyone can still POST there directly without touching peanut.me. Nothing we do on this route closes that.
- **Our invocation + egress cost: genuinely new.** Up to 20 outbound fetches (3s timeout each) per inbound request, on a route an attacker controls, with `maxDuration: 300` from `vercel.json` applying to `app/api/**`.

The right control is a **Vercel WAF rate-limit rule on `/api/csp-report`** — it runs before the function is invoked and holds across instances. Low priority; the fan-out cap (`MAX_FORWARDS_PER_REQUEST`) already bounds the amplification per request.

### 2. Route-level test coverage

`src/utils/csp-report.utils.ts` is well covered (38 tests). `src/app/api/csp-report/route.ts` has **none**, and it holds the subtlest logic in the feature — logic that has now been reworked three times on that branch:

- cap-before-de-duplicate ordering (getting this backwards silently loses a group first sighting — the original CodeRabbit Major on #2519)
- intra-batch de-duplication
- oldest-eviction at `SEEN_GROUPS_MAX`
- the 1% duplicate sampling
- content-type gating and the always-204 contract

Cleanest route: extract the selection loop into a pure function in the utils file that takes the `shouldForward` predicate as a parameter, leaving `seenGroups` in the route. That makes the whole invariant unit-testable without moving module-level mutable state into a utils file.

Not urgent — this is telemetry sampling, it does not move money — but it is the part most likely to regress unnoticed.

PR #2563, opened 2026-07-29
fix(deploy): reload documents left running on a superseded deployment

innolope-dev:
## Summary

Investigated as a "CSP regression removing RPC endpoints from `connect-src`". **That turned out not to exist** — but the investigation surfaced a real defect underneath it, which is what this PR fixes.

### The reported CSP regression is not real

- Production's `connect-src` contains **all 14 chain RPC hosts**. Nothing was removed; `main` and `dev` are byte-identical in `next.config.js`.
- The **enforcing** header is only `frame-ancestors 'self'; object-src 'none'; base-uri 'self'` — no `connect-src` at all. Every violation is `disposition: report`. **No request has ever been blocked by CSP.**
- The gaps that did exist were already closed by #2479 (RPC hosts) and #2519 (`wss:` + the same-origin collector).

It also was not a spike that started 22 hours ago: `PEANUT-UI-R79` has been open since 2026-07-22 (8,002 events / 406 users). Sentry's issue search reports "first seen" relative to the query window, which is what made a week-old issue look new.

### What those 7,800 events actually are

Every sampled violation carries a **superseded** `original-policy`:

| Issue | policy vintage | `report-uri` |
|---|---|---|
| `PEANUT-UI-R79` (chainstack, 8,002 occ / 406 users) | pre-#2479 — no RPC hosts at all | Sentry-direct |
| `PEANUT-UI-R7M` (`wss://api.peanut.me`, 5,083 occ / 1,603 users) | post-#2479, pre-#2519 — no `wss:` | Sentry-direct |

They come from documents loaded days earlier and still open. Those documents also carry the **old** `report-uri` pointing straight at Sentry's ingest endpoint, so #2519's collector (`/api/csp-report`, de-dup + 1% sampling) is structurally unable to filter them. No code change suppresses that traffic.

### The actual defect — and what this PR fixes

**1,603 users were running weeks-old bundles.** A loaded document can outlive arbitrarily many deployments, keeping its original JS chunks, API assumptions and response headers, because three things compound:

- `src/app/layout.tsx` skips `window.location.reload()` on `controllerchange` **in standalone PWA** — deliberate, an Android standalone session can bounce out to Chrome.
- App Router navigations fetch RSC payloads, never a document, so headers are never refreshed.
- `src/app/sw.ts` spreads serwist's `defaultCache`; browser navigations fall through its `others` NetworkFirst rule, and a cached `Response` replays the headers it was stored with.

There was no deployment-version check anywhere. The only reload-on-stale machinery, `src/utils/chunk-error-recovery.ts`, is reactive (fires once an asset 404s), rate-limited, and *also* PWA-disabled.

## Changes

| File | |
|---|---|
| `src/app/api/version/route.ts` | new — `force-dynamic`, `no-store`, returns the live deployment's commit |
| `src/hooks/useStaleDeploymentReload.ts` | new — detect, gate, purge, reload |
| `src/components/Global/StaleDeploymentReload/index.tsx` | new — null-rendering mount point |
| `src/app/ClientProviders.tsx` | mount it inside the provider tree |
| `src/constants/cache.consts.ts` | add `DOCUMENT_CACHE_PATTERNS` |
| `src/utils/cache.utils.ts` | new — `purgeCaches` / `isStandalonePwa` |
| `src/context/authContext.tsx` | reuse `purgeCaches` instead of the inline copy |

**Detection.** `NEXT_PUBLIC_GIT_COMMIT_HASH` is already inlined at build time (`next.config.js`), so a stale bundle holds the old hash while `/api/version` — always executed by the live deployment — returns the current one. Checked on mount (catches a document served back out of the SW cache), on tab re-focus, and on a 30-minute poll. The poll matters: a tab that is never backgrounded and never navigated is exactly the shape of the sessions that stayed stale for days, and mount-only detection would miss them entirely.

**Reload timing is deliberately conservative.** A forced reload mid-flow can destroy state that only lives in component memory (a Sumsub WebSDK session, a half-filled card application) or leave a user unsure whether a payment went through. Staleness is latched on detection and acted on only at a safe boundary — a route change, or a flow finishing — with the gate clear:

| Signal | Source |
|---|---|
| `hasPendingTransactions` | `usePendingTransactions` (`BALANCE_DECREASE` mutations) |
| `isSendingUserOp` | zerodev redux slice |
| `isLoading` | loading-state context (covers `'Executing transaction'`, `'Withdrawing'`, `'Paying'`, `'Submitting Offramp'`, `'Awaiting KYC confirmation'`) |
| path deny-list | `card`, `withdraw`, `setup`, `add-money`, `kyc` — matched per segment, so `/en/card` is covered too |

**Caches are purged before reloading.** Otherwise the reload can be answered out of Cache Storage with the same stale document and its stale headers.

**Loop guard.** A one-attempt `sessionStorage` latch: if a reload does not clear the mismatch, the reload is not the fix and the hook stands down for the session rather than looping. The latch is cleared once the bundle is current again, so a later deploy can still act.

Mounted inside the providers rather than called in `ClientProviders`' body like `useOtaUpdates`, because it reads the query client, redux and the loading-state context. Inert on Capacitor — native serves local files and updates through Capgo.

## Verification

- 22 new unit tests, covering each gate signal, the deny-list, the loop guard, the standalone path, the poll, and both failure modes of the endpoint.
- `tsc --noEmit`: no errors in any touched file.
- `pnpm build` succeeds; `/api/version` builds as `ƒ` (dynamic).
- Probed against the built server: `{"commit":"134cf01"}` with `Cache-Control: no-store, no-cache, must-revalidate, proxy-revalidate`, matching `git rev-parse --short=7 HEAD`.
- Full suite 2,300 passing. The 2 failing suites (`countryCurrencyMapping`, `lib/content`) are fresh-worktree artifacts — uninitialized `src/content` submodule and missing generated `public/flags` — not touched here.

## Risks

- **Standalone PWA needs a real-device check before merge.** This intentionally does *not* copy the blanket PWA skip — that skip is the main reason documents survive. It uses `window.location.replace()` rather than `reload()` to avoid the documented Android PWA→Chrome bounce, but that has only been reasoned about, not observed on hardware. If the bounce reproduces, the fallback is to degrade standalone to a "New version available — Reload" prompt and keep the automatic path for browser tabs.
- **This helps future deploys only.** Already-stale documents cannot receive this code, so the current ~1,400 CSP events/day will decay as those sessions die, not drop on merge. That is expected, not a failed fix.
- Adds one small serverless invocation per open tab per 30 minutes.

## Not in scope

#2543's collector follow-ups (Vercel WAF rate-limit on `/api/csp-report`, route-level tests) are unchanged and still open.

### DIFF
--- a/src/app/ClientProviders.tsx
+++ b/src/app/ClientProviders.tsx
@@ -9,6 +9,7 @@
 import { ConsoleGreeting } from '@/components/Global/ConsoleGreeting'
 import RainCooldownIntroModal from '@/components/Global/RainCooldown/IntroModal'
 import StaleCardApprovalReEnableModal from '@/components/Global/StaleCardApproval/ReEnableModal'
+import StaleDeploymentReload from '@/components/Global/StaleDeploymentReload'
 import BadgeEarnToast from '@/components/Badges/BadgeEarnToast'
 import { AppLockGate } from '@/components/Global/AppLock'
 import { ScreenOrientationLocker } from '@/components/Global/ScreenOrientationLocker'
@@ -61,6 +62,11 @@ export function ClientProviders({ children }: { children: React.ReactNode }) {
                                 {/* Non-intrusive "badge unlocked" toast on /home (TASK-19791).
                                 Global so it surfaces wherever the user lands after earning. */}
                                 <BadgeEarnToast />
+                                {/* Mounted inside the providers (not called in this
+                                component's body like useOtaUpdates) because it
+                                reads the query client, redux and loading-state
+                                context to know when a reload is safe. */}
+                                <StaleDeploymentReload />
                                 {HarnessBootstrap && (
                                     <Suspense fallback={null}>
                                         <HarnessBootstrap />
--- a/src/app/api/version/route.ts
+++ b/src/app/api/version/route.ts
@@ -1,0 +1,23 @@
+import { NextResponse } from 'next/server'
+
+/**
+ * The commit of the deployment serving this request.
+ *
+ * NEXT_PUBLIC_GIT_COMMIT_HASH is inlined at build time (next.config.js), so a
+ * client bundle carries the hash of the deployment it was built from while this
+ * route — always executed by whichever deployment is live — returns the current
+ * one. The mismatch is what useStaleDeploymentReload keys off.
+ *
+ * force-dynamic + no-store on purpose: a cached response would report a stale
+ * deployment as current and defeat the whole check. Next already emits its own
+ * no-store for dynamic routes and wins on the wire; the explicit header is the
+ * declaration of intent and the fallback if this route ever stops being dynamic.
+ */
+export const dynamic = 'force-dynamic'
+
+export async function GET() {
+    return NextResponse.json(
+        { commit: process.env.NEXT_PUBLIC_GIT_COMMIT_HASH ?? 'unknown' },
+        { headers: { 'Cache-Control': 'no-store, max-age=0' } }
+    )
+}
--- a/src/components/Global/StaleDeploymentReload/index.tsx
+++ b/src/components/Global/StaleDeploymentReload/index.tsx
@@ -1,0 +1,13 @@
+'use client'
+
+import { useStaleDeploymentReload } from '@/hooks/useStaleDeploymentReload'
+
+/**
+ * Renders nothing. Exists so the hook runs inside the provider tree — it reads
+ * the query client, the redux zerodev slice and the loading-state context, none
+ * of which are available in ClientProviders' own body.
+ */
+export default function StaleDeploymentReload() {
+    useStaleDeploymentReload()
+    return null
+}
--- a/src/constants/cache.consts.ts
+++ b/src/constants/cache.consts.ts
@@ -7,6 +7,10 @@ const CACHE_NAMES = {
     USER_API: 'user-api',
     TRANSACTIONS: 'transactions-api',
     KYC_MERCHANT: 'kyc-merchant-api',
+    PAGES: 'pages',
+    PAGES_RSC: 'pages-rsc',
+    PAGES_RSC_PREFETCH: 'pages-rsc-prefetch',
+    OTHERS: 'others',
 } as const
 
 /**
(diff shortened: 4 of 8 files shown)


kushagrasarathe:

Read the full diff (all 8 files) and verified the load-bearing claims against the codebase.

**Verified (all hold):**
- serwist `defaultCache` names `pages` / `pages-rsc` / `pages-rsc-prefetch` / `others` are real (`@serwist/next/dist/index.worker.js:4-6,232`); the `others` NetworkFirst rule (line 230) is the one matching browser navigations, and `pages` keys off a `Content-Type` request header navigations don't send — the "effectively dead" comment is accurate.
- `NEXT_PUBLIC_GIT_COMMIT_HASH` is inlined in `next.config.js:36`, so the stale-bundle-vs-live-endpoint mismatch mechanism is sound.
- authContext refactor is behavior-preserving: `purgeCaches(USER_DATA_CACHE_PATTERNS)` == the old inline filter/delete.
- 18 test cases cover every gate signal, deny-list, loop guard, standalone, poll, and both endpoint failure modes.

```
TIER: T1-treatment (external/unfamiliar author -> T1 rules) & T2 (reload machinery can wipe in-progress KYC/card/withdraw state)
PAIRED PR: none - self-contained. /api/version is an in-repo Next route; the env var already exists on dev. No deploy-order constraint.
A. BREAKAGE:  pass - no counterpart to gate on. One live risk (author-flagged): the standalone-PWA window.location.replace() path is reasoned-about, NOT device-tested. If it still bounces Android PWA -> Chrome, installed-app users get kicked. Verify on hardware before merge.
B. PERF:      pass - 1 serverless GET per open tab per 30min + on-focus (throttled 5min) + mount. Reasonable.
C. QUALITY:   pass, 2 minor non-blocking notes:
              - cache.consts.ts DOCUMENT_CACHE_PATTERNS lists pages/pages-rsc/pages-rsc-prefetch, but purgeCaches uses name.includes('pages') which already matches all three. Redundant (defensive), harmless.
              - useStaleDeploymentReload.ts RELOAD_GUARD_MS/RELOAD_AT_KEY time-throttle is nearly subsumed by the one-attempt RELOAD_ATTEMPTED latch set in the same block. Near-dead, not wrong.
D. SECURITY:  pass - /api/version returns only the commit hash, already public in the client bundle. No auth needed, no PII, no-store correct. No secrets.
TASK LINK:    MISSING (gate) - body references #2479/#2519/#2543 but no tracking task.
```

**Two gates before merge:**
1. Add a tracking task link (untracked PRs are how paired-PR mistakes slip).
2. Real-device standalone-PWA check on Android — the one unverified assumption, flagged by the author.

**One design point to confirm (not a bug):** `send` is deliberately absent from `RELOAD_UNSAFE_SEGMENTS` — correct *if* send-flow state lives in the URL (per the URL-as-state rule) so a reload restores it. Worth a reviewer nod that send-flow is URL-backed end to end.

Solid PR — thorough investigation, honest about what it fixes vs. doesn't, tests match the claims.

<sub>🤖 AI advisory review (pr-review rulebook). Not a final approval — routes to a human peanut-ui owner.</sub>

kushagrasarathe (approved):
Approving — no blocking code findings. Diff read in full; serwist cache names, env inlining, and the authContext refactor all verified. 18 tests match the claims.

Non-blocking follow-ups tracked in the review comment above: add a tracking task link, and a real-device Android standalone-PWA check on the replace() path before this rides to prod.

PR #2564, opened 2026-07-29
fix(csp): close report-only allow-list gaps for the SW, RPC tail and Google Ads

innolope-dev:
## Summary

Five gaps in the report-only `connect-src`/`script-src`/`style-src`/`frame-src` allow-list. **Nothing breaks today** — the enforcing header is still only `frame-ancestors`/`object-src`/`base-uri`, and everything here is `disposition: report`. Each of these would break something the moment the policy is promoted to enforcing.

Companion to #2563, which fixes *why* the report-only signal is currently unreadable (stale documents replaying superseded policies). Independent of it — different files, no conflict.

Every host below was read out of a source, not guessed: Google's own published domain list, `viem/chains`, and the Typeform SDK bundle.

### 1. Google Ads remarketing — 187 domains, and why it can't be a wildcard

GA4 sends its audience/remarketing beacon (`/ads/ga-audiences`) to the user's **own country Google domain** — `www.google.com.ar` for a user in Argentina, `www.google.it` in Italy — picked by Google at runtime from the user's locale, not by us.

CSP host-sources allow `*` only as the **leftmost label**. So there is no wildcard that spans TLDs:
- `https://*.google.com` does **not** match `google.com.ar` — different registrable domain.
- `https://www.google.*` is not valid CSP grammar.

Enumeration is the only option that keeps remarketing working under enforcement. All 187 domains live in `csp-google-domains.js`, taken **verbatim from Google's own list** at `https://www.google.com/supported_domains`, with the fetch date and refresh command recorded in the file. Only the `www.` host is allowed — the only form present in the violation data.

Observed live under the **current** policy (so these are real gaps, not stale-client noise): `PEANUT-UI-R8F` (`.com.ar`, 54 users), `PEANUT-UI-R8T` (`.com.bd`, 35 users), `PEANUT-UI-R8Z` (`.it`, 8 users) — all still firing.

This supersedes the former lone `https://www.google.com` entry, which is the first element of the generated list.

> **Tradeoff worth a conscious decision:** the policy header grows from ~2.6 KB to **~6.5 KB**, on every response. Comfortably within browser and Vercel limits, but it is real bytes on every request. That is the unavoidable price of the enumeration; the alternative is remarketing silently breaking per-country at enforcement.

### 2. Service worker — `pbs.twimg.com`

The landing page's testimonial avatars. They render as ``, so `img-src https:` would normally cover them — but **serwist's runtime cache re-fetches them from inside `sw.js`, and in a service worker every `fetch()` is `connect-src`** regardless of what the response is used for. The reports carry `document-uri: https://peanut.me/sw.js` for exactly this reason; `/sw.js` is served by the same `/:path*` header rule, so it inherits this policy.

Same mechanism behind the existing `www.googletagmanager.com` and `api.onesignal.com` violations reported from `sw.js` / `OneSignalSDKWorker.js` — those hosts were already allow-listed, so they needed no change.

### 3. RPC tail — 10 hosts

`getPublicClient` (`src/app/actions/clients.ts`) hands any chain absent from `rpcUrls` a bare `http()`, and viem then uses that chain's own default endpoint. That path is reachable from client code via `fetchTokenSymbol` on the token selector, whose chain list is **API-driven** — so it covers the whole of `chainRegistry.consts.ts`, not just the ten chains wired into `wagmi.config`.

Taking the registry as the boundary rather than picking "popular" by feel: it is exactly the set a user can select. Gnosis, Celo and Linea were already covered; the remaining ten, each resolved from `viem/chains`:

| chain | viem default |
|---|---|
| 43114 Avalanche | `api.avax.network` |
| 5000 Mantle | `rpc.mantle.xyz` |
| 999 HyperEVM | `rpc.hyperliquid.xyz` |
| 8217 Kaia | `public-en.node.kaia.io` |
| 57073 Ink | `rpc-gel.inkonchain.com`, `rpc-qnd.inkonchain.com` |
| 747474 Katana | `rpc.katana.network` |
| 9745 Plasma | `rpc.plasma.to` |
| 988 Stable | `rpc.stable.xyz` |
| 4217 Tempo | `rpc.presto.tempo.xyz` |

Silent in the report-only data because these chains are rarely picked — **not** because they are unreachable. Same shape as the `wss://` gap, which stayed invisible until #2519 went looking for it.

### 4. Typeform — `frame-src` / `style-src` / `connect-src`

The feedback `Widget` in `Global/Layout` needs `form.typeform.com` (the SDK's `DEFAULT_DOMAIN`, i.e. the iframe), `embed.typeform.com` (injected stylesheet from `/next/css/`) and `api.typeform.com` (single-embed `fetch`). Directives confirmed by reading the SDK bundle.

Currently **unreachable** — its `Modal` is never opened, so it produces zero violations — but re-arming it is a one-line change, and the failure mode under enforcement is a blank iframe.

### 5. jsdelivr — `script-src`

`cdn.jsdelivr.net` serves the mermaid bundle for `/dev/kyc-flows`. Dev-only route, but a real violation on a real deployment.

## Deliberately NOT added

`www.googleadservices.com` and `googleads.g.doubleclick.net` — the other two standard Google Ads remarketing hosts. Neither appears anywhere in 14 days of violation data; the current setup beacons via the GA4 `ga-audiences` path only. Adding them would be speculative widening, against the precedent #2519 set. **Add them if a Google Ads remarketing tag is ever added to GTM.**

## Verification

- `pnpm build` succeeds.
- Served the production build and asserted against the **real** response header: all 187 Google domains present, and every one of `pbs.twimg.com`, `api.typeform.com`, `api.avax.network`, `rpc.mantle.xyz`, `public-en.node.kaia.io`, `cdn.jsdelivr.net`, `embed.typeform.com`, `form.typeform.com` appears exactly once.
- Rendered policy parses into all 13 expected directives; `connect-src` has 241 sources with **zero duplicates**.
- `prettier --check` clean on both files.

## Risks

- Larger response header on every request (see the tradeoff note above).
- The Google domain list is a point-in-time snapshot. Google adds/retires country domains rarely; the refresh command is in the file header. A missing future domain degrades remarketing in that country only — it cannot break the app.


kushagrasarathe (approved):
**Approving — READY.** Reviewed against: code quality, DRY, bug risk, security, funds/trust.

**What it does:** Adds ~187 Google country domains (new `csp-google-domains.js`), 10 exact RPC hosts, and Typeform/jsdelivr/twimg hosts to the `Content-Security-Policy-Report-Only` allow-list. All additions are specific hosts or provider-scoped wildcards; nothing loosens the enforced policy.

**Security — clean (verified against the branch, not a stale checkout):**
- **Genuinely report-only.** The enforced header is still only `frame-ancestors 'self' https://hugo0.com` — the entire touched allow-list feeds `Content-Security-Policy-Report-Only`. None of these additions can break resource loading today, and none widens any *enforced* boundary. Security/funds risk from this diff is effectively nil.
- **No directive loosened.** `script-src` gains only `https://cdn.jsdelivr.net` (exact host) — no new `unsafe-*`, no `data:`, no wildcard. `style-src` gains exact `embed.typeform.com`. Pre-existing `'unsafe-inline'/'unsafe-eval'` and broad `img-src … https:` are untouched (not this PR's job to fix).
- **RPC tail: all 10 are exact hosts**, no wildcards.

**DRY — good.** The Google list is *generated* from a suffix list (187 entries, zero duplicates verified by executing the module) rather than hand-typed, with a documented refresh command. The former standalone `https://www.google.com` source entry is fully removed — no double-listing.

**Minor (non-blocking):**
- The Typeform additions are for a `Widget` that is currently unreachable (Modal never opened) — speculative allow-listing, but harmless under report-only and explicitly flagged with a re-arm rationale.
- Header grows ~2.6 KB → ~6.5 KB per response (disclosed). Within limits, negligible; inherent to the enumeration (CSP grammar can't wildcard across TLDs).

**Funds/trust:** No risk — report-only, no enforced boundary touched.

PR #2568, opened 2026-07-29
feat(security): biometric-guarded session token (Keychain/Keystore)

### LINKED ISSUE #2472: Native app lock: move the session token into biometric-guarded Keychain/Keystore (real control behind #2461)

innolope-dev:
Continuation of #2489, retargeted at `dev`. Same feature, same reviewed content — see #2489 for kushagrasarathe's adversarial review (design approved; the one blocking code item, the `authReady()` gate in `onramp-quote.ts`, and the follow-up test fixes are all included here).

**Why a new PR:** #2489's branch was based on the `mobile-release` lineage. Repository rules forbid force-pushing the branch, and merging `dev` into it would drag the whole release-branch ancestry into `dev` on merge (breaking future backmerges). This branch is the same seven commits transplanted onto `dev`, plus one commit restoring the mobile-release-lineage pieces `dev` hasn't received yet (`clearEpoch` sliding-refresh guard, awaited `clearAuthToken` on 401/404, lock-screen message keys).

The feature is already merged directly into `mobile-release` (87f14c4e7) per our release-branch flow; this PR lands it in `dev`.

Original description follows.

---

Closes #2472 — the follow-up to #2461, which shipped the app lock as a privacy screen only.

## What changes

**Three session modes on native, detected once per launch** (`src/utils/auth-token.ts`):

- **guarded** — the JWT is stored via `@capgo/capacitor-native-biometric` (pinned **8.6.0**, clears the 14-day dependency floor) with `AccessControl.BIOMETRY_CURRENT_SET`: Keychain `SecAccessControl` on iOS, a `BiometricPrompt.CryptoObject`-bound Keystore key on Android. **The unlock ceremony IS the token read** — one OS biometric prompt both proves presence and releases the credential. No separate WebAuthn assertion.
- **plain** — byte-for-byte the previous behavior, kept for older binaries running OTA'd JS (plugin is feature-detected) and for devices without enrolled biometrics. After the legacy gate opens, the session migrates to guarded storage; the plain copy is deleted only after the next cold start's guarded read proves the round-trip.
- **none** — signed out, nothing to protect.

**Fail closed:** the lock decision derives from a non-secret presence marker (`guarded-token-present`), never from the user query. Deleting the marker or any local pref yields a signed-out app — never an open session — because the JWT itself is unreadable without a biometric.

**Session paused while locked:** `authReady()` parks every API caller (a locked app never emits an unauthenticated request that would 401 → `clearAuthToken()`); the `[USER]` query is disabled (which also kills `refetchOnWindowFocus` on resume); the auto-refresh poller skips its tick; a sliding-refresh token landing after suspension is dropped. On unlock, react-query's stale refetch doubles as the post-unlock refresh.

**Re-enrollment = session expired, never a stuck lock:** `BIOMETRY_CURRENT_SET` invalidates the item when biometrics change (by design). Both platforms surface this as not-found → clean logout to /setup.

**card.ts fix (from the #2463 re-review):** `services/card.ts` read the `jwt-token` web cookie directly, which never worked on native — now routed through `authReady()` + `getAuthHeaders()`.

## Android write-prompt nuance

Keystore writes prompt outside a post-auth window (per-op auth keys). Handled with `authValidityDuration: 60`: the unlock read opens a 60s window that silently covers the re-mint `/users/me` ships right after unlock; writes outside the window are skipped (memory-only, server re-mints later). iOS Keychain writes never prompt. Consequence: on Android, login and one-time migration each show one extra 'Protect Credentials' prompt.

## Deliberate trade-offs

- Lock-screen **Log out** in guarded mode is local-only (`skipBackendCall`): there's no token in memory to authenticate the revocation POST. tokenVersion isn't bumped — acceptable since the attacker can't extract the guarded JWT.
- Strict biometric-only (no device-passcode fallback): the plugin's gated path exposes only `BIOMETRY_CURRENT_SET`/`BIOMETRY_ANY` on iOS and `AUTH_BIOMETRIC_STRONG` keys on Android. After an OS biometric lockout the escape hatch is Log out → fresh login.
- OTA rollback after migration cleanup looks signed-out (re-login) — accepted, worth a release-note line.

## Testing

- 168 suites / 2206 tests green, including new coverage: mode-detection truth table (incl. fail-closed uncertainty rule), authReady park/release, suspend-without-epoch-bump, locked `setAuthToken` no-op, silent-window write policy, migration round-trip lifecycle, plugin error-code mapping, lock registry, and a D7 regression test (gate locks without waiting for the user query — the permanent-white-screen trap).
- Typecheck: same 228 pre-existing errors before/after (fresh-worktree asset-decl noise), zero introduced.
- **Needs a binary release** (new plugin) + the on-device matrix from #2472 before store rollout: unlock→token released; cancel→stays locked; 6-min background; kill/relaunch; re-enroll→session-expired; old-build→new-build migration.

## Plugin gate verified by code-read

iOS: `SecAccessControlCreateWithFlags(kSecAttrAccessibleWhenPasscodeSetThisDeviceOnly, .biometryCurrentSet)` + `SecItemCopyMatching` with `LAContext`. Android: Keystore AES-GCM with `setUserAuthenticationRequired(true)`, `setUserAuthenticationValidityDurationSeconds(-1)` (per-op), `setInvalidatedByBiometricEnrollment(true)`, reads/writes through `BiometricPrompt.CryptoObject` ciphers. The ungated `getCredentials()`/`verifyIdentity()` paths are never used.


kushagrasarathe (approved):
**Approving — READY.** Reviewed against: code quality, DRY, bug risk, security, funds/trust.

**What it does:** Adds biometric-guarded session-token infrastructure (Keychain/Keystore), shipped **dormant** behind `OPEN_GATED` (default off). A default production build is byte-identical to current behavior with zero on-device biometric-binding risk until the flag + a binary release deliberately enable it.

**Supersedes #2489 — all CHANGES_REQUESTED addressed:**
- The blocking `await authReady()` in `onramp-quote.ts` is present; `apiFetch` and `getSessionTokenForSocket` also await it, so every authenticated caller (incl. the charges WebSocket) parks while locked — no un-gated caller remains.
- The two accepted residuals (migration-window plaintext, legacy plain-mode fail-open) are moot while `OPEN_GATED=false` — `guardedModeEnabled()` short-circuits every guarded path.

**Security — clean:**
- Token never logged / sent to Sentry; error paths carry the plugin's message, never the secret.
- Gate fails **closed**: the "can't tell" branch in `detectSessionMode` returns `guarded` (stays locked), never falls open; lock decision derives from a non-secret presence marker, and the JWT is unreadable without a biometric.
- Access control `BIOMETRY_CURRENT_SET` (invalidates on re-enrollment → clean logout); iOS `kSecAttrAccessibleWhenPasscodeSetThisDeviceOnly`, Android per-op auth key + `setInvalidatedByBiometricEnrollment(true)`.
- Lifecycle (store/read/clear) consistent; `clearAuthToken` wipes memory + Keychain + Preferences + cookie jar + localStorage; `clearEpoch` guard prevents a sliding-refresh token resurrecting a cleared session.

**Funds/trust:** Cannot hurt funds — money movement stays passkey-gated at the transaction layer independently of this flag; the guarded path is dead code until an env flag + binary release enable it.

**Minor (disclosed & accepted, non-blocking):** OTA-toggling `OPEN_GATED` off after migration cleanup orphans the Keychain-only token → user looks signed-out (must re-login). Not a funds/security risk; just ensure the release note lands if/when the flag is flipped.

On-device iOS+Android matrix + residual sign-off correctly gate flipping the flag on, not this merge.

PR #2570, opened 2026-07-29
fix(card): force collateral-only routing on lock/cancel

abalinda:
## What broke (prod)

Locking or cancelling a card fails with **"Unexpected withdrawal strategy — please contact support"** whenever the user's smart-wallet USDC ≥ their card spending power. The card stays `ACTIVE`; the user burns a passkey tap and cannot secure or close their card. No money at risk — the signed UserOp is never broadcast.

Reported via Crisp on 2026-07-24 (bug bounty paid). **3 prod users are currently blocked**, including one with $25k wallet / $2k collateral; 235 distinct users routed `smart-only` in the last 6 weeks and were transiently exposed (auto-balance keeps wallets swept near-empty, which is why the standing count is low).

## Root cause — regression trail

The lock/cancel withdrawal exists to drain Rain collateral back to the wallet, so it must always route `collateral-only`:

- `2eb1da396` (2026-05-08): forced this by passing `smartBalance: 0n` to `signSpend`.
- `cb302d35a` (2026-06-16): removed the `smartBalance` input repo-wide and deleted the `0n` line from both modals **without replacement** — routing fell through to `computeSpendStrategy`, which returns `smart-only` whenever the live wallet balance covers the amount. Both modals then reject their own artifact and throw.
- `ca79b66a7`: added `forceStrategy: 'collateral-only'` for exactly this need, but wired it only into `useReturnExcessCollateral`.

A stale comment ("smart=0n eliminates the smart-only and mixed branches") kept describing the deleted parameter, which is why later reviews missed it.

## The fix (frontend-only, backend unchanged)

1. **`LockCardModal` / `CancelCardModal`**: pass `forceStrategy: 'collateral-only'` to `signSpend` — same pattern as `useReturnExcessCollateral`. Replaced the stale comment with one stating the actual constraint. The `artifact.strategy !== 'collateral-only'` throw stays: unreachable at runtime now, but load-bearing for TS narrowing to the `rainWithdrawal` variant.
2. **Fail closed on unloaded overview** (same block, second known failure): `rainCentsToUsdcUnits(undefined)` → `0n`, so an unresolved `useRainCardOverview` silently skipped the withdrawal and the backend rejected with "Withdrawal signature required to return collateral before this action" (2 users, 07-09 and 07-13). Both modals now throw a friendly retry message (new i18n key `card.errors.cardDetailsLoading`, all 3 locales) instead of signing nothing.

## Tests

New `src/components/Card/__tests__/LockCardModal.test.tsx` covering both modals (6 tests):

- `signSpend` is called **with** `forceStrategy: 'collateral-only'` (exact-args assertion — this is the one that catches the next `SignSpendBundleInput` refactor; the regression already happened once).
- Unloaded overview → no signing, no backend call, error phase.
- Zero spending power → proceeds without signing, in both modals (guards the guard).

The suite was mutation-tested pre-merge: 6 line-level mutants (drop `forceStrategy` ×2, drop the overview guard ×2, `> 0n` → `>= 0n` ×2) all killed by exactly the intended assertion. The existing `useSignSpendBundle.test.tsx` forced-routing contract covers the layer this suite mocks.

Gate: typecheck clean, full jest suite green, prettier clean.

## Known residual (pre-existing, deliberately not touched)

A loaded overview with `balance: null` but real positive Rain spending power would still read as `0n` and hit the server-side rejection. Not covered because `balance: null` normally means no collateral contract — extending the guard to `!overview.balance` would block legitimate no-collateral cancels. Needs a backend-truth check if it ever shows up in prod.

## Verification after deploy

PostHog: `card_withdraw_attempted` with `flow=sign-only` must never report `strategy=smart-only` again; `card_lock_failed` with "Unexpected withdrawal strategy" must stop. Then reply to the affected users' tickets.

## Screenshots

⚠️ NONE — no visual or layout change; the only new user-visible surface is the transient "Card details are still loading" error copy, which requires an unresolved overview query mid-slide and is not cheaply reproducible in the harness. Reviewer flies blind on that one string by design.


jjramirezn (approved):

abalinda:
Superseded by #2571 — same fix rebased onto `main` as a prod hotfix (dev is 88 commits ahead, so repointing the base was not viable). This branch is kept (not deleted): it holds the i18n-ified variant of the fix, needed to resolve the modal/test conflicts when main back-merges into dev.

PR #2578, opened 2026-07-29
fix(watchdog): don't treat a mid-run mirror as an emergency

Hugo0:
The watchdog false-alarmed in Discord today. Content was fine — it caught the pipeline **mid-flight**:

```
16:03:52  push f3008839
16:04:13  watchdog: "the mirror is behind mono ... Stuck for 5m"   ← 21s after the push
16:04:25  mirror completes                                          ← 12s after the watchdog looked
16:43:38  green
```

## Cause

Mirror-behind skipped the 45-minute grace period, on my reasoning that *"a failed mirror will not fix itself."* **That reasoning was wrong.** A mirror that is merely mid-run fixes itself in ~30 seconds, and from outside it is indistinguishable from a dead one.

So any content push landing shortly before a `*/15` tick produces a false alarm — roughly **1 in 20 pushes, indefinitely**. That is exactly the alert-fatigue pattern that makes a watchdog worthless.

## Fix

Skipping the grace period is now reserved for states that genuinely cannot recover: **red CI**, and **a watchdog that cannot evaluate itself**. A dead mirror still alerts — it just waits out the window first, which is what the window is for.

## Decision table, verified

| Case | Before | After |
|---|---|---|
| mirror mid-flight, 5m | ALERT ❌ | **SILENT** ✅ |
| mirror genuinely stuck, 60m | ALERT | ALERT ✅ |
| prod behind, 5m | SILENT | SILENT ✅ |
| prod behind, 60m | ALERT | ALERT ✅ |
| prod behind, CI red | ALERT now | ALERT now ✅ |
| watchdog degraded | ALERT now | ALERT now ✅ |
| healthy | silent | silent ✅ |

Also ran the real step under `bash -e` against live APIs — reports `production content is current`, exit 0.

## Note

This is a false-alarm fix, not a pipeline fix. Content publishing itself is working: three consecutive autonomous publishes today (#2561, #2575, #2576), all merged by the automation, production in sync.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pipeline monitoring during degraded conditions.
  * Watchdog evaluation failures are now escalated immediately.
  * Mirror delays receive the full grace period before being marked as failed, reducing premature failure alerts.

### DIFF
--- a/.github/workflows/content-pipeline-watchdog.yml
+++ b/.github/workflows/content-pipeline-watchdog.yml
@@ -116,13 +116,20 @@ jobs:
                   fi
 
                   STALE=""
-                  MIRROR_BROKEN=false
+                  # URGENT = skip the grace period. Reserve it for states that cannot
+                  # resolve on their own. A mirror that is merely mid-run is NOT one of
+                  # them: it finishes in ~30s, and treating it as urgent turned every
+                  # push landing shortly before a tick into a false alarm. That fired
+                  # for real (16:04:13 run, 21s after a push, 12s before the mirror
+                  # completed — "Stuck for 5m", and green again on the next tick).
+                  # A genuinely dead mirror still alerts; it just waits out the grace,
+                  # which is exactly what the grace period is for.
+                  URGENT=false
                   if [ "$DEGRADED" = true ]; then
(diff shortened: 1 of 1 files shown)


Hugo0:
Back to draft.

PR #2587, opened 2026-07-29
feat(native): make the deferred-link match rate measurable

innolope-dev:
## Summary

Follow-up to #2560, stacked on `feat/deferred-deep-link-main`. Adds the telemetry the deferred-link restore path is missing.

**⚠️ Base:** targets `feat/deferred-deep-link-main`, so the diff shown includes #2560's commits until that merges. Retarget to `main` once #2560 lands — this branch's own change is 3 files / +157.

## Why

The restore path currently emits nothing, and the iOS clipboard `catch {}` swallows the one distinction that matters:

```ts
} catch {
    // user declined the paste prompt, or clipboard unavailable
}
```

A declined paste prompt and an organic install both end up as `raw === null`. So a hand-off that **never works on iOS** is indistinguishable from one **nobody used** — and the only signal either way is the absence of restores, which looks the same as success. The download modal (TASK-20769) is about to drive real volume through this, and the plan was always to measure the match rate before escalating to a blocking interstitial.

## What

`DEFERRED_LINK_RESTORED`, fired once per fresh install with `channel` (`referrer` / `clipboard` / `none`) and one of four outcomes:

| outcome | meaning |
|---|---|
| `restored` | payload found and applied |
| `no_handoff` | nothing there — organic install |
| `marker_missing` | payload present but no `pnutdl=1` — organic Play referrer, or unrelated clipboard text that passed the URL gate |
| `clipboard_unavailable` | **paste prompt declined, or plugin missing** — the broken-hand-off signal |

On `restored` it also reports `has_dest` / `has_locale` / `has_invite` / `has_campaign`.

Plus `trackDeferredHandoffCreated(platform)` as the seam for the modal to report the denominator. Deliberately **not** fired inside `playStoreUrlWithReferrer` — that's a pure URL builder invoked on render, so it would count impressions as taps.

## Privacy

Only booleans and the channel leave the device. The invite code, campaign tag and destination never do — putting a user's inviter and intended screen into analytics would be a real regression. There's a test asserting the payload values are absent from the captured events.

## Safety

`posthog.capture` is wrapped in `try {} catch {}` at both call sites. This runs on the first-launch path, where a telemetry failure must not cost the user their restored context — covered by a test that makes capture throw and asserts the restore still returns.

## Tests

4 added to the existing suite (25 pass total, including all 21 from #2560 unchanged):

- successful restore reports the right fields, and no payload content
- organic install vs declined paste prompt land on **different** outcomes
- present-but-unmarked referrer is `marker_missing`, not `no_handoff`
- a throwing posthog doesn't lose the restored context

`tsc --noEmit` clean on the touched files; prettier clean.

### DIFF
--- a/src/constants/analytics.consts.ts
+++ b/src/constants/analytics.consts.ts
@@ -270,7 +270,33 @@ export const ANALYTICS_EVENTS = {
     DELETE_ACCOUNT_INITIATED: 'delete_account_initiated',
     DELETE_ACCOUNT_CONFIRMED: 'delete_account_confirmed',
     DELETE_ACCOUNT_FAILED: 'delete_account_failed',
+
+    // ── Deferred deep linking (TASK-20772) ──
+    // Fired once per fresh install, on the one-shot restore. `outcome` is a
+    // DEFERRED_LINK_OUTCOMES value and is the whole point of the event: it's the
+    // only way to tell a working store→install hand-off from one that silently
+    // matches nothing, since a declined iOS paste prompt and an organic install
+    // both simply produce no payload.
+    DEFERRED_LINK_RESTORED: 'deferred_link_restored',
+    // Fired on the web side when a store bounce writes the hand-off.
+    DEFERRED_LINK_HANDOFF_CREATED: 'deferred_link_handoff_created',
 } as const
+
+/**
+ * Outcomes for DEFERRED_LINK_RESTORED — why a first-launch restore did or did
+ * not recover context. Distinguishing the empty cases is the reason this event
+ * exists: `clipboard_unavailable` (user declined the paste prompt, or the
+ * plugin is missing) and `no_handoff` (organic install) are indistinguishable
+ * without it, and only the first indicates a broken hand-off.
+ */
+export const DEFERRED_LINK_OUTCOMES = {
+    RESTORED: 'restored',
+    NO_HANDOFF: 'no_handoff',
+    MARKER_MISSING: 'marker_missing',
+    CLIPBOARD_UNAVAILABLE: 'clipboard_unavailable',
+} as const
+
+export type DeferredLinkOutcome = (typeof DEFERRED_LINK_OUTCOMES)[keyof typeof DEFERRED_LINK_OUTCOMES]
 
 /**
  * Valid modal_type values for MODAL_SHOWN / MODAL_DISMISSED / MODAL_CTA_CLICKED events.
--- a/src/utils/__tests__/deferred-link.test.ts
+++ b/src/utils/__tests__/deferred-link.test.ts
@@ -47,6 +47,14 @@ jest.mock('@capacitor/preferences', () => ({
     Preferences: { set: (o: unknown) => preferencesSet(o) },
 }))
 
+const posthogCapture = jest.fn()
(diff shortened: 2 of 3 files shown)


kushagrasarathe (commented):
Reviewed — the direction is right and I'd like to keep this PR (don't close it). Matches the repo's direct `posthog.capture` convention (`useZeroDev`, `webauthn.utils`), booleans-only payload is exactly the right privacy line, and the outcome taxonomy (`no_handoff` / `marker_missing` / `clipboard_unavailable`) is the diagnostic we'd otherwise wish we had after a silent iOS rollout. The posthog-throw isolation test is appreciated.

Two things before this can land, both fallout from the base branch moving under you today:

**1. Stale base — the block you instrument was rewritten in `ffc9e79e8` (Jota's review fixes).** Your diff context still shows the old "one shot, even on failure" consumption. On the current base: Android only consumes on a definitive read (transient `null` retries next launch), and iOS consumes *before* the prompt-raising read. Needs a rebase onto current `feat/deferred-deep-link-main`.

**2. After that rebase, the Android transient path will fire `no_handoff` on every retry launch** — one install with a flaky first boot emits N events and inflates `no_handoff`, which skews the exact match-rate metric this PR exists to protect. Suggest a distinct `read_failed` outcome for the android transient-null path (it's also a signal worth seeing on its own — a rising `read_failed` means the 5s timeout is too tight), or alternatively only capture once consumption happens.

One non-blocking check: on the very first cold launch, `capture` may run before `posthog.init` — posthog-js buffers pre-init calls on the singleton so it should be fine, but worth one on-device confirmation via `/dev/deferred` when we run the internal-track test.

Plan-wise: #2560 merges first, then retarget this to `main`, rebase, adapt per above. Happy to do the rebase myself if you prefer — say the word.

innolope-dev:
**Conflict resolved + retargeted to `main`.**

#2560 merged, so the old base (`feat/deferred-deep-link-main`) is now part of `main` — base changed to `main`, and the branch merges the base-branch tip rather than `main` itself (`main` carries unsigned content-submodule commits that the signature ruleset re-flags on a feature ref, GH013).

The conflict wasn't purely textual. #2560's review pass replaced the unconditional one-shot burn with `markConsumed()`, so **a transient android referrer read no longer consumes** — the next launch retries. Under that behaviour the telemetry as written would have fired `no_handoff` on every retry, i.e. double-counting the install *and* filing a broken read as an organic one, which is the exact conflation this PR exists to remove. Resolution:

- `doRestore` now tracks whether the one-shot was actually burned, and the empty-outcome capture is gated on it — an unconsumed restore is a pending retry and reports nothing. `DEFERRED_LINK_RESTORED` keeps its documented "once per fresh install" contract.
- The android `referrer: null` test moved to the iOS empty-clipboard case: with the hardening, an *organic* android install arrives as Play's `utm_source=google-play&utm_medium=organic` string (already covered by the `marker_missing` test), while `null` means the read failed. A new test pins the silent-retry behaviour.

`deferred-link` + `useNativePlugins` suites: 34 passed. `mergeable` is green.

kushagrasarathe (approved):

innolope-dev:
Superseded by the dev-based re-author (new PR references this one).

PR #2588, opened 2026-07-30
perf: fix Android WebView jank (confetti, instrumentation, wrapper)

innolope-dev:
## Why

The Android (Capacitor) build feels jittery vs the same app as a PWA — confetti on reward claim stutters on a Pixel 8 Pro, and low-end devices (Realme Narzo 50) see random slowness on navigation and camera open. Investigation found multiple independent causes rather than one:

1. **Confetti**: up to 400 star particles rendered on the main thread on the qr-pay success screen (payment-success + perk-claim bursts stack), no worker, no native scaling.
2. **Instrumentation**: PostHog rrweb recording was *enabled* on capable devices (the Pixel class), autocapture walked the DOM on every tap, Sentry BrowserTracing instrumented 100% of sessions (sampling only gates sending), and reactComponentAnnotation stamped every DOM node for a replay product that isn't running (Sentry replay was dead config — `replayIntegration()` was never registered).
3. **Wrapper**: MainActivity replaced Capacitor's WebViewClient with a bare one that dropped `onRenderProcessGone` (renderer OOM = app death, reads as 'random freeze'), pulltorefreshjs put non-passive touch listeners on `body` (every touch blocked compositor scrolling on the main thread), release APKs shipped ~23MB of per-read-inflated JS with remote debugging left on, and the QR scanner decoded 8×/sec on the app-shell critical path.

## What

- **Confetti**: worker-backed OffscreenCanvas rendering, single half-sized burst on native, 1.5s duplicate-burst throttle, reduced-motion support, lazy import.
- **PostHog**: recording + autocapture off on native (explicit `posthog.capture` events unchanged — heads-up to analytics owners). **Sentry native**: BrowserTracing dropped, console.warn capture dropped; error capture untouched. **Sentry web**: dead replay config deleted. GA no longer ships into the WebView.
- **Android**: MainActivity now extends `BridgeWebViewClient` (all 8 callbacks restored, incl. renderer-crash recovery; also picks up the OTA-aware `findPageHtml` from mobile-release so post-OTA navigation can't serve stale APK HTML), `noCompress` for web assets, `NODE_ENV=production` in the release script (kills `webContentsDebuggingEnabled` in release; `WEBVIEW_DEBUG=true` overrides for profiling), splash `backgroundColor`, ~8MB of web-only assets pruned from the bundle.
- **UI**: pull-to-refresh rewritten with passive listeners + compositor-only indicator (pulltorefreshjs removed), QR scanner lazy-loaded and tuned (4Hz, no outline overlay, 300ms close), Crisp native SDK configured on first support open instead of every launch, Capgo update check deferred past first paint.
- **Testing**: `NEXT_PUBLIC_PERF_BARE=true` builds strip all instrumentation, for A/B-ing a bare APK against a control on the affected devices.

## Verification

- `pnpm typecheck` ✅, full jest suite 182/182 suites ✅, `:app:compileDebugJavaWithJavac` ✅
- Needs on-device regression before release (deep links, dynamic-route `_` fallbacks, offline nav, pull-to-refresh feel, QR scan, support chat open) — profiling protocol and before/after metrics tracked internally.

## Notes

- All changes ride the next Play release; the JS-only commits are additionally OTA-able via Capgo afterward.
- Left alone on purpose: `minifyEnabled` (R8 shrinks Java only, plugin-reflection risk), flag sprites (runtime-resolved by country code).


innolope-dev:
Closing: 4 of 5 commits landed on dev individually; the remaining Capgo log-level commit was re-authored onto dev in the replacement PR (references this one).

PR #2590, opened 2026-07-30
fix(seo): submit only changed URLs to IndexNow, and fix the workflow

innolope-dev:
## The workflow has never run successfully

`.github/workflows/indexnow.yml` resolves Node from `node-version-file: '.node-version'`, but that file does not exist in this repo. Every production deploy since the workflow landed in February has failed at `actions/setup-node`, before `pnpm install`:

> `The specified node version file at: .../.node-version does not exist`

**0 successes out of 2750 runs** (348 failures, the rest correctly skipped preview deploys). IndexNow has never actually been pinged from CI. The key at `public/054e10e6239a45cb2d06e92d669f5b6f.txt` is in place, so nothing else was wrong — the job just never got that far.

## Behind that, it was pinging far too much

`deployment_status` fires on every successful **Production** deploy (~14/day), and the script rebuilt and submitted the entire URL list each time, unconditionally. IndexNow is for URLs that were **added, updated or deleted** — resubmitting an unchanged site burns the daily quota and is the pattern that gets a host's submissions deprioritised. Nothing changes on `/en/argentina` because a mobile jank fix merged.

## And the URL list had drifted from the sitemap

`scripts/ping-indexnow.ts` kept its own copy of the page list. It had fallen behind `sitemap.ts` by 205 URLs — missing blog posts, stories, use-cases, withdraw, deposit rails, pricing, supported-networks and the content hub. Those are exactly the new-article pages worth submitting.

## Changes

- **`node-version: '20'`** (matching `tests.yml`) so the job runs at all.
- **URLs come from `sitemap.ts`** instead of a second, drifting copy — 916 → 1121 URLs, and no way to drift again.
- **Delta by default.** The sitemap's URL set is diffed against the set submitted last time (kept in the Actions cache), and content files changed since then are mapped back to the pages they render. Nothing to submit → no API call.
- **Skip the install** when nothing feeding the sitemap changed (content submodule, `src/data/seo`, `sitemap.ts`, `src/i18n`, `src/lib/{content,blog}.ts`).
- **`workflow_dispatch` gains a `full` input** for a deliberate full resubmission.
- Sitemap URLs are rewritten onto the production origin — `BASE_URL` is env-dependent, and a stray `NEXT_PUBLIC_BASE_URL` would otherwise submit preview URLs under `peanut.me`.
- `concurrency: indexnow` serialises overlapping deploys so they can't race on the cache.

## Effect

| deploy | before | after |
|---|---|---|
| no content change | 916 URLs | none — job skips before install |
| one blog post added | 916 URLs (post not even included) | 5 (one per locale) |
| help article + corridor edited | 916 URLs | 10 |
| first run / cache evicted / manual `full` | 916 URLs | 1121 (full, deliberate) |

## Verification

Ran locally against a mock IndexNow endpoint, and the workflow's resolve step against real git history:

- bootstrap with no cached state → full 1121
- immediately after → `Nothing changed — skipping IndexNow submission.`
- 5 blog URLs removed from cached state → submits exactly those 5
- `content/help/passkeys/*.md` + `content/send-to/argentina/from/spain/en.md` changed → submits exactly `/{locale}/help/passkeys` and `/{locale}/send-money-from/spain/to/argentina`, 10 URLs
- singleton `content/help/en.md` matches `/{locale}/help` only, not all 37 help articles
- resolve step returns `run=false` when nothing changed, and `run=true` for commit `032683ba9` ("register Faster Payments + SPEI as deposit rails") — a data-only SEO change that adds URLs without moving the content submodule
- `pnpm typecheck` clean; `prettier --check` clean

The first run after merge will submit the full sitemap once (no cached state), which is correct — it's also the first ping this site has ever sent.

## Summary by CodeRabbit

* **New Features**
  * IndexNow submissions now support full-site mode or changed-content-only mode.
  * Added a manual option to submit every sitemap URL.
  * Automatically detects newly added and updated pages for targeted notifications.

* **Bug Fixes**
  * Prevented overlapping IndexNow runs from causing submission-state conflicts.
  * Improved behavior when prior submission history is unavailable.
  * Failed request batches no longer abort the entire run; errors are reported after processing.

* **Performance / Reliability**
  * Skips submissions when no sitemap URLs have changed and batches requests for more dependable delivery.

* **Chores**
  * Updated ignore rules for persisted IndexNow submission state.

### DIFF
--- a/.github/workflows/indexnow.yml
+++ b/.github/workflows/indexnow.yml
@@ -56,13 +56,24 @@ jobs:
                       exit 0
                   fi
 
+                  # checkout clones submodules at the superproject's depth; the commit we last
+                  # submitted for can sit outside it, so take the whole content history.
+                  if [ "$(git -C src/content rev-parse --is-shallow-repository)" = 'true' ]; then
+                      git -C src/content fetch --unshallow --quiet || true
+                  fi
+
                   prev_content_sha=$(jq -r '.contentSha // empty' "$state")
                   if [ -n "$prev_content_sha" ] && git -C src/content cat-file -e "$prev_content_sha^{commit}" 2>/dev/null; then
                       changed=$(git -C src/content diff --name-only "$prev_content_sha" "$content_sha")
+                      force_full=false
                   else
-                      echo "Previous content commit unavailable — new URLs are still detected via the sitemap diff."
+                      # Without the diff we cannot tell which existing pages were edited, and the
+                      # sitemap diff only surfaces new URLs. Resubmit everything rather than skip.
+                      echo "Previous content commit unavailable — falling back to a full submission."
                       changed=""
+                      force_full=true
                   fi
+                  echo "force_full=$force_full" >> "$GITHUB_OUTPUT"
 
                   {
                       echo 'changed_files<<EOF'
@@ -81,7 +92,7 @@ jobs:
                   fi
 
                   if [ -z "$changed" ] && [ -z "$sources_changed" ] && [ "$prev_content_sha" = "$content_sha" ] \
-                      && [ "${{ inputs.full }}" != "true" ]; then
+                      && [ "$force_full" != 'true' ] && [ "${{ inputs.full }}" != 'true' ]; then
                       echo 'Nothing that affects the sitemap changed since the last submission — skipping.'
                       echo "run=false" >> "$GITHUB_OUTPUT"
                   else
@@ -105,7 +116,7 @@ jobs:
               env:
                   # Not a secret by design — IndexNow requires it to be served at public/<key>.txt.
                   INDEXNOW_KEY: '054e10e6239a45cb2d06e92d669f5b6f'
-                  INDEXNOW_FULL: ${{ inputs.full }}
+                  INDEXNOW_FULL: ${{ inputs.full == true || steps.changes.outputs.force_full == 'true' }}
                   INDEXNOW_CHANGED_FILES: ${{ steps.changes.outputs.changed_files }}
                   INDEXNOW_CONTENT_SHA: ${{ steps.changes.outputs.content_sha }}
               run: pnpm tsx scripts/ping-indexnow.ts
--- a/.github/workflows/indexnow.yml
+++ b/.github/workflows/indexnow.yml
@@ -6,9 +6,19 @@ on:
     deployment_status:
     # Allow manual trigger
     workflow_dispatch:
+        inputs:
+            full:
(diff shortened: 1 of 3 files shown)


innolope-dev:
Both CodeRabbit findings addressed in c7d84d64e.

**`submit()` didn't catch `fetch` rejections** — real. A transport error or the 30s abort firing escaped the loop, skipping remaining batches with a stack trace. Now caught and counted like an HTTP failure. Kept the non-zero exit deliberately: it leaves the state file unwritten, so the next run retries those URLs rather than recording them as submitted.

**Unresolvable previous content commit** — real, though narrower than described: the sitemap diff still catches *new* pages, so what was lost was re-pings for *edited* existing pages in that window. Fixed at the root as well as the symptom:

- `actions/checkout` clones submodules at the superproject's depth, so the commit we last submitted for could fall outside it. The resolve step now unshallows `src/content` first, which makes the diff resolvable in practice.
- If it still can't resolve, `force_full=true` now ORs into `INDEXNOW_FULL` and we resubmit everything instead of skipping — as suggested.

Verified against real git history: unresolvable `contentSha` → `force_full=true run=true`; content moved 5 commits → `force_full=false run=true` with 33 changed files; nothing changed → `run=false`. Transport failure → `Batch 1: request failed`, exit 1, no state written. Delta behaviour unchanged (bootstrap 1121, no-op 0, new page + two edits 15, forced full 1121). `typecheck` and `prettier --check` clean.

The `eslint` failure is the pre-existing red baseline on `main` — no hits for `ping-indexnow` anywhere in that log.

0xkkonrad:
Closing as superseded by #2671, which rewrote this workflow and script wholesale. Your diagnosis here was right and valuable — the job had 0 successes in 2750 runs because setup-node resolved a `.node-version` file that doesn't exist — and #2671 fixes that (Node 22 pinned directly, pnpm setup corrected). The goal of this PR (actually getting root-sitemap URLs submitted) is delivered by commit 8faede4d on #2671: the run is gated on the root sitemap only, with Split URLs gated separately on the split-sitemap advertisement. After #2671 merges, one `workflow_dispatch` with `bootstrap=true, dry_run=false` records the baseline and the cron takes over. If #2671 fails to land for any reason, reopen this.

0xkkonrad:
Follow-up: #2671 will not land. Keep this PR closed: the Split integration was rolled back and PeanutSplit now owns its app and SEO at https://peanutsplit.com. The root-site IndexNow diagnosis in this PR remains valid, but it is saved as a standalone Peanut UI TODO and should return only as a fresh root-site-only PR.

PR #2592, opened 2026-07-30
Localize the marketing site: three locales, localized landing, language switcher

innolope-dev:
## Why

App UI localization shipped in #2447 (`next-intl`, `en / es-419 / pt-BR`); marketing runs a separate URL-driven i18n system that had drifted: a bloated locale set generating ~450 near-duplicate indexed URLs, and a landing page with zero i18n. The prose itself was already ~95% translated in the content submodule — the work is the locale set, the landing page, the leaks, and the SEO plumbing.

Since the first revision, the target locale set changed: **es-AR is back in** as a first-class locale (deltas-over-es-419 in the app, own routes on marketing), while es-ES stays retired. The near-duplicate problem that motivated removing es-ar is solved structurally instead (content-aware hreflang/sitemap + fallback canonicals), so a sparse locale no longer pollutes the index.

Localizes the marketing site and landing page, now with the full target locale set: **en · es-419 · es-AR · pt-BR** (es-ES stays retired, 301 → es-419).

## What's in here

**Locale set**
- Marketing locales: `en`, `es-419`, `es-ar`, `pt-br` — es-AR restored as a first-class locale (own landing route, catalog, switcher entry, hreflang `es-AR`). Content fallback `es-ar → es-419 → en`: ~34/240 slugs have real AR prose today; the rest serve es-419 under the es-ar URL without being advertised to crawlers (see SEO below).
- App locales: `en`, `es-419`, `es-AR`, `pt-BR`. `es-AR.json` is **deltas-only** (voseo/Argentine terms) layered over es-419 via deepMerge — shared keys are never re-translated. Parity test asserts the delta file stays a subset of `en` while the loader output stays complete.
- An Argentine device resolves to `es-AR` (app) / `es-ar` (marketing); the shared `app-locale` cookie carries the choice both ways.

**Landing language UI**
- The `LocaleSuggestion` banner is switched off (kill switch in the component; mounts stay wired for re-enable). The footer `LocaleSwitcher` stays.
- First visit now respects the browser locale: the proxy redirects `/` per `Accept-Language` (q-ordered, explicit English wins) and pre-sets the cookie. **Crawlers are exempt** (UA check + missing-UA treated as bot), so bots always see the English `/` and hreflang keeps routing them — the concern that made the previous revision cookie-only.
- Locale-cookie redirect now **preserves the query string** (UTM/campaign params survive), and the promo-link branch runs before it, so `?promo=…&id=…` links work for cookie'd users.

**SEO**
[...]

PR #2593, opened 2026-07-30
Give /privacy and /terms the marketing page header

innolope-dev:
Stacked on #2592 — review that one first; this branch targets it, so the diff here is just the two page components plus two catalog strings.

## Problem

`/en/privacy` and `/en/terms` render as bare prose with no header, unlike every other marketing page.

The pink hero and yellow marquee come from the MDX `<Hero>` component. Help articles get it because their **content markdown** contains a `<Hero title="…" />` tag, and so do the five card-* legal pages (`card-esign`, `card-privacy`, `card-terms-us`, `card-terms-international`, `card-prohibited-activities`). Only `privacy` and `terms` never had one.

## Why the fix is in the page component, not the content

Legal markdown is `content_mode: verbatim` and authored upstream in `peanutprotocol/mono`. `content/legal/README.md` is explicit that content generation must never overwrite those files, and `peanut-ui/src/content` is a read-only mirror — anything written there is reverted by the next mirror run.

Rendering `<Hero>` from `privacy/page.tsx` and `terms/page.tsx` sidesteps that entirely and applies to every locale without touching legal text.

## Localization

- **Title** comes from `mdxSource.frontmatter.title`, with the `| Peanut` suffix already stripped by the existing `displayTitle`. Legal is English-only today (translations are gated on legal review), so it renders English via the fallback chain — and picks up translations automatically if they ever ship, with no further code change.
- **Subtitle** is a new catalog string, translated for all three locales:
  - privacy — "How we handle your data" / "Cómo tratamos tus datos" / "Como tratamos seus dados"
  - terms — "The agreement between you and Peanut" / "El acuerdo entre tú y Peanut" / "O acordo entre você e o Peanut"

## Screenshot
image-1785440863769

## Verification

`tsc --noEmit` clean, `prettier --check .` clean, production build green. Hero confirmed rendering on `/en/privacy`, `/en/terms`, `/es-419/privacy` and `/pt-br/terms`.

Note: stacked PRs in this repo don't always trigger the full CI matrix, since some workflows only fire for `main`/`dev`-based PRs. If checks look sparse here they should be kicked manually rather than read as green.


innolope-dev:
Pushed `deb6d5db7` — fixes the duplicate `<h1>` this PR introduced on /privacy and /terms: the new `<Hero title>` sat above verbatim markdown that already opens with `# Privacy Policy` / `# Terms of Service`, so the same string rendered in two h1s. Since the content mirror is read-only (`content_mode: verbatim`), the fix is render-time: an opt-in **positional** `stripLeadingH1` remark plugin in `renderContent` — it removes only the FIRST node when it's an h1 (the bodies use `##` for sections; a blanket strip would swallow authoring signals), enabled on exactly these two pages. The card-* legal pages already follow the Hero-instead-of-h1 convention and are untouched.

Also merged the updated base (`feat/marketing-localization` — es-AR restore) and added the two legal Hero subtitle keys to the restored es-ar catalog.

Verified in the browser: exactly one h1 on /en/privacy, /en/terms and /es-419/privacy. Reminder for merge: this PR is stacked — after #2592 lands, retarget the base to dev in the UI (no force-push needed).

PR #2596, opened 2026-07-31
feat: add NITA badge for the Nita Cervio creator collab

0xkkonrad:
## Summary

Frontend half of the `NITA` badge — a paid creator collab with Nita Cervio. Anyone who signs up through her link gets the badge, which also skips the card waitlist. The backend half (badge code, whitelists, waitlist skip) is https://github.com/peanutprotocol/peanut-api-ts/pull/1263 — **merge and deploy that one first.**

Locked by Konrad — title, description, and artwork are final:

| | |
|---|---|
| Badge code | `NITA` |
| Campaign slug | `nita` |
| Title | Nita's Recommendation |
| Description | Nita sent you. You skipped the card line on her word. |

Three changes:

1. **`public/badges/nita.svg`** — the approved artwork. Named for the badge code, lowercased, matching the neighbours (`arbitrum.svg`, `dunbar.svg`, `verified.svg`).
2. **`src/components/Badges/badge.utils.ts`** — the `NITA` entry in `BADGES`. This is load-bearing: the backend's `syncBadgeDefinitions` does not set `iconUrl`, so without it the awarded badge renders the Peanutman fallback and the raw backend name. `displayName` is set so the locked title renders even before the backend seeds the definition.
3. **`src/components/Invites/campaign-maps.ts`** — `nita: 'NITA'` in `UTM_CAMPAIGN_TO_BADGE_MAP`.

## Why `UTM_CAMPAIGN_TO_BADGE_MAP` and not the invite-code map

Nita's link is `peanut.me/invite?code=<her personal invite code>&campaign=nita`.

`resolveCampaigns` reads three sources. Source 1 is the explicit `?campaign=` / `?campaignTag=` param, mapped through `UTM_CAMPAIGN_TO_BADGE_MAP` before anything else. That is where `nita` arrives, so that is the map it belongs in. One entry also covers `?utm_campaign=nita` (source 3), because both param spellings read the same map — so a UTM-shaped variant of her link works without a second entry.

`INVITE_CODE_TO_CAMPAIGN_MAP` (source 2) is keyed on the **invite code itself**, for links shaped `?code=nita`. Her invite code is her own personal code, not the string `nita`, so an entry there would never be read. It would also be actively wrong: `/invite?code=nita` runs `validateInviteCode('nita')`, which fails unless a real invite code `nita` exists, and the visitor lands on the "Invalid Invite Code" screen. A stale entry in that map is also how campaign attribution gets hijacked when an unrelated code later matches.

`campaign-maps.test.ts` asserts every value in both maps exists in `BADGES`, so change 3 depends on change 2. Both are in this PR.

## Open question for Konrad — returning users

**This is a product decision, not a code gap. Nothing here blocks the merge; the primary path works.**

Her link carries an invite code, so `classifyBareCampaigns` treats it as **not bare** and the normal invite-validation path owns the flow. What each visitor gets today:

| Visitor | Outcome |
|---|---|
| New signup | ✅ Badge + skip, and Nita gets the referral. `handleClaim` saves the campaign tag to a cookie, `useZeroDev` redeems it on registration. |
| Returning user, logged in, has app access | ✅ Badge + skip. The auto-claim effect fires `/badge/award` for every resolved campaign. No referral credit — they already have an account. |
| Returning user, logged in, **still app-waitlisted** | ❌ Nothing. `isInviteAutoClaim` needs `hasAppAccess`, so the effect returns early and the CTA sends them to `/setup?step=signup` — a dead end for an account that already exists. |
| Returning user, logged out, taps the primary CTA instead of "Log in" | ❌ Nothing. The cookie is set, but only the **registration** path reads it. Log in afterwards and the tag is never redeemed. |

Nita's audience is ~107k and some already have Peanut accounts, so the two ❌ rows are real people.

Note what a **bare** `?campaign=nita` link (no `code=`) does today: nothing at all. `nita` is in neither `WAITLIST_SKIP_CAMPAIGNS` nor `BARE_VANITY_CAMPAIGNS`, so `classifyBareCampaigns` returns not-claimable, `isDeadBareLink` fires, and the visitor is silently redirected to `/`. If a shortener or a repost strips the `?code=`, the link dies quietly.

The available lever is adding `nita` to `WAITLIST_SKIP_CAMPAIGNS`. That makes a bare `?campaign=nita` link auto-claim for any logged-in user regardless of `hasAppAccess`, bypass the invalid-invite gate, and show skip-the-waitlist copy — which matches the badge description, and the backend PR already has `NITA` in `POSTLAUNCH_SKIP_BADGE_CODES`.

What it costs:

- **No referral attribution on that link.** No invite code means no inviter link, no `invites_accepted` for Nita, and no `invite_code` on the PostHog event. If she is paid or measured on referrals, a bare link is invisible.
- **Uncapped and shareable.** A bare link needs no invite to redeem, so anyone who reposts the URL grants themselves the badge and the card-waitlist skip.
- `WAITLIST_SKIP_CAMPAIGNS` must stay in sync with peanut-api-ts `SKIP_BADGE_CODES`.

Three ways to go, pick one:

1. **Ship as-is (this PR).** One attributed link. New signups convert; returning users in the two ❌ rows get nothing.
2. **Two links.** Keep the attributed link for new signups, give Nita a bare `?campaign=nita` link for her existing-user audience. Best coverage, but the bare one is uncapped and unattributed, and she has to use the right link with the right people.
3. **Bare link only.** Everyone gets the badge, Nita gets no referral data.

I did not pick — it trades her attribution against her existing-user reach, and both sides are yours.

## Design notes

- `displayName` duplicates the backend `name` in peanut-api-ts#1263 rather than overriding it. Kept deliberately: it pins the agreed title on the FE side, and the same redundancy already exists for `MANICERO`. `path` is the field that actually has to be here — `syncBadgeDefinitions` never sets `iconUrl`.
- No new test was strictly required (the suite is parametrized over `UTM_CAMPAIGN_TO_BADGE_MAP`, so `nita` is auto-covered), but the production URL shape — personal invite code plus campaign tag — was not asserted anywhere. Added, since that exact string goes to a paid creator.

## Risks

- **Cross-repo deploy order.** This PR alone renders the badge; it does not award it. `NITA` must exist in peanut-api-ts (`BADGE_CODES`, `ALLOWED_BADGE_CODES`, `INVITE_FLOW_ALLOWED_BADGES`, `POSTLAUNCH_SKIP_BADGE_CODES`) or `/badge/award` 400s and the invite-accept path logs a non-whitelisted-badge warning. peanut-api-ts#1263 covers all four. **Merge and deploy it first.** Shipping this one early is harmless — an unused map entry.
- Nothing else on `/invite` changes. Adding one key to a lookup table cannot affect any other campaign.
- Pre-existing, not touched here: `campaign-maps.ts` looks up `UTM_CAMPAIGN_TO_BADGE_MAP[utmCampaignParam]` without `.toLowerCase()`, unlike the explicit-param branch above it. Harmless in-app because `InvitesPage` lowercases at the call site, but a direct caller passing `?utm_campaign=NITA` gets nothing. Follow-up, not this PR.

## QA

1. Backend deployed with `NITA` whitelisted.
2. New account: open `/invite?code=<nita's code>&campaign=nita`, tap the CTA, complete signup. Profile shows "Nita's Recommendation" with the pink artwork, and the card waitlist is skipped.
3. Existing account with app access, logged in: open the same URL. Auto-claims and lands on `/home` with the badge.
4. Rendering only, no backend: `/dev/share-builder` and `/dev/debug` iterate `BADGE_CODES`, so `NITA` shows up there with the right icon and title.

Screenshots: N/A — no visible change to any existing surface. The new badge only renders once the backend awards the code, which is not yet deployed.


0xkkonrad:
@coderabbitai review

0xkkonrad:
@coderabbitai review

0xkkonrad:
Replaced by https://github.com/peanutprotocol/peanut-ui/pull/2597, which cuts the same two commits against `main`.

Nita starts posting within days, so the badge has to be on production rather than wait for the next `dev` release. Retargeting this PR in place was not viable — `dev` is 118 commits ahead of `main`, so the base flip would have turned a 4-file change into a full release PR.

#2597 carries the same diff (4 files, 41 insertions) and the same open returning-user question. Closing this one so the change cannot land twice.

PR #1249, opened 2025-09-24
[TASK-14942] feat(manteca-deposit): fix manteca deposit

jjramirezn:
Video Demo:

- It took a little longer than previous tests hmmmm
- Status is updated via webhook and manteca only supports one webhook (right now pointed at staging) so thats why the history doesn't update

https://github.com/user-attachments/assets/522df697-df01-44af-8990-059de20aab8e

PR #2604, opened 2026-08-04
fix(i18n): localize backend-sourced status strings across all domains

innolope-dev:
## Why

Backend-sourced status strings showed in English for es-419 / pt-BR users — a transaction reading "pending", KYC status text, reason messages. The audit found the raw-English problem was narrower than expected (KYC status rendering, rail status and card status were already keyed or never rendered as text), but the sites that did leak were high-traffic: the shared `StatusBadge` (transaction drawer, receipts, KYC drawer, success views) had nine hardcoded English literals plus a raw-passthrough default — sitting in a directory excluded from the `jsx-no-literals` guard, which is exactly how it shipped unnoticed.

## What

Everything keys off **stable enums/codes, never backend prose** (#2554's rule), following the existing `TYPE_LABEL_KEYS` / `rejectLabelCode()` patterns:

- **`StatusBadge`** → exhaustive `STATUS_LABEL_KEYS` (`as const satisfies Record<StatusType, string>`) rendered via `t()` under `common.status.*`; unknown values collapse to `status.unknown` instead of rendering the raw string. The `Badges/**` lint ignore is removed. (Backend statuses were already normalized to the FE union by `mapEntryStatusToUiStatus` — no wire changes needed.)
- **`PerkRewardReceipt`** — the one genuine raw `{status}` interpolation in product code, now translated through the same table.
- **Reaper fail copy + intent-strategy display names** (`REAPER_FAIL_COPY`, 'Sent via Link', 'Card payment', …) — the transformer emits a `nameKey` (+ ICU params for dynamic parts like usernames and bank names, which are data, not copy); render sites translate. The English-literal `displayName` comparisons in `TransactionDetailsHeaderCard` are replaced with discriminants (the `isTestTransaction()` precedent) so translating the labels can't break the logic.
- **`CapabilityReason.code` → key table** with a `rejectLabelCode()`-style fallback, covering the KYC/card/region reason-message sites (ApplicationStatusScreen, InitiateKycModal, UnlockedRegions, useWaitingOnProviderModal, ActivationCTAs). Where the wire carries no code (`identity.actionMessage`, Rain dispute prompts) the backend prose remains as fallback — adding those codes is an api-ts follow-up (#1249 pattern).
- **Prop-literal stragglers** keyed: ActivationCTAs spend prompts, the 'Adjustment' card row, the PIX maintenance badge, Sumsub error fallbacks (server actions now also return a `code`, translated client-side — no server-side next-intl in this app).

Out of scope, documented: `/receipt/[entryId]` server-component titles (structurally blocked on server-side next-intl).

## Verification

- Full jest: 183 suites, 2,416 passed (includes a regression the suite caught during development: `customText=''` means "no override" — truthiness preserved).
- `tsc --noEmit` clean; eslint on all touched files: 0 errors; catalog parity + ICU compilation tests green for all three locales.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -235,7 +235,7 @@ module.exports = [
             // fallback reached through the shared 0_Bruddle/Button.
             'src/components/Global/{Layout,AnimateOnView,MarqueeWrapper,FAQs,FooterVisibilityObserver,ExchangeRateWidget,Modal,Loading}/**',
             'src/components/Global/{Layout,AnimateOnView,MarqueeWrapper,FAQs,FooterVisibilityObserver,ExchangeRateWidget,Modal,Loading}.tsx',
-            'src/components/Global/{PeanutLoading,Icons,Badges}/**',
+            'src/components/Global/{PeanutLoading,Icons}/**',
             // InvitesGraph is a /dev-only debug visualization, not user-facing UI.
             'src/components/Global/InvitesGraph/**',
             // Hidden support tool — never linked in-app; support DMs the URL to
--- a/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
+++ b/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
@@ -11,7 +11,7 @@ import { formatAmount } from '@/utils/general.utils'
 import { countryData } from '@/components/AddMoney/consts'
 import { useAuth } from '@/context/authContext'
 import { useCapabilities } from '@/hooks/useCapabilities'
-import { resolveKycModalVariant, getGateUserMessage } from '@/utils/capability-gate'
+import { resolveKycModalVariant, getGateUserMessage, getGateReasonCode } from '@/utils/capability-gate'
 import { useModalsContext } from '@/context/ModalsContext'
 import { useCreateOnramp, GENERIC_ONRAMP_ERROR } from '@/hooks/useCreateOnramp'
 import { useParams, useRouter, useSearchParams } from 'next/navigation'
@@ -511,6 +511,7 @@ function BridgeBankOnrampPage() {
                     error={sumsubFlow.error}
                     variant={resolveKycModalVariant(gate)}
                     providerMessage={getGateUserMessage(gate)}
+                    reasonCode={getGateReasonCode(gate)}
                     regionName={selectedCountry?.title}
                 />
 
--- a/src/app/(mobile-ui)/card/page.tsx
+++ b/src/app/(mobile-ui)/card/page.tsx
@@ -655,13 +655,14 @@ const CardPage: FC = () => {
                         </div>
                     )
                 }
-                const cardRailReason = poaSubmitted
-                    ? 'We received your proof of address — it’s being reviewed.'
-                    : railsForProvider('rain')[0]?.reason?.userMessage
+                const cardRail = railsForProvider('rain')[0]
+                const cardRailReasonCode = poaSubmitted ? 'proof_of_address_review' : cardRail?.reason?.code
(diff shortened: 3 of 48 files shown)


innolope-dev:
Follow-up pushed: KYC action messages now key off `identity.status` (the backend's `actionMessage` is a pure function of status — the discriminant was already on the wire, no API change needed), and `DocsLink` re-points `/en/…` hrefs at the app locale's marketing twin, covering all ~15 app-side docs/help links incl. the wallet nav. Rain's dispute `evidenceRequestedMessage` stays verbatim deliberately: it's case-specific free prose from the processor's dispute agents — no enum exists to key off. Full suite: 184 suites / 2,419 passed.

kushagrasarathe (approved):
## Agent review (pr-review rulebook) — full-diff read, all four checklists

**Scope reviewed:** all 48 files (+805/−137), plus the backend code this PR's claims depend on (`actionMessageFor`, the capability resolver, the self-heal classifier).

### What checked out

- **`actionMessageFor` is genuinely a pure function of status** (api-ts `identity.ts:81`) — gating on presence and rendering catalog copy is safe; the updated KycStates test encodes the contract.
- **Catalog parity:** identical key sets across en / es-419 / pt-BR; every new key present in all three.
- **`REAPER_FAIL_KEYS` exactly mirrors `REAPER_FAIL_COPY`** (9/9) with a safe generic fallback.
- **English-literal discriminants eliminated, not just translated** — `nameKey === TRANSACTION_NAME_KEYS.sentViaLink` replaces string comparison; zero literal comparisons remain at head.
- **eslint `Badges/**` ignore removal is safe** — the directory contains only StatusBadge, now literal-free.
- **`customText=''` truthiness contract preserved**; raw `userName` kept for data uses (test-tx marker, profile URL).
- **No deploy-order dependency:** `reason.code` already ships from the resolver on `dev`; unknown codes fall back to BE prose.
- **Security:** no new routes, deps, or secrets; diff matches the description exactly.

### Finding (fixed in `121117b`)

`REASON_CODE_KEYS` mapped `document_rejected` to a generic catalog line, but the backend only emits that code with the self-heal classifier's **fixable-specific** prose ("Your ID photo was blurry. Please upload a clearer photo." — `classifier.ts`, preferred at `resolver.ts:1346`). The generic line masked the fix instruction. Resolved by removing the mapping so those sites keep the BE prose, until the classifier's stable `action` code (`REUPLOAD_ID`, …) reaches the wire — api-ts follow-up, #1249 pattern.

Kept as-is (accepted): requirement-code entries stay mapped — their catalog copy is byte-identical to the resolver's `messageForRequirement` prose, so localization is a pure win in the common case; the rare fixable-classifier override on those codes is covered by the same api-ts follow-up. `region_block` also stays mapped — the BE's state/country/region word choice is lost, but the loss is cosmetic and localization wins.

### Smell flags (fifth-bug rule, advisory)

Fix-commit density in the last 20 commits: `useSumsubKycFlow.ts` 12/20, `transactionTransformer.ts` 9/20, `card/page.tsx` 8/20, `ActivationCTAs.tsx` 8/20. The KYC-flow hook and the transformer are standing deep-dive candidates.

### Verdict

```
TIER: T0 (external author; spans T2 KYC-gating surfaces)
PAIRED PR: none required — api-ts follow-up (classifier action codes on the wire) named, no deploy ordering
A. BREAKAGE:  pass — document_rejected masking fixed in 121117b
B. PERF:      pass — static table lookups, no new log/alert emission
C. QUALITY:   pass with smell flags (fifth-bug rule, above)
D. SECURITY:  pass — diff-vs-description exact match
TASK LINK:    waived by reviewer
VERDICT:      approve
```

Reviewed and approved by @kushagrasarathe (human, accountable); agent review advisory. Merge once CI is green on `121117b`.

PR #2606, opened 2026-08-04
fix: real anchor for 'Open Peanut app' — untrusted synthetic clicks skip WebAPK handoff

abalinda:
## Summary

Follow-up on the TASK-21050 journey (sibling of #2603). With the middleware loop fixed, logged-out Android users reach the "Open Peanut app" screen — but tapping the button re-opened **the same screen in a browser tab** instead of the installed PWA, boomeranging them (reporter session: 4 identical `/setup` loads at ~1.4s per tap, then a manual in-browser login to escape).

The button fabricated an anchor and called `.click()` — an **untrusted synthetic event**. The popup opens (transient activation covers that), but the WebAPK handoff never engages. Fix: a real `<a href="/setup" target="_blank" rel="noopener">` the user taps directly — the canonical trusted-link path Chrome's WebAPK link capturing is built around — styled with the exact class string Bruddle `Button` composes for this usage (in-repo precedent: `MarketingHero.tsx` uses the identical anchor+`btn`-classes pattern).

**Honesty note on the mechanism:** adversarial review traced Chromium's `ExternalNavigationHandler` and could not conclusively pin the old failure on event trustedness (the gesture gate keys off transient activation, which the synthetic click had). The real anchor is strictly-not-worse in every browser and is the documented capture path — but it is a bet. That's why the tap is now instrumented: `pwa_open_app_clicked` (device_type tagged). **Repeat fires within one session = a device where capture is still not engaging.**

`/setup` as the target is deliberate and load-bearing: the native Android app's App Links manifest (`android/app/src/main/AndroidManifest.xml`) verifies `/home`, `/card`, `/claim` etc. but **excludes `/setup`**, so the WebAPK remains the *sole* verified intent handler for it — the condition `launchWebApkIfSoleIntentHandler` requires. Pointing at `/home` would kill the handoff on any device that also has the native beta installed.

## Risks / breaking changes

- Visual: same class recipe Button emits (verified against `tailwind.config.js` `.btn`/`.btn-purple`/`.btn-shadow-primary-4`); `justify-center`/`no-underline` are no-ops kept for clarity. Cosmetic deltas: no `notranslate` guard and no haptic on this one control (minor, noted).
- If capture still declines (e.g. Firefox/Samsung Internet, no WebAPK): opens `/setup` in a new tab — same boomerang as before, not worse, now measurable.
- **Dead-end shape is retained deliberately**: a "Continue here instead" fallback was added in `6cf5d5b77` and reverted next day in `5413e9dbe` ("keep automatic PWA open behavior"); re-adding it was offered and descoped by Hugo for this PR ("ship 1"). Receipts left here so the reviewer decides with history in hand.
- Hotfix to `main` → back-merge main→dev debt (stacked with #2603's).

## QA

- **Device test required** (link capturing is unverifiable headless/CI): Android Chrome, PWA installed, logged out of both → `peanut.me/setup` → tap **Open Peanut app** → expect the WebAPK to open (landing/login inside the app), not a new browser tab of this screen.
- Post-deploy signal: PostHog `pwa_open_app_clicked` — sessions with >1 fire = handoff still failing on that device profile.
- `typecheck` + prettier clean; no unit test (browser-handoff behavior is not reachable from jsdom/Playwright).

## Screenshots

N/A — pixel-identical intent (same composed classes); the change is which element receives the tap.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved opening the installed Peanut app from the setup experience.
  * The “Open Peanut app” link now opens reliably in a new browser tab after a direct user tap.
  * Added analytics tracking for Open Peanut app interactions.

### DIFF
--- a/src/components/Setup/Views/InstallPWA.tsx
+++ b/src/components/Setup/Views/InstallPWA.tsx
@@ -180,6 +180,11 @@ const InstallPWA = ({
                         href="/setup"
                         target="_blank"
                         rel="noopener"
+                        // capture without preventDefault — repeat fires in one session
+                        // mean link capturing is still not engaging on that device
+                        onClick={() =>
+                            posthog.capture(ANALYTICS_EVENTS.PWA_OPEN_APP_CLICKED, { device_type: deviceType })
+                        }
                         className="btn btn-purple btn-shadow-primary-4 flex w-full items-center justify-center gap-2 no-underline transition-all duration-100 active:translate-x-[3px] active:translate-y-[4px] active:shadow-none"
                     >
                         Open Peanut app
--- a/src/components/Setup/Views/InstallPWA.tsx
+++ b/src/components/Setup/Views/InstallPWA.tsx
@@ -168,24 +168,22 @@ const InstallPWA = ({
                 return null
             }
 
-            // for other browsers, try to open the pwa in a new tab
+            // for other browsers, try to open the pwa in a new tab.
+            // Must be a REAL anchor: a synthesized link.click() fires an untrusted
+            // event, which Chrome excludes from WebAPK link capturing — the tap
+            // opened a plain browser tab on this same screen instead of the
+            // installed app (the TASK-21050 "Open Peanut app" boomerang). Only a
+            // trusted tap on an in-scope target="_blank" anchor hands off.
             return (
                 <div className="flex flex-col gap-4">
-                    <Button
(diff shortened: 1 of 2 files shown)


abalinda:
@coderabbitai review

abalinda:
@coderabbitai review

jjramirezn (approved):

PR #2613, opened 2026-08-05
feat(badges): consume the capability platform, generate the asset map

0xkkonrad:
## Outcome

Badge acquisition becomes badge-agnostic in the UI, and the frontend stops keeping its own badge catalog.

```text
?badge_campaign=<slug>  →  UI intent  →  POST /badge/claims  →  permanent BadgeAward
```

Four hand-maintained maps in `campaign-maps.ts` and a 48-entry asset map in `badge.utils.ts` are gone. Adding a badge no longer touches this repo except to drop in the SVG.

Paired with peanut-api-ts#1282 and mono#69. **Deploy this first**, so the generic asset/fallback path exists during the rolling deploy.

## What changed

**The asset map is generated, not written.** `badge.utils.ts` carried 48 hand-written `code → path` entries, a parallel catalog that could drift from the backend silently. It now reads `src/types/badge-assets.json`, generated from the API badge catalog.

That turns the existing "every fallback asset is checked in" test into the UI half of a cross-repo contract: peanut-api-ts CI asserts the manifest still matches its catalog, and this test asserts every path in it resolves to a real file. Adding a badge without its artwork fails CI in the repo that can fix it — without API CI needing a checkout of this repo, which would have forced the API to merge before the UI and inverted the rollout order.

**Campaign maps deleted.** `INVITE_CODE_TO_CAMPAIGN_MAP`, `UTM_CAMPAIGN_TO_BADGE_MAP`, `WAITLIST_SKIP_CAMPAIGNS` and `BARE_VANITY_CAMPAIGNS` are replaced by the canonical `badge_campaign` namespace plus the typed claim contract. The backend owns which campaigns exist, which are live, and what they grant.

**`utm_campaign` no longer mints badges.** Values are still forwarded, but the backend resolves only its own dated aliases — today just `offramp`, expiring 2026-09-07. A marketing link can no longer award a badge as a side effect.

**Presentation is API-driven** with a generic fallback, so an earned badge never disappears when local artwork lags the catalog.

## Merged with main

`main` moved 247 commits ahead of this branch's base and shipped two things that overlap it: a full i18n pass, and the root-domain/stackable-campaign work in `3e7dda652`, which solved the same problem this platform replaces. 21 conflicts. Note that GitHub does not run `pull_request` workflows while a PR is `CONFLICTING`, so CI had gone silent rather than red.

Four things would have been lost silently and are preserved:

- **Comma-separated campaign values** (`?campaign=a,b`). The new parser only read repeated params, so a published CSV link would have awarded just the first badge. Splitting happens at the URL boundary only — a stored queue identity may itself contain a comma and must round-trip unchanged.
- **42 per-badge `shareLine` strings**, now their own frontend-owned map. Share copy is first-person English presentation, deliberately not backend catalog data, which owns third-person descriptions.
- **The root-domain redirect's dead-bare-link guard**, retargeted at the badge campaign namespace.
- **Localized copy on `InvitesPage`.** The CTA, login label, title and description had reverted to hardcoded English. `main` localized all four, and the existing `invites` keys map exactly — a bare badge-campaign link is the localized vanity-claim case.

`useZeroDev` also keeps emitting the singular `campaign_tag` alongside the plural `campaign_tags`, so the `INVITE_ACCEPTED` funnel does not step at deploy.

## Verification

- Typecheck clean.
- Prettier clean.
- **2,820 tests passing across 222 suites.**
- CI fully green: `ci-success`, `e2e`, `eslint`, `format`, `Deploy-Preview`, CodeQL. No failures.
- OpenAPI snapshot and generated client regenerated from the paired API branch: badge claims no longer return 503, so the readiness error shape is gone.

## Known gap

The pending-campaign cookie queue keeps its epoch/generation machinery from the original branch. It works and is covered by tests, but it is more apparatus than the problem needs. Simplifying it touches auth-adjacent retry logic, so it is deliberately left for a follow-up rather than bundled here.

## Related

- API: https://github.com/peanutprotocol/peanut-api-ts/pull/1282
- Growth runbook: https://github.com/peanutprotocol/mono/pull/69
- Retire the stale skill: https://github.com/peanutprotocol/chip/pull/9


0xkkonrad:
@coderabbitai review

Hugo0 (approved):
Approved by Hugo. CI fully green (ci-success, e2e, eslint, format, Deploy-Preview, CodeQL). Merged as part of the badge platform release: UI first so badge artwork exists during the rolling deploy.

PR #2626, opened 2026-08-06
feat(i18n): emit device_language + platform for localization OKR

kushagrasarathe:
## Summary
The localization **Fit** OKR measures: of users whose **phone is set to ES/PT**, how many actually use the app in ES/PT (i.e. did not override back to English). To read that from PostHog we need two signals per user — the language they use (`app_locale`, already shipped in #2621) and the language their device asks for (`device_language`).

This adds `device_language` + `platform` as PostHog **super properties**, emitted once per session from `AppIntlProvider` startup:
- `device_language` — the **raw** device/browser tag (`@capacitor/device` on native, `navigator.language` on web), lowercased. Deliberately **not** run through `resolveLocale`: an unsupported language (e.g. `fr-FR`) must stay `fr-fr`, not collapse to `en` and pollute the "phone set to ES/PT" denominator.
- `platform` — reuses the existing `getPlatform()` helper (`web` / `ios-native` / `android-native` / `ios-pwa` / `android-pwa`) so the OKR can be split web vs native.

The OKR then filters `device_language` starting with `es`/`pt` and reads the `app_locale = en` override rate. **No KYC/nationality join** — this supersedes the residence-property approach (closed peanut-api-ts#1295).

## Task
TASK-20776 — https://app.notion.com/p/peanutprotocol/Build-the-localization-OKR-tracking-query-dashboard-3a583811757981e1a7ecfce4b57a2287

## Risks / breaking changes
Analytics-only. No route, schema, or FE contract change. The emit is fire-and-forget and fenced (a posthog throw can never break i18n or the app). Works unchanged in the native app — the Capacitor static-export webview runs `posthog-js` exactly like the browser.

Minor pre-existing inconsistency (not fixed here, follow-up): a few `capture()` calls already pass an ad-hoc `platform` event property with a different vocabulary (`'native'`, `'unknown'`). Event properties override the super property per-event, so no data breaks; consolidating the vocabulary is out of this PR's scope.

## QA
- `npm run typecheck`, `pnpm prettier --check`, unit tests — green locally.
- New tests in `locale-store.test.ts` cover: raw tag registered, unsupported language kept as-is (OKR-denominator guard), once-per-session, posthog-throw fenced.
- Manual: load the app with the browser/device language set to `es-*` / `pt-*` / `fr-*`, confirm the `device_language` + `platform` super properties attach to events in PostHog.

Screenshots: N/A (analytics wiring, no visible change).

## Summary by CodeRabbit

* **New Features**
  * Added automatic device language and platform context reporting during app startup.
  * Native device language is used when available, with browser language as a fallback.
  * Language values are normalized for consistent analytics reporting.
  * Device context is restored after signing out and back in.

* **Bug Fixes**
  * Analytics and platform integration errors no longer interrupt app startup or locale handling.

### DIFF
--- a/src/context/authContext.tsx
+++ b/src/context/authContext.tsx
@@ -16,7 +16,7 @@ import {
 } from '@/utils/general.utils'
 import { apiFetch } from '@/utils/api-fetch'
 import { useAppLocked } from '@/hooks/useAppLocked'
-import { currentAppLocale } from '@/i18n/app/locale-store'
+import { currentAppLocale, currentDeviceContext } from '@/i18n/app/locale-store'
 import { isCapacitor } from '@/utils/capacitor'
 import { clearAuthToken } from '@/utils/auth-token'
 import { resetCrispProxySessions } from '@/utils/crisp'
@@ -267,6 +267,9 @@ export const AuthProvider = ({ children }: { children: ReactNode }) => {
             // locale so logout-window events keep carrying app_locale
             const locale = currentAppLocale()
             if (locale) posthog.register({ app_locale: locale })
+            // same for the localization-OKR device context (device_language + platform)
+            const deviceContext = currentDeviceContext()
+            if (deviceContext) posthog.register(deviceContext)
         } catch (e) {
             console.warn('posthog reset failed:', e)
         }
--- a/src/i18n/app/AppIntlProvider.tsx
+++ b/src/i18n/app/AppIntlProvider.tsx
@@ -4,7 +4,14 @@ import { NextIntlClientProvider, IntlErrorCode, type IntlError } from 'next-intl
 import { createContext, useCallback, useContext, useEffect, useRef, useState } from 'react'
 import { DEFAULT_APP_LOCALE, type AppLocale } from './config'
 import { loadMessages, type AppMessages } from './messages'
-import { currentAppLocale, emitLocaleToAnalytics, localeReady, markLocaleApplied, persistLocale } from './locale-store'
+import {
+    currentAppLocale,
+    emitDeviceContextToAnalytics,
+    emitLocaleToAnalytics,
+    localeReady,
+    markLocaleApplied,
+    persistLocale,
+} from './locale-store'
 import en from './messages/en.json'
 
 interface AppLocaleContextValue {
@@ -41,6 +48,9 @@ export function AppIntlProvider({ children }: { children: React.ReactNode }) {
 
     useEffect(() => {
         let cancelled = false
+        // device_language + platform super properties for the localization OKR;
+        // independent of which locale resolves, fire-and-forget
+        void emitDeviceContextToAnalytics()
         localeReady().then(async (resolved) => {
             startupLocale.current = resolved
             if (resolved === DEFAULT_APP_LOCALE) {
--- a/src/i18n/app/__tests__/locale-store.test.ts
+++ b/src/i18n/app/__tests__/locale-store.test.ts
@@ -41,7 +41,9 @@ function freshStore(): LocaleStore {
 }
 
 beforeEach(() => {
-    jest.clearAllMocks()
+    // resetAllMocks (not clearAllMocks) so a mockImplementation set in one test
+    // — e.g. the posthog-throw cases — can never leak into the next.
+    jest.resetAllMocks()
     mockIsIdentified.mockReturnValue(true)
     mockIsCapacitor.mockReturnValue(false)
     mockGetPlatform.mockReturnValue('web')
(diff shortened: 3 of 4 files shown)


kushagrasarathe:
@coderabbitai review

kushagrasarathe:
### Review dispositions

**Fixed** (commit `6be67443c`):
- **Once-guard set before `await` + lost on `posthog.reset()` at logout** — the emit could permanently drop `device_language`/`platform` for a session. Now caches the resolved context and re-registers it on logout (mirrors `app_locale`); the guard is set only after a successful `register`, so a throw can retry.
- **Duplicate native bridge call** — `Device.getLanguageTag()` was fired twice on the splash-gated startup path. Memoized the raw tag; `resolveStartupLocale` and the analytics emit now share one round-trip.

**Added test** (commit `c50033838`): native `Device.getLanguageTag()` path (CodeRabbit nitpick).

**Won't fix — correct for our native setup:**
- *native `navigator.language` fallback gives app locale, not OS language* — that fallback only runs when `@capacitor/device` is absent (older OTA binary). We don't override the webview locale, so `navigator.language` there ≈ OS language. Best available; matches `resolveStartupLocale`.
- *`getPlatform()` may return `web` for a native user* — only when `window.Capacitor` is absent / `NEXT_PUBLIC_CAPACITOR_BUILD` unbaked. Our native app is **static export with `window.Capacitor` injected, no `server.url`**, so `getPlatform()` is reliable. Fixing this is a pre-existing `getPlatform()` concern, out of scope.

kushagrasarathe:
@coderabbitai review

kushagrasarathe:
### ✅ Ready for review

**What & why:** adds `device_language` + `platform` PostHog super-properties so the localization **Fit** OKR can read the "phone set to ES/PT but app in English" override rate — no KYC/nationality join, works web + native.

**Risk:** analytics-only. No route/schema/FE-contract change. Emit is fenced — a posthog throw can never break i18n.

**Gates:** CI green (typecheck · unit 10/10 · e2e · format) · CodeRabbit 1 nitpick addressed, 0 open · /code-review 4 findings (2 fixed, 2 documented won't-fix) · Capgo/Deploy-Preview 🟡 known-amber.

**Tested:** unit (web + native tag paths, unsupported-language guard, logout re-register, retry-on-throw). No visible UI → Screenshots N/A.

**Smells:** none added. Noted: a few `capture()` calls pass an ad-hoc `platform` event-prop with a different vocabulary — event props override the super-prop per-event so no data breaks; consolidation is a follow-up, not this PR.

**Docs:** no customer-facing fact changed.

kushagrasarathe:
## Agent review — advisory (T3)

**Verdict: approve-advisory.** Clean, well-fenced, well-tested analytics wiring with a genuine perf improvement folded in. No breakage, perf, security, or quality concerns. Task link present (TASK-20776).

### Risk tier
- **T3** — analytics wiring only. No money, auth-gating, routes, schema, or FE contract. The `authContext.tsx` touch is analytics re-register inside the existing try/catch, not auth logic.

### Paired PR
- **none required.** Supersedes/closes peanut-api-ts#1295 (residence-property approach); consumes `app_locale` from #2621 (already shipped). No cross-repo dependency.

### A. Breakage — pass
- The `resolveStartupLocale` refactor is behavior-preserving: old code fell back to `navigatorLocale()` on a `Device.getLanguageTag()` failure; new `resolveLocale(await rawDeviceTag())` does the same — `rawDeviceTag()` internally falls through to `navigator.language` on the Capacitor import failure. typecheck + unit green, path directly tested.
- Memoized `deviceTag` promise makes the locale resolver and the analytics emit share **one** native bridge round-trip instead of two.
- Logout re-register is correct: the `if (deviceContext) return` guard is permanent, but the post-`reset()` re-register goes through the cached `currentDeviceContext()` in authContext (same pattern as `app_locale`), not the guarded emit.
- Fencing is real: `deviceContext` is set **only** after a successful `register()`, so a posthog throw leaves it null and a later call retries instead of silently disabling the emit forever — explicitly tested.

### B. Performance — pass
- Net improvement (one shared bridge call). Fire-and-forget emit, no loops/queries.

### C. Quality — pass (exemplary)
- Comments explain every non-obvious decision: why the raw tag is **not** run through `resolveLocale` (OKR-denominator integrity), why memoize, why re-register after reset, why fenced.
- Reuse: `getPlatform()` + shared `rawDeviceTag()`; no reinvention.
- Test hygiene: `beforeEach` switched `clearAllMocks` → `resetAllMocks` with a comment explaining the mockImplementation-leak it prevents. Tests assert exact registered payloads (non-vacuous).
- Honest scope: flags the pre-existing ad-hoc `platform` event-property vocabulary (`'native'`/`'unknown'`) and defers consolidation.

### D. Security — pass
- Emits only a BCP-47 language tag + platform bucket to PostHog. Low-sensitivity preference data, no secrets. No auth behavior touched.

### Small, low-risk changes worth naming (all covered by tests, none blocking)
1. **`resolveStartupLocale` refactored** to share the memoized `rawDeviceTag()` — behavior-preserving, and it removes a duplicate native bridge call.
2. **`authContext` logout path** re-registers `device_language` + `platform` after `posthog.reset()`, mirroring the existing `app_locale` re-register.
3. **Test `beforeEach` `clearAllMocks` → `resetAllMocks`** to stop a `mockImplementation` leaking between cases.

### CI
- eslint **FAIL** — identical 4 pre-existing errors (`ReConsentModal` + `AdvisoryPreemptModal`, untouched here). **0 new** from this PR; matches the claim. Advisory; `ci-success` green. typecheck / unit / e2e / format all pass.

### Bottom line
Genuinely ready. The OKR-denominator reasoning (keep `fr-fr` raw instead of collapsing to `en`) is correct and tested, the memoization is a real win, and the fencing is airtight. Nothing to fix.

<sub>Automated advisory review. Sign-off recorded by the author (@kushagrasarathe).</sub>

PR #2627, opened 2026-08-06
fix: reset the withdraw flow when a send method click enters it

abalinda:
## Summary

Browser/device back from `/withdraw?method=crypto` never runs the in-app NavHeader reset, so `selectedMethod` survives in the app-wide `WithdrawFlowContext`. The next **Bank** click on `/send` then skips method selection, lands on the **crypto amount step**, and Continue routes to `/withdraw/crypto?method=bank` — the crypto flow wearing a bank marker. The mirror direction (stale bank method hijacking an **Exchange or Wallet** click) breaks the same way.

**Fix:** a fresh method click is a fresh statement of intent. `SendRouterView.handleMethodClick` is the only click-site that mints the `?method=` marker, so call the pre-existing `resetWithdrawFlow()` there before navigating, in all four withdraw-bound cases (`bank`, `exchange-or-wallet`, `mercadopago`, `pix`). Plus one completeness fix the adversarial pass surfaced: `resetWithdrawFlow()` now also closes `showCompatibilityModal`, which every reset call site (home mount, crypto-page unmount, these clicks) was leaving armed.

`withdraw/page.tsx` is untouched: inside the withdraw flow, `selectedMethod` stays the routing source of truth (the `7edcf0919` invariant and its pinned test survive unmodified) — we only guarantee the context is fresh at entry.

## Task

TASK-21203 — https://app.notion.com/p/3b483811757981418f7ff0efe4d33221

## Verification (3-agent adversarial panel, all lenses traced to code)

- **Refute-the-fix:** HOLDS on all axes. The click-handler reset flushes in the sync lane before the `startTransition`-wrapped route render, the provider never remounts, and the step-sync effect self-heals as a third safety net. Mirror direction traced clean.
- **Regression hunt:** no real regressions. The "preserved typed amount" a reset might wipe never existed (`rawTokenAmount` is component-local; context amount is only written on Continue). The manteca page reads zero context fields and self-resets on mount. `/home` has always reset the flow on mount, so "state survives leaving the flow" was never a supported invariant, and every re-entry path guards against cleared context.
- **Test attack (empirical mutation pass):** revert, reorder-past-push, and over-eager-reset mutants all caught; the initially-untested manteca one-liners got their own test after the pass flagged them.

- **Reset at the click site, not self-healing inside `/withdraw`:** the within-flow back-navigations that echo `?method=` (`AddWithdrawCountriesList`) *want* state preserved and are correctly untouched. Browser-forward cannot recreate a mismatch (a fresh click truncates forward history).
- **Known residual, same bug family, lower severity (follow-up, not this PR):** marker-less `/withdraw` entries (home tile, bottom nav, KYC modal, exchange-rate widget) still resume a stale `selectedMethod` at the amount step. That path is internally consistent (no marker/context mismatch — Continue routes where the UI says) and matches the codebase's blessed resume semantics.
- **Known larger refactor (follow-up):** the `?method=` URL marker and `WithdrawFlowContext` remain two sources of truth — this is the third bug from that split.

## Risks / breaking changes

- FE-only, no API changes, no cross-repo deploy order.
- Blast radius: the four send-method click paths + the one-line reset completeness fix (strictly closes a modal that should never be open at reset time).
- Hotfix to `main` → **back-merge debt main→dev after merge.** No conflict expected: `handleMethodClick` is byte-identical on `main` and `dev`, and open PR #2625 does not touch `SendRouter.view.tsx`.

## QA

1. `/send` → **Exchange or Wallet** → **browser back** (swipe/hardware/browser — not the in-app `‹`) → **Bank** → must land on bank method selection, not the crypto amount step.
2. Reverse: `/send` → **Bank** → pick a saved account → **browser back** → **Exchange or Wallet** → must land on the crypto amount step cleanly.
3. In-app `‹` back paths unchanged. Plain `/withdraw` (no marker) unchanged.
4. Note: do **not** QA by asserting `/withdraw/crypto?method=bank` can never occur — it is legitimately mintable (send → bank → "show all methods" → Crypto card → Continue).

Screenshots: ⚠️ NONE — the change is which pre-existing screen appears after a click given abandoned state; both screens are unchanged pixel-wise. Manual repro steps above are the visual evidence.

## Summary by CodeRabbit

* **Bug Fixes**
  * Withdrawal flow state is now reset before starting bank, crypto, Mercado Pago, or Pix transfers.
  * Resetting the withdrawal flow now also closes any open compatibility modal.
  * Contact navigation remains unaffected.

### DIFF
--- a/src/components/Send/__tests__/send-states.test.tsx
+++ b/src/components/Send/__tests__/send-states.test.tsx
@@ -344,6 +344,27 @@ describe('GROUP 4: Method Selection', () => {
         fireEvent.click(screen.getByTestId('action-card-Exchange or Wallet'))
         expect(mockResetWithdrawFlow).toHaveBeenCalledTimes(1)
         expect(mockRouterPush).toHaveBeenCalledWith('/withdraw?method=crypto')
+        expect(mockResetWithdrawFlow.mock.invocationCallOrder[0]).toBeLessThan(
+            mockRouterPush.mock.invocationCallOrder[0]
+        )
+    })
+
+    test('Mercado Pago and Pix also reset the withdraw flow before navigating', () => {
+        mockUseGeoFilteredPaymentOptions.mockReturnValue({
+            filteredMethods: [
+                { id: 'mercadopago', title: 'Mercado Pago', description: '', icons: [], soon: false },
+                { id: 'pix', title: 'Pix', description: '', icons: [], soon: false },
+            ],
+        })
+        renderSend()
+
+        fireEvent.click(screen.getByTestId('action-card-Mercado Pago'))
+        expect(mockResetWithdrawFlow).toHaveBeenCalledTimes(1)
+        expect(mockRouterPush).toHaveBeenCalledWith('/withdraw/manteca?method=mercado-pago&country=argentina')
+
+        fireEvent.click(screen.getByTestId('action-card-Pix'))
+        expect(mockResetWithdrawFlow).toHaveBeenCalledTimes(2)
+        expect(mockRouterPush).toHaveBeenCalledWith('/withdraw/manteca?method=pix&country=brazil')
     })
 
     test('Clicking Peanut contacts navigates to /send?view=contacts without touching the withdraw flow', () => {
--- a/src/components/Send/__tests__/send-states.test.tsx
+++ b/src/components/Send/__tests__/send-states.test.tsx
@@ -169,6 +169,12 @@ jest.mock('../views/Contacts.view', () => ({
     default: () => <div data-testid="contacts-view">Contacts View</div>,
 }))
 
+// withdraw-flow context — SendRouterView resets it when a click enters the withdraw flow
+const mockResetWithdrawFlow = jest.fn()
+jest.mock('@/context/WithdrawFlowContext', () => ({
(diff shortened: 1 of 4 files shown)


abalinda:
@coderabbitai review

abalinda:
@coderabbitai full review

The single nitpick (method coverage + call order) is addressed: 7e73e9f21 added the Mercado Pago / Pix reset test + exchange ordering, 1aee6e755 added ordering assertions to both manteca clicks. Every withdraw-bound method now pins reset + route + reset-before-push ordering; the contacts test pins no-reset. Also since your review: 8e5347a00 makes resetWithdrawFlow close showCompatibilityModal (adversarial-pass finding).

abalinda:
@coderabbitai do the review now bro, 11 minutes passed

abalinda:
@coderabbitai review

r2 nitpick addressed in the latest commit: `src/context/__tests__/WithdrawFlowContext.test.tsx` mounts the real provider, arms the flow state (method, amounts, compatibility modal), invokes `resetWithdrawFlow()`, and asserts every field clears — including `showCompatibilityModal`.

jjramirezn (approved):

PR #2634, opened 2026-08-07
fix(ci): handle content guard diffs above 300 files

kushagrasarathe:
## Root cause
`gh pr diff --name-only` returns HTTP 406 for pull requests with more than 300 files. Under `set -e`, that harmless non-content PR becomes a false-red `approve-and-merge` check.

## Fix
Use GitHubs paginated pull-request Files API. The guard still auto-merges only when the complete file list is exactly `src/content`.

## Verification
- YAML parses with Ruby Psych.
- The API returned 759 paths for release PR #2624 and correctly evaluated it as not content-only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automated content-only change detection for large pull requests.
  * Ensured validation continues to require exactly one eligible file change.

### DIFF
--- a/.github/workflows/content-publish-automerge.yml
+++ b/.github/workflows/content-publish-automerge.yml
@@ -61,7 +61,10 @@ jobs:
                   REPO: ${{ github.repository }}
               run: |
                   set -euo pipefail
-                  FILES=$(gh pr diff "$PR" --repo "$REPO" --name-only)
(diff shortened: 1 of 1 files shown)


kushagrasarathe:
Superseded by #2636, which combines this change with the Capgo workflow fix and targets dev.

PR #2637, opened 2026-08-07
fix(websocket): stop shipping raw frames to Sentry on a parse error

Hugo0:
## Summary

The WebSocket parse-error catch logged the whole frame:

```ts
console.error('Error parsing WebSocket message:', error, event.data)
```

`console.error` is not local. Both `instrumentation-client.ts:60` and `sentry.client.config.ts:28` register `captureConsoleIntegration({ levels: ['error', 'warn'] })`, so **every `console.error` becomes a Sentry event**.

`beforeSendHandler` does not save us. It scrubs `request.headers`, `request.data`, `extra`, `contexts` and breadcrumb `data` — **by key name**, and it never touches `event.message`. Key-name redaction does nothing to a raw serialized blob.

The frames this handler receives are `kyc_status_update`, `sumsub_kyc_status_update`, `manteca_kyc_status_update`, `history_entry`, `rain_card_balance_changed`, `user_rail_status_changed` — user KYC state and financial data. A malformed one carried all of it straight to Sentry.

Now it logs the byte length. That still separates a truncated frame from a malformed one, which is the only thing this catch ever needed.

## Why this is worth a PR rather than a dismissal

This is CodeQL alert **#145** (`js/log-injection`, medium), open on `main` since 2026-07-28. I initially read it as negligible — "it's just a browser console". That was wrong: it's negligible in a repo that doesn't pipe console into Sentry, and this repo does.

The irony is that it leaks PII into Sentry past `scrubObject`, the function that exists to stop exactly that.

## Test

`src/services/__tests__/websocket-parse-error-pii.test.ts` — feeds a malformed frame carrying a name, document number and email, then asserts none of it reaches `console.error`, and that the size still does.

Verified the test is real: it **fails** against the old line and **passes** against the new one.

## Risks

None to behaviour. One log line; no control flow, no API surface, no user-visible change. The diagnostic value of the log is preserved (error + frame size).

## Follow-ups — NOT in this PR

Same class, different sites, found while grepping the blast radius. Both log an object rather than a status code, so both could carry PII into Sentry the same way. Neither is CodeQL-flagged and neither is in this PR's scope:

- `src/context/authContext.tsx:194` — `console.error('Unexpected error adding account', response)`
- `src/services/perks.ts:68` — `console.error('claimPerk: API request failed', response.status, data)`

Also worth its own ticket: there are **two** `sentry.utils.ts` (root, 10 KB — what the Sentry configs import; and `src/utils/`, 19 KB), both defining `scrubObject`. Duplicated redaction logic means adding a sensitive key to one silently leaves the other leaking.

## QA

`npm test` — 211 suites, 2693 passed. `npm run typecheck` clean. `prettier --check` clean.

To see it by hand: break a WS frame in devtools and confirm the console line reads `Error parsing WebSocket message: <err> (frame bytes: N)` with no payload.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket error logging to prevent sensitive message contents from appearing in logs.
  * Parse errors now record only the message size or identify non-string data, while retaining useful error details.

* **Tests**
  * Added regression coverage to verify that malformed messages containing sensitive information are safely handled.

### DIFF
--- a/src/services/__tests__/websocket-parse-error-pii.test.ts
+++ b/src/services/__tests__/websocket-parse-error-pii.test.ts
@@ -1,0 +1,62 @@
+import { PeanutWebSocket } from '@/services/websocket'
+
+/**
+ * console.error is wired to Sentry through
+ * captureConsoleIntegration({ levels: ['error', 'warn'] }), and
+ * beforeSendHandler scrubs headers/request.data/extra/contexts/breadcrumbs
+ * by key name — it never touches event.message. So anything handed to
+ * console.error leaves the browser verbatim.
+ *
+ * A malformed WebSocket frame carries the same shapes the good ones do
+ * (kyc_status_update, history_entry, rain_card_balance_changed), which is
+ * user KYC and financial data. This pins the parse-error path so nobody
+ * reintroduces the raw frame into that log line.
+ */
+describe('PeanutWebSocket — malformed frame never reaches Sentry via console', () => {
+    // A frame that fails JSON.parse but still carries recognisable PII.
+    const PII_FRAME =
+        '{"type":"kyc_status_update","data":{"status":"approved","fullName":"ALEKSEI SOKOLOV",' +
+        '"documentNumber":"AB1234567","email":"aleksei@example.com"}' // truncated → invalid JSON
+
+    const SECRETS = ['ALEKSEI SOKOLOV', 'AB1234567', 'aleksei@example.com', 'kyc_status_update']
+
+    let socket: { onmessage: ((event: MessageEvent) => void) | null }
+    let errorSpy: jest.SpyInstance
+
+    beforeEach(() => {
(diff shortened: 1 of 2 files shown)


Hugo0:
@coderabbitai review

Hugo0:
Good catch — fixed in `533397402`, but by relabelling rather than by measuring.

`String.length` is UTF-16 code units, so `frame bytes` was simply the wrong word. Getting a true byte count means running the whole frame through a `TextEncoder` inside an error path, and a malformed frame can be large. Only the magnitude matters here — it exists to tell a truncated frame from a malformed one — so the log now says `frame length` and the comment states explicitly that it is code units, not bytes, and why.

Deliberately not using `TextEncoder`: paying an allocation over the full payload to make a diagnostic log marginally more precise is the wrong trade in a catch block.

PR #2639, opened 2026-08-07
feat(support): unread badge on the Support nav icon

abalinda:
Notion: TASK-21141 — notify users when support replies. Frontend half.

Backend: peanutprotocol/peanut-api-ts#1303.

## ⛔ Merge order

**Do not merge before peanut-api-ts#1303 is deployed.** An old backend ignores the `category` query param, so the badge would show the *global* unread count — a dot on the Support icon for any unread notification at all, and opening support would never clear it.

**This PR is deliberately left as a draft** so that ordering cannot be lost to a stray click. The code is finished and green — CI, review and tests are all done. Mark it ready once #1303 is live.

## What this does

1. A pink dot on the Support icon in the mobile nav while support has replied and the user has not opened the chat.
2. The dot clears when the chat opens.
3. `/home?support=open` opens the support drawer — the deep link a support-reply push carries.

The count is server truth: `GET /notifications/unread-count?category=support`. The client cannot work it out for itself — the web Crisp widget is a sandboxed iframe that mounts only after the drawer is first opened, and the native plugin exposes no message events.

## Files

| File | Role |
|---|---|
| `hooks/useSupportUnread.ts` | `hasUnread`. Fetches on mount, on `notifications:updated`, and when the app returns to the foreground. No polling. |
| `components/Global/SupportDeepLink/index.tsx` | Reads `?support=open`, opens the drawer, clears the param. Renders null. |
| `components/Global/IndicatorDot/index.tsx` | The pink dot, now one component. |
| `components/Global/SupportDrawer/index.tsx` | Clears the badge on open. |
| `components/Global/WalletNavigation/index.tsx` | Renders the badge. |
| `services/notifications.ts` | `unreadCount(category?)`, new `markAllRead(category)`. |

## Why clearing hangs off `isSupportModalOpen`

It is the one flag every entry point sets before anything opens — the nav tap, `openSupportWithMessage()`, the push deep link, and the Capacitor path (which sets it before handing over to the native messenger). One effect covers all four, so there is no entry that lights the chat without clearing the badge.

`markAllRead` takes a category rather than ids because the drawer never sees notification ids. The conversation lives in Crisp, not in our inbox.

## The dot extraction

The same pink dot was copy-pasted in three places and the badge would have made a fourth, so it is now `IndicatorDot`. Approved by Aleks before the work started.

All three call sites render the same as before — `twMerge` resolves the size and animation overrides:

- `ProfileMenuItem` — the `animate-pulse` wrapper collapses onto the dot. Same visual; `aria-label="highlight-indicator"` preserved.
- `CarouselCTA` — the positioned wrapper stays (it owns `CAROUSEL_CLOSE_BUTTON_POSITION`, `z-10` and the aria-label); only the inner div is replaced.
- `TransactionCard` — `h-2 w-2 animate-pulsate` passed through as className.

The name is neutral on purpose: on a transaction card the dot means pending, on the perk carousel it means claimable, on the nav it means unread.

## Verification

- `npm test` — 212 suites, 2706 tests pass. New: 7 for `useSupportUnread` (category scoping, both refetch triggers, listener cleanup, failure path), 5 for `IndicatorDot` (class resolution at all four call sites + the announced badge), and 3 for the drawer (clears once the chat renders, does NOT clear on the Crisp-failure fallback, does NOT clear for a guest).
- `npm run typecheck` clean, `prettier --check .` clean, `npm run build` clean.
- The existing `SupportDrawer` suite needed a `@/services/notifications` mock — the drawer now makes a request on open, and `serverFetch` reaches for Capacitor Preferences, which jsdom has no shim for.

## Screenshots

⚠️ **None, and here is why.** A truthful shot of the badge needs the backend deployed — the count comes from `?category=support`, which only exists on peanut-api-ts#1303. Against today's backend the dot would light for *any* unread notification, so a screenshot would show the wrong thing and read as proof.

The dot extraction is the part that actually carries visual risk, and it is covered by something stronger than a screenshot: `IndicatorDot.test.tsx` asserts the resolved class string for all four call sites, so a twMerge override that stopped winning (leaving `h-2.5` next to `h-2`) fails CI rather than needing an eye. That check is permanent; a screenshot is a one-off.

**Still to do by hand after the backend deploys:** visual check at 375×667 on the three migrated dots (profile menu highlight, perk carousel dot, pending transaction dot) next to the new nav badge. Then the live path: real support reply → push + badge; second reply without opening → no second push; open chat → badge clears.

## Known gaps, not in scope

- The desktop sidebar still links to a dead `/support` route. Pre-existing.
- `notificationsApi.list()` already sends `filter` and `category` params the backend ignores. Pre-existing.
- Peanut-ui gets **no CodeRabbit pass** on a dev-base PR (the check reports "reviews are disabled for this base branch"), so `/code-review` was the only automated reviewer here.
- The green `e2e` check is vacuous — that job aborts before running any test. Do not read it as e2e coverage.

## Review fixes applied

`/code-review medium` returned 6 findings. The deploy-ordering one is the ⛔ at the top of this description. The other five are fixed:

- **The badge cleared even when the reply was never seen.** When the Crisp bundle fails to load, this component shows the email fallback instead — and the badge cleared anyway, burying the reply. The web path now waits for `CRISP_READY`; the native path has no such signal, so it clears right after `openMessenger()`.
- **A reply arriving while the drawer was open was never marked read** — the normal case in a live conversation. The badge would light with nothing new behind it and stay lit until the next open. Clearing now also fires on the closing edge.
- **Concurrent refreshes could resurrect a cleared badge.** Tapping a push fires a foreground refetch (count 1); the deep link then clears and refetches (count 0); if the first response lands last it won, and with no polling nothing corrected it. Responses now carry a request id and stale ones are dropped.
- **Guests were sending an unauthenticated mark-read** on every support open (claim and pay links mount this layout). Gated on a resolved userId.
- **The nav badge announced nothing.** `aria-label` on a bare span is ignored by assistive tech and is an `aria-prohibited-attr` violation. It now carries `role="status"` with a translated label. `es-AR` has no `navigation` block and falls back, so only the three locales that have one were touched.

### DIFF
--- a/src/app/(mobile-ui)/layout.tsx
+++ b/src/app/(mobile-ui)/layout.tsx
@@ -10,11 +10,12 @@ import BackendErrorScreen from '@/components/Global/BackendErrorScreen'
 import { useAuth } from '@/context/authContext'
 import classNames from 'classnames'
 import { usePathname } from 'next/navigation'
-import { useCallback, useEffect, useRef, useState } from 'react'
+import { Suspense, useCallback, useEffect, useRef, useState } from 'react'
 import { twMerge } from 'tailwind-merge'
 import '../../styles/globals.css'
 import QRScannerOverlay from '@/components/Global/QRScannerOverlay'
 import SecurityVerificationOverlay from '@/components/Global/SecurityVerificationOverlay'
+import SupportDeepLink from '@/components/Global/SupportDeepLink'
 import SupportDrawer from '@/components/Global/SupportDrawer'
 import JoinWaitlistPage from '@/components/Invites/JoinWaitlistPage'
 import { useRouter } from 'next/navigation'
@@ -253,6 +254,12 @@ const Layout = ({ children }: { children: React.ReactNode }) => {
             <ReConsentModal />
 
             <SupportDrawer />
+
+            {/* Suspense is required: nuqs reads useSearchParams, which triggers
+                a client-side-rendering bailout without a boundary. */}
+            <Suspense fallback={null}>
+                <SupportDeepLink />
+            </Suspense>
 
             <QRScannerOverlay />
 
--- a/src/components/Global/IndicatorDot/__tests__/IndicatorDot.test.tsx
+++ b/src/components/Global/IndicatorDot/__tests__/IndicatorDot.test.tsx
@@ -42,4 +42,11 @@ describe('IndicatorDot', () => {
             expect.arrayContaining(['absolute', '-right-1', '-top-1', 'h-2.5', 'w-2.5', 'bg-primary-1'])
         )
     })
+
+    it('announces the support badge to assistive tech', () => {
+        // aria-label on a bare span (generic role) is ignored, so the nav badge
+        // pairs it with role="status".
+        const { getByRole } = render(<IndicatorDot role="status" aria-label="New support reply" />)
+        expect(getByRole('status')).toHaveAttribute('aria-label', 'New support reply')
+    })
 })
--- a/src/components/Global/IndicatorDot/__tests__/IndicatorDot.test.tsx
+++ b/src/components/Global/IndicatorDot/__tests__/IndicatorDot.test.tsx
@@ -1,0 +1,45 @@
+/**
+ * Visual-parity guard for the dot extraction.
+ *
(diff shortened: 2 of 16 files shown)


abalinda:
@coderabbitai review now

PR #2642, opened 2026-08-07
Prod Release Sprint 156 — shared FX API · Sumsub canonical identity · withdraw/ledger reliability (2026-08-18)

Hugo0:
## Payload

**FX — consume the shared backend rate policy** (#2607)
Replaces the local conversion implementation with `GET /fx/rate`, so the wallet and Peanut Split share one contract.
- Device-clock tolerance widened from 5 min to 6 h. At 5 min a phone whose clock had drifted rejected **every** response the backend could send, permanently, with no local fallback left. The checks that actually bound staleness compare two backend timestamps and stay tight.
- `503` added to the Sentry skip list for `/fx/rate` — peanut-api already reports the upstream cause; reporting it client-side multiplies one incident by every mounted hook and its retries.
- `/fx/rate` gets a real demo-api handler; without one a failed passthrough answered `200 {}`.

**Support — unread badge** (#2639) — FE half of peanut-api-ts #1303. Ships together.

**Observability** (#2637) — stop shipping raw websocket frames to Sentry on a parse error.

**CI** (#2636) — workflow consolidation. No runtime effect.

## Order — merge this AFTER the backend

peanut-api-ts #1308 must be deployed first. This PR drops `api.frankfurter.app` from CSP, so if the FE ships before `/fx/*` is live on prod, every rate surface blanks and the old path cannot be hot-restored.

## Verification

FX verified against staging (which tracks `dev`): **43/43** on the shadow-compare, every Manteca pair resolving from Manteca rather than the reference feed.

## Summary by CodeRabbit

* **New Features**
  * Added localized press content, profile viewing, region and country labels, and expanded badge translations.
  * Added invite and referral sharing across profiles, badges, QR payments, transactions, and guest onboarding.
  * Added unread support indicators and automatic badge clearing when support conversations are viewed.
* **Bug Fixes**
  * Prevented duplicate history entries and stalled pagination.
  * Improved timeout messaging, deep-link routing, and app-store handoffs.
  * Protected sensitive data from error logs and allowed replacing existing app bundles during deployment.
* **SEO & Navigation**
  * Added redirects, canonical URLs, sitemap improvements, and safer indexing for invalid routes.

PR #2643, opened 2026-08-07
chore: back-merge main → dev (post badge-platform, 2026-08-07)

Hugo0:
Second back-merge today — the badge-platform UI work (#2613) merged straight to `main` after the first one, which made the release PR #2642 dirty.

**Conflict:** `src/types/api.openapi.json` and `src/types/api.generated.ts`. Both are generated artifacts, and both sides added endpoints — FX on `dev`, badges on `main`. Hand-merging either would have dropped one side's paths, so instead the openapi snapshot was replaced with the **merged backend `openapi.json`** (which already carries both) and `pnpm gen:api` regenerated the types.

Verified: `/fx/rate` and the badge paths are both present, typecheck clean, 2890 tests pass.


Hugo0:
Closing in favour of #2652.

This PR was opened 2026-08-07 17:04Z — minutes *before* the badge-platform merge landed on `main` — so it never contained the work it was named for, and it has been `mergeable=false`/`dirty` since. Rebasing it would have meant resolving conflicts against a base that had moved substantially.

The badge platform has since reached `dev` by another route, so the remaining gap was small. #2652 is a fresh branch off current `dev` with `main` merged in: **zero conflicts**, full suite green.

PR #2649, opened 2026-08-08
fix(sentry): stop one console.warn from swallowing every network error

Hugo0:
**One `console.warn` was bucketing every non-2xx in the app into a single Sentry issue.**

`fetchWithSentry` logged each failed request with `console.warn`, and `captureConsoleIntegration` listens on `['error','warn']`. So every non-2xx produced a **second** Sentry event, grouped by that call site rather than by request.

That is `PEANUT-UI-60Y`: **~32,000 events / 2,494 users**, spanning every endpoint, titled after whichever request happened to fail last.

## Why this matters beyond noise

It actively misled triage. Yesterday it read as *"`invites/accept` 400 — 80 users"* and was ranked the top production issue. Real volume for that endpoint is **1 event in 7 days** (corroborated by PostHog: 8 `invite_accept_failed` in 60 days, and prod API logs: 0–2 validation 400s/day). The bucket was hiding what it contained and mislabelling itself with a specific URL.

## The signal is kept

The explicit `Sentry.captureMessage` below it is unchanged and is the real report — it fingerprints on `[method, sanitizeUrl(url), status]` and carries headers, body and response. Dropping the log to `console.info` removes the **duplicate**, not the signal. `PEANUT-UI-R0Y` is the correctly-split sibling and stays.

## Also in here

`/tokens/price` 404 added to `SKIP_REPORTING`. Those are Mobula 429s surfacing as 404 (`PEANUT-UI-QKY`, 60 users). The UI falls back to token denomination — a degraded display, never a wrong number — and the backend already downgraded its own log for exactly this reason (PEANUT-API-75). Same class of noise, same file, so it travels with this.

## Tests

- a non-2xx reports **exactly once**, via `captureMessage`, never via `console.warn`
- a `/tokens/price` 404 reports **not at all**

Typecheck clean, 2886 tests green.

## Expected effect

`PEANUT-UI-60Y` stops accumulating. Its ~32k events redistribute into per-endpoint issues that are actually triageable — expect the issue list to get longer and far more useful.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unnecessary error reporting for expected `/tokens/price` not-found responses.
  * Improved handling of failed requests by sending a single Sentry event without duplicate warning messages.
  * Updated request diagnostics to use informational logging for non-success responses.

### DIFF
--- a/src/utils/__tests__/sentry.utils.test.ts
+++ b/src/utils/__tests__/sentry.utils.test.ts
@@ -90,6 +90,31 @@ describe('fetchWithSentry — expected-response suppression', () => {
         expect(warnSpy).not.toHaveBeenCalled()
     })
 
+    it('does not report a /tokens/price 404 (upstream price provider declined)', async () => {
+        // Mobula 429s surface here as a 404. The UI falls back to token
+        // denomination, so it is a degraded display and never a wrong number —
+        // and the backend already downgraded its own log for this (PEANUT-API-75).
+        global.fetch = jest.fn().mockResolvedValue(mockResponse(404, { error: 'Token price not available' }))
+
+        const response = await fetchWithSentry('https://api.peanut.me/tokens/price?address=0xaf88&chainId=42161', {})
+
+        expect(response.status).toBe(404)
+        expect(Sentry.captureMessage).not.toHaveBeenCalled()
+    })
+
+    it('reports a non-2xx exactly once, via captureMessage and never via console.warn', async () => {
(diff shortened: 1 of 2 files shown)


Hugo0:
Fixed in `722bd76d5` — bounded to the path end or a query/fragment.

Correct as stated: the bare prefix also matched `/tokens/price-history` and any nested path, so a 404 from one of those would have skipped its `captureMessage` and gone unreported. Given this PR exists to stop a rule from swallowing reports, a skip rule that over-matches is the same bug in miniature.

PR #2656, opened 2026-08-10
fix(withdraw): map Rhino NoRouteFoundError to actionable route-not-found copy

abalinda:
## Summary

Companion to peanut-api-ts#1320 (TASK-21284). When Rhino cannot route a pair, the backend surfaces the provider tag (`NoRouteFoundError`) verbatim and the withdraw confirm screen shows the raw string. Map it to `ROUTE_NOT_FOUND_ERROR` in the shared `calculate()` catch, so all three consumers (withdraw, pay-request, claim-xchain) inherit the copy the Confirm view already special-cases — its Retry button calls `onBack()` for this constant, which returns the user to the amount input. No new UI code; one mapping line plus the hook's first test file.

## Task

Notion TASK-21284 — Fix Solana USDC withdrawal preview failure (DoD bullet 2: unsupported routes show a specific error and leave the user able to correct the input).

## Risks / breaking changes

- None cross-repo. Substring match on the thrown message; every other error passes through verbatim (pinned by test).
- Independent of api#1320 deploy order — before it, previews fail with the api-side message either way; after it, unroutable pairs get the actionable copy.

## QA

- New `useCrossChainTransfer.test.ts`: NoRouteFoundError → ROUTE_NOT_FOUND_ERROR; any other message passes through verbatim.
- Full local gate: typecheck clean, 226 suites / 2889 tests green.

Screenshots: N/A (no new surface — the existing ErrorAlert renders different copy on an error path that today shows a raw provider string).

- `/code-review medium` (1 finding, applied-as-follow-up): the `ROUTE_NOT_FOUND_ERROR` English constant bypasses the `classifyError`/next-intl localization path. Pre-existing pattern (same constant + equality check in `Confirm.withdraw.view.tsx` and `Claim/Link/Initial.view.tsx` before this PR); migrating it to a `FriendlyErrorCode` touches the Confirm view's `error === ROUTE_NOT_FOUND_ERROR` special-case, so it goes in a separate localization follow-up rather than this hotfix. No behavior regression — users previously saw a raw JSON blob.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-chain transfer error handling by showing a clear route-not-found message when no transfer route is available.
  * Preserved the original error details for other preview and fee-estimation failures.
  * Enabled the appropriate fee-estimation error state when a route cannot be found.

### DIFF
--- a/src/features/payments/shared/hooks/useCrossChainTransfer.test.ts
+++ b/src/features/payments/shared/hooks/useCrossChainTransfer.test.ts
@@ -1,0 +1,79 @@
+/**
+ * useCrossChainTransfer — calculate() error mapping.
+ *
+ * Rhino tags an unroutable pair `NoRouteFoundError`; the backend surfaces the
+ * tag verbatim inside the thrown message. The hook maps it to
+ * ROUTE_NOT_FOUND_ERROR so the Confirm view's existing special-case (Retry →
+ * onBack) sends the user back to fix the input. Any other failure passes
+ * through verbatim.
+ */
+import { act, renderHook } from '@testing-library/react'
+import { useCrossChainTransfer } from './useCrossChainTransfer'
+import { ROUTE_NOT_FOUND_ERROR } from '@/constants/general.consts'
+import { previewSdaTransfer } from '@/services/rhino-sda'
+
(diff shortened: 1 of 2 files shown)


abalinda:
@coderabbitai review

abalinda:
@coderabbitai review

PR #2663, opened 2026-08-11
feat(split): prove isolated edge transport canary

0xkkonrad:
Task: https://app.notion.com/p/3b783811757981b481f1d9528c6f0600?pvs=204

## Summary
- adds an exact preview-only Split edge canary and namespace 404 firewall
- strips caller credentials and spoofed marker headers before the trusted Split renderer rewrite
- keeps Split paths NetworkOnly in the parent service worker and deletes upstream Set-Cookie at the Vercel response layer
- adds disposable Vercel proof routes with a strict non-secret request-header allowlist

## Verification
- 63 focused tests pass
- TypeScript typecheck passes
- changed-file Prettier and git diff checks pass
- local two-port seam passes HTML, RSC, assets, sitemap, credential stripping, marker validation, and real 404 propagation
- full production build previously passed before the disposable proof-route addition; the final local retry was SIGTERM under unrelated workspace memory pressure, so CI is the isolated build gate

## Safety
**DO NOT MERGE. Disposable A2 preview proof only; close after Vercel evidence.** The workflow injection is exact-branch-only, the canary is disabled in production, real renderer paths are proof-only 404s in this preview, and no production route ownership changes. The httpbingo echo receives only fixed non-secret canary headers; the public preview must be removed after evidence.

### DIFF
--- a/.github/workflows/preview.yaml
+++ b/.github/workflows/preview.yaml
@@ -29,6 +29,27 @@ jobs:
             - name: Pull Vercel Environment Information
               run: vercel pull --yes --environment=preview --scope=squirrellabs --token=${{ secrets.VERCEL_TOKEN }}
             - name: Build Project Artifacts
+              if: github.head_ref != 'feat/split-a2-edge-canary'
+              run: vercel build --target=preview --token=${{ secrets.VERCEL_TOKEN }}
+            - name: Build A2 Edge Proof Artifacts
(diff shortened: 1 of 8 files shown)


0xkkonrad:
Disposable A2 transport proof completed and independently verified. Closing without merge; exact preview teardown and temporary-secret deletion are in progress.

PR #2665, opened 2026-08-11
chore: back-merge main into dev (2026-08-11)

abalinda:
Routine back-merge. `main` was 20 commits ahead of `dev`.

Carries #2658 (the `/fx/card-markup` consumer, which deletes both duplicate card-comparison modules) plus the hotfixes that shipped straight to main.

No conflicts. Verified locally: `typecheck` clean, 229 suites / 2925 tests green.


abalinda:
Superseded by #2668, which is `CLEAN` and ready to merge.

This PR was never failing CI — every check on its head SHA was green (`ci-success` included), `dev` requires 0 approvals, and there were no unresolved review threads. GitHub still reported `BLOCKED`.

The cause: this branch carries three **unsigned** commits from the content-publish bot (`Update content submodule to latest main` at 10:05 / 10:22 / 10:33Z). The `All branches` ruleset (6457257) requires signed commits on every ref and has **no bypass actors**. The ruleset audit log shows the bot reaches `main` by *bypassing* that rule — a back-merge PR has no such bypass, so those commits cannot flow into `dev` through an ordinary merge.

#2668 is the same back-merge collapsed into one signed commit, off the current `dev` tip. Proof the diagnosis is right: identical content, `mergeStateStatus: CLEAN`.

**Follow-up worth filing:** the content-publish workflow should sign its commits (or author them through the GitHub API, which signs automatically). Until then every main→dev back-merge will hit this, and each one has to be squashed — which costs the merge ancestry and makes the next back-merge re-present the same delta.

PR #1325, opened 2025-10-15
[TASK-15768] feat: Clickable PFPs in txn receipt

Zishan-7:

PR #2666, opened 2026-08-11
fix: fetch Crisp support token from the API instead of deriving it client-side

jjramirezn:
## Summary

Frontend half of the Crisp support-token IDOR hotfix. **Pairs with peanut-api-ts #1325 — that API PR (and its migration) must deploy first.**

The Crisp session token was derived **in the browser** as `SHA-256("peanut-crisp-session-v1:" + userId)`. The salt ships in the bundle and `userId` is readable, so anyone could reproduce any user's token and open their support thread via `/crisp-proxy?crisp_token_id=<token>`.

This PR removes the client-side derivation. `useCrispTokenId` now fetches the token from the authenticated `GET /user/crisp-token` (the API issues a stored per-user random token — Crisp's documented safe pattern). The hook keeps the same contract — `undefined` until it resolves, in-memory cache per userId — so `SupportDrawer`'s `isAwaitingToken` gate and reset logic are unchanged.

Hardening from code review is included:
- **Reset on account switch** — the token is reset to the current user's value before fetching, so it never serves the previous user's token while the new one loads.
- **Identity check** — the route echoes `userId`; the hook discards any token whose `userId` doesn't match the account it is fetching for (stale-bearer guard).
- **Bounded retry** — a transient failure doesn't strand the support drawer for the session.

## Task

TASK link pending — user security report. Will attach.

## Risks / breaking changes

- **Hard dependency on peanut-api-ts #1325.** The endpoint (and its migration) must be live first, or `GET /user/crisp-token` 404s and support chat stays on its loading state.
- No visible UI change — same drawer, same states; only the token's source changed.
- *Residual (follow-up, not here):* on a persistent endpoint failure the drawer still shows the spinner with no fallback UI — a small SupportDrawer change, kept out of this security PR.

## QA

- Unit (`src/hooks/__tests__/useCrispTokenId.test.ts`): fetches from the authed endpoint, passes no userId parameter, discards a token minted for a different user, drops the previous user's token on account switch, and stays undefined (with retry) on failure.
- `SupportDrawer` suite still green (it mocks the hook).

## Screenshots

N/A (no visible change) — the drawer renders identically; only the token's source changed.

## Summary by CodeRabbit

* **New Features**
  * Crisp tokens are now securely retrieved from the authenticated service.
  * Added validation to ensure tokens match the correct account and token type.
  * Added automatic retry handling for temporary token request failures.
  * Tokens are cleared when switching accounts or when authentication is unavailable.

* **Tests**
  * Added coverage for authentication states, account changes, token validation, and retry behavior.

### DIFF
--- a/src/hooks/__tests__/useCrispTokenId.test.ts
+++ b/src/hooks/__tests__/useCrispTokenId.test.ts
@@ -1,0 +1,60 @@
+import { renderHook, waitFor } from '@testing-library/react'
+import { useCrispTokenId } from '@/hooks/useCrispTokenId'
+
+const mockUseAuth = jest.fn()
+jest.mock('@/context/authContext', () => ({
+    useAuth: () => mockUseAuth(),
+}))
+
+const apiFetchMock = jest.fn()
+jest.mock('@/utils/api-fetch', () => ({
+    apiFetch: (...args: unknown[]) => apiFetchMock(...args),
+}))
+
+const jsonResponse = (body: unknown, ok = true) => ({ ok, json: async () => body }) as unknown as Response
+
+describe('useCrispTokenId', () => {
+    beforeEach(() => {
+        apiFetchMock.mockReset()
+        mockUseAuth.mockReset()
+    })
+
+    it('returns undefined and never calls the API when unauthenticated', () => {
+        mockUseAuth.mockReturnValue({ userId: undefined })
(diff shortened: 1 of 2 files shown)


jjramirezn:
@coderabbitai review

jjramirezn:
@coderabbitai review

abalinda (approved):

PR #2673, opened 2026-08-12
fix: open profiles for every real receipt peer

kushagrasarathe:
## Summary

Receipt profile navigation trusted the peer identity signal, but then narrowed it with a presentation-type allowlist. This left valid Peanut users inert on bank-fulfilled request receipts and other real-user flows.

This PR removes that stale allowlist. The shared `hasUserProfile` predicate now allows any non-link transaction whose peer is a real Peanut user with a valid username. Raw addresses, UUID fallbacks, guests, links, merchants, and generated system labels stay inert.

The bank-fulfilled request strategy now derives peer identity only from `recipientAccount`. It no longer treats the logged-in sender as the peer.

The receipt header separates name and avatar navigation. A real counterparty name always links to the profile. User initials or a real user avatar also link. Bank flags, rail icons, and the rest of the receipt card stay inert.

The predicate is shared by the activity row and receipt header. The fix therefore covers the history drawer and standalone `/receipt/[entryId]` page without separate peer-identity logic.

## Task

TASK-21010

## Design decision: “To” row

The “To” detail row stays unchanged. It only renders for `claim_external`, where the recipient is not a Peanut user and no profile exists. Expanding that row to other receipt types would be separate product work.

## Risks / breaking changes

Low, frontend-only. No API, contract, dependency, or cross-repo change. The risk is an incorrect profile link if peer identity is wrong. Username, generated-label, link, crypto-address, and UUID checks remain as defense in depth.

## QA

- Added strategy coverage that a bank fulfillment never treats the logged-in sender as the peer.
- Added predicate coverage that generated fallback labels stay inert even if the peer flag is wrong.
- Added coverage for `bank_request_fulfillment`, real-user `add`, unknown future presentation types, and non-user QR/card/external flows.
- Added receipt-header coverage for linked names, linked real-user avatars/initials, inert bank flags, and an inert card body.
- Added activity-row click coverage for a bank-fulfilled request peer.
- `pnpm prettier --check .`
- `npm run typecheck`
- `npm test -- --runInBand` — 230 suites, 2,939 passed, 3 skipped.
- `npm run build`

## Screenshots

N/A — no visible change. This PR changes click targets only.

## Follow-up after release

Verify a real bank-fulfilled request receipt on production in both the history drawer and standalone receipt page. Then mark `product/feedback/problems/cant-tap-name-to-open-profile.md` as shipped in a separate content-only change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved profile availability across supported transaction types, including bank-fulfilled requests.
  * Users with valid profiles can now be selected even when transaction presentation details are unavailable.
  * Bank-fulfilled request payers now navigate directly to the payer’s profile without opening the details drawer.
  * Improved navigation consistency between clickable profile names and avatars.

* **Tests**
  * Expanded coverage for profile eligibility, navigation, and bridge-fulfilled requests.

### DIFF
--- a/src/components/TransactionDetails/TransactionDetailsHeaderCard.tsx
+++ b/src/components/TransactionDetails/TransactionDetailsHeaderCard.tsx
@@ -37,6 +37,7 @@ interface TransactionDetailsHeaderCardProps {
     transactionType?: TransactionType
     avatarUrl?: string
     haveSentMoneyToUser?: boolean
+    isNameClickable?: boolean
     isAvatarClickable?: boolean
     showProgessBar?: boolean
     progress?: number
@@ -205,6 +206,7 @@ export const TransactionDetailsHeaderCard: React.FC<TransactionDetailsHeaderCard
     transactionType,
     avatarUrl,
     haveSentMoneyToUser = false,
+    isNameClickable = false,
     isAvatarClickable = false,
     showProgessBar = false,
     progress,
@@ -232,10 +234,8 @@ export const TransactionDetailsHeaderCard: React.FC<TransactionDetailsHeaderCard
     const isTest = isTestTransaction(userName)
     const icon = getIcon(direction, isLinkTransaction, isTest)
 
-    const handleUserPfpClick = () => {
-        if (isAvatarClickable) {
-            router.push(profileUrl(userName))
-        }
+    const handleUserProfileClick = () => {
+        router.push(profileUrl(userName))
     }
 
     const isNoGoalSet = isRequestPotTransaction && goal === 0
@@ -253,7 +253,10 @@ export const TransactionDetailsHeaderCard: React.FC<TransactionDetailsHeaderCard
                 </div>
             ) : (
                 <div className="flex items-center gap-3">
-                    <div className={twMerge(isAvatarClickable && 'cursor-pointer')} onClick={handleUserPfpClick}>
+                    <div
+                        className={twMerge(isAvatarClickable && 'cursor-pointer')}
+                        onClick={isAvatarClickable ? handleUserProfileClick : undefined}
+                    >
                         {avatarUrl ? (
                             <div className="flex h-16 w-16 items-center justify-center rounded-full">
                                 <Image
@@ -297,7 +300,7 @@ export const TransactionDetailsHeaderCard: React.FC<TransactionDetailsHeaderCard
                                 className="flex items-center gap-1"
                                 haveSentMoneyToUser={haveSentMoneyToUser}
                                 iconSize={18}
-                                onNameClick={isAvatarClickable ? handleUserPfpClick : undefined}
+                                onNameClick={isNameClickable ? handleUserProfileClick : undefined}
                             />
 
                             <div className="ml-auto">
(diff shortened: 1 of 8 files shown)


kushagrasarathe:
@coderabbitai review

kushagrasarathe:
@coderabbitai review

abalinda:
### `isPeerActuallyUser` is true for every viewer on the bank-fulfilled branch

Verified against head `287027b`. The widened `hasUserProfile` is correct in intent, but it now rests on a flag that is not trustworthy on one branch.

**`src/components/TransactionDetails/strategies/intent/p2p-send.ts:29`**

```ts
isPeerActuallyUser: !!entry.recipientAccount?.isUser || !!entry.senderAccount?.isUser,
```

This branch is already gated on `entry.userRole === EHistoryUserRole.SENDER`, so `senderAccount` is the viewer, not the peer. The viewer is always a logged-in Peanut user. The `||` therefore makes the flag unconditionally true, whatever the recipient is. Every sibling branch in this file reads a single account, which is the correct shape.

**`src/components/TransactionDetails/transaction-predicates.ts:131`**

Dropping the card-type list removes the last thing that masked the flag above:

```ts
return (
    !!transaction.isPeerActuallyUser &&
    !transaction.extraDataForDrawer?.isLinkTransaction &&
    !!userName &&
    !isCryptoAddress(userName) &&
    !isUuid(userName)
)
```

`nameKey` is not checked, so a front-end label passes. On the bank branch `userName` falls back to `recipientAccount.identifier`, then to the literal `'Recipient'`. Both reach `router.push(profileUrl(userName))` in the header card and land on a 404 profile.

**Reachability is unconfirmed.** A request is created by a Peanut user, so `recipientAccount.isUser` is probably true in the common case. This may be latent rather than live. A query for `P2P_REQUEST_FULFILL` with `fulfillmentType = 'bridge'` and a non-user recipient would settle it. The type list was doing this job by accident before, so the PR should not rely on the data staying that way.

**Suggested fix — two lines.**

```ts
// p2p-send.ts:29 — match the sibling branches
isPeerActuallyUser: !!entry.recipientAccount?.isUser,
```

```ts
// transaction-predicates.ts — same class as the address and uuid filters
&& !transaction.nameKey
```

### The tests cannot catch this

**`src/components/TransactionDetails/__tests__/transaction-predicates.test.ts:183`**

Every rewritten "non-user counterparty" case sets `isPeerActuallyUser: false`, so the assertion reduces to `false === false`. The card-type list is no longer covered. A strategy that sets the flag on a merchant row would turn merchant names into profile links and the suite would stay green.

A case with `isPeerActuallyUser: true` plus a `nameKey`, expecting `hasUserProfile` to be `false`, would fail today and pass after the fix.

### Note on CI

The `e2e` check on this PR is a green empty run. The job aborts before any test, under `continue-on-error` with no harness secret. Real coverage here is `unit` and `typecheck`.

kushagrasarathe:
@abalinda Valid finding, thanks. The bank branch incorrectly included the logged-in sender in peer identity, and `hasUserProfile` did not reject generated labels. Fixed in `564f15db5`: the strategy now trusts only `recipientAccount.isUser`, and the shared predicate rejects any `nameKey`. I added regression tests for both cases; they failed before the fix and pass now. The full local gate is green: Prettier, typecheck, 230 suites with 2,939 passing tests, and the production build.

kushagrasarathe:
@coderabbitai review

PR #2675, opened 2026-08-12
fix(qr): recognize scanned Solana and Tron addresses

abalinda:
## Summary

The QR scanner rejected valid Solana deposit addresses as "Unrecognized QR code".

`processQRCode` lowercased the scanned payload and passed that copy to `recognizeQr`. Base58 carries meaning in its case, so the lowercase pass destroyed the address:

- **Solana** — an uppercase `L` is a valid base58 character, a lowercase `l` is not. Any address holding an `L` stopped matching. That is **~53% of addresses** (measured over random 44-char base58 strings).
- **Tron** — every Tron address starts with an uppercase `T`, and the pattern anchors on `^T`. Lowercasing broke **100%** of them, always.

Hex addresses were unaffected, which is why this stayed invisible: EVM survives `.toLowerCase()` because hex is case-insensitive.

## The fix

Recognize the **raw** scan first, and retry lowercased **only for an all-uppercase payload**:

```ts
const recognized = recognizeQr(data) ?? (data === data.toUpperCase() ? recognizeQr(normalized) : null)
```

The two halves encode when case is information and when it is an artifact:

| Payload | Case is… | Behavior |
|---|---|---|
| Solana / Tron base58 | **data** — `L` ≠ `l`, `T` ≠ `t` | raw pass matches (**the fix**) |
| All-uppercase bech32 / BOLT-11 / hex | **artifact** — QR alphanumeric mode encodes uppercase only, and is denser for it | retried lowercased |
| Mixed-case hex | **the user's** — EIP-55 checksum | honoured; viem rejects a bad one |

That last row is a deliberate tightening. The old blanket lowercase laundered a mixed-case address with a **broken** EIP-55 checksum into a valid-looking payment target — which is the exact corruption EIP-55 exists to catch. It now reaches the unrecognized modal instead of the send flow.

`normalized` still drives the routing below, where hex and ENS are case-insensitive anyway, and `recognizeQr` lowercases the branches that need it (Peanut URL, PIX) itself.

## Task

[TASK-21111](https://app.notion.com/p/peanutprotocol/P2-QR-scanner-does-not-recognize-Solana-deposit-addresses-3b183811757981e48e89e667fbd03ab7) — reported twice via support; reporter received the bug bounty.

## Evidence

`recognizeQr` was never the problem. Its 635-line suite passes raw mixed-case data and even asserts case-sensitivity on purpose (`'LNBC…'` must not match `BITCOIN_INVOICE`). The single call site lowercased one line earlier and had no test at all — so the new test covers the **component**, not the parser.

PostHog, last 60 days of `qr_scanned`:

| qr_type | scans | users |
|---|---|---|
| `null` (unrecognized) | 457 | 215 |
| `SOLANA_ADDRESS` | 6 | 4 |
| `TRON_ADDRESS` | **0** | **0** |

Tron has never once been recognized in production. One unrecognized scan in that window is a confirmed base58 Solana address holding an `L`.

Production sanity checks behind the design:
- **All 28** distinct mixed-case addresses scanned in 90 days carry a **valid** EIP-55 checksum — the tightening costs nothing real.
- All-uppercase payloads are genuinely common (93 in 180 days: EMVCo, FIDO, PIX), which is why the uppercase retry stays.

## Risks

Low. Frontend only, one expression, no backend or contract surface.

The one behavior change beyond recognition is the bad-checksum tightening described above — measured at zero production instances.

Recognized Solana/Tron scans now reach the existing "not supported yet / get notified" modal instead of the unrecognized one. See Design notes.

**Routing is unchanged, deliberately.** After this fix a scanned Solana address lands on the existing `QR_NOT_SUPPORTED` modal. Solana and Tron *are* live withdraw destinations (`chainRegistry.consts.ts`, Rhino, behind `chain-rollout-solana` / `chain-rollout-tron`), so that copy is arguably stale — but wiring scan-to-send would need a withdraw deep-link with chain preselection plus flag gating. That is a feature, not this bug, and the reported user was scanning **Peanut's own deposit address**, for which "send to it" is not a meaningful destination anyway. Flagging it rather than scope-creeping.

**Revealed, not introduced — uppercase URL schemes.** The `PEANUT_URL` branch strips the protocol with a case-sensitive regex (`index.tsx:286`), so `HTTPS://PEANUT.ME/satoshi` routes to `//PEANUT.ME/satoshi` and 404s. Same QR-uppercasing reason as above, so it is reachable. Pre-existing and untouched by this diff — filed for a follow-up rather than bundled here.

**Solana/Bitcoin label ambiguity.** A Solana address starting with `1` or `3` (~3.4%) matches the Bitcoin pattern first and is labelled "Bitcoin". Cosmetic, on a modal that says "not supported" either way, and unfixable without real checksum validation. Pre-existing.

## QA

`npm test` — 228 suites, 2925 tests, green.

New pin: `src/components/Global/QRScannerOverlay/__tests__/base58-case.test.tsx`. It discriminates all three candidate designs — only the shipped one passes all 8:

| variant | failures |
|---|---|
| original (lowercase everything) | 3 — both base58 cases + checksum laundering |
| raw-only (no uppercase retry) | 3 — all uppercase cases |
| **shipped** | **0** |

Manual: open the scanner, scan any Solana address containing an uppercase `L` (e.g. the QR on Add money → Crypto → Solana). Before: "Unrecognized QR code". After: "Solana not supported yet."

## Screenshots

N/A — no visible change. The modal that renders was already built; this PR only changes which one is reached.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved QR code scanning for case-sensitive addresses and payment payloads.
  * Preserved valid uppercase and mixed-case formats, including checksummed EVM addresses.
  * Added fallback handling for fully uppercase QR contents.
  * Improved routing for Peanut URLs.
  * Improved validation and handling of invalid address checksums.
  * Enhanced support for Solana, Tron, bech32, and BOLT11 QR payloads.

### DIFF
--- a/src/components/Global/QRScannerOverlay/__tests__/base58-case.test.tsx
+++ b/src/components/Global/QRScannerOverlay/__tests__/base58-case.test.tsx
@@ -13,7 +13,7 @@
  * the wiring in this component was wrong, so the guard has to live here.
  */
 import React from 'react'
-import { screen, waitFor } from '@testing-library/react'
+import { act, screen, waitFor } from '@testing-library/react'
 import { renderWithIntl } from '@/test-utils/intl'
 
 const mockPush = jest.fn()
@@ -64,10 +64,19 @@ import QRScannerOverlay from '../index'
 const SOLANA_WITH_UPPERCASE_L = '9WzDXwBbmkg8ZTbNMqUxvQRAyrZzDsGYdLVL9zYtAWWM'
 const TRON = 'TJRyWwFs9wTFGZg3JbrVriFbNfCug5tDeC'
 const EVM_CHECKSUMMED = '0x5aAeb6053F3E94C9b9A09f33669435E7Ef1BeAed'
+// EIP-55 says case only carries a checksum when the address is neither all-lower
+// nor all-upper, so this one is valid and must survive.
+const EVM_UPPERCASE = '0X5AAEB6053F3E94C9B9A09F33669435E7EF1BEAED'
+// Mixed case, so the checksum is real — and wrong. viem must keep rejecting it.
+const EVM_BAD_CHECKSUM = '0xAbCdEf1234567890123456789012345678901234'
+const BECH32_UPPERCASE = 'BC1QAR0SRRR7XFKVY5L643LYDNW9RE59GTZZWF5MDQ'
+const BOLT11_UPPERCASE = 'LNBC1230N1PJJ2LX9PP5ABC123'
 
 const scan = async (data: string) => {
     renderWithIntl(<QRScannerOverlay />)
-    await onScan(data)
+    await act(async () => {
+        await onScan(data)
+    })
 }
 
 describe('QRScannerOverlay case handling', () => {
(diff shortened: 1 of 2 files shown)


abalinda:
@coderabbitai review

abalinda:
@coderabbitai review

abalinda:
@coderabbitai full review

abalinda:
@coderabbitai full review

kushagrasarathe (approved):
Reviewed via pr-review rulebook (T2, full-diff read). Fix verified: all recognizeQr branches checked case-safe on raw input (ENS, PIX email/UUID, EMV regexes all case-tolerant); bad-checksum tightening consistent through the clipboard-extract recursion path. Tests pin the component wiring, not the parser. One accepted edge: lowercase-0x + all-caps-hex EVM address now rejected — measured zero in prod. Reminder: base is main, back-merge to dev after landing.

PR #2671, opened 2026-08-11
fix(seo): harden Split public content boundary

0xkkonrad:
[Task / rollout tracker](https://app.notion.com/p/3b9838117579812e860ad0dc7ee6c67d)

## Outcome

This is the single Peanut UI integration PR for the planned Peanut Split SEO work. Peanut UI owns only the public-host boundary: routing/firewall, privacy headers, Split asset and sitemap forwarding, parent service-worker isolation, robots exposure, and IndexNow. Split content, messaging, localization, rendering, and calculators remain in Mono and PeanutSplit.

The matching PeanutSplit renderer/schema PR is merged at `625203cf564df09d602e6fc5cb538bbf2cdb89b7`. This PR contains no production release controller or production mutation. Routes and indexing remain dark until a separate reviewed release document is deliberately configured.

## Included

- canonical manifest-backed stage 0/1/2/3 route authorization;
- fail-closed malformed, encoded, mixed-case, credential, recursion, and product-route boundaries;
- bodyless private/noindex failures, request-header minimization, and upstream `Set-Cookie` stripping;
- `/split-static/**` and `/split-sitemap.xml` transport with parent SW `NetworkOnly`;
- future-locale route grammar plus safe legacy-redirect carve-outs;
- robots exposure only for a complete index-enabled manifest;
- fail-closed IndexNow discovery, page/canonical/noindex validation, bounded reads, deletion notifications, and durable state.

## Explicitly excluded

- production release/deployment controller, Vercel promotion logic, or environment mutation;
- Split copy, localization, MDX/layouts, renderer, calculators, and generated content;
- C1-C3 content migration;
- any route release, indexing flip, Split-URL IndexNow submission, DNS change, or redirect cutover.

## Verification — exact head `11457f846bc688f60a1c5496826f516b2b59d417`

- full hosted test matrices, TypeScript, format, lint, E2E, supply-chain and reports: green;
- CodeQL and code analysis: green;
- Vercel Preview build/deploy: green;
- independent exact-head security/architecture review: pass;
- focused Split edge/proxy/redirect/robots/SW/Vercel/IndexNow: 707 tests passed;
- full Jest: exit 0 across 233 discovered suites;
- live official Vercel schema and diff/secret checks: passed;
- substantive CodeRabbit findings were triaged and valid items fixed; its final incremental run was skipped because the PR remained draft.

The configured stage-0/1/2 Preview matrix remains a release gate before setting the production release document; it is not required to merge this dark infrastructure.

[...]

PR #2676, opened 2026-08-12
seo: fix Googlebot disallows, missing canonicals, and 10 dropped receive-from pages

0xkkonrad:
Three independent SEO hygiene defects, batched because each is a few lines and they share no code.

## T1 — Googlebot ignored every disallow

**Defect.** The `Googlebot` group in `src/app/robots.ts` declared only `allow: ['/api/og']` with no `disallow` key. A crawler obeys the single most specific group that matches it, never the union — so Googlebot read that group *instead of* `*` and treated every auth-gated route (`/home`, `/profile`, `/settings`, `/kyc`, `/claim`, …) as fair game. The one crawler that matters was the one exempted. Live `robots.txt` confirms this shipped.

**Fix.** Extracted the `*` group's disallow list into a shared `DISALLOWED_PATHS` const and applied it to the Googlebot group too — mirroring how the AI-crawler group in the same file is built. The narrower `/api/og` allow still wins over `/api/` by longest-match, so OG images keep rendering in link previews.

## T2 — `/lp/card` and `/careers` declared the homepage as canonical

**Defect.** Both pages call the metadata helper without `canonical`. `src/app/layout.tsx` sets `alternates: { canonical: '/' }`, and `src/app/metadata.ts` only emits `alternates` when a canonical is passed — so both pages inherited the root value and told Google their canonical was `/` while sitting in the sitemap as indexable URLs. Self-canonical was never declared.

**Fix.** Pass the real path in each `generateMetadata()` call, matching the existing `src/app/exchange/layout.tsx` workaround (comment carried over). Neither route has a `layout.tsx`, so the fix belongs in `page.tsx`.

## T6 — 10 authored countries silently unreachable

**Defect.** `loadReceiveSources()` in `src/data/seo/corridors.ts` seeded from `CORRIDORS.map(c => c.from)` and then intersected with the receive-from content tree. But receive-from is authored independently of corridors, so the intersection dropped every article whose country was not also a corridor origin — 10 of them: australia, india, kenya, malaysia, netherlands, pakistan, philippines, saudi-arabia, singapore, united-arab-emirates. Written, published, and unreachable.

**Fix.** Enumerate published receive-from content directly via the existing `listPublishedSlugs('receive-from')` helper (same semantics the loader already applied: `en.md` present and `published !== false`). `RECEIVE_SOURCES` goes **9 → 19**. colombia and mexico stay excluded, correctly — they are corridor origins with no article, the original May-2026 404 case.

### The test that had to change

`src/data/seo/corridors.test.ts` is a checked-in regression guard from the May 2026 no-content incident, and two of its three assertions encoded the *old* invariant (`RECEIVE_SOURCES ⊆ CORRIDORS.from`). That subset relation was an artifact of how the list was built, not the thing that protected us. The real invariant is "every entry has a published article". Rewritten to assert exactly that, in both directions — nothing rendered without content (no 404s), nothing authored left behind (no orphans) — plus a duplicate check.

The new guard re-derives the expected set straight off the filesystem rather than reusing the loader's own helper, so it cannot pass by tautology. **I verified it has teeth** by temporarily restoring the old loader and confirming it goes red, naming the exact dropped countries:

```
✕ contains every published receive-from article (no orphaned content)
    -   "philippines",
        "portugal",
    -   "saudi-arabia",
    -   "singapore",
        "spain",
    -   "united-arab-emirates",
```

## Verification

```
$ npx pnpm@10.30.1 exec jest src/data/seo/corridors.test.ts --verbose
PASS src/data/seo/corridors.test.ts
  RECEIVE_SOURCES
    ✓ only contains slugs that have a published receive-from article (no 404s)
    ✓ contains every published receive-from article (no orphaned content)
    ✓ has no duplicate slugs

Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
```

```
$ npx pnpm@10.30.1 run test          # full suite
Test Suites: 229 passed, 229 total
Tests:       3 skipped, 2925 passed, 2928 total
```

```
$ npx pnpm@10.30.1 run lint
✖ 64 problems (0 errors, 64 warnings)   # all pre-existing, none in the 5 changed files

$ npx pnpm@10.30.1 exec prettier --check <the 5 changed files>
All matched files use Prettier code style!

$ npx pnpm@10.30.1 run typecheck     # tsc --noEmit
(clean, exit 0)
```

### T1 — generated `robots.txt` (served from this branch)

```
User-Agent: Googlebot
Allow: /api/og
Disallow: /api/
Disallow: /sdk/
Disallow: /home
Disallow: /profile
Disallow: /settings
...
Disallow: /add-money
Disallow: /withdraw
```

All 24 disallows now present on the Googlebot group, `/api/og` allow retained, `*` group unchanged.

### T2 — canonical `<link>` grepped out of the rendered HTML body

```
$ curl -s http://localhost:4137/careers | grep -o '<link[^>]*rel="canonical"[^>]*>'
<link rel="canonical" href="https://peanut.me/careers"/>     # HTTP 200, 59930 bytes

$ curl -s http://localhost:4137/lp/card | grep -o '<link[^>]*rel="canonical"[^>]*>'
<link rel="canonical" href="https://peanut.me/lp/card"/>     # HTTP 200, 101428 bytes
```

Checked the response headers too — the only `Link:` header is font preloads, so the canonical genuinely lives in the HTML, not a header. For contrast, production today:

```
$ curl -s https://peanut.me/careers | grep -o '<link[^>]*rel="canonical"[^>]*>'
<link rel="canonical" href="https://peanut.me"/>             # the bug
$ curl -s https://peanut.me/exchange | grep -o '<link[^>]*rel="canonical"[^>]*>'
<link rel="canonical" href="https://peanut.me/exchange"/>    # the precedent this fix copies
```

### T6 — the 10 recovered countries

```
RECEIVE_SOURCES count = 19
RECEIVE_SOURCES = argentina, australia, brazil, france, germany, india, italy, kenya,
  malaysia, netherlands, pakistan, philippines, portugal, saudi-arabia, singapore, spain,
  united-arab-emirates, united-kingdom, united-states
corridor origins = argentina, brazil, colombia, france, germany, italy, mexico, portugal,
  spain, united-kingdom, united-states
NEWLY REACHABLE (has article, not a corridor origin) = australia, india, kenya, malaysia,
  netherlands, pakistan, philippines, saudi-arabia, singapore, united-arab-emirates
origins WITHOUT article (correctly still excluded) = colombia, mexico
```

The user-visible effect is confirmable on production right now. `receive-money-from/[country]` calls `notFound()` for anything outside `RECEIVE_SOURCES`, and this site's custom 404 returns HTTP 200 — so the dropped pages are *soft* 404s and status codes tell you nothing. Comparing titles instead:

```
$ curl -s https://peanut.me/en/receive-money-from/singapore  → <title>Peanut - Instant Global P2P Payments in Digital Dollars</title>
$ curl -s https://peanut.me/en/receive-money-from/zzznotacountry → <title>Peanut - Instant Global P2P Payments in Digital Dollars</title>
$ curl -s https://peanut.me/en/receive-money-from/brazil     → <title>Receive Money from Brazil | Peanut</title>
```

singapore is byte-for-byte the same shell as a garbage URL; brazil is a real article. That is the defect, live.

**e2e was not run.** The Playwright suite needs a live API on `:5000` serving `/dev/test-session` for its `globalSetup`, which isn't available in this environment. No new e2e specs were written.

## Caveats for the reviewer

- **Expect a conflict with #2671** (base `main`), which also edits `robots.ts` — different hunks (sitemap field + an import), but it lands in `dev` via back-merge on its own schedule. My hunks are deliberately minimal.
- `RECEIVE_SOURCES` growing 9 → 19 adds 10 URLs to the sitemap. That is the intended effect: the content already exists and is already published, it just had no route. Worth a glance at the sitemap diff on the preview deploy.
- **`next build` never ran to completion in my sandbox, so CI is the first green signal on the production build specifically.** Not a code problem, and I checked rather than assumed: this box runs `earlyoom` with `--prefer (^|/)(node|next-server|esbuild)`, so under memory pressure from other agents building concurrently it SIGTERMs whichever node process is largest. Its own log names the victim — `sending SIGTERM to process ... "next-server": badness 1128, VmRSS 3383 MiB` — which is exactly the `Next.js build worker exited with code: null and signal: SIGTERM` reported across three build attempts. The dev server behind the render evidence above was killed the same way twice before it survived long enough to answer. Everything else (typecheck, lint, jest, prettier, and the live render evidence) is green locally.

## Summary by CodeRabbit

* **SEO Improvements**
  * Added canonical URLs for the Careers and Card landing pages.
  * Updated crawler guidance to better control access to API, SDK, and authentication routes while preserving image endpoint access.
  * Improved sitemap and route discovery for published receive-money content.

* **Bug Fixes**
  * Ensured all published receive-money source pages are represented accurately without duplicates or missing entries.

### DIFF
--- a/scripts/verify-content.ts
+++ b/scripts/verify-content.ts
@@ -74,14 +74,17 @@ function listDirs(dir: string): string[] {
 }
 
 /**
- * Receive-money-from pages render only for corridor origins that actually have
- * a receive-from article. Mirrors RECEIVE_SOURCES in src/data/seo/corridors.ts.
+ * Receive-money-from pages render for every published receive-from article.
+ * Mirrors RECEIVE_SOURCES in src/data/seo/corridors.ts (listPublishedSlugs):
+ * publication is gated on the article existing, not on corridor membership.
  * Without this gate, both the route index and the sitemap "expected URLs" would
  * agree with each other on a slug that 404s at runtime (e.g. colombia, mexico).
  */
-function gateReceiveSources(corridors: Array<{ from: string; to: string }>): string[] {
-    const origins = [...new Set(corridors.map((c) => c.from))]
-    return origins.filter((slug) => fs.existsSync(path.join(CONTENT_DIR, 'receive-from', slug, 'en.md')))
+function gateReceiveSources(): string[] {
+    return listDirs(path.join(CONTENT_DIR, 'receive-from')).filter((slug) => {
+        const en = path.join(CONTENT_DIR, 'receive-from', slug, 'en.md')
+        return fs.existsSync(en) && isPublished(fs.readFileSync(en, 'utf-8'))
+    })
 }
 
 function getAllMdFiles(dir: string): string[] {
@@ -193,7 +196,7 @@ function discoverRoutes(): Set<string> {
             corridors.push({ to: dest, from: origin })
         }
     }
-    const receiveSources = gateReceiveSources(corridors)
+    const receiveSources = gateReceiveSources()
 
     // Check which routes actually have page.tsx files
     const hasRoute = (routePath: string) => {
@@ -705,7 +708,7 @@ function expectedSitemapUrls(): string[] {
         const fromDir = path.join(CONTENT_DIR, 'send-to', dest, 'from')
         for (const origin of listDirs(fromDir)) corridors.push({ from: origin, to: dest })
     }
-    const receiveSources = gateReceiveSources(corridors)
+    const receiveSources = gateReceiveSources()
 
     for (const locale of SUPPORTED_LOCALES) {
         for (const slug of countrySlugs) {
--- a/src/app/careers/page.tsx
+++ b/src/app/careers/page.tsx
@@ -6,6 +6,9 @@ export const metadata = generateMetadata({
     description:
         'Explore career opportunities at Peanut. Join our team to build the future of fast, global peer-to-peer payments with digital dollars.',
     keywords: 'careers, jobs, employment, Peanut careers, P2P payments jobs, fintech jobs, crypto jobs, tech jobs',
+    // Without this the page inherits the root layout's `canonical: '/'` and
+    // declares the homepage as its canonical while sitting in the sitemap.
+    canonical: '/careers',
 })
 
 export default function CareersPage() {
--- a/src/app/lp/card/page.tsx
+++ b/src/app/lp/card/page.tsx
@@ -9,8 +9,9 @@ export const metadata = generateMeta({
         'Join Card Pioneers for early access to the Peanut Card. Reserve your spot with $10, earn $5 for every friend who joins, and spend your dollars globally.',
     keywords:
         'peanut card, card pioneers, crypto card, digital dollars, global spending, early access, referral rewards, international card',
-    // Without this the page inherits the root layout's `canonical: '/'` and
-    // declares the homepage as its canonical while sitting in the sitemap.
+    // Without this the page inherits lp/layout.tsx's deliberate `canonical: '/'`
+    // (the /lp subtree aliases the root landing page). That policy is wrong for
+    // /lp/card specifically: it's a distinct product page sitting in the sitemap.
     canonical: '/lp/card',
 })
 
--- a/src/app/lp/card/page.tsx
+++ b/src/app/lp/card/page.tsx
@@ -9,6 +9,9 @@ export const metadata = generateMeta({
         'Join Card Pioneers for early access to the Peanut Card. Reserve your spot with $10, earn $5 for every friend who joins, and spend your dollars globally.',
     keywords:
         'peanut card, card pioneers, crypto card, digital dollars, global spending, early access, referral rewards, international card',
+    // Without this the page inherits the root layout's `canonical: '/'` and
+    // declares the homepage as its canonical while sitting in the sitemap.
+    canonical: '/lp/card',
 })
 
 export default function CardLPPage() {
--- a/src/app/robots.ts
+++ b/src/app/robots.ts
@@ -4,6 +4,38 @@ import { SUPPORTED_LOCALES } from '@/i18n/types'
 
 const IS_PRODUCTION_DOMAIN = BASE_URL === 'https://peanut.me'
 
+// Paths kept out of the index: the API surface, the SDK bundle, and the
+// auth-gated app routes. Shared by every named crawler group — a group that
+// omits these silently opts that crawler out of ALL disallows, because a
+// user-agent only ever obeys the single most specific group that matches it.
+const DISALLOWED_PATHS = [
+    '/api/',
+    '/sdk/',
+    // Auth-gated app routes
+    '/home',
+    '/profile',
+    '/settings',
+    '/send',
+    '/request',
+    '/setup',
+    '/claim',
+    '/pay',
+    '/dev/',
+    '/qr',
+    '/history',
+    '/points',
+    '/rewards',
+    '/invite',
+    '/kyc',
+    '/maintenance',
+    '/quests',
+    '/receipt',
+    '/crisp-proxy',
+    '/card-payment',
+    '/add-money',
+    '/withdraw',
+]
+
 export default function robots(): MetadataRoute.Robots {
     // Block indexing on staging, preview deploys, and non-production domains
     if (!IS_PRODUCTION_DOMAIN) {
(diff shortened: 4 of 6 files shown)


0xkkonrad:
**Adversarial review pass (agent) — verdict: fixable, fixed in 622977ce9.**

Main catch: `scripts/verify-content.ts` `gateReceiveSources()` was a stale second copy of the loader this PR rewrites — its docstring claims it mirrors `RECEIVE_SOURCES` but it still did the corridor-origin intersection, so Pass 11 (the blocking sitemap-coverage CI check, added after the May-2026 mass-404 incident) silently stopped covering the 10 recovered slugs (script said 9, app said 19, both sides agreed with themselves). Now enumerates published receive-from articles directly; `validate-links` green locally, Pass 11 coverage 1115 → 1165 URLs, and script/app both return the same 19 slugs.

Also fixed two misleading comments the review flagged (robots.ts scope overclaim; lp/card canonical misattribution — it inherits `lp/layout.tsx`'s deliberate alias policy, not the root layout's).

Follow-up noted, not in this PR: the 10 recovered pages have zero inbound internal links from published content — tracked in the content workstream.

abalinda:
@coderabbitai review

abalinda:
## `/code-review medium`

The three fixes are correct. I checked each one:

- The Googlebot group now carries the full disallow list, and `/api/og` still wins by longest match.
- Both `canonical` additions flow through `generateMetadata` → `alternates.canonical` and override the layout-level `canonical: '/'`. Page metadata replaces the `alternates` key, and neither layout sets `languages`.
- `listPublishedSlugs('receive-from')` reproduces the old gate (`en.md` present, `published !== false`) without the corridor intersection. All 10 newly reachable articles have a `content/countries/` hub for `getCountryName`, use only MDX components registered in `createMdxComponents`, and their internal links resolve to real routes. No sitemap URL is caught by a prefix in `DISALLOWED_PATHS`.

Six findings below. All low. None blocking.

---

### 1. `src/app/robots.ts:54` — Twitterbot has the same footgun this PR fixes for Googlebot

The Twitterbot group sets `disallow: []`, so Twitterbot obeys that group **instead of** `*`. Every auth-gated route (`/home`, `/claim`, `/kyc`, `/card-payment`) stays crawlable for it. The new `DISALLOWED_PATHS` const makes this a one-line fix.

As it stands, the new comment ("a named group that omits a path silently opts that crawler out of it") documents a bug that is still live two lines below it. Impact is lower than Googlebot, because Twitterbot fetches link previews and does not index.

### 2. `src/app/robots.ts:80` — AI-crawler group omits the money-link routes

The narrower list leaves out `/claim`, `/qr`, `/receipt`, `/history`, `/invite`, `/card-payment`, `/add-money`, `/withdraw`. GPTBot, ClaudeBot, and PerplexityBot can crawl claim links and receipts that every other crawler is blocked from. This is pre-existing, but this PR is the moment the intent gets written down.

### 3. `src/app/robots.ts:66` — confirm the Googlebot behavior change before merge

Not a defect. Closing Googlebot's access also blocks `/invite`, which is the CTA target of every marketing article (`peanut.me/invite?code=…`), plus `/pay`, `/qr`, and `/claim`. Googlebot had no disallow list before, so these prefixes go from crawlable to blocked. If any of them have indexed pages or organic traffic today, they drop out of the index after this ships. Worth a Search Console check on those prefixes.

### 4. `src/data/seo/corridors.ts:95` — no guard against a meta directory

`listPublishedSlugs('receive-from')` takes every directory in the tree. The other consumers guard against this: `sitemap.ts:165` and `sitemap.ts:220` both `continue` on `slug === 'index'`.

The content submodule is bumped on its own schedule. An `index/` or template directory added there with an `en.md` becomes a live route and a sitemap URL with no code change here. The new test blesses it instead of catching it. The old corridor intersection prevented this by accident.

### 5. `src/data/seo/corridors.test.ts:17` — frontmatter regex diverges from the loader

The test parses frontmatter with `/^published:\s*false\s*$/m`. The loader uses gray-matter, which parses YAML. The two disagree on `published: False`, on `published: false # note`, and on a `published: false` nested under another key.

Concretely: an author writing `published: False` makes the loader drop the slug while the test still expects it. CI turns red on a content-only change, with a confusing set diff as the only clue. Deriving the set independently of the loader is the right call — the parser just needs to match YAML semantics for these few cases.

### 6. `src/data/seo/corridors.test.ts:12` — raw ENOENT when the content submodule is not initialized

`fs.readdirSync(RECEIVE_FROM_DIR)` throws when `src/content` is empty, which is a documented worktree trap. The loader swallows this through try/catch and returns `[]`, so the suite used to pass (vacuously) in that state. It now fails with a raw ENOENT that does not name the real cause. CI checks out submodules, so this only hits local worktrees.

---

*Automated review. Findings verified against the branch head.*

0xkkonrad:
Consolidated into #2685 (one commit per fix, review catches included) at the CTO's request — this PR's evidence and review thread remain the reference for its slice. Branch kept until #2685 merges.

PR #2684, opened 2026-08-12
seo: report real content dates as sitemap lastmod

0xkkonrad:
## Defect (T5)

`src/app/sitemap.ts` set `lastModified` to `BUILD_DATE` for **every** URL:

```ts
const BUILD_DATE = new Date()
...
lastModified: page.lastModified ?? BUILD_DATE   // no call site ever passed lastModified
```

`SitemapEntry.lastModified` existed but was dead — nothing populated it. So every deploy
republished all 709 URLs claiming they had just changed. A `lastmod` that always equals the
deploy timestamp is noise: crawlers learn nothing from it, and the pages that genuinely did
change lose the signal they should have had.

## Fix

Content-backed URLs now report the `generated_at` of the exact file that serves them.

- `contentGeneratedAt()` in `src/lib/content.ts` — coerces a content file's `generated_at`
  frontmatter to a `Date`, returning `undefined` when missing or unparseable.
- Three one-line lookups in `sitemap.ts` (`pageDate` / `corridorDate` / `singletonDate`) pair
  the right reader with that coercion. The readers were **not** all imported before; this adds
  `readPageContent`, `readCorridorContent`, `readSingletonContent`.
- 15 of the 20 `pages.push()` call sites now pass a real date. Hand-built pages (homepage,
  `/lp/card`, `/careers`, `/exchange`, legal) and index pages with no single backing file keep
  `BUILD_DATE` via the existing `?? BUILD_DATE` fallback — untouched.

The lookups read through the same cache the `has*Content()` guards already populate, so they
add no extra file reads.

### One subtlety worth reviewing

`ContentFrontmatter` declares `generated_at?: string`, but that is **wrong at runtime**:
gray-matter runs js-yaml, which parses unquoted YAML timestamps into JS `Date` objects. Proven
against the real submodule:

```
src/content/content/blog/earn-with-peanut-3min-setup/en.md
  YAML source : generated_at: 2026-03-27
  typeof      : object   instanceof Date: true

src/content/content/send-to/argentina/from/colombia/en.md
  YAML source : generated_at: 2026-02-21T15:00:00Z
  typeof      : object   instanceof Date: true
```

A naive `new Date(frontmatter.generated_at)` would still work, but the helper handles `Date`,
quoted string, missing and unparseable inputs explicitly so this cannot regress silently.

I deliberately did **not** correct the declared type: `generated_at` is consumed as JSON-LD
`datePublished` in 12 page files that expect a string, so retyping it belongs in its own PR.
[...]

PR #2682, opened 2026-08-12
seo: route /pricing /stories /content, fix /help subpaths + locale trailing-slash dupes

0xkkonrad:
Part of the SEO code track (12 Aug 2026). Redirect/reserved-route layer only — no page or content changes. Sibling PRs in the same track touch `robots.ts`/canonicals/corridors, the username catch-all's `robots` meta, and `sitemap.ts`.

## Defects

**1. `/pricing`, `/stories`, `/content` render a payment profile, not the page they name**

All three ship as real pages under `src/app/[locale]/(marketing)/`, but only at their locale-prefixed paths. The bare paths are seven lowercase letters, so `couldBeRecipient()` (src/constants/routes.ts) accepted them as usernames and `src/app/[...recipient]/page.tsx` served a payment-profile shell — HTTP 200, wrong content, indexable.

```
$ curl -so /dev/null -w '%{http_code}\n' https://peanut.me/pricing   # today, production
200
```

**2. `/help/:path*` soft-404s — including the Google Play account-deletion URL**

`redirects.json` redirected bare `/help` but never its subpaths, so every help article at its bare path fell into the same catch-all shell. `/help/delete-account` is the account-deletion URL required by Google Play's data-safety policy, and it returns a 200 payment shell today:

```
$ curl -so /dev/null -w '%{http_code}\n' https://peanut.me/help/delete-account   # today, production
200
$ curl -so /dev/null -w '%{http_code}\n' https://peanut.me/en/help/delete-account
200
```

**3. Trailing-slash duplicates across the whole locale tree**

`skipTrailingSlashRedirect: true` (next.config.js) turns off Next's global trailing-slash redirect, so `/en/help/` and `/en/help` both return 200 — two URLs, one page, for every marketing page in every locale.

```
$ curl -so /dev/null -w '%{http_code}\n' https://peanut.me/en/help/   # today, production
200
```

**4. `docs.peanut.to` has no host rule.** `docs.peanut.me` has had one since the docs site was retired; the `.to` twin was never added. It is currently limping along on a Namecheap URL-forward that only works over plain HTTP — `https://docs.peanut.to/` times out, because Namecheap forwarding has no certificate for it:

```
$ curl -sI http://docs.peanut.to/
HTTP/1.1 301 Moved Permanently
Location: https://peanut.me/en/help
X-Served-By: Namecheap URL Forward

$ curl -sI https://docs.peanut.to/
curl: (28) Connection timed out after 20002 ms
```

## Fixes

[...]

PR #2681, opened 2026-08-12
seo: return real HTTP 404s from the username catch-all

0xkkonrad:
## Defect (T4) — the site cannot return a real 404

`src/app/[...recipient]/loading.tsx` (5 lines, rendered `<PeanutLoading />`) created a Suspense boundary on the username catch-all segment. `Page()` suspends on `use(props.params)`, so React rendered the fallback, **flushed the shell with HTTP 200**, and only *then* resolved `notFound()` — which arrived as a `NEXT_HTTP_ERROR_FALLBACK;404` marker inside an already-committed 200 response.

The catch-all owns every unmatched path, so *every* 404 on peanut.me was a soft 404: HTTP 200 + noindex. That breaks Google's 404 signal, GSC coverage reporting and every external link checker.

Verified on production before the fix:

```
$ curl -s -o body.html -w "status=%{http_code}\n" https://peanut.me/zzzfake-user-98765
status=200
$ grep -o "NEXT_HTTP_ERROR_FALLBACK;404" body.html
NEXT_HTTP_ERROR_FALLBACK;404
```

## Fix

Delete `src/app/[...recipient]/loading.tsx`. With no boundary at that segment, the suspension propagates to the shell, so `notFound()` throws **before** headers flush and Next sets a real 404 status.

Why this is safe to remove rather than replace:

- The route's server render is otherwise synchronous — `Page()` only runs the `isReservedRoute` / `couldBeRecipient` guards after `use(params)`.
- `PaymentPage` is a client component (`./client`), so no server data fetch is left uncovered; its own loading states are unchanged.
- No PPR / `cacheComponents` in `next.config.js`, and no `src/app/loading.tsx` — this was the only boundary in the route's ancestry, and no middleware sits in front.
- Nothing imports the file (`grep -rn "recipient\]/loading" src` → empty). `PeanutLoading` itself stays and is used in 15+ other places.

Cost: the loading spinner no longer paints while the (fast, guard-only) server render resolves. That is the price of a correct status code.

## Verification evidence

`next build` cannot run on the dev box, so everything below is measured against the **Vercel Deploy Preview for this PR** (`peanut-wallet-git-seo-real-404s-squirrellabs.vercel.app`, deployment `2eSJX7qLtkbjijSfdiSwL8Fq3jF3`), compared side by side with production.

### CI — all green

[...]

PR #2678, opened 2026-08-12
seo: noindex the username catch-all route

0xkkonrad:
## Defect (T3, layer 1)

`src/app/[...recipient]/page.tsx` is the username catch-all. Its `generateMetadata` builds a full, **indexable** "X on Peanut" shell for *any* username-shaped string — `USERNAME_PATTERN = /^[a-z][a-z0-9]{3,11}$/` in `src/constants/routes.ts:155` is the only filter. Whether the account exists is checked **client-side**, after the HTML has already been served with HTTP 200.

Two consequences, both live on peanut.me right now:

```
$ curl -s https://peanut.me/kkonrad | grep -o '<meta name="robots"[^>]*>'
<meta name="robots" content="index, follow"/>
$ curl -s https://peanut.me/kkonrad | grep -o '<title>[^<]*</title>'
<title>kkonrad on Peanut</title>

$ curl -s -o /dev/null -w '%{http_code}\n' https://peanut.me/zzqx9      # nobody owns this handle
200
$ curl -s https://peanut.me/zzqx9 | grep -o '<meta name="robots"[^>]*>'
<meta name="robots" content="index, follow"/>
$ curl -s https://peanut.me/zzqx9 | grep -o '<title>[^<]*</title>'
<title>zzqx9 on Peanut</title>
```

So Google is offered an unbounded space of soft-404 "profile" pages, plus every real profile, address, ENS name and request/receipt link — all app surface, none of it a search landing page.

The `index, follow` comes from the root layout (`src/app/layout.tsx:24`, `robots: IS_PRODUCTION_DOMAIN ? { index: true, follow: true } : …`). That is why the empty `return {}` guard branches are *not* already safe: with nothing set at page level they inherit `index, follow` from the root.

## Fix

`robots: { index: false, follow: false }` on **all four** `generateMetadata` return paths (count re-verified on current `origin/dev` @ `ad5b61b6` — `grep -n return` shows exactly 3 early returns plus the metadata object):

| line (pre-patch) | branch |
|---|---|
| 26 | reserved route |
| 31 | no recipient segment |
| 37 | `!couldBeRecipient(...)` |
| 160–186 | full metadata object (profiles, addresses, ENS, amounts, receipts) |

Extracted as one `NOINDEX` const so the four branches cannot drift apart.

Per the 12 Aug decision: **noindex everything, no carve-outs** — real profiles and `.eth` ENS pages included.

[...]

PR #2685, opened 2026-08-12
seo: technical fix bundle — robots, canonicals, receive-from, noindex catch-all, real 404s, redirects, sitemap lastmod

0xkkonrad:
Consolidates the five SEO fix PRs (#2676, #2678, #2681, #2682, #2684) into one PR at the CTO's request — **one commit per fix**, so any single change can be reverted surgically. Every fix was implemented and then adversarially reviewed by an independent agent that re-derived the evidence itself; all review catches are already folded in as their own commits. Full evidence lives on the five closed PRs; the load-bearing parts are summarized below.

## What's in here (commit order)

**1. `2648b5aaf` — robots.txt, canonicals, receive-from loader** (from #2676)
- `robots.ts`: the Googlebot group had only `Allow: /api/og`, so per the robots spec Googlebot ignored all 24 `*` disallows. Now shares the disallow list.
- `/careers` + `/lp/card` emitted `rel=canonical` → homepage (inherited); both now self-canonicalize. Verified by grepping the canonical `<link>` from rendered preview HTML vs the production defect.
- `corridors.ts`: `loadReceiveSources()` intersected receive-from content with corridor origins, silently dropping **10 fully-authored countries** (australia, india, kenya, malaysia, netherlands, pakistan, philippines, saudi-arabia, singapore, UAE). Now enumerates published content directly: `RECEIVE_SOURCES` 9 → 19. `corridors.test.ts` rewritten to the invariant it actually protects (every slug has published content — the May-2026 incident was the reverse). Review re-derived the counts and confirmed the new guard goes red on the old loader.

**2. `936b012f9` — review catch on the above:** `scripts/verify-content.ts` had a stale second copy of the old loader — Pass 11 (the **blocking** sitemap-coverage CI check) would have silently stopped covering the 10 recovered slugs. Script and app now both return the same 19; Pass 11 coverage 1115 → 1165 URLs.

**3. `153c6bae5` — noindex the username catch-all** (from #2678)
Any username-shaped string (`/brazil`, `/pricing`, `/zzzfake1`) rendered an indexable "X on Peanut" shell — the existence check is client-side, invisible to crawlers, and the root layout injects `index, follow`, so Google indexed thousands. All 4 `generateMetadata` return paths now ship `robots: {index:false, follow:false}`. Decision on record (12 Aug): no carve-outs, including real profiles and `.eth` pages.

**4. `69b3e5dbf` — HIGH review catch on the above:** the noindex was shipping alongside the inherited `rel=canonical` → homepage — a combination Google's guidance warns can attribute the noindex to the canonical cluster head (the homepage, sitemap priority 1.0). `alternates: {canonical: null}` on all four branches; noindex pages now carry no canonical.

**5. `5cb458a46` — real HTTP 404s** (from #2681)
`[...recipient]/loading.tsx` made the catch-all a Suspense boundary, so Next streamed an HTTP 200 shell before `notFound()` resolved — the site never returned a real 404 (breaks GSC soft-404 reporting, link tools, crawl budget). One file deleted; the route's render path was confirmed to have no other boundary. Preview-verified: garbage URL → **404**, real pages → 200, profiles → 200. Review verdict SHIP after independently re-probing.

**6. `74ca6e544` + **7.** `f5034b218` — redirects bundle** (from #2682)
- `/pricing`, `/stories`, `/content` added to `DEDICATED_ROUTES` + bare→`/en/…` redirects (they were being squatted by the username catch-all).
- `/help/:path*` bare→locale redirect — `/help/delete-account` (the **Google Play account-deletion compliance URL**) soft-404'd; it now resolves to the real article.
- `docs.peanut.to` host-redirect rule (inert until ops attaches the domain — see asks).
- Trailing-slash 308s scoped to the locale content tree; `skipTrailingSlashRedirect` untouched (protects the PostHog `/relay/*` proxy — probed: still proxies).
- Review caught a genuine blocker in the first cut: `:path*` matched zero segments, so `/es-ar/`, `/es-419/`, `/pt-br/` 308-looped to themselves (measured on the preview; browsers cache 308s permanently). Fixed with `:path+`; re-probed — bare locales 200, subpaths still redirect, `follow /es-ar/` = 0 hops.

**8. `8ce2b769e` — real sitemap lastmod** (from #2684)
All 709 URLs reported the deploy timestamp as `lastmod`, so Google discounts the signal. Dates now come from each page's `generated_at` frontmatter; static/legal pages keep BUILD_DATE. Review independently cross-checked **686 URLs against the content tree: 0 mismatches**; dev deploy serves 1 distinct lastmod, this branch serves 26, identical URL set.

**9. `d2765ee81`** — stale comment cleanup flagged by review.

## Verification status
- Local (this branch): jest (corridors + content) green, `validate-links` all passes (Pass 11 = 1165 URLs), `tsc --noEmit` exit 0, eslint 0 errors, all 9 commits SSH-signed.
- `next build` cannot run on the dev box (earlyoom) — CI + the Vercel preview are the build gates; each constituent branch already had them green. A fresh probe pass runs on THIS PR's preview after CI; evidence will be posted as a comment.
- Note preview limitations: `robots.txt` and the `noindex` meta can't be observed on previews (non-production BASE_URL noindexes everything) — those two rest on code + the per-branch evidence.

## Asks / follow-ups (not in this PR)
- **Backend:** add `stories` and `content` to the server-side reserved-username list (`pricing` is already reserved; the other two are claimable today, and frontend routing alone doesn't stop a signup claim).
- **Ops (needs access):** attach `docs.peanut.to` to this Vercel project + repoint DNS; flip the redirect.pizza rule for `peanut.to` deep paths to path-preserving.
- **Post-deploy checks:** X/Twitter card unfurls on payment-request/receipt links (they now carry noindex; most unfurlers ignore it, Twitterbot historically sometimes doesn't). Legal pages still report deploy-time lastmod (`last_updated` frontmatter isn't read — small follow-up).
- **Expected conflicts:** #2671 (base main) touches `robots.ts`/`redirects.json` in non-overlapping hunks; #2605 adds `creator-contest` to `DEDICATED_ROUTES` in the same region (trivial).
- The 10 recovered receive-from pages have zero inbound internal links — tracked in the content workstream.

Replaces #2676, #2678, #2681, #2682, #2684 (closed with pointers; branches kept until this merges).

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -298,6 +298,18 @@ let nextConfig = {
         return config
     },
     reactStrictMode: false,
+    // Do NOT remove. Next's built-in trailing-slash redirect is global, and the
+    // PostHog reverse proxy below (/relay/*) is hit with trailing slashes by the
+    // SDK's POSTs (/relay/decide/, /relay/e/). A 308 on those either drops the
+    // body or costs every event an extra round trip, so the automatic redirect
+    // stays off.
+    //
+    // The SEO problem it leaves behind — /en/help/ and /en/help both returning
+    // 200 — is solved narrowly instead: redirects.json ends with a
+    // `/:locale(en|es-419|es-ar|pt-br)/:path*/` -> slashless permanent (308)
+    // redirect, which only covers the locale-prefixed marketing tree and cannot
+    // touch /relay, /monitoring, /passkeys or the recipient catch-all. Keep that
+    // locale list in sync with SUPPORTED_LOCALES (src/i18n/types.ts).
     skipTrailingSlashRedirect: true,
     async rewrites() {
         return {
--- a/redirects.json
+++ b/redirects.json
@@ -207,8 +207,8 @@
         "permanent": true
     },
     {
-        "source": "/:locale(en|es-419|es-ar|pt-br)/:path*/",
-        "destination": "/:locale/:path*",
+        "source": "/:locale(en|es-419|es-ar|pt-br)/:path+/",
+        "destination": "/:locale/:path+",
         "permanent": true
     }
 ]
--- a/redirects.json
+++ b/redirects.json
@@ -45,6 +45,11 @@
         "permanent": false
     },
     {
+        "source": "/help/:path*",
+        "destination": "/en/help/:path*",
+        "permanent": true
+    },
+    {
         "source": "/terms",
         "destination": "/en/terms",
         "permanent": false
@@ -53,6 +58,26 @@
         "source": "/privacy",
         "destination": "/en/privacy",
         "permanent": false
+    },
+    {
+        "source": "/pricing",
+        "destination": "/en/pricing",
+        "permanent": true
+    },
+    {
+        "source": "/stories",
+        "destination": "/en/stories",
+        "permanent": true
+    },
+    {
+        "source": "/stories/:path*",
+        "destination": "/en/stories/:path*",
+        "permanent": true
+    },
+    {
+        "source": "/content",
+        "destination": "/en/content",
+        "permanent": true
     },
     {
         "source": "/:slug(card-terms-us|card-terms-international|card-esign|card-privacy|card-prohibited-activities)",
@@ -75,6 +100,17 @@
             {
                 "type": "host",
                 "value": "docs.peanut.me"
+            }
+        ],
+        "destination": "https://peanut.me/en/help",
+        "permanent": true
+    },
+    {
+        "source": "/:path*",
+        "has": [
+            {
+                "type": "host",
+                "value": "docs.peanut.to"
             }
         ],
         "destination": "https://peanut.me/en/help",
@@ -168,6 +204,11 @@
     {
         "source": "/:locale/deposit/from-spei",
         "destination": "/:locale/deposit/via-spei",
+        "permanent": true
+    },
+    {
+        "source": "/:locale(en|es-419|es-ar|pt-br)/:path*/",
+        "destination": "/:locale/:path*",
         "permanent": true
     }
 ]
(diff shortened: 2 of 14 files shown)


0xkkonrad:
**Response to @abalinda's six findings from #2676** (they all transfer here — this PR contains that code):

**1. Twitterbot (fixed — `8057851c2`):** the empty disallow is *deliberate* — Twitterbot fetches user-shared claim/payment links to render X cards and doesn't index; restricting it would break unfurls on exactly the links users share. Now documented in the file so the comment no longer denies the exemption.

**2. AI-crawler group (fixed — `8057851c2`):** agreed and applied — the AI group now shares `DISALLOWED_PATHS`. AI engines should read and cite content pages, not claim links and receipts. Note the trade-off accepted here: `ChatGPT-User` (user-initiated fetches) also loses access to transactional URLs.

**3. Googlebot newly blocks `/invite` `/pay` `/qr` `/claim` `/home` — ruled INTENDED, ship as-is (Konrad).** GSC can't be checked yet (access grant is a pending workstream item), so we pulled PostHog instead: google-referred visitors on those prefixes over 90 days = **586 total** (`/home` 350 — navigational, those users will land on the homepage; `/invite` 231 ≈ 77/mo — likely indexed referral URLs; `/qr-pay` 5). The disallow list was always the site's declared policy — Googlebot ignoring it was the defect — and user-specific invite URLs in the index are junk surface. If we ever want `/invite` crawlable it's a one-line carve-out.

**4. `index/` meta-directory guard (fixed — `31bced6ba`):** loader, verify-content gate, and test all skip `index`, matching sitemap.ts's existing guard.

**5. Frontmatter regex vs YAML (fixed — `31bced6ba`):** the test now parses with gray-matter directly (still not via the loader's helper, so independence is preserved) — `published: False`, trailing comments, and nesting now agree with the loader.

**6. Submodule ENOENT (fixed — `31bced6ba`):** the suite now fails with `run: git submodule update --init` instead of a raw ENOENT.

Gates on the new head: jest 3/3, `validate-links` all passes, `tsc` clean, eslint 0 errors, prettier clean, both commits SSH-signed.

PR #2687, opened 2026-08-12
fix(history): dedupe entries when flattening infinite-query pages

abalinda:
## Summary

The history page concatenates infinite-query pages with a bare `flatMap`. Anything the API cursor serves on two pages therefore renders twice — byte-identical rows, one under the other, which reads as a double charge.

Adds `dedupeHistoryEntriesByUuid()` and applies it where the pages get flattened. Keeps the first position (rows must not jump on scroll) and the last copy (a later page is the fresher fetch, so a boundary row that was `PENDING` on page 1 renders as `COMPLETED`).

## Task

TASK-21039 — https://app.notion.com/p/3ae83811757981348353f498ee69eeb5

## Evidence

Reported by a user via support (Crisp thread linked on TASK-21039), with a screenshot of two identical rows for one PIX payment. The prod ledger has exactly **one** debit for it. Nothing was double-charged; the feed drew one payment twice.

## Why here and not only in the API

The API-side cursor bug is documented and guarded in peanut-api-ts#1334, but page overlap has more than one source:

1. the API cursor is **deliberately over-inclusive** — it is keyed on the later of `timestamp`/`createdAt` so that no entry can fall through the crack between pages, and re-serving the boundary row is the price. Tightening it silently drops payments, which peanut-api-ts#1334 documents and locks with a test;
2. `openRequestLinks` in the intent fetcher re-queries with `createdAt <= dateCursor` and no tie-break exclusion at all;
3. websocket-pushed entries racing a REST page.

Deduping at the flatten covers all three, needs no deploy-order coupling with the API, and the websocket path in the same file already dedupes by uuid for the same reason — this makes the REST path consistent with it.

## Risks

Low. Pure function, no network or state change. Two entries only collide if they share a uuid, which by contract means they are the same entry.

## Screenshots

N/A — no visible change. The fix removes a duplicated row; nothing new renders.

## QA

`npm test` — 4 new cases in `src/utils/__tests__/history.utils.test.ts` (boundary repeat, first-position/freshest-copy, distinct-but-similar entries, empty feed). Full suite green.

### DIFF
--- a/src/app/(mobile-ui)/history/page.tsx
+++ b/src/app/(mobile-ui)/history/page.tsx
@@ -33,7 +33,7 @@ import { useInfiniteScroll } from '@/hooks/useInfiniteScroll'
 import { TRANSACTIONS } from '@/constants/query.consts'
 import type { HistoryEntry, HistoryResponse } from '@/hooks/useTransactionHistory'
 import { AccountType } from '@/interfaces/interfaces'
-import { completeHistoryEntry } from '@/utils/history.utils'
+import { completeHistoryEntry, dedupeHistoryEntriesByUuid } from '@/utils/history.utils'
 import { formatUnits } from 'viem'
 import { PEANUT_WALLET_TOKEN_DECIMALS } from '@/constants/zerodev.consts'
 
@@ -161,7 +161,12 @@ const HistoryPage = () => {
         },
     })
 
-    const allEntries = useMemo(() => historyData?.pages.flatMap((page) => page.entries) ?? [], [historyData])
+    // Deduped: page overlap on the API cursor would otherwise render the same
+    // payment twice in a row, which reads as a double charge.
+    const allEntries = useMemo(
+        () => dedupeHistoryEntriesByUuid(historyData?.pages.flatMap((page) => page.entries) ?? []),
+        [historyData]
+    )
 
     const combinedAndSortedEntries = useMemo(() => {
         if (isLoading) {
--- a/src/utils/__tests__/history.utils.test.ts
+++ b/src/utils/__tests__/history.utils.test.ts
@@ -175,8 +175,8 @@ describe('getAvatarUrl', () => {
 /**
  * The history page concatenates infinite-query pages. If the API cursor serves
  * an entry on two pages, the row renders twice — byte-identical, so it reads
- * as a double charge. Real report: a BRL 909.00 PIX payment shown twice with
- * one debit in the database (2026-07-27).
+ * as a double charge. Users have reported exactly that on a PIX payment the
+ * ledger had only debited once.
  */
 describe('dedupeHistoryEntriesByUuid', () => {
     const entry = (uuid: string, amount = '1') => ({ uuid, amount })
--- a/src/utils/__tests__/history.utils.test.ts
+++ b/src/utils/__tests__/history.utils.test.ts
@@ -12,7 +12,13 @@
  * If a future refactor reintroduces a wei assumption here, this fails.
  */
 
-import { completeHistoryEntry, getAvatarUrl, getReceiptUrl, getTransactionSign } from '../history.utils'
+import {
+    completeHistoryEntry,
+    dedupeHistoryEntriesByUuid,
+    getAvatarUrl,
+    getReceiptUrl,
+    getTransactionSign,
+} from '../history.utils'
 import type { HistoryEntry } from '../history.utils'
 import type { TransactionDetails } from '@/components/TransactionDetails/transactionTransformer'
 
@@ -163,5 +169,37 @@ describe('getAvatarUrl', () => {
 
     it('falls back to undefined (generic avatar) for historical DEPRECATED_SIMPLEFI rows', () => {
         expect(getAvatarUrl(tx('DEPRECATED_SIMPLEFI', 'ARS'))).toBeUndefined()
+    })
+})
+
+/**
+ * The history page concatenates infinite-query pages. If the API cursor serves
+ * an entry on two pages, the row renders twice — byte-identical, so it reads
+ * as a double charge. Real report: a BRL 909.00 PIX payment shown twice with
+ * one debit in the database (2026-07-27).
+ */
+describe('dedupeHistoryEntriesByUuid', () => {
+    const entry = (uuid: string, amount = '1') => ({ uuid, amount })
+
+    it('drops a boundary row the API served on two pages', () => {
+        const page1 = [entry('a'), entry('b'), entry('c')]
+        const page2 = [entry('c'), entry('d')] // 'c' repeated at the page boundary
+
+        expect(dedupeHistoryEntriesByUuid([...page1, ...page2]).map((e) => e.uuid)).toEqual(['a', 'b', 'c', 'd'])
+    })
+
+    it('keeps the first copy, so newer page-1 data wins over a stale refetch', () => {
+        const deduped = dedupeHistoryEntriesByUuid([entry('a', '909.00'), entry('a', '0')])
+
+        expect(deduped).toHaveLength(1)
+        expect(deduped[0].amount).toBe('909.00')
+    })
+
+    it('keeps distinct entries that only look alike', () => {
+        expect(dedupeHistoryEntriesByUuid([entry('a', '5'), entry('b', '5')])).toHaveLength(2)
+    })
+
+    it('handles an empty feed', () => {
+        expect(dedupeHistoryEntriesByUuid([])).toEqual([])
     })
 })
--- a/src/utils/__tests__/history.utils.test.ts
+++ b/src/utils/__tests__/history.utils.test.ts
@@ -188,20 +188,30 @@ describe('dedupeHistoryEntriesByUuid', () => {
         expect(dedupeHistoryEntriesByUuid([...page1, ...page2]).map((e) => e.uuid)).toEqual(['a', 'b', 'c', 'd'])
     })
 
-    it('keeps the first position but the freshest copy', () => {
-        // Page 2 is fetched after page 1, so its copy of a repeated entry has
-        // the newer status. Position must not move, or rows jump on scroll.
-        const page1 = [
-            { uuid: 'a', status: 'PENDING' },
+    it('keeps a websocket-prepended copy over a stale one from a later page', () => {
+        // The websocket prepends full updated entries to page 0, so for that
+        // source the FRESHEST copy is the first one. An in-flight page-2
+        // response carrying the pre-update row must not overwrite it.
+        const page0 = [
+            { uuid: 'a', status: 'COMPLETED' }, // just pushed over the socket
             { uuid: 'b', status: 'COMPLETED' },
         ]
-        const page2 = [{ uuid: 'a', status: 'COMPLETED' }]
+        const page2 = [{ uuid: 'a', status: 'PENDING' }] // already in flight, stale
 
-        const deduped = dedupeHistoryEntriesByUuid([...page1, ...page2])
+        const deduped = dedupeHistoryEntriesByUuid([...page0, ...page2])
 
         expect(deduped).toHaveLength(2)
         expect(deduped[0]).toEqual({ uuid: 'a', status: 'COMPLETED' })
-        expect(deduped[1].uuid).toBe('b')
+    })
+
+    it('keeps the fuller request-pot rollup from the earlier page', () => {
+        // Rollups reuse link.uuid on every page holding any of that pot's
+        // charges, and each copy aggregates only its own page window. The
+        // earlier copy holds the most recent charge and the larger total.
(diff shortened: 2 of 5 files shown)


abalinda:
@coderabbitai review

src/app/(mobile-ui)/history/page.tsx
```diff
@@ -161,7 +161,12 @@ const HistoryPage = () => {
         },
     })
 
-    const allEntries = useMemo(() => historyData?.pages.flatMap((page) => page.entries) ?? [], [historyData])
+    // Deduped: page overlap on the API cursor would otherwise render the same
+    // payment twice in a row, which reads as a double charge.
+    const allEntries = useMemo(
+        () => dedupeHistoryEntriesByUuid(historyData?.pages.flatMap((page) => page.entries) ?? []),
```
jjramirezn, line 167: **[CONFIRMED] Infinite auto-fetch loop when a page adds zero new uuids**

Deduping here removes the list growth that paced infinite scroll. If the API serves a page whose entries are all already-seen uuids and the cursor does not advance, `allEntries` stops growing, the loader stays in the viewport, and `useInfiniteScroll` refires `fetchNextPage` in an unbounded automatic loop.

Concrete trigger: a user with >= 20 open request links (or 20+ timestamp-drifted rows). The `openRequestLinks` re-query (api `transaction-intent/history.ts:1381-1392`) re-serves up to `limit` rollups on every page with no exclusion, so the sorted top-20 can repeat exactly. Then `buildHistoryCursor` emits the same cursor, `hasMore` stays true (`routes/user/history.ts:191`), and `useInfiniteScroll.ts` recreates its IntersectionObserver on every `isFetchingNextPage` flip — a fresh observer fires immediately for a still-intersecting target. Pre-PR the duplicate rows pushed the loader off-screen, so the loop was user-paced; now it is hot.

Guard: stop auto-fetching when a page contributes zero new uuids, or when the cursor does not advance.

jjramirezn, line 167: **[PLAUSIBLE] Stale later-page copy can overwrite a fresher websocket-prepended copy**

Last-copy-wins assumes "later page = fresher fetch", but websocket entries are prepended to page 0 — there the freshest copy is FIRST. Race: a Bridge offramp completes while the user scrolls; the WS pushes a full COMPLETED entry (api `charges-ws/index.ts:280-289`) onto page 0, and a `fetchNextPage` request already in flight returns the same uuid still PENDING on a later page. flatMap order = [fresh ws copy, ..., stale copy], so the Map keeps the stale PENDING copy and the row visibly regresses from Completed to Pending until a refetch (staleTime 30s). The `existsAnywhere` guard only blocks the WS insert; nothing protects the reverse direction.

Possible guard: when a uuid repeats, keep the copy with the more advanced status/timestamp instead of blind last-wins.

src/utils/history.utils.ts
```diff
@@ -463,6 +463,30 @@ export async function completeHistoryEntry(entry: HistoryEntry): Promise<History
...
+ * the two timestamps so that no entry can fall through the crack between
+ * pages, and re-serving the boundary row is the price. Page overlap has other
+ * sources too (`openRequestLinks` is re-queried on every page), so deduping
+ * where the pages get flattened covers all of them at once.
+ *
+ * Keeps the FIRST position but the LAST copy: a later page is fetched later,
+ * so its copy of a repeated entry carries the fresher status. `Map` gives both
+ * — re-setting a key updates the value without moving it.
+ */
+export function dedupeHistoryEntriesByUuid<T extends { uuid: string }>(entries: T[]): T[] {
```
jjramirezn, line 500: **[CONFIRMED] Request-pot rollups: last-copy-wins keeps the staler, smaller aggregate**

Pot rollups reuse `uuid = link.uuid` on every page that contains any of that pot's charges, and each copy aggregates only that page's window (`mapRequestLinkRollup`, api `transaction-intent/history.ts:1322-1366`: timestamp = latest fulfilment in window, total = sum over window). A pot with a recent charge on page 1 and an older charge on page 2 emits two rollups with the same uuid but different contents. This dedupe keeps the page-2 copy: the collected total and timestamp regress, and the re-sort in `page.tsx` moves the row down the feed mid-scroll.

For window-scoped aggregates the "later page = fresher status" premise is inverted. Rollup entries need first-copy-wins or a merge.

src/utils/history.utils.ts
```diff
@@ -463,6 +463,30 @@ export async function completeHistoryEntry(entry: HistoryEntry): Promise<History
...
+ * `useInfiniteQuery` concatenates pages, so anything the cursor hands back
+ * twice renders twice — and the rows are byte-identical, which reads as a
+ * double charge. Users have reported exactly that on a PIX payment the ledger
+ * had only debited once.
+ *
+ * The API cursor is over-inclusive on purpose — it is keyed on the later of
+ * the two timestamps so that no entry can fall through the crack between
+ * pages, and re-serving the boundary row is the price. Page overlap has other
+ * sources too (`openRequestLinks` is re-queried on every page), so deduping
+ * where the pages get flattened covers all of them at once.
```
jjramirezn: **[CONFIRMED gap] "covers all of them at once" overclaims: send-link claim WS duplicates survive**

The send-link claim WS push uses `uuid = fullSendLink.pubKey` (api `routes/claim.ts:417`) while the REST history row for the same link uses the intent id (`extraData.parentSendLinkPubKey` documents the split at lines 177-178 of this file). The pair never collides in the Map, so the same transaction still renders twice for the sender until the next refetch — the exact double-render class this PR fixes for the cursor source.

Fix options: normalize the WS payload uuid to the intent id, or also key the dedupe on `extraData.parentSendLinkPubKey`. At minimum, soften this doc claim.

src/utils/__tests__/history.utils.test.ts
```diff
@@ -165,3 +171,44 @@ describe('getAvatarUrl', () => {
...
+    const entry = (uuid: string, amount = '1') => ({ uuid, amount })
+
+    it('drops a boundary row the API served on two pages', () => {
+        const page1 = [entry('a'), entry('b'), entry('c')]
+        const page2 = [entry('c'), entry('d')] // 'c' repeated at the page boundary
+
+        expect(dedupeHistoryEntriesByUuid([...page1, ...page2]).map((e) => e.uuid)).toEqual(['a', 'b', 'c', 'd'])
+    })
+
+    it('keeps the first position but the freshest copy', () => {
```
jjramirezn: **[CONFIRMED, minor] Test locks insertion order with a rationale the consumer contradicts**

The rationale "Position must not move, or rows jump on scroll" is not enforced by Map order: the only consumer re-sorts by timestamp (`page.tsx` `combinedAndSortedEntries`), and the kept fresher copy's NEWER timestamp legitimately moves the row after the sort. The test locks an implementation detail with a misleading justification.

Suggest asserting the observable property instead: one row per uuid, freshest copy kept.

jjramirezn:
**Review finding outside the diff** (line not part of this PR's hunks, so it cannot be an inline comment):

**[CONFIRMED, pre-existing] `src/app/(mobile-ui)/history/page.tsx:120` — the websocket `existsAnywhere` guard drops status-UPDATE pushes for entries already in cache**

The backend pushes full updated entries on status change (`onOfframpTransferStatusChange` → `mapBridgeIntentToHistoryEntry`, api `charges-ws/index.ts:280-289`) with the same intent uuid already rendered PENDING in cached pages. The FE logs `[History] Duplicate transaction ignored` and returns `oldData` — no merge, no invalidation — so the user keeps seeing PENDING until a refetch trigger (staleTime 30s).

This predates the PR, but the PR codifies "last copy = freshest" semantics one hook below, and the new dedupe helper makes a replace-in-place trivial to express now (with a shallow merge for the minimal charge payloads, which carry only `uuid`/`status`/`txHash`). Worth a follow-up.

src/app/(mobile-ui)/history/page.tsx
abalinda: Confirmed and fixed in `14ca8831` — this was a regression I introduced, and the mechanism is exactly as you describe. I checked `useInfiniteScroll`: the effect lists `isFetchingNextPage` in its deps, so every flip tears down and rebuilds the `IntersectionObserver`, and a fresh observer fires immediately for a still-intersecting target. Before the dedupe the repeated rows grew the list and pushed the loader out of view, so it was user-paced; after it, nothing moves and it spins.

Guard is in `getNextPageParam` rather than at the scroll site: it returns `undefined` when `lastPage.cursor === lastPageParam`. An unchanged cursor means the API is re-serving the same window, so there is nothing further to fetch and React Query stops paginating — that kills the loop for every trigger, not just the request-link one, and needs no extra state in the page.

One case it does not cover, deliberately: if the cursor *does* advance but every row is a duplicate, the list still does not grow and auto-fetch continues. That one is bounded by the length of the history and terminates on `hasMore: false`, so it is a burst rather than an infinite loop. Catching it properly needs the deduped count back in the hook, which felt like the wrong shape for this fix. Covered by a test that drives the real hook against a never-advancing cursor and asserts pagination stops at two pages.

abalinda: Confirmed, and it decided the fix — reverted to first-copy-wins in `14ca8831`.

Your case and the rollup one below both invert the premise last-wins rested on. "Later page = fresher" is only true when array position tracks fetch time, and it does not: the websocket **prepends** to page 0, so for that source the freshest copy is the first one. Last-wins let an in-flight page-2 response overwrite a just-pushed COMPLETED row with its stale PENDING copy — a visible Completed → Pending regression, which is worse than the staleness last-wins was introduced to fix.

I did not go with status-ranking. It needs a status lattice that has to stay in sync with every provider's terminal states, and first-wins already gets both known repeat sources right. The case it loses — a page-1 row PENDING at fetch time, COMPLETED by the time page 2 lands — self-heals inside the 30s `staleTime`, and is not a regression: before this dedupe both rows rendered and the stale one was already on screen. Written into the docblock so the trade-off is visible rather than folklore. Test now covers your exact race.

src/utils/history.utils.ts
abalinda: Confirmed, fixed by the same revert to first-copy-wins in `14ca8831`.

This is the sharper of the two, because a shrinking `totalAmountCollected` is a money number regressing on screen, not just a status. You are right that window-scoped aggregates invert the premise completely: each rollup copy sums only its own page window, so "later" is strictly *less* information, never fresher. Nothing about wall-clock fetch order helps there.

Kept it simple rather than special-casing rollups — first-wins is correct for both this and the websocket race, so one rule covers both. Test added with a pot whose page-1 copy carries 250 and page-2 copy carries 40, asserting the 250 survives.

src/utils/history.utils.ts
abalinda: Fair — the claim was wrong and I softened it in `14ca8831`. The docblock now names the gap explicitly: the send-link claim websocket push keys by `fullSendLink.pubKey` while the REST row for the same transaction uses the intent id, those two never collide in the Map, and that pair still renders twice until the next refetch.

I did not add `parentSendLinkPubKey` as a second dedupe key. It makes the helper carry knowledge of one provider's id split and would still be a workaround for two systems disagreeing on an identity — the real fix is aligning the websocket payload's uuid with the intent id on the API side, which is a backend change and does not belong in this diff. Filed on TASK-21039 alongside the other follow-up.

src/utils/__tests__/history.utils.test.ts
abalinda: You are right and I have replaced the test in `14ca8831`. "Position must not move" was never enforced by Map order — `combinedAndSortedEntries` re-sorts by timestamp — so it locked an implementation detail behind a rationale its only consumer contradicts.

The two replacements assert observable behaviour instead, and they double as the regression tests for your other two findings: a websocket-prepended COMPLETED copy surviving a stale later-page PENDING one, and a request-pot rollup keeping the larger page-1 total. Both state one row per uuid and which copy wins, with no claim about ordering.

abalinda:
On the out-of-diff finding (websocket `existsAnywhere` guard at `page.tsx:120` dropping status-UPDATE pushes) — confirmed, and agreed it is a follow-up rather than this PR.

It is genuinely pre-existing: the guard has always returned `oldData` on a uuid it already has, so a status push for a cached entry is logged and discarded, and the row stays PENDING until a refetch trigger. Nothing in this diff touches that path.

Two reasons I am not fixing it here. It is a behaviour change to the websocket cache-update contract, so it wants its own diff and its own test rather than riding along in a dedupe PR. And it needs the shallow-merge you mention — the charge-completed payload carries only `uuid`/`status`/`txHash`, so a blind replace-in-place would blank the fields the rendered row needs. Filed on TASK-21039.

Worth noting it interacts with the first-copy-wins rule I just landed: once the WS guard is fixed to replace in place, page 0 becomes authoritative for status, which is the direction first-wins already assumes. The two changes agree.

jjramirezn (approved):

PR #2690, opened 2026-08-13
feat(landing): homepage rebuild — persona manifesto, card beat high, /lp/card retired

0xkkonrad:
> Continues #2686 — same commits, branch renamed `feat/lp-refresh` → `lp-refresh` per Konrad (org ruleset blocks in-place renames, so this is a successor PR; all QA history lives on #2686).

## Summary

Rebuild of the peanut.me homepage per mockup **v3.1** (mono `projects/lp-refresh/mockups/mockup.html`, Konrad-approved 2026-08-13). Doctrine: **reuse the live page heavily** — the homepage is the live sections reordered, plus two new folds. The earlier v1 implementation on this branch was rejected as too busy and is fully unwound (additive commits, no force-push).

New fold order (bg): live hero (pink) → **ShhhhhFold** (black, NEW) → **ProblemFold** (cream, NEW) → NoFees (blue) → yourMoney (yellow) → TweetCarousel (pink) → RegulatedRails (cream) → Manteca (blue ← was cream) → securityBuiltIn (pink) → sendInSeconds (yellow) → FAQs → footer. Yellow marquee strips between every fold as on live; the strip after ShhhhhFold carries the /shhhhh words (IYKYK · WORD TRAVELS · …).

- **ShhhhhFold** — distills /shhhhh §1 onto black: wordmark, tagline, body with shared `ScarcityCounter` (51→50), TRY THE DOOR + waitlist link, blurred `PixelatedCardFace` at −12°. Every click routes to `/shhhhh` — no inline join logic. Copy comes from the existing `shhhhh.*` next-intl catalog (already localized, 4 locales). Pink drop shadows on this fold only (black shadows vanish on black).
- **ProblemFold** — "You already know the problem." + three named cards (Living across borders / Sending money home / Paid from abroad) with inline line icons. Landing i18n keys added EN-first in all catalogs.
- **Restored verbatim from dev** (v1 had deleted/edited them): TweetCarousel + tweets.consts, hero, noFees, StickyMobileCTA, ExchangeRateWidget, landingStrings/types/content, all `src/i18n` catalogs. sendInSeconds keeps the full scroll-freeze grow-button machinery.
- **Kept from v1:** `/lp/card` retired (redirect → `/shhhhh` in redirects.json, route deleted, sitemap + robots cleaned), shared `ScarcityCounter`, 20→50/week admission copy across /shhhhh + metadata per `product/card.md` ruling 2026-08-12.
[...]

PR #2692, opened 2026-08-14
fix(landing): give the country lockups the partner-tile treatment

0xkkonrad:
Stacked on `lp-refresh` (#2690), which is where this fold's background moved to blue and where `Manteca.tsx` is already being changed.

## The bug

The Pix wordmark inside `pix-brz.svg` is mint `#32BCAD`. Against the homepage's `#90A8ED` that is **1.01:1** — the same luminance as the ground. It reads as a rendering fault, not a logo.

It was already failing on cream (2.15:1); the blue only made it obvious. And a lighter card does not rescue it — the mint measures 2.35:1 on white and 1.17:1 on pink. Only the black mark clears 4.5:1 anywhere.

| against `#90A8ED` | ratio | |
|---|---|---|
| Pix mint `#32BCAD` | 1.01:1 | fail |
| Argentine flag blue `#3F96D7` | 1.37:1 | fail |
| Brazilian flag green `#177D32` | 2.24:1 | fail |
| MercadoPago navy `#0A0080` | 6.80:1 | pass |
| black — what this PR uses | 9.01:1 | pass |

## What changed

Each lockup now uses what `#regulated-rails` already ships one fold up: the pink partner tile, 1px black border, 4px hard shadow, holding the black mark on the **white inner card** that fold gives PIX and MercadoPago. That inner card is not new here — `RegulatedRails.tsx` already flags those two marks `onWhite` because they are line art rather than heavy wordmarks and go weak straight on the pink. Black on white in the tile is 21:1.

**The flags stay**, but as their own assets drawn to one rule. The old SVGs baked flag and logo into a single file, so the logo could not move into a tile without splitting them. `mepa-arg.svg` also carried no black outline at all, which is why Argentina dissolved into the background; both flags now share the same 3px black edge and hold their silhouette on either ground.

**Asset fix, independent of this fold:** `pix-logo.svg` declared `viewBox="0 0 88 38"` while its ink spans x = −15.7…87.7, so the left petal of the diamond was sheared off. That is visible on the rails marquee today. Widened to `-15.5 0 102 38`.

**Affordance:** both links keep their hrefs and now look like links — the tile lifts on hover and an underlined `Learn more →` sits under each, reusing the existing `landingLearnMore` key, so **no locale files change**.

**Mobile:** the pair sits above the phone, so the sticky CTA covers the screenshot instead of a link. Mobile and desktop now share one block via `md:contents`, which drops the duplicated markup from four link elements to two.

`mepa-arg.svg` and `pix-brz.svg` are deleted — `Manteca.tsx` was their only consumer.

## Checks

- `pnpm typecheck` clean
- `pnpm test` — 232 suites, 2973 passed
- `pnpm prettier --check .` clean

**Not verified locally:** the dev server would not serve on this sandbox (inotify watch limit exhausted, 36s+ compiles). The fold needs an eyes-on check on the Vercel preview before merge.

## Open for review

- The PIX mark carries a "powered by Banco Central" strapline that is unreadable below ~40px. Left as-is here so the mark matches the rails fold exactly; a descriptor-free asset would be the real fix.
- The flags are hand-drawn simplifications. Worth a designer's eye at other sizes.


0xkkonrad:
## Verified on the preview

https://peanut-wallet-git-fix-qrpay-lockup-house-tiles-squirrellabs.vercel.app/en

Checked at 1280 and 390 against the deployed build, not the mockup:

- Fold background resolves to `rgb(144, 168, 237)` — the homepage still passes blue, unchanged by this PR.
- **2 links in the fold**, down from 4 — the mobile and desktop blocks are now one via `md:contents`.
- Both marks render on the white inner card inside the pink tile; the PIX diamond shows all four petals, so the `viewBox` fix lands.
- Both flags hold their silhouette against the blue — the Argentine flag has an outer edge for the first time.
- On mobile the lockups sit above the phone, so the sticky CTA overlaps the screenshot instead of a link.
- No page or console errors on either viewport.

CI: e2e, eslint, format, typecheck, unit, report, ci-success all pass.

Not checked: `/quests`, which renders this same component on cream. The mockup held on cream, but the deployed cream case has not been opened.

PR #2693, opened 2026-08-14
fix(connectivity): windowed failure counting + connection-aware timeout copy

kushagrasarathe:
## Summary

The connectivity banner (`ConnectivityBanner`, shipped 07-03) could never fire: failures were counted **consecutively** and reset by **any** successful response. The app fires requests in parallel (`/users/me`, `/rain/cards`, `/users/history`, …), so on a flaky connection successes interleave with timeouts and the threshold of 2 was never reached. Sentry shows ~500 timeout events / 40+ users in 7d who saw nothing but raw error toasts.

- **Windowed failure counting** — `connectivity.ts` now keeps failure timestamps in a 60s sliding window; entries age out on their own and a success no longer clears them. `useConnectivity` re-reads when the oldest failure expires so the banner drops after recovery without needing a new request. `reportNetworkOk` is deleted (its reset was the bug).
- **Connection-aware copy** — a timeout/DNS/refused failure means the request **never reached the server**, but users were told "Service temporarily unavailable" / "The network is busy". `ServiceUnavailableError` now maps to a new `connectionTimeout` friendly-error code: *"We couldn't reach Peanut — check your internet connection and try again."* (en, es-419, es-AR, pt-BR). `networkBusyTimeout` stays for viem/bundler timeouts, where the user's connection is fine.
- The task's third fix (retry GETs once on `AbortError`) already shipped on dev (PEANUT-UI-R44) — no change here.

## Task

[TASK-21108 — Bad-internet UX](https://app.notion.com/p/peanutprotocol/Bad-internet-UX-3b18381175798008ac2de11dd2778198)

- After connectivity recovers, the banner can linger up to 60s (until the failures age out) — acceptable: the copy says "retrying…", and any success-based clearing reintroduces the interleaving bug.
- `FAILURE_THRESHOLD` is 2 **distinct endpoints** within 60s (was "2 consecutive failures"), and lives next to `FAILURE_WINDOW_MS` so the policy reads in one place.
- The raw English message on the thrown timeout error intentionally duplicates `en.json`'s `connectionTimeout` copy: a few surfaces toast `error.message` directly (e.g. Setup landing, card PIN), and a human-readable fallback beats a technical sentinel there. Localized surfaces go through `friendlyError` by name.
- **Monitoring note:** flows that tag errors with the friendly-error code will see fetch timeouts move from `networkBusyTimeout` to `connectionTimeout` — update any dashboard/alert keyed on the old bucket.
- Pre-existing, unchanged: `ConnectivityBanner` renders above `MaintenanceBanner`, so sustained timeouts during a flagged maintenance can mask the maintenance notice for up to the 60s window. Narrow case, noted not fixed.

## Risks / breaking changes

- FE-only, no API/cross-repo impact. Blast radius: connectivity banner visibility + user-facing timeout copy.
- The banner will now actually appear for users on bad connections — expected behavior change, not a regression.

## QA

- `npm test` — 232 suites green, incl. new windowed-counting tests (interleaved-success case pinned in `useConnectivity.test.tsx`).
- Manual: throttle to offline-ish flakiness in DevTools → two failed requests within 60s show the banner even with interleaved successes; banner clears ≤60s after recovery.

Screenshots: N/A (banner renders identically — only its trigger condition and error-toast copy change; copy strings visible in the diff)

### DIFF
--- a/sentry.utils.test.ts
+++ b/sentry.utils.test.ts
@@ -13,6 +13,10 @@ describe('shouldIgnoreError — alreadyReported (fetchWithSentry wrapper)', () =
         expect(shouldIgnoreError(eventWith({ type: 'ServiceUnavailableError', value: 'upstream 503' }))).toBe(true)
     })
 
+    it('ignores a re-thrown ConnectionTimeoutError (already captured at the fetch site)', () => {
+        expect(shouldIgnoreError(eventWith({ type: 'ConnectionTimeoutError', value: 'timed out' }))).toBe(true)
+    })
+
     it('does not ignore an unrelated application error', () => {
         expect(shouldIgnoreError(eventWith({ type: 'TypeError', value: 'x is not a function' }))).toBe(false)
     })
--- a/sentry.utils.ts
+++ b/sentry.utils.ts
@@ -42,10 +42,11 @@ const IGNORED_ERRORS = {
     // failure is already captured at the fetch site with full context, so the
     // re-thrown ServiceUnavailableError bubbling to global handlers (or being
     // console.error'd by a consumer) would only double-count it (PEANUT-UI-QDJ).
-    // Substring-matching this pattern is safe only because ServiceUnavailableError
-    // is our own internal fetchWithSentry wrapper name, not a generic string that
-    // could appear in an unrelated third-party error message.
-    alreadyReported: ['ServiceUnavailableError'],
+    // Substring-matching these patterns is safe only because both are our own
+    // internal fetchWithSentry wrapper names, not generic strings that could
+    // appear in an unrelated third-party error message. ConnectionTimeoutError
+    // is the timeout-path wrapper; ServiceUnavailableError the generic one.
+    alreadyReported: ['ServiceUnavailableError', 'ConnectionTimeoutError'],
 
     // Third-party SDK internal errors (not actionable)
     thirdPartySdkErrors: [
--- a/src/components/Home/HomeHistory.tsx
+++ b/src/components/Home/HomeHistory.tsx
@@ -350,7 +350,9 @@ const HomeHistory = ({
     if (isError) {
         const isNetworkError =
             error instanceof Error &&
-            (error.name === 'ServiceUnavailableError' || (typeof navigator !== 'undefined' && !navigator.onLine))
+            (error.name === 'ServiceUnavailableError' ||
+                error.name === 'ConnectionTimeoutError' ||
+                (typeof navigator !== 'undefined' && !navigator.onLine))
         // Network timeouts are already captured at the fetch layer — don't
         // re-report them here (and not on every re-render). Only surface
         // genuinely unexpected errors to Sentry.
--- a/src/hooks/__tests__/useConnectivity.test.tsx
+++ b/src/hooks/__tests__/useConnectivity.test.tsx
@@ -1,10 +1,10 @@
 import { act, renderHook } from '@testing-library/react'
 import { useConnectivity } from '../useConnectivity'
-import { FAILURE_WINDOW_MS, reportNetworkError, resetConnectivity } from '@/utils/connectivity'
+import { __resetConnectivityForTests, FAILURE_WINDOW_MS, reportNetworkError } from '@/utils/connectivity'
 
 beforeEach(() => {
     jest.useFakeTimers()
-    resetConnectivity()
+    __resetConnectivityForTests()
     Object.defineProperty(navigator, 'onLine', { configurable: true, value: true })
 })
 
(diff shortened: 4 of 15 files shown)


kushagrasarathe:
@coderabbitai review

kushagrasarathe:
@coderabbitai review

abalinda:
@kushagrasarathe — `/code-review medium` on this branch. Four findings, nitpicks left out.

---

### 1. `connectionTimeout` copy blames the user for our slowness — HIGH
`src/utils/sentry.utils.ts:519`

The new copy tells the user to check their internet. But the 20s abort comes from our own `AbortController` (`CLIENT_FETCH_TIMEOUT_MS = 20_000`), not from the network. It fires just as readily on a healthy connection talking to a slow backend — cold start, DB lock, overloaded route. That is the class of timeout we already track (PEANUT-UI-QH9 on `/confirm`).

During a backend slowdown, every user reads "check your internet connection and try again". Support then gets tickets from people whose internet works.

`AbortError` cannot tell "never reached the server" from "server was slow". Only a `TypeError`/DNS-class rejection can, so the inline comment is not accurate.

Fix: either soften the copy ("Peanut is taking too long to respond — check your connection and try again"), or map `connectionTimeout` only on the non-abort branch.

### 2. The `online` event never clears the failure window — MEDIUM
`src/hooks/useConnectivity.ts:22`

`reportNetworkOk` is gone, so nothing resets the count when the browser comes back online.

Walk it through: the user loses connection in a lift. Two requests fail. `navigator.onLine` goes false, so the 📡 "No internet connection" banner shows. Connection returns. `handleOnline` sets `isOnline` true, but `failures >= 2` still holds, so `isApiUnreachable` turns true. The banner switches to ⚠️ "Trouble reaching Peanut — retrying…" for up to 60s, while `refetchOnReconnect` is already succeeding.

Fix: clear the window inside `handleOnline`. One line. It does not bring back the interleaving bug, because the `online` event is not a per-request success.

### 3. `setTimeout` expiry does not survive freeze or sleep — MEDIUM
`src/utils/connectivity.ts:36`

The PR describes a 60s sliding window of failure timestamps. The code is a counter plus one `setTimeout` per failure. Browsers throttle `setTimeout` in background tabs and suspend it on mobile page-freeze, bfcache, and laptop sleep.

Walk it through: two failures land. The user backgrounds the PWA and returns an hour later on good WiFi. The expiry timers never ran while frozen. `getRecentFailures()` still returns 2, so the banner renders on resume with no failing request behind it.

Fix: keep an array of `Date.now()` stamps and prune on read. Immune to clock suspension, and about the same amount of code.

### 4. One slow endpoint now trips the app-wide banner — MEDIUM
`src/utils/connectivity.ts:22`

`FAILURE_THRESHOLD = 2` was tuned for *consecutive* failures. Under "2 within 60s" it means something different.

The default query client uses `RETRY_STRATEGIES.FAST` (`retry: 2`, so 3 attempts). Each attempt is its own `fetchWithSentry` call, and each lands in the catch and calls `reportNetworkError()`. So one persistently slow route — say `/users/history` — produces 3 failures in a few seconds. The global "we can't reach Peanut" banner shows while every other request succeeds.

The old interleaved-success reset masked this. Nothing masks it now.

Fix: raise the threshold, or de-duplicate failures per endpoint inside the window.

---

Findings 2, 3 and 4 share one root cause: removing the success-reset took away the only thing that cleared the state, and nothing replaced it.

kushagrasarathe:
@abalinda all four addressed in 8f196d5 — agreed on the shared root cause (removing the success-reset left nothing to clear state; there are now three clearing paths: per-entry window expiry, the `online` event, and prune-on-read).

1. **Copy** — took your softer wording: *"Peanut is taking too long to respond — check your connection and try again."* (4 locales + the raw fallback message). Kept the mapping on the abort branch: the non-abort branch also wraps CORS/CSP failures, which can be our outage, so it stays on the fully neutral `networkBusyTimeout`. Inline comments corrected — no more "provably never reached the server" claim.
2. **`online` clear** — `handleOnline` now calls `clearRecentFailures()`. Test pins the lift scenario: offline with 2 failures → online → no banner.
3. **Freeze/sleep** — truth is now `Date.now()` stamps pruned on read; the per-failure timer only emits so subscribers re-read. A tab resumed after an hour reads 0 regardless of whether timers ran. Test simulates suspended timers via `jest.setSystemTime`.
4. **Retry amplification** — the count is now **distinct failing endpoints** (keyed by sanitized url) inside the window, so FAST-retry ×3 on one slow route stays at 1. A genuinely dead connection fails multiple endpoints at once, so the threshold of 2 still trips. Test pins both.

CI green, 2988 tests passing.

abalinda (approved):

abalinda (approved):

abalinda (approved):

PR #2695, opened 2026-08-14
fix(i18n): translate countries, regions and badge copy for pt-BR

abalinda:
## Summary

QA of the app in Portuguese (TASK-20958) found four surfaces still rendering English, plus one term the **Cobrar** section names two different ways. This fixes all five.

**Countries** and **regions** were untranslated because they are static English catalog data rendered straight to the screen.

- Countries now resolve through `Intl.DisplayNames` from the ISO-2 code `countryData` already carries. No locale ships ~250 country names, and every locale we add gets country names for free. The English `title` stays searchable, so typing "Brazil" still finds "Brasil"; sorting and collation follow the active locale.
- Regions keep their English `name` — `deriveRegionAccess` branches on it, and the code comment there already warned that localizing it would silently kill that branch. Display copy resolves from the stable `path` against a new shared `common.regions` catalog, which also absorbs the old `limits.restOfWorld` so the label lives in one place.

**Badge names and reasons** come from the backend catalog, which ships English only. A `badges.catalog` namespace keyed by badge code now supplies localized copy for all 48 codes (pt-BR + es-419). The backend prose stays the fallback, so a badge added after this build still renders rather than showing a raw key.

**"Cobrar" terminology**: in Portuguese the money request a user creates from that section is a *cobrança*, not a *solicitação* (which reads as a support ticket). Aligned across the `request`, `payment`, `transaction`, `global` and `loadingStates` namespaces — one name for one thing.

## Task

TASK-20958 — https://www.notion.so/63f83811757982508d5a8131c25f12bf

## Findings → fix

| Finding (from the task) | Fix |
|---|---|
| Countries are not translated | `Intl.DisplayNames` at every country render site |
| Regions are not translated (North America, Europe, LATAM, Rest of the world) | new `common.regions` catalog, resolved by region `path` |
| Badge "Motivos" are in EN (shh = "You know the secret") | `badges.catalog` — SHHHHH now reads "Quem sabe, sabe." |
| "Deixe vazio para quem paga escolher o valor" | "Deixe sem valor para o pagador escolher." |
| "Criar solicitação" → "Criar cobrança" | done, plus the 20 sibling strings that used the same wrong term |

- **`Intl.DisplayNames` over a translated country list.** The platform already ships CLDR country names for every locale. Two guards are load-bearing and tested: `of()` echoes the input back for a valid-but-unassigned code (`QM`), and throws `RangeError` on an alpha-3 code — both fall back to the English catalog title.
- **The FE badge catalog now wins over the backend name for known codes.** That is the point — the backend cannot localize. Unknown codes still fall through to backend prose, and there is a test for it. One existing test that asserted backend-name precedence was updated, and a fallback test added alongside it.
- **es-419 badge copy is deliberately neutral between tuteo and voseo**, so the deltas-only `es-AR` overlay does not have to re-translate 48 badges. The glossary test enforces this.
- **`limits.restOfWorld` was removed**, not duplicated — its one consumer now reads the shared catalog.

## Risks

- Copy-only in behavior, but it touches shared render paths (`CountryList`, region lists, all five badge surfaces). No API, no money path, no state machine.
- Country names in the picker change for every non-English locale — that is the intended fix, and the English title stays searchable so no user loses a search term they were used to.
- No cross-repo action needed. The backend badge catalog is unchanged and stays authoritative for English.

## QA

- `npm run typecheck`, `npm test` (236 suites / 3033 tests), `pnpm prettier --check .`, `npm run build` — all green locally.
- New tests: `src/utils/__tests__/country-name.utils.test.ts`, `src/hooks/__tests__/useRegionLabel.test.ts`, plus a badge-fallback case in `BadgeEarnToast.test.tsx`.
- The catalog gates already in the repo cover the rest: key parity across locales, duplicate-value drift, ICU compilation, and the pt-BR/es glossary rules.
- Manual: switch the app language to Português in Settings, then check the country picker (Add money / Withdraw), Profile → Limits, Profile → Badges, and the Cobrar flow.

## What the diff cannot show

Country names are not strings in this diff — they come out of `Intl.DisplayNames` at render time. This is what the picker actually renders in Portuguese:

| ISO-2 | before (catalog) | after (pt-BR) |
|---|---|---|
| US | United States | Estados Unidos |
| BR | Brazil | Brasil |
| DE | Germany | Alemanha |
| GB | United Kingdom | Reino Unido |
| ES | Spain | Espanha |
| NL | Netherlands | Países Baixos |
| CH | Switzerland | Suíça |
| AE | United Arab Emirates | Emirados Árabes Unidos |
| ZA | South Africa | África do Sul |
| JP | Japan | Japão |
| SE | Sweden | Suécia |
| PL | Poland | Polônia |

Regions: Europa · América do Norte · América Latina · Resto do mundo.

**Screenshots: ⚠️ NONE.** No layout, spacing, control or component changed — this is text substitution inside existing surfaces. I did try the capture pipeline: the sandbox API needs the known `TSX_TSCONFIG_PATH` workaround to boot at all, and after that the seeded users bounce to `/setup` before the target routes render. Rather than keep fighting local state for a copy-only diff, the table above covers the one thing the diff genuinely hides. Reviewing in the app: Settings → Idioma → Português, then Add money (country picker), Profile → Limits, Profile → Badges, Cobrar.

## Follow-ups (not in this PR)

- **The English badge catalog is now a second hand-synced mirror of the backend copy**, alongside the existing `src/types/badge-assets.json`. The generator that comment names (`pnpm badge:check --write-manifest`) no longer exists in peanut-api-ts, so both mirrors drift silently today. Worth one generator that emits both — a peanut-api-ts change, so it does not belong here.
- **`publicDescription` stays English for every badge.** The backend's third-person copy still wins when you view someone else's profile (that is the fix for the review finding), so those tooltips are not localized. Localizing them means a second catalog field × 48 badges — separate PR if we want it.
- **`BADGE_SHARE_LINES` in `badge.utils.ts`** is a third badge-copy mechanism (English-only bespoke share brags with a `locale !== 'en'` bail-out). It could fold into `badges.catalog`.

## Docs impact

None. This changes how existing facts are worded in the pt-BR app UI; it changes no documented customer-facing fact (no fee, limit, supported country, KYC or card rule). Checked `product/`, `content/` and `docs/` for the badge, region and request copy this PR touches — no hits.

### DIFF
--- a/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
+++ b/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
@@ -46,7 +46,8 @@ import { addMoneyCountryUrl, rewriteMethodPath } from '@/utils/native-routes'
 import { isMantecaSupportedCountryCode } from '@/constants/manteca.consts'
 import { useSafeBack } from '@/hooks/useSafeBack'
 import { getRegionIntent } from '@/utils/regions.utils'
-import { useTranslations } from 'next-intl'
+import { useLocale, useTranslations } from 'next-intl'
+import { localizedCountryTitle } from '@/utils/country-name.utils'
 
 // Step type for URL state
 type BridgeBankStep = 'inputAmount' | 'showDetails'
@@ -57,6 +58,7 @@ type BridgeBankStep = 'inputAmount' | 'showDetails'
 function BridgeBankOnrampPage() {
     const params = useParams()
     const _searchParams = useSearchParams()
+    const locale = useLocale()
     const t = useTranslations('addMoney')
     const tCommon = useTranslations('common')
 
@@ -514,7 +516,7 @@ function BridgeBankOnrampPage() {
                     variant={resolveKycModalVariant(gate)}
                     providerMessage={getGateUserMessage(gate)}
                     reasonCode={getGateReasonCode(gate)}
-                    regionName={selectedCountry?.title}
+                    regionName={selectedCountry && localizedCountryTitle(locale, selectedCountry)}
                 />
 
                 <AdvisoryPreemptModal {...advisoryModalProps} />
--- a/src/app/(mobile-ui)/withdraw/[country]/bank/page.tsx
+++ b/src/app/(mobile-ui)/withdraw/[country]/bank/page.tsx
@@ -49,11 +49,13 @@ import { ANALYTICS_EVENTS } from '@/constants/analytics.consts'
 import { withdrawCountryUrl } from '@/utils/native-routes'
 import { useSafeBack } from '@/hooks/useSafeBack'
 import { useSendFlowOrigin } from '@/hooks/useSendFlowOrigin'
-import { useTranslations } from 'next-intl'
+import { useLocale, useTranslations } from 'next-intl'
+import { localizedCountryTitle } from '@/utils/country-name.utils'
 
 type View = 'INITIAL' | 'SUCCESS'
 
 export default function WithdrawBankPage() {
+    const locale = useLocale()
     const t = useTranslations('withdraw')
     const tNav = useTranslations('navigation')
     const tCommon = useTranslations('common')
@@ -97,6 +99,7 @@ export default function WithdrawBankPage() {
     // PENDING rail in an unrelated jurisdiction can't block this page.
     const { gateFor } = useCapabilities()
     const bankCountry = useMemo(() => railJurisdictionForBank(getCountryFromPath(country)?.id), [country])
+    const countryFromPath = getCountryFromPath(country)
     const gate = useMemo(() => gateFor('withdraw', { channel: 'bank', country: bankCountry }), [gateFor, bankCountry])
     // bridge re-verification ("we're reviewing your details") modal for the
     // waiting-on-provider gate — keeps the status poll alive + auto-dismisses.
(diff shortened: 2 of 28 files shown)


abalinda:
@coderabbitai review

kushagrasarathe (approved):
Reviewed the full diff plus mechanical checks (agent-assisted review, T2 — copy-only but on add-money/withdraw gating surfaces).

**Verified, not just read:**

- Catalog integrity re-derived: en / es-419 / pt-BR in exact key parity (0 missing, 0 extra). `badges.catalog` covers exactly the 48 codes in `badge-assets.json` in all three locales. `common.regions` covers all 7 region paths from `regions.utils.ts`. The removed `limits.restOfWorld` has zero remaining consumers.
- Caller sweep clean: the only `getBadgeDisplayName` call left outside `useBadgeCopy` is the dev-only share-builder page. The remaining `country.title` in `AddWithdrawRouterView.tsx:355` feeds URL slugging / context state, not display — keeping it English is correct.
- The `deriveRegionAccess` hazard is handled right: internal `Region.name` stays English, display resolves from `path`.
- Terminology claim holds: every remaining "solicitação" in pt-BR is a different sense (card application, KYC docs, wallet prompts, generic errors).
- Ran the 4 new/updated suites locally: 22/22 pass, including the `QM` echo and alpha-3 `RangeError` guards and the publicDescription-precedence regression test. CI green on head.

**Two non-blocking notes:**

1. Fifth-bug signal: `LimitsPageView.tsx` (5 fixes in last 20 commits), `UnlockedRegions.view.tsx` (4), `regions.utils.ts` (3). This PR does not touch rails reads, but the limits/regions cluster stays a hotspot — more weight behind the TASK-20629 refactor.
2. The en `badges.catalog` is now a second hand-synced mirror of backend badge copy, next to `badge-assets.json`. Since the FE catalog now wins for known codes, an English copy edit on the backend silently stops reaching users. Please file the generator follow-up you named in the PR body so it does not get lost — one script in peanut-api-ts that emits both mirrors, plus a drift check.

Approving.

PR #2696, opened 2026-08-14
fix(qr-drawer): stop browser hijacking the expand swipe

kushagrasarathe:
## Summary

AB reported the "My QR" bottom drawer on the scanner often needs two upward swipes and feels laggy (TASK-20720). Root cause: the drawer runs with `modal={false}`, which disables vaul's scroll prevention, and nothing set `touch-action` on the drawer content. The browser is free to claim an upward swipe as a scroll and fire `pointercancel` — vaul aborts the drag mid-way and the drawer springs back. The second swipe works because the scroll chain is already exhausted. The abort + spring-back is also most of the perceived lag.

Three targeted changes in `QRBottomDrawer`:

1. `touch-none` on the drawer content — the browser never takes over the gesture, so the pointer stream reaches vaul intact and one swipe expands the drawer. Content fits the drawer at full snap, so no inner scrolling is lost.
2. `dismissible={false}` restored — drag-down at the first snap point was calling vaul's `closeDrawer()` against the forced `open={true}` (the exact collision #2005/#2008/#2012 fixed), causing a close/reopen flicker plus a 500ms window after reopen where vaul ignores all drags. It was dropped by accident in 80d930cf8.
3. `snapPoints` hoisted to module scope — a per-render array changes identity every render, making vaul's snap-sync effect refire and re-apply the transform transition on every parent re-render.

The title swap between "My QR" and "Show QR to Get Paid" across snap states is intended behavior and unchanged.

## Task

TASK-20720 — https://app.notion.com/p/peanutprotocol/39f83811757980a9bfa2fea6236092a7

## Risks / breaking changes

- Scoped to one component; no API or cross-repo impact.
- `touch-none` disables inner content scrolling only while the drawer is collapsed. At full snap the scroll wrapper scrolls normally, so overflowing content (SE-class viewports, large OS text) stays reachable.
- `dismissible={false}` means vaul ignores drag-down at the collapsed snap — the drawer does not move. Intended, since the drawer was never actually dismissible (`open={true}`).

## QA

Touch-cancel semantics don't reproduce in desktop emulation — verify on a real phone:

1. Open the scanner (scan button), drawer appears collapsed at "My QR".
2. One upward swipe expands it to "Show QR to Get Paid" — first try, no bounce-back.
3. Drag down from collapsed: the drawer does not move — no flicker or duplication.
4. Share button and QR render unchanged.

Screenshots: N/A (no visible change — gesture/interaction fix only, rendered output is identical).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the QR drawer’s stability when it is forced open.
  * Prevented accidental dismissal and drag-related flickering during touch interactions.

- Forced `open={true}` fighting vaul's dismissal model is the recurring root of this drawer's bug history (#2005, #2008, #2012, `80d930cf8`, this PR) — flagged as a deep-dive follow-up, out of scope here.
- On SE-class viewports at full snap, a drag-down that starts on overflowing content can still race the inner scroller's overscroll. Rare, degrades to a second swipe, and the alternative (blanket `touch-none`) made content unreachable — this side of the trade-off was chosen deliberately.
- vaul 1.1.2 pinned-behavior note: with `dismissible={false}`, a fast down-flick at the collapsed snap hits a benign out-of-bounds snap index inside vaul (clamped internally). Re-test on any vaul upgrade.

### DIFF
--- a/src/components/Global/Drawer/index.tsx
+++ b/src/components/Global/Drawer/index.tsx
@@ -26,10 +26,12 @@ DrawerOverlay.displayName = DrawerPrimitive.Overlay.displayName
 type DrawerContentProps = React.ComponentPropsWithoutRef<typeof DrawerPrimitive.Content> & {
     /** Screen-reader-only DialogTitle for drawers without a visible DrawerTitle (Radix a11y requirement). */
     accessibleTitle?: string
+    /** Merged onto the inner scroll wrapper — the element that owns panning when content overflows. */
+    scrollAreaClassName?: string
 }
 
 const DrawerContent = React.forwardRef<React.ElementRef<typeof DrawerPrimitive.Content>, DrawerContentProps>(
-    ({ className, children, accessibleTitle, ...props }, ref) => (
+    ({ className, children, accessibleTitle, scrollAreaClassName, ...props }, ref) => (
         <DrawerPortal>
             <DrawerOverlay />
             <DrawerPrimitive.Content
@@ -45,7 +47,12 @@ const DrawerContent = React.forwardRef<React.ElementRef<typeof DrawerPrimitive.C
                 {accessibleTitle && <DrawerTitle className="sr-only">{accessibleTitle}</DrawerTitle>}
                 <div className="mx-auto my-4 h-1.5 w-10 rounded-full bg-black" />
                 <div className="flex w-full justify-center">
-                    <div className="max-h-[80vh] w-full overflow-auto pb-[env(safe-area-inset-bottom)] md:max-w-xl">
+                    <div
+                        className={twMerge(
+                            'max-h-[80vh] w-full overflow-auto pb-[env(safe-area-inset-bottom)] md:max-w-xl',
+                            scrollAreaClassName
+                        )}
+                    >
                         {children}
                     </div>
                 </div>
--- a/src/components/Global/QRBottomDrawer/index.tsx
+++ b/src/components/Global/QRBottomDrawer/index.tsx
@@ -14,12 +14,14 @@ interface QRBottomDrawerProps {
     className?: string
 }
 
+// module scope: a per-render array changes identity every render, which makes
+// vaul's snap-sync effect refire and re-apply the transform transition
+const snapPoints = [0.75, 1]
+
 const QRBottomDrawer = ({ url, collapsedTitle, expandedTitle, text, buttonText, className }: QRBottomDrawerProps) => {
     const t = useTranslations('global')
     const tCommon = useTranslations('common')
     const contentRef = useRef<HTMLDivElement>(null)
-
-    const snapPoints = [0.75, 1]
     const [activeSnapPoint, setActiveSnapPoint] = useState<number | string | null>(snapPoints[0])
 
     const handleSnapPointChange = (snapPoint: number | string | null) => {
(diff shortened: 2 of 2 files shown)


kushagrasarathe:
@coderabbitai review

jjramirezn:
## `/code-review high` — 2 findings

Reviewed against the branch head (`fix/qr-drawer-gesture-hijack`) and the vaul 1.1.2 internals.

### 1. MEDIUM — unconditional `touch-none` on the drawer content also blocks the inner scroller at full snap

`src/components/Global/QRBottomDrawer/index.tsx:52`

`touch-action` is resolved by intersecting the values of the touched element and all its ancestors. `touch-none` on `DrawerPrimitive.Content` therefore stops touch panning of the `overflow-auto` wrapper at every snap point, not only the collapsed one. Two results:

- The new `scrollAreaClassName={activeSnapPoint === snapPoints[0] ? 'touch-none' : undefined}` on line 53 is dead code.
- The stated mitigation — "at full snap the wrapper must scroll again, or overflowing content is unreachable" — does not hold.

Concrete case the PR itself concedes: on an SE-class viewport, or with large OS text, title + QR + text + divider + Share exceed the wrapper's `max-h-[80vh]`. The Share button then cannot be reached by touch.

Fix: make the outer `touch-none` conditional on the collapsed snap too, and drop the redundant inner copy.

### 2. LOW — the `dismissible={false}` comment and QA step 3 describe behavior vaul does not produce

`src/components/Global/QRBottomDrawer/index.tsx:43`

In vaul 1.1.2, `onDrag` returns early for a downward drag at the first snap point when `dismissible` is false (`node_modules/vaul/dist/index.mjs:1039-1041`). The drawer does not move at all — it does not "rubber-band".

On release, `onRelease` then calls `snapToPoint(snapPointsOffset[-1])`, that is `snapToPoint(undefined)`, which writes `transform: translate3d(0, undefinedpx, 0)`.

I confirmed this is inert: the invalid value is ignored, `overlayRef.current` is null because `Overlay` returns null when `modal={false}`, and `setActiveSnapPoint` re-sets the same `0.75` so React bails. The change is safe. But the comment and the QA script should say "does not move" rather than "rubber-bands smoothly", or the next person chases a phantom regression.

### Clean

No issues found in the `Drawer` `twMerge` plumbing.

With finding 1 fixed, this one looks mergeable. Finding 2 is a comment edit.

kushagrasarathe:
Thanks — finding 2 was right and is fixed in `350447595`. Finding 1 I tested empirically before touching the code, and the ancestor-intersection model doesn't hold; details below.

### Finding 1 — declined, with a repro

`touch-action` is not intersected across **all** ancestors. Per the pointerevents spec, the UA only considers the touched element and its ancestors **up to and including the nearest element that implements the panning** — a scroll container stops the walk. Chromium touch emulation confirms it, and adds one surprise: the walk stops at an `overflow-auto` container **even when nothing overflows**.

Test (Playwright/Chromium, `hasTouch` + CDP `Input.synthesizeScrollGesture`), fixture `outer > scroller(overflow-auto, 200px) > content(2000px)`:

| case | result |
|---|---|
| outer `auto`, scroller `auto` (control) | scrolled |
| **outer `none`, scroller `auto`** | **scrolled** — outer does not block the inner scroller |
| outer `auto`, scroller `none` | blocked |
| outer `none`, scroller fits (page behind scrollable) | page scrolled — outer not consulted even without overflow |

Consequences:

- The inner `scrollAreaClassName` conditional is not dead code — it is the piece that actually blocks the collapsed-snap hijack for touches on the drawer body. Dropping it would reintroduce the reported bug.
- "Share unreachable at full snap" doesn't happen: at full snap the wrapper's touch-action is `auto`, and the outer `touch-none` is not consulted for touches inside it, so overflowing content scrolls.
- The outer `touch-none` only governs the drag-handle area (outside the wrapper) — the comment now says so instead of the previous overflow-based explanation, which this test also disproved.

Repro script if you want to run it yourself:

```js
const ctx = await chromium.launch().then(b => b.newContext({ hasTouch: true, isMobile: true }))
const page = await ctx.newPage()
await page.setContent(`<div style="touch-action:none">
  <div id="s" style="height:200px;overflow:auto"><div style="height:2000px"></div></div></div>`)
const cdp = await ctx.newCDPSession(page)
await cdp.send('Input.synthesizeScrollGesture', { x: 187, y: 100, yDistance: -300, gestureSourceType: 'touch' })
// #s.scrollTop ≈ 300 — the outer touch-action:none did not block it
```

### Finding 2 — accepted, fixed

Correct on both points: with `dismissible={false}` vaul's `onDrag` returns early at the first snap, so the drawer does not move — no rubber-band. The code comment, the PR body risk note, and QA step 3 now all say "the drawer does not move", and the vaul note now ties the benign `snapToPoint(undefined)` to that release path.

PR #2707, opened 2026-08-17
fix(fx): stop the add-money and withdraw screens freezing on a rate fetch

### LINKED ISSUE #1848: fix: add timeout/error handling for currency rate and limits API calls
## Problem
Both the Manteca add-money flow and Bridge bank flow call:
1. `useCurrency()` → `getCurrencyPrice()` for exchange rates
2. `useLimitsValidation()` → `useLimits()` → `/users/limits` API

If either API call hangs or is slow (common on budget Android with spotty connections):
- `InputAmountStep` shows `<PeanutLoading />` indefinitely while `currencyData.isLoading`
- The page appears blank/frozen with no way to interact

Neither call has a timeout, and there's no fallback UI for failed fetches.

## Fix
- Add `AbortController` timeout (e.g. 10s) to `getCurrencyPrice()` in `useCurrency.ts`
- Add error/timeout state to `InputAmountStep` — show the amount input with a retry button for the rate instead of blocking the entire page
- Consider adding `retry: 1` and `retryDelay: 2000` to the `useQuery` in `useLimits.ts`

## Context
User report: app freezes on add-money/withdraw screen. Reproducible on Motorola G31 (Android 12) and G23 (Android 14).
Investigation: https://discord.com/channels/972435984954302464/1486336087747199059/1486337803708141588

innolope-dev:
Fixes #1848.

## What was already handled

Worth stating up front so the diff reads in context — most of the issue's suggested fix already landed:

- **Timeouts on both calls.** `fetchWithSentry` aborts at 20s (client) with one silent GET retry, and both `/bridge/exchange-rate` and `/users/limits` go through it.
- **`useLimits` retry.** It inherits `RETRY_STRATEGIES.FAST` (2 retries, exponential backoff) from the app `QueryClient`.
- **The null-rate crash** in `InputAmountStep` was fixed in #2325.

So the remaining freeze was not a missing timeout.

## The actual freeze

`useCurrency` seeded `code` from the prop on the first render and then ignored the prop for the rest of its life — there was no sync effect. Both callers derive the currency from `useSearchParams()`:

- `withdraw/manteca/page.tsx` → `useCurrency(selectedCountry?.currency ?? null)`
- `MantecaAddMoney.tsx` → `useCurrency(selectedCountry?.currency ?? 'ARS')`

On the first render of a statically exported page the search params are empty, so the withdraw page passed `null`, the hook latched it, never fetched, and `price` stayed null forever. Its gate is `if (isCurrencyLoading || !currencyPrice || …) return <PeanutLoading />` — an unresolvable loader with no header and no way out. That's a deterministic hang rather than a slow network, which fits "reproducible on G31 and G23" better than a timeout would.

The same latch also meant add-money would price a Brazil flow in ARS (its fallback) whenever the country arrived a render late.

## Changes

**`useCurrency`**
- Sync `code` when the prop changes, tracking the previous prop so the unsupported-currency branch (which deliberately nulls `code`) isn't clobbered on the next render.
- Report the gap between the prop landing and the fetch starting as `isLoading`, so consumers can't read it as a settled "no rate" and flash an error.
- Add `refetch()`. Previously the effect only re-ran on a currency change, so a rate outage stranded the user until they left the screen entirely.
- Ignore a superseded in-flight response so a slow request can't overwrite a newer currency's rate.

**Failure states** — new shared `RateUnavailable` (alert + retry), wired into all three rate-dependent flows:
- `InputAmountStep`: retry button instead of a bare alert; header stays mounted while loading so back always works.
- `withdraw/manteca`: had no error branch at all — a failed fetch fell into the same infinite loader. Now an error screen with back and retry.
- `MantecaReviewStep`: rendered `1 USD = undefined ARS`; now shows the retry state.

**i18n** — moved `rateUnavailable` from `addMoney.errors` to the shared `errors` namespace across all four catalogs rather than duplicating the string.

## Verification

- Full Jest suite: 238 suites / 3068 passing, including 6 new `useCurrency` tests (late-arriving currency, refetch recovery, out-of-order response) and 2 new `InputAmountStep` tests.
- `next build` succeeds; 1089 static pages generated.
- `tsc`, `eslint`, and `prettier` clean on the touched files.

## Not covered

No page-level test for `withdraw/manteca` — it has no existing suite and would need a large mock surface. The hook test covers the root cause; the page's new branch is not directly exercised.

### DIFF
--- a/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
+++ b/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
@@ -19,6 +19,7 @@ import { useCallback, useEffect, useMemo, useState } from 'react'
 import countryCurrencyMappings, { isNonEuroSepaCountry, isUKCountry } from '@/constants/countryCurrencyMapping'
 import { formatUnits } from 'viem'
 import PeanutLoading from '@/components/Global/PeanutLoading'
+import RateUnavailable from '@/components/Global/RateUnavailable'
 import EmptyState from '@/components/Global/EmptyStates/EmptyState'
 import AddMoneyBankDetails from '@/components/AddMoney/components/AddMoneyBankDetails'
 import { getCurrencyConfig, getCurrencySymbol, getMinimumAmount, railJurisdictionForBank } from '@/utils/bridge.utils'
@@ -192,7 +193,12 @@ function BridgeBankOnrampPage() {
     // deposit-side price (localCurrency per USD) for limits validation — deposits
     // execute at buy, so the USD equivalent must derive from buy, not the sell-side
     // display quote served by /fx/rate (useCurrency handles USD as 1:1)
-    const { price: localPrice, isLoading: isRateLoading, isError: isRateError } = useCurrency(localCurrency)
+    const {
+        price: localPrice,
+        isLoading: isRateLoading,
+        isError: isRateError,
+        refetch: refetchRate,
+    } = useCurrency(localCurrency)
 
     // convert input amount to USD for limits validation
     // bridge limits are always in USD, but user inputs in local currency
@@ -471,9 +477,7 @@ function BridgeBankOnrampPage() {
                     {error.showError && !!error.errorMessage && !limitsValidation.isBlocking && (
                         <ErrorAlert description={error.errorMessage} />
                     )}
-                    {localCurrency !== 'USD' && isRateError && (
-                        <ErrorAlert description="We couldn't load the exchange rate. Please try again in a moment." />
-                    )}
+                    {localCurrency !== 'USD' && isRateError && <RateUnavailable onRetry={refetchRate} />}
                 </div>
 
                 <OnrampConfirmationModal
--- a/src/app/(mobile-ui)/add-money/__tests__/add-money-states.test.tsx
+++ b/src/app/(mobile-ui)/add-money/__tests__/add-money-states.test.tsx
@@ -1896,6 +1896,52 @@ describe('GROUP 8: InputAmountStep Component', () => {
         expect(screen.getByText('Continue')).toBeDisabled()
     })
 
+    // #1848: the error state used to be a dead end — useCurrency only refetches
+    // when the currency changes, so the user had to leave the screen to recover.
+    test('rate fetch failure offers a retry that refetches the rate', () => {
+        const refetch = jest.fn()
+        renderWithProviders(
+            <InputAmountStep
+                tokenAmount="100"
+                setTokenAmount={jest.fn()}
+                onSubmit={jest.fn()}
+                isLoading={false}
+                error={null}
+                setCurrencyAmount={jest.fn()}
+                currencyData={{ isLoading: false, isError: true, symbol: null, price: null, refetch }}
+                limitsValidation={{ isBlocking: false, isWarning: false, currency: 'USD' }}
+                limitsCurrency="USD"
+                onBack={jest.fn()}
+            />
+        )
+
+        fireEvent.click(screen.getByText('Retry'))
+
+        expect(refetch).toHaveBeenCalledTimes(1)
+    })
+
+    // A slow rate fetch can hold this screen for tens of seconds on a bad mobile
+    // connection. The header has to stay mounted or the page reads as frozen.
+    test('currency data loading keeps the back button available', () => {
+        renderWithProviders(
+            <InputAmountStep
+                tokenAmount=""
+                setTokenAmount={jest.fn()}
+                onSubmit={jest.fn()}
+                isLoading={false}
+                error={null}
+                setCurrencyAmount={jest.fn()}
+                currencyData={{ isLoading: true, symbol: null, price: null }}
+                limitsValidation={{ isBlocking: false, isWarning: false, currency: 'USD' }}
+                limitsCurrency="USD"
+                onBack={jest.fn()}
+            />
+        )
+
+        expect(screen.getByTestId('peanut-loading')).toBeInTheDocument()
+        expect(screen.getByTestId('nav-header')).toBeInTheDocument()
+    })
+
     test('onSubmit called when Continue clicked', async () => {
         const onSubmit = jest.fn()
         renderWithProviders(
--- a/src/app/(mobile-ui)/withdraw/manteca/page.tsx
+++ b/src/app/(mobile-ui)/withdraw/manteca/page.tsx
@@ -17,6 +17,7 @@ import NavHeader from '@/components/Global/NavHeader'
 import ErrorAlert from '@/components/Global/ErrorAlert'
 import { Icon } from '@/components/Global/Icons/Icon'
 import PeanutLoading from '@/components/Global/PeanutLoading'
+import RateUnavailable from '@/components/Global/RateUnavailable'
 import { mantecaApi, type WithdrawPriceLock } from '@/services/manteca'
 import { useCurrency } from '@/hooks/useCurrency'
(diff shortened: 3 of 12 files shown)


kushagrasarathe (changes requested):
Agent review (advisory, rulebook pass — full diff read, callers swept, CI checked). The diagnosis and fix are correct and well-tested. One blocking issue, two minor notes.

## Blocking

**The i18n key move breaks the rate-error message on the Bridge bank flow.** This PR deletes `addMoney.errors.rateUnavailable` from all four catalogs, but `src/app/(mobile-ui)/add-money/[country]/bank/page.tsx:474` still reads it:

```tsx
// line 62: const t = useTranslations('addMoney')
{localCurrency !== 'USD' && isRateError && <ErrorAlert description={t('errors.rateUnavailable')} />}
```

After merge, a rate failure on that screen renders the raw key path instead of the message. No test covers that branch, so CI stays green.

Fix: point that line at the new shared `errors.rateUnavailable` key — or better, wire `RateUnavailable` into that screen too. It is the 4th `useCurrency` caller, and right now it shows an error with no retry button, the exact dead-end this PR removes everywhere else.

## Minor (fine as follow-up)

1. **Stale rate on failed refetch.** The hook's `catch` does not clear `price`/`symbol`, so after an in-place currency change whose fetch fails, the previous currency's rate stays in memory. The withdraw page's new error branch gates on `!currencyPrice` (not `isError`), so it would fall through and render with the stale rate. Narrow window, but it is a money-display path — one line in the `catch` closes it.
2. **Architecture smell (rulebook flag, not on this PR).** `withdraw/manteca/page.tsx` trips the fifth-bug rule: 8+ fix commits in its last 20, and code-analysis rates it critical (CC 162, 66 commits in 6 months). This PR adds 16 careful lines, which is fine — the file itself is a deep-dive candidate.

## Verdict

```
TIER: T2 (frontend money flows). Diff matches description, no surplus changes.
PAIRED PR: none — frontend-only.
A. BREAKAGE:  FLAG — bank/page.tsx:474 reads the deleted addMoney.errors.rateUnavailable key.
              Callers sweep: 4 useCurrency callers, 3 fixed, this one missed.
B. PERF:      pass — no new requests; kills the stale-response race.
C. QUALITY:   pass + smell flag (withdraw/manteca/page.tsx hotspot). Root-cause fix, good reuse, tests cover the bug.
D. SECURITY:  pass — no new endpoints, deps, or secrets.
TASK LINK:    present — #1848.
VERDICT:      request-changes (advisory) — one small fix, then good to go.
```

innolope-dev:
Thanks — addressed in bcb78db5b.

**On the blocking item:** the break does not exist as described. `bank/page.tsx:474` does not call `t('errors.rateUnavailable')` — it renders a hardcoded English literal:

```tsx
<ErrorAlert description="We couldn't load the exchange rate. Please try again in a moment." />
```

`grep -rn rateUnavailable src/` returns no hit in that file, so the deleted `addMoney.errors.rateUnavailable` key had exactly one reader (`InputAmountStep`), which this PR migrated. Nothing renders a raw key path after merge.

The *other* half of the note was right, though, and is the better fix you suggested anyway: it is the 4th `useCurrency` caller, and its error state was both untranslated and a dead end. It now renders the shared `RateUnavailable`, so all four rate-dependent flows offer the same retry in every locale.

**Minor 1 (stale rate):** valid, fixed. `useCurrency`'s `catch` now clears `price` and `symbol` before setting `isError`, so an in-place currency change whose fetch fails can't leave the previous currency's number on screen. Covered by a new test (`drops the previous currency rate when the next fetch fails`).

**Minor 2 (`withdraw/manteca` hotspot):** agreed and out of scope here, as you noted — leaving it for a separate deep dive.

Verification: `useCurrency` suite 7/7, `add-money-states` 56/56, `tsc` and `eslint` clean on the touched files, prettier applied.

kushagrasarathe (approved):
Re-review of bcb78db5b — all three findings addressed. One mechanical step left before merge.

## Resolving the disputed blocking item

We were both right, about different trees. At this branch's merge-base (b2001c743, Aug 17), `bank/page.tsx:474` had the hardcoded English literal you quoted. On **current dev**, later i18n work replaced that literal with `t('errors.rateUnavailable')` under the `addMoney` namespace — which is the tree the PR merges into, and where the break was real: a dry-run merge (`git merge-tree`) shows the i18n deletions auto-merge cleanly, so dev's version of that line would have read a deleted key. Your fix resolves it regardless: the line no longer references the key on either side.

## Findings check

1. **Bridge bank flow (blocking)** — fixed. `bank/page.tsx` now renders the shared `RateUnavailable` with `refetchRate`. All four `useCurrency` callers verified consistent; `git grep` on the merged result shows zero remaining readers of the deleted `addMoney.errors.rateUnavailable`. Bonus: the flow's error is now translated and retryable instead of a hardcoded English dead end.
2. **Stale rate (minor)** — fixed. `catch` clears `price`/`symbol` before setting `isError`, with a test that pins the exact scenario (ARS success → BRL failure → price null).
3. **Hotspot (minor)** — deferred to a separate deep dive, as agreed.

CI green on bcb78db5b (unit, typecheck, eslint, e2e, format).

## Required before merge

The branch now **conflicts with dev** in exactly one file: `bank/page.tsx` (dev turned the literal into `t('errors.rateUnavailable')`; this branch turned it into `RateUnavailable`). Merge `origin/dev` into the branch and **keep this branch's side** (`<RateUnavailable onRetry={refetchRate} />`) — resolving toward dev's side would recreate the deleted-key break, since the key deletions auto-merge. Also confirmed: dev's shared `errors` block has no `rateUnavailable` yet, so the added key can't duplicate. Re-run CI on the merged head.

## Verdict

```
TIER: T2 (frontend money flows)
A. BREAKAGE:  pass — 4/4 callers migrated; no orphaned i18n key readers post-merge
B. PERF:      pass
C. QUALITY:   pass (hotspot flag stands as follow-up)
D. SECURITY:  pass
VERDICT:      approve-advisory — merge dev in, keep the RateUnavailable side of the
              one conflict, CI green, then good for the human pass.
```

innolope-dev:
Correction to my previous comment: **you were right about the blocking issue, and I was wrong to push back.**

I checked `bank/page.tsx` on the PR branch, where the line was a hardcoded English literal, and concluded the break did not exist. You were evidently reading `dev`, which is the state that actually matters — and `dev` has since moved. Commit 23c47f79f (`fix(i18n): guard copy passed as props...`) changed that exact line to:

```tsx
{localCurrency !== 'USD' && isRateError && <ErrorAlert description={t('errors.rateUnavailable')} />}
```

with `t = useTranslations('addMoney')` — i.e. `addMoney.errors.rateUnavailable`, the key this PR deletes. So post-merge it would have rendered the raw key path, exactly as you said. It surfaced as the *only* merge conflict in the whole diff, on that one line.

The fix I already pushed resolves it, since `RateUnavailable` reads the shared `errors` namespace — the resolution keeps that and supersedes dev's version.

**This PR is being replaced rather than updated.** Resolving the conflict needs a merge commit, and merging `dev` into this branch drags in 7 pre-2026-08-18 unsigned `src/content` bumps that the signature ruleset rejects (`GH013`); force-push is blocked too. Replacement, cut fresh from `dev` with the resolved tree and identical content: (link in the next comment).

Both your minor notes are carried over — the stale-rate clear is in, the `withdraw/manteca` hotspot is filed as TASK-21934.

innolope-dev:
Replacement: https://github.com/peanutprotocol/peanut-ui/pull/2843 — cut fresh from `dev`, identical resolved tree, all review follow-ups carried over.

PR #2708, opened 2026-08-17
fix(capgo): give every OTA upload a unique bundle version

### LINKED ISSUE #2638: Capgo OTA: --version-exists-ok greens the check but ships nothing (version never bumped)
`#2636` added `--version-exists-ok` to both Capgo upload steps. That was the right call to stop a red CI check — the `deploy` job had been failing every release with:

```
■  Version 1.0.8 already exists
■  uploadBundle failed: Version 1.0.8 already exists
```

But it fixes the check, not the delivery. `--version-exists-ok` makes the upload a **silent no-op** when the version already exists. `package.json` is pinned at `1.0.8` and is not bumped per release, so:

- before `#2636`: every OTA upload after the first **failed loudly**
- after `#2636`: every OTA upload after the first **succeeds and ships nothing**

Native users get no OTA update either way. The difference is that CI no longer tells us.

## What needs to happen

Give each release a unique bundle version so the upload is real — either bump `package.json` version per release, or pass a derived version to the Capgo CLI (`1.0.8-<short-sha>` or the commit count). Then `--version-exists-ok` goes back to meaning "a retry of the same commit is safe", which is what it should mean.

## Why this is not urgent-urgent

Capgo OTA has been chronically broken for a while, so this is not a new regression — it is a known-broken path that just got quieter. But quiet is worse than loud for something nobody is watching, so it needs a real fix rather than staying like this.

## Verify when fixed

A release lands → the Capgo dashboard shows a new bundle version matching that release → a native client actually receives the update.

Found while reviewing the CI/CD PRs after Prod Release Sprint 154 (2026-08-07).

innolope-dev:
Closes #2638

## Problem

`package.json`'s version only moves on a native release, so every OTA upload after the first reused bundle version `1.0.8`. Capgo failed those uploads loudly until #2636 added `--version-exists-ok`, which turned them into silent no-ops: green check, nothing shipped.

## One correction to the issue's suggestion

`1.0.8-<short-sha>` would have made it worse. Capgo's update endpoint rejects any bundle that semver-sorts **below** the device's native version (`disable_auto_update_under_native`, on by default — [update.ts](https://github.com/Cap-go/capgo/blob/main/supabase/functions/_backend/plugin_runtime/utils/update.ts)), and a prerelease sorts under plain `1.0.8`. Uploads would succeed, the dashboard would show new bundles, and every device on the 1.0.8 binary would refuse them. The commit-count variant the issue also proposes is what this PR uses.

## Changes

- `scripts/capgo-bundle-version.mjs` — prints `<major>.<minor>` from `package.json` + the git commit count (today: `1.0.9798`). Anchoring major/minor to `package.json` keeps the version above the native floor when a native release bumps it; the commit count makes it unique per commit. It refuses a shallow clone (would emit `1.0.1`), a prerelease override, and any `CAPGO_BUNDLE_VERSION` override below the native version.
- Both Capgo workflows — `fetch-depth: 0`, a "Resolve bundle version" step, `--bundle <version>` on the upload, and a post-upload assertion that `channel currentBundle` actually serves the version the run just built. `--version-exists-ok` stays, but now only covers a re-run of the same commit; any other no-op leaves the channel stale and turns the job red. That check is the issue's "verify when fixed" criterion, enforced in CI.
- `capgo:upload:dev|staging` package scripts get the same derived version — they had the identical collision.

## Verification

- 6 new jest tests (`scripts/__tests__/capgo-bundle-version.test.js`), including a real shallow-clone repo.
- The verification step was simulated under `bash -e` for match / stale-channel / CLI-error; it is green only on a real match.
- eslint and `prettier --check` clean on the touched files; both workflows parse.

## Before the first release

Check on the Capgo dashboard that the channels' update policy is not `patch` — that setting blocks any patch-differing bundle, i.e. all OTA, regardless of the version scheme. Also note a native release that overrides `versionName` without bumping `package.json` would strand OTA bundles under the shipped binary.

### DIFF
--- a/.github/workflows/capgo-deploy-ios.yml
+++ b/.github/workflows/capgo-deploy-ios.yml
@@ -32,6 +32,8 @@ jobs:
               with:
                   submodules: true
                   token: ${{ secrets.SUBMODULE_TOKEN }}
+                  # Full history: the bundle version's patch is the commit count.
+                  fetch-depth: 0
 
             - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6
 
@@ -68,6 +70,13 @@ jobs:
                     --apikey ${{ secrets.CAPGO_API_KEY }} \
                     --ios --no-android --self-assign
 
+            - name: Resolve bundle version
+              id: version
+              run: |
+                  VERSION="$(node scripts/capgo-bundle-version.mjs)"
+                  echo "name=$VERSION" >> "$GITHUB_OUTPUT"
+                  echo "Uploading as bundle version $VERSION"
+
             - name: Upload iOS bundle to Capgo
               # Pass the commit message via env, never inline — a multi-line message
               # (or one containing quotes) injected into the run script breaks the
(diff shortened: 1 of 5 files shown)


innolope-dev:
Verified this against today's live OTA failure and it does fix it — but there's a latent gap in the floor check worth closing before it bites.

**It fixes the current outage.** Devices are on binary `1.0.53` while Capgo serves `1.0.51`, so every update check returns `disable_auto_update_under_native` — 105 events/day, OTA effectively dead for that build. With this PR, `node scripts/capgo-bundle-version.mjs` derives `1.0.9800`, comfortably above `1.0.53`. Confirmed by running it on this branch.

**The gap.** `deriveVersion` guards with `count <= Number(patch)`, where `patch` comes from `package.json` — currently **`1.0.8`**. The shipped binary is **`1.0.53`**. The guard is therefore comparing against a version that is already 45 patches stale, and only survives because the commit count (9800) happens to dwarf both.

That drift is structural, not accidental: `android-release.yml` and `ios-release.yml` both take a `versionName` dispatch input that overrides the binary version without touching `package.json`. So the anchor this script relies on is expected to be wrong.

It holds as long as releases only bump the patch. It breaks the first time one bumps the **minor**:

| | value |
|---|---|
| binary ships as | `1.1.0` |
| `package.json` still | `1.0.8` |
| derived bundle | `1.0.9801` |
| `1.0.9801` vs `1.1.0` | **sorts below** → every device refuses it |

`assertAtLeastNative` would catch that, but it only runs on the `CAPGO_BUNDLE_VERSION` override path, never on the derived one. So the failure is silent, and it's the exact scenario the file's own comment warns about — *"a release that overrides versionName without bumping package.json would strand OTA bundles under the shipped binary."*

**Suggestion:** compare against the real shipped version rather than `package.json` — read `ANDROID_VERSION_NAME` / `MARKETING_VERSION` where available, or run `assertAtLeastNative` on the derived path too so a minor drift fails the job loudly instead of shipping a bundle nobody can install.

Not a blocker for landing — this is strictly better than the status quo and unblocks the live outage. Filing it so the trap is written down.

Context: #2735 deliberately keeps `disable_auto_update_under_native` reportable while filtering the rest of the Capgo updater noise, precisely so this stays visible if it recurs.

innolope-dev:
Superseded by #2795 — same change, rebuilt on current `dev`.

This branch could not be fixed in place. `dev` has since deleted `capgo-deploy-ios.yml` (folded into `capgo-deploy.yml`), moved production OTA onto `ota-*` tags, and started appending the short SHA on non-production lanes — which is the prerelease shape this PR exists to rule out. Merging `dev` in resolves cleanly locally, but the push is rejected (`GH013`): it makes 7 pre-2026-08-18 unsigned `src/content` bumps newly reachable from this ref, and `non_fast_forward` blocks the rebase escape hatch.

#2795 carries the script, the tests, the `capgo-deploy.yml` wiring and the channel-serves-the-new-bundle assertion, reconciled with dev's consolidated lane; production's tag-derived version is left as dev wrote it.

PR #2710, opened 2026-08-17
fix(i18n): card copy that bypassed the catalog, and the prop guard that let it

innolope-dev:
## Why

The card page reads as untranslated for es-419 / pt-BR users, and it is not the plumbing: every screen under `src/components/Card/` calls `useTranslations`, the `card` namespace has 232 genuinely translated keys, and there is exactly one `NextIntlClientProvider` — above every route. What leaks is copy that never reaches the catalog.

The common cause is a guard gap. `react/jsx-no-literals` runs with `ignoreProps: true`, so it only inspects JSX **children**. Copy handed to a component as a prop is invisible to it, and that is precisely the shape that shipped English:

```tsx
<InfoCard
    title={`$${(balanceDueCents / 100).toFixed(2)} will be debited based on your next deposit`}
    description="A recent card payment ended up higher than the amount held at checkout. …"
/>
```

`ignoreProps` cannot simply be flipped — `variant="warning"`, `icon="info"` and `type="button"` are string literals too.

## What changed

**Card surface**

| Fix | Where |
|---|---|
| Balance-due notice → `card.yourCard.balanceDueTitle` / `balanceDueBody` | `YourCardScreen` |
| Card-receipt adjustment notice → `transaction.cardRows.settlementAdjustedNotice` | `CardAdjustmentNotice` |
| Legal links follow the user's locale instead of hardcoded `/en/` | `CardTermsScreen` |
| Loading-overview error → `card.errors.cardDetailsLoading` | `LockCardModal`, `CancelCardModal` |

The five card-terms links pointed at `peanut.me/en/card-esign`, `/en/card-terms-us`, `/en/card-privacy` and friends. Those marketing pages are locale-routed and fall back to English prose when a document has no translation, so linking the user's own locale is strictly better than pinning `/en/`. The marketing locale set spells its tags differently from the app's (`pt-br` vs `pt-BR`), so the href goes through `toMarketingLocale` rather than the raw locale — pinned by a new test.

The two modal errors are worth a look: they sat two lines above siblings already using `t('errors.*')`, and unlike most throws in this codebase they are rendered straight into the modal's error slot rather than collapsed by the friendly-error mapper.

**The guard**

A local ESLint rule (`eslint-rules/copy-props-from-catalog.js`) covers the gap. It checks only props that carry prose, and only values that read as prose — two or more words, with each interpolation standing in as one word. `label="CUIT"` and ``title={`$${cents}`}`` stay legal; ``title={`${amount} will be debited …`}`` does not. Both edges are pinned by RuleTester cases.

It is a **named rule** rather than another `no-restricted-syntax` selector for a reason worth flagging in review: that array lives in the repo-wide `src/**` block, and flat config *replaces* rule options instead of merging them. Scoping a second `no-restricted-syntax` to the localized surface would have silently dropped the `router.back`, nuqs and toast guards exactly where they matter most.

**Everything the rule found is fixed** — no exception list, no follow-up debt:

- `add-money/[country]/bank` reuses `addMoney.errors.rateUnavailable`, which already said this in three languages
- `ExchangeRate` had two hardcoded labels **plus three module-level English constants the rule structurally cannot see**; all five now come from the catalog
- `"Exchange rate"` existed twice once `ExchangeRate` needed it, so the transaction-row key moved to `common` — the duplicate-value drift test allows one key per string, and this is one string
- `MaintenanceBanner`, the `ReConsentModal` title, and the dismiss `aria-label` on pending-task cards

## Deliberately not changed

- **`MantecaDepositInfo` keeps `"Razón Social"`** behind a documented one-line disable. It is the field name the user's Argentine banking app displays; translating it breaks the match they are transcribing. Same precedent as the glossary's verbatim Apple Wallet quote.
- **`recover-wallet` stays outside the guard's globs.** It has zero `useTranslations` — an English-only recovery tool, like the `fix-card-signature` page already exempted right above it. Localizing it is its own job, not a silent glob widening that reds the build.
- **Backend reason prose on the application-status screen.** Every code the API resolver emits already maps to `identity.reasons.*` except `document_rejected`, which is unmapped on purpose: it only ships with the self-heal classifier's specific instruction ("Your ID photo was blurry…"), and a generic catalog line would mask it. Closing that needs the classifier's stable action code on the wire — an api-ts change.

## Locales

Keys added to en / es-419 / pt-BR. es-AR takes voseo overrides only where the es-419 string it inherits carries a tuteo verb (`vuelve`→`volvé`, `reconoces`→`reconocés`, `contacta`→`contactá`) — the glossary test checks es-AR **resolved**, so tuteo leaking through the fallback would fail.

## Verification

- `pnpm jest` — all suites green (note: `git submodule update --init src/content` is required or three suites fail on missing content)
- `tsc --noEmit` clean
- `eslint .` — 0 errors
- `prettier --check .` — clean

### DIFF
--- a/eslint-rules/__tests__/copy-props-from-catalog.test.js
+++ b/eslint-rules/__tests__/copy-props-from-catalog.test.js
@@ -1,0 +1,73 @@
+/**
+ * @jest-environment node
+ *
+ * RuleTester needs structuredClone, which this jsdom version does not provide.
+ *
+ * The rule's value is entirely in where it draws the prose/not-prose line: too
+ * loose and every `variant="warning"` becomes an error, too tight and the bug it
+ * exists for (card balance-due notice, card-receipt adjustment) walks straight
+ * through. Both edges are pinned here.
+ */
+const { RuleTester } = require('eslint')
+const tsParser = require('@typescript-eslint/parser')
+const rule = require('../copy-props-from-catalog')
+
+const ruleTester = new RuleTester({
+    languageOptions: {
+        parser: tsParser,
+        parserOptions: { ecmaVersion: 'latest', sourceType: 'module', ecmaFeatures: { jsx: true } },
+    },
+})
+
+ruleTester.run('copy-props-from-catalog', rule, {
+    valid: [
+        // not copy props
+        '<InfoCard variant="warning" icon="credit-card" />',
+        '<button type="button" className="flex items-center" />',
+        '<a href="https://peanut.me/en/card-terms-us" />',
+        // copy props holding a single token, an id or a code
+        '<PaymentInfoRow label="CUIT" />',
+        '<PaymentInfoRow label="USD" />',
+        '<Row title="card-terms-us" />',
+        // copy props fed from the catalog
+        "<InfoCard title={t('card.yourCard.balanceDueTitle')} />",
+        "<InfoCard description={t('balanceDueBody', { amount })} />",
+        // template literals that interpolate values without prose around them
+        '<InfoCard title={`$${(cents / 100).toFixed(2)}`} />',
+        '<InfoCard title={`${a}${b}`} />',
+        '<Row label={`${count}%`} />',
+    ],
+    invalid: [
+        {
+            code: '<InfoCard description="A recent card payment ended up higher than the amount held at checkout." />',
+            errors: [{ messageId: 'literal' }],
+        },
+        {
+            code: '<ActionModal title="A small update to our terms" />',
+            errors: [{ messageId: 'literal' }],
+        },
+        {
+            code: '<PaymentInfoRow label="Exchange Rate" />',
+            errors: [{ messageId: 'literal' }],
+        },
+        {
+            // the shape that shipped: prose with one interpolated amount
+            code: '<InfoCard title={`${amount} will be debited based on your next deposit`} />',
+            errors: [{ messageId: 'literal' }],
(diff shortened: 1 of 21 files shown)


innolope-dev:
### Alternative: #2711 lands the same change on `main`

This PR and **#2711** carry identical changes; pick one, close the other.

- **This one (→ `dev`)** is the normal route. It reaches users after the next `dev` → `main` promotion, and reaches the **mobile app only after a further `main` → `mobile-release` sync** — the native build ships from `mobile-release`. Since the bug was reported *in the mobile app*, that is two hops of latency.
- **#2711 (→ `main`)** is the same two commits cherry-picked onto `origin/main`, so it skips the first hop. It applied cleanly and was re-verified from scratch on the main base (231 suites / 2953 tests, `tsc` clean, `eslint .` 0 errors, prettier clean) — worth noting because `main` and `dev` have diverged by 94/63 commits and the i18n catalogs differ by ~220 keys between them, so a clean cherry-pick was not a given.

Whichever merges, the other's branch should be deleted rather than merged too — merging both would replay the same catalog inserts and conflict.

One thing neither PR does: **get this into the app**. Both stop at a web branch. The card fix only reaches the native binary once the change is on `mobile-release`, so if this is worth shipping to the users who reported it, it needs that sync as a follow-up (direct merge — per repo rules `mobile-release` does not take PRs).

PR #2721, opened 2026-08-18
fix(i18n): pt-BR landing framing + <html lang> on localized pages

0xkkonrad:
Two localization fixes found during the pt-BR launch prep, from `mono/inbox/localization-launch-2026-08-18/HANDOFF-FIXES.md`. Same domain, same file tree, one review.

- **Fix 1 — pt-BR landing copy breaks the Brazil framing rule.** Launch-gating. Copy only.
- **Fix 2 — `<html lang>` stays `en` on localized pages.** Low priority, assistive tech only.

Task: [fix pt-BR framing](https://app.notion.com/p/3c083811757981d997f9fe6c892f1d8a) · [`lang="en"` on localized pages](https://app.notion.com/p/3c083811757981d88766d8c39b9ea2cf)

---

# Fix 1 — positive passport framing on `/pt-br`

Brazil has high internet fraud rates, so copy implying you can get around a government ID requirement reads as *illegal*, not convenient — and the tab closes. `content/_system/context/localization.pt-br.md` §6 bans that framing in **all headings, subtitles and H-tags**; §7 names the prescribed replacements.

The page already contradicted itself: `landingPayLocalBody`, one line below the offending `<h2>`, already reads *"Só com seu passaporte."*

| Key | Surface | Before | After |
|---|---|---|---|
| `landingHeroNoLocalId` | hero subtitle | `Sem documento local nem banco.` | `Funciona com qualquer passaporte.` |
| `landingPayLocalSubheading` | Manteca `<h2>` | `RECEBA DE QUALQUER LUGAR. SEM DOCUMENTO LOCAL.` | `RECEBA DE QUALQUER LUGAR. COM SEU PASSAPORTE.` |
| `landingPixAria` | PIX link `aria-label` | `Como funciona o Pix sem CPF` | `Como funciona o Pix com seu passaporte` |

**pt-BR only.** No other locale file changes.

### Wording notes

- **`landingPixAria` was not in the original findings.** It is a literal `sem CPF` — the phrase §7 names first — rendered twice as the PIX link's accessible name, live on prod. Leaving it would have failed this fix's own acceptance check.
- **The `<h2>` says `COM SEU PASSAPORTE`, not the handoff's proposed `COM QUALQUER PASSAPORTE`.** That repeats `QUALQUER` four words after `DE QUALQUER LUGAR`, which reads as a translation artifact. `COM SEU PASSAPORTE` is §6's canonical phrasing and is one character *shorter* than the string it replaces.
- **The `<h2>` now echoes the `<h3>`'s passport beat.** Deliberate — the English source does the same thing (`NO LOCAL ID NEEDED` → `Just your passport.`). The alternative (`NA HORA E EM DÓLARES`) was rejected: it drops the access claim *and* echoes `landingPayLocalSettles` ("dólares digitais") directly below instead.
- **The hero drops the "bank" half** of `No local ID or bank required.` Keeping it needs `não precisa de conta em banco local`, which pushes a 14px muted line to 70 chars / 2 lines at 375px.

### Spanish deliberately untouched

`es-419` / `es-ar` use the same construction (`SIN DOCUMENTO LOCAL`) and it stays. Their own rule files mandate it — `localization.es-419.md` §1 says to emphasize "no-ID barrier removal", `localization.es-ar.md` §4 says "focus on the access unlock (no DNI needed)". `content/_system/HUMAN-AUDITS.md` shows both locales audited in the same round: the Brazilian reviewer raised this, three Argentine native reviewers did not, and a human scoped the rule to pt-BR holding both sets of feedback. 95 shipped Spanish content files use `sin DNI`, 61 in titles or heroes — changing the landing page alone would desync it from all of them.

### The fix is now enforced

`src/i18n/app/__tests__/glossary.test.ts` already banned `sem CPF` and the rest of the circumvention framing — but it only read `src/i18n/app/messages/*`. The landing copy lives in `src/i18n/*.json`, which **nothing checked**. That is how `Pix sem CPF` reached prod and stayed green through CI.

This extends that suite to the three marketing catalogs, reusing the same rule table rather than a second copy. All three pass every existing rule as-is. Restoring the old value fails by name:

```
✕ "sem CPF" framing is banned (trust rules)
+   "landingPixAria",
```

The Spanish rules do not ban `sin documento local`, and this does not add them.

---

# Fix 2 — `<html lang>` on localized pages

`peanut.me/pt-br` reports `<html lang="en">` seconds after hydration. Same for `/es-419` and `/es-ar`.

**The cause is not a missing mount.** `HtmlLang` *is* mounted on the landing routes — [`src/app/pt-br/page.tsx:19`](https://github.com/peanutprotocol/peanut-ui/blob/dev/src/app/pt-br/page.tsx#L19). The problem is that **two components write `document.documentElement.lang`**:

| Writer | Locale it writes | Where it sits |
|---|---|---|
| `HtmlLang` | **page** locale, from the route | inside each landing/marketing route |
| `AppIntlProvider` | **app** locale — cookie / localStorage / `navigator`, **never the URL** | `ClientProviders`, above *every* route |

React commits child effects before parent ones, so `AppIntlProvider` always ran **last** and undid the page locale. A direct hit on `/pt-br` from an English browser resolves the app locale to `en`, so the page ends up labelled English.

Fix, in two halves:

1. **`HtmlLang` claims the attribute while mounted; `AppIntlProvider` stands down while a claim is held.** A small shared module holds the claim so `src/i18n/*` doesn't import from `src/components/*`.
2. **On release, the provider re-applies the app locale.** Restoring the snapshot `HtmlLang` took at mount is not enough — on a direct hit that snapshot is the root layout's pre-hydration `en`, so a pt-BR user leaving a landing page would run the rest of the session under `lang="en"`. The provider's own effect keys on the app locale, which doesn't change on navigation, so nothing else would put it back.

`src/app/page.tsx` (the English landing) had no `HtmlLang` at all — it now claims too, otherwise a visitor with a `pt-BR` cookie would have the English page relabelled Portuguese. It sits **inside** `LandingPageCapacitorGate`: on native this route is only a bootstrap shell that redirects away, and the device locale is what `<html lang>` should report there.

### Impact — assistive tech only, not SEO

An earlier write-up called this an SEO bug. It isn't. `hreflang` is emitted correctly and is what crawlers use — verified live on `peanut.me/pt-br`: 5 correct `link rel="alternate"` tags. Unchanged here. What was broken is screen-reader pronunciation and in-browser tooling.

### Design notes

- **A claim counter, not effect reordering.** Parent passive effects always run after child ones, and `useLayoutEffect` doesn't help — all layout effects flush before any passive effect. One writer has to yield.
- **Counter, not a boolean** — paired increment/decrement survives StrictMode double-invoke and overlapping route transitions. `reactStrictMode` is `false` today, so this is future-proofing.
- **Rejected alternative:** having the landing seed the app locale via `persistLocale`. That writes the `app-locale` cookie, which `src/proxy.ts` then uses to 307 `/` → `/{locale}` permanently. Far larger blast radius for a `lang` attribute.
- **Case skew is expected.** Marketing tags are lowercase (`pt-br`), app tags are `pt-BR`. BCP 47 is case-insensitive; tests don't assert exact case.

---

# Risks

Fix 1 is copy-only: one locale file, three string values, no key renames. Fix 2 is wider — `AppIntlProvider` is on every route — but the guard is one boolean check that only ever *skips* a write; it cannot write a wrong value. Non-marketing routes claim nothing, so the provider keeps full control there.

No `headers()`, no `params`, no dynamic API added — the native static export (`output: 'export'`) is unaffected.

# QA

Local gate: `prettier --check .` ✅ · `typecheck` ✅ · `npm test` **239 suites / 3086 tests** ✅. Baseline measured on `origin/dev` before any change: 237 suites / 3060 tests.

**Both fixes are mutation-checked** — each one, removed, fails its own test:

| Mutation | Result |
|---|---|
| restore the old `landingPixAria` | glossary suite fails, naming the key |
| remove the claim guard | 2 `html-lang` tests fail |
| remove the release listener | release test fails — `Expected "pt-BR", Received "en"` |

The `<html lang>` tests render the *real* `AppIntlProvider`, not a stand-in. They live in two files on purpose: the locale store memoizes the resolved startup locale per module, so a preceding test in the same file shifts the timing the release case depends on — which is exactly how the release gap survived the first round of tests here.

Real browser, 375×667, `Accept-Language: en-US` (the failing condition), probed 4s after `networkidle`:

| Route | `lang` before | `lang` after | `sem documento local` | `sem CPF` |
|---|---|---|---|---|
| `/pt-br` | `en` ❌ | `pt-br` ✅ | gone ✅ | gone ✅ |
| `/es-419` | `en` ❌ | `es-419` ✅ | — (kept, deliberate) | — |
| `/es-ar` | `en` ❌ | `es-ar` ✅ | — (kept, deliberate) | — |
| `/pt-br/help` | `en` ❌ | `pt-br` ✅ | — | — |
| `/` | `en` | `en` ✅ | — | — |

With a **pt-BR** app locale (the case a first round of QA missed, because an English app locale makes a stale snapshot and the correct value the same string):

| Step | before | after |
|---|---|---|
| `/es-419` | `pt-BR` ❌ (app locale won) | `es-419` ✅ |
| then navigate to `/setup` | `en` ❌ (stale snapshot) | `pt-BR` ✅ |

Console errors on `/pt-br` are identical to the untouched `/es-419` control route — nothing introduced.

# Screenshots

375×667, mobile. "Before" is **live prod** `peanut.me/pt-br`; "after" is this branch. Fix 2 changes an HTML attribute, so it has no visual output.

| Before (prod) | After (this branch) |
|---|---|
| before | after |
| before hero | after hero |

> The "after" shots were taken before fix 2 was folded in, so their marquee/CTA still render English from the default app locale. The landing strings under test are all pt-BR.

The `pr-assets-ptbr` branch is deleted after merge.

# Follow-ups found but deliberately NOT bundled

Found while sweeping `pt-br.json` against the same rule — all filed, none blocking:

1. `landingPayLocalSettles` — `"A liquidação é em dólares digitais…"`. To a Brazilian consumer *liquidação* reads first as a clearance sale; it's bank-desk register. Naturalness, not framing.
2. `landingSecurityControlDesc` — `"…sem KYC obrigatório nem burocracia."` §3 puts "KYC" on the avoid list. Judged **not** a §6 violation (the sentence explicitly says the card *does* require verification, which is the prescribed positive framing), and any rewrite drops a product claim — needs product sign-off, not a copy PR.
3. `pressSubtitle` — `"sobre a Peanut"` (feminine) where every other string uses `o Peanut`. One-word consistency fix.
4. The card block implies open availability, contradicting §5's canonical "beta fechado / lista de espera" line. **`en.json` says the same thing** — a source-copy issue across every locale, not a translation defect.
5. Four mono pt-BR **content** pages carry §6 violations in headings (`send-to/united-kingdom`, `receive-from/argentina`, `use-cases/digital-nomads`, `use-cases/remote-workers`). Content repo, separate path.

# Note for the reviewer

The handoff doc this came from states the landing route "never mounts `HtmlLang` at all" and that no route serves `/[locale]`. Both are wrong, and worth knowing: the routes are **literal segments** (`src/app/pt-br/`, `src/app/es-419/`, `src/app/es-ar/`) — a `[locale]/page.tsx` would shadow `[...recipient]` and break every `peanut.me/{username}` profile, which the files' own header comments explain. The middleware is `src/proxy.ts` (Next 16 renamed it from `middleware.ts`), which is likely why it wasn't found. The missing `<main>` on the landing page is real but unrelated — the landing tree renders a plain `div` by design. The handoff has been corrected in mono.


0xkkonrad:
@coderabbitai review

0xkkonrad:
@coderabbitai review

0xkkonrad:
@coderabbitai review

jjramirezn (approved):

PR #2722, opened 2026-08-18
fix(i18n): stop the app locale overwriting <html lang> on localized pages

0xkkonrad:
## Summary

`peanut.me/pt-br` reports `<html lang="en">` in a real browser, seconds after hydration. Same for `/es-419` and `/es-ar`.

**The cause is not a missing mount.** `HtmlLang` *is* mounted on the landing routes — [`src/app/pt-br/page.tsx:19`](https://github.com/peanutprotocol/peanut-ui/blob/dev/src/app/pt-br/page.tsx#L19). The problem is that **two components write `document.documentElement.lang`**:

| Writer | Locale it writes | Where it sits |
|---|---|---|
| `HtmlLang` | **page** locale, from the route | inside each landing/marketing route |
| `AppIntlProvider` | **app** locale — cookie / localStorage / `navigator`, **never the URL** | `ClientProviders`, above *every* route |

React commits child effects before parent ones, so `AppIntlProvider` always ran **last** and undid the page locale. A direct hit on `/pt-br` from an English browser resolves the app locale to `en`, so the page ends up labelled English.

Fix, in two halves:

1. **`HtmlLang` claims the attribute while mounted; `AppIntlProvider` stands down while a claim is held.** A small shared module holds the claim so `src/i18n/*` doesn't have to import from `src/components/*`.
2. **On release, the provider re-applies the app locale.** Restoring the snapshot `HtmlLang` took at mount is not enough — on a direct hit that snapshot is the root layout's pre-hydration `en`, so a pt-BR user leaving a landing page would run the rest of the session under `lang="en"`. The provider's own effect keys on the app locale, which doesn't change on navigation, so nothing else would put it back.

`src/app/page.tsx` (the English landing) had no `HtmlLang` at all — it now claims too, otherwise a visitor with a `pt-BR` cookie would have the English page relabelled Portuguese. It sits **inside** `LandingPageCapacitorGate`: on native this route is only a bootstrap shell that redirects away, and the device locale — not this page's English — is what `<html lang>` should report there.

## Task

- Notion: [`lang="en"` on localized pages](https://app.notion.com/p/3c083811757981d88766d8c39b9ea2cf)
- Source: `mono/inbox/localization-launch-2026-08-18/HANDOFF-FIXES.md` (fix 2)

**Not launch-blocking.** Sibling PR #2721 is the one that gates the localization launch.

## Impact — assistive tech only, not SEO

The handoff's earlier framing called this an SEO bug. It isn't. `hreflang` is emitted correctly and is what crawlers use — verified live on `peanut.me/pt-br`: 5 correct `link rel="alternate"` tags (`x-default`, `en`, `es-419`, `es-AR`, `pt-BR`). Unchanged by this PR.

What was actually broken is screen-reader pronunciation and in-browser tooling (translate prompts, spellcheck) on the localized landings.

- **A claim counter, not effect reordering.** Parent passive effects always run after child ones, and `useLayoutEffect` doesn't help — all layout effects flush before any passive effect. There is no ordering trick; one writer has to yield.
- **Counter, not a boolean.** Paired increment/decrement survives StrictMode double-invoke and overlapping route transitions. `reactStrictMode` is `false` today, so this is future-proofing, not current need.
- **Rejected alternative:** having the landing seed the app locale via `persistLocale`. That writes the `app-locale` cookie, which `src/proxy.ts` then uses to 307 `/` → `/{locale}` permanently. Much larger blast radius for a `lang` attribute.
- **Case skew is expected.** Marketing tags are lowercase (`pt-br`), app tags are `pt-BR`. BCP 47 is case-insensitive, so the tests compare lowercased and don't assert exact case.

## Risks

Low, but wider than a copy change — `AppIntlProvider` is on every route. The guard is one boolean check and it only ever *skips* a write; it cannot write a wrong value. Non-marketing routes claim nothing, so the provider keeps full control there (verified below).

No `headers()`, no `params`, no dynamic API added — the native static export (`output: 'export'`) is unaffected.

## QA

Local gate: `prettier --check .` ✅ · `typecheck` ✅ · `npm test` **239/239 suites, 3064 tests** ✅ (237 suites on the `origin/dev` baseline + two new ones).

**The new tests reproduce the live bug**, rendering the *real* `AppIntlProvider` rather than a stand-in. Both fixes are mutation-checked — each one, removed, fails its own test:

| Mutation | Result |
|---|---|
| remove the claim guard (the original bug) | 2 tests fail |
| remove the release listener | release test fails — `Expected "pt-BR", Received "en"` |

They live in two files on purpose: the locale store memoizes the resolved startup locale per module, so a preceding test in the same file shifts the timing the release case depends on. That is exactly how the release gap survived the first round of tests here.

Real browser, 375×667, `Accept-Language: en-US` (the failing condition), probed 4s after `networkidle`:

| Route | `documentElement.lang` before | after |
|---|---|---|
| `/pt-br` | `en` ❌ (live prod today) | `pt-br` ✅ |
| `/es-419` | `en` ❌ | `es-419` ✅ |
| `/es-ar` | `en` ❌ | `es-ar` ✅ |
| `/pt-br/help` | `en` ❌ | `pt-br` ✅ |
| `/` | `en` | `en` ✅ (unchanged) |

With a **pt-BR** app locale (the case the first round of QA missed, because every row above uses `en` where a stale snapshot and the correct value are the same string):

| Step | before this PR | after |
|---|---|---|
| `/es-419` | `pt-BR` ❌ (app locale won) | `es-419` ✅ |
| then navigate to `/setup` | `en` ❌ (stale snapshot) | `pt-BR` ✅ |

## Screenshots

`N/A (no visible change)` — this changes an HTML attribute, not rendered output.

## Note for the reviewer

The handoff doc this came from states the landing route "never mounts `HtmlLang` at all" and that no route serves `/[locale]`. Both are wrong, and worth correcting for anyone reading it: the routes are **literal segments** (`src/app/pt-br/`, `src/app/es-419/`, `src/app/es-ar/`) — a `[locale]/page.tsx` would shadow `[...recipient]` and break every `peanut.me/{username}` profile, which the files' own header comments explain. The middleware is `src/proxy.ts` (Next 16 renamed it from `middleware.ts`), which is likely why it wasn't found. The missing `<main>` on the landing page is real but unrelated — the landing tree renders a plain `div` by design.

### DIFF
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -13,9 +13,12 @@ export const metadata = landingMetadata(DEFAULT_LOCALE)
 export default function RootPage() {
     return (
         <>
-            <HtmlLang locale={DEFAULT_LOCALE} />
             <LocaleSuggestion locale={DEFAULT_LOCALE} />
             <LandingPageCapacitorGate>
+                {/* Inside the gate: on native this route is only a bootstrap
+                    shell that redirects away, and the device locale — not this
+                    page's English — is what <html lang> should report there. */}
+                <HtmlLang locale={DEFAULT_LOCALE} />
                 <LandingPageShell>
                     <LandingPageContent locale={DEFAULT_LOCALE} />
                 </LandingPageShell>
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -1,6 +1,7 @@
 import { LandingPageShell } from '@/components/LandingPage/LandingPageShell'
 import { LandingPageCapacitorGate } from '@/components/LandingPage/LandingPageCapacitorGate'
 import { LandingPageContent } from '@/components/LandingPage/LandingPageContent'
+import { HtmlLang } from '@/components/Marketing/HtmlLang'
 import { LocaleSuggestion } from '@/components/Marketing/LocaleSuggestion'
 import { DEFAULT_LOCALE } from '@/i18n/types'
 import { landingMetadata } from '@/lib/seo/landing'
@@ -12,6 +13,7 @@ export const metadata = landingMetadata(DEFAULT_LOCALE)
 export default function RootPage() {
     return (
         <>
+            <HtmlLang locale={DEFAULT_LOCALE} />
             <LocaleSuggestion locale={DEFAULT_LOCALE} />
             <LandingPageCapacitorGate>
                 <LandingPageShell>
--- a/src/components/Marketing/HtmlLang.tsx
+++ b/src/components/Marketing/HtmlLang.tsx
@@ -2,6 +2,7 @@
 
 import { useEffect } from 'react'
 import type { Locale } from '@/i18n/types'
+import { claimHtmlLang, releaseHtmlLang } from '@/i18n/htmlLangClaim'
 
 /**
  * Stamps the page locale onto <html lang>. The root layout sits above the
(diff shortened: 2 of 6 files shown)


0xkkonrad:
@coderabbitai review

0xkkonrad:
@coderabbitai review

0xkkonrad:
Folded into #2721 — one PR for both localization fixes, to save reviewer time. Same four commits, same base (`dev`), all signed and cherry-picked onto `fix/ptbr-landing-framing`; nothing was dropped, and the review findings that landed here are included.

Closing this one. Review #2721 instead.

PR #2728, opened 2026-08-18
feat(mascot): move the mascot to Lottie and drop 6 MB of rasters

0xkkonrad:
Moves the mascot from animated WebP/GIF to Lottie. Vector art scales to any size, the files are far smaller, and playback becomes something we control at runtime instead of something baked into a sprite.

Batch 2 of the rigs landed last week, so the set now covers all ten animated mascots one-for-one and the whole surface can move at once.

## What changed

**New component** — `src/components/Global/PeanutMascot/`

```tsx
<PeanutMascot pose="cheering" className="size-24" alt="…" />
```

- lottie-web's light SVG build, loaded with a dynamic `import()` inside the mount effect, so nothing enters the SSR graph.
- Each pose's JSON is its own dynamic import, so a screen downloads only the pose it shows.
- The 1930s stutter is applied at runtime — the clock is quantized to held frames rather than played smoothly. `MASCOT_HOLD_FRAMES = 2`, `MASCOT_SPEED = 1.30`, the values picked in July. Nothing is baked into the files, so re-tuning the feel is a constant, not a re-export.
- One `requestAnimationFrame` loop drives every instance on the page and stops when the last one unmounts. Ticking is gated by `IntersectionObserver`, and `prefers-reduced-motion` holds a static frame.

**Sizing comes from the artwork, not the canvas.** The comps carry different per-pose padding, so sizing the SVG to its 1050×1000 canvas renders every pose at a different apparent size — measured at −20% to +23% against the WebP each one replaces. `MASCOT_ART_BOXES` holds each pose's art bounding box and the component places by that instead. `MASCOT_ART_FILL` is now the single knob for mascot size across the entire app.

**Call sites.** Most were a direct swap. Six passed the image URL through a data structure and needed the structure to carry a pose:

- Setup steps now describe an illustration as `{ src }` or `{ pose }` — three steps use a still, five use a mascot — and `SetupWrapper` renders whichever it gets.
- `InvitesPageLayout` takes a pose rather than an image URL; both callers only ever passed mascots.
- Home carousel CTAs take a `mascotPose`.

**`RejectionAssetD3` deliberately stays on rasters.** It is drawn with D3 and captured into a shareable image; a live Lottie cannot be serialized through that pipeline. The three poses it uses keep their WebP behind a dedicated barrel export, commented on both sides.

**Removed** — 10 GIFs and 7 WebPs. The GIF fallbacks existed only because old WebKit cannot animate WebP; lottie-web's SVG renderer has no such problem, so `isLegacyWebKit` goes with them. `getIOSMajorVersion` stays, still used by `instrumentation-client` for Sentry replay gating.

**Added to `/dev/components`** with all ten poses, per the house rule that the showcase carries every component and variant.

## Payload

| | |
|---|---|
| Raster removed | **5.97 MB** across 17 files |
| Lottie added | **90 KB** gzipped, 10 rigs |
| Player added | **45 KB** gzipped (`lottie_light`, SVG renderer only) |

Lottie JSON compresses about nine times over; WebP and GIF are already compressed and do not shrink further.

## Verification

Local gate, measured against `origin/dev` as the baseline:

| | baseline | this branch |
|---|---|---|
| `tsc --noEmit` | 0 errors | **0 errors** |
| `npm test` | 237 suites / 3060 | **238 suites / 3068** |
| `prettier --check` | clean | **clean** |
| `next build` | — | **passes** |

Nothing was skipped or deleted to get green. The extra suite is the component's own tests.

### Rendered size, real screens, before vs after

Both builds were served side by side and the **rendered artwork height** measured on each — opaque-pixel bounds for the raster, `getBBox` for the vector:

| screen | pose | before | after | delta |
|---|---|---|---|---|
| landing hero | waving-chill | 227.6 | 239.3 | +5% |
| careers | too-cool | 234.3 | 239.7 | +2% |
| maintenance | worried | 230.5 | 235.6 | +2% |
| setup signup | thinking | 232.4 | 248.8 | +7% |
| invalid invite | sad | 89.4 | 87.9 | −2% |
| setup landing | waving-chill | 244.3 | 210.7 | −14% |
| setup finish | waving-chill | 260.7 | 225.6 | −13% |

The first measurement pass caught two real regressions, both now fixed in `38e751f`: the setup and invites heroes had swapped a square sprite's box for an unrelated `dvh` height and lost 14–18%, and `MASCOT_ART_FILL` sat at the median of the old per-pose fills rather than the top of the range.

The two remaining `waving-chill` rows are expected and not a defect: that rig is genuinely wider than the whistling sprite it replaces (art aspect 1.21 vs 1.01), so in the same square box it trades height for width.

### Canary screenshots

24 before/after pairs — 8 screens × 6 mobile widths (320 / 360 / 375 / 390 / 414 / 430), captured against production builds of both branches. Every after-shot mounts a Lottie and none still render a raster. Screens covered: landing hero, careers, maintenance, setup landing, setup signup, setup finish, invite, invalid invite.

## Worth a reviewer's judgement

1. **Two poses are forced remaps.** `PeanutWhistling` → `waving-chill` and `PeanutCrying` → `worried`; there is no `whistling` or `crying` rig in the set. `worried` is the teary pose and now carries the maintenance screen, claim errors and empty states. Worth an opinion on whether teary is the right register there.
2. **Four poses are not covered by the canary** — cheering, pointing, pointing-down and walking live behind auth or seeded state (claim success, payment success, physical card, marketing CTA). They are covered by the component tests and by `/dev/components`, but no reviewer has seen them on their real screen yet.
3. **`MASCOT_ART_FILL = 0.93`** was chosen so no screen ends up smaller than before. Raising it is a one-line change if we want the mascots larger overall.


0xkkonrad:
## Canary screenshots — 24 before/after pairs

📸 **[Full canary — all 24 pairs, side by side](https://claude.ai/code/artifact/a3740689-cd83-46a8-b8ca-53693ea4cc8b)**

Captured against **production builds of both branches**, served locally and driven with Playwright at device scale 2. Left of each pair is the WebP shipping today; right is the same route, same viewport, on this branch.

**8 screens × 6 mobile widths:**

| | |
|---|---|
| Screens | landing hero, careers, maintenance, setup landing, setup signup, setup finish, invite, invalid invite |
| Widths | 320, 360, 375, 390, 414, 430 |
| Poses seen | waving-chill, too-cool, worried, thinking, sad |
| Result | **24/24 mount a rig, 0 still render a raster, 0 page errors** |

### What the canary caught

It earned its keep — the first capture pass exposed two regressions that the type checker and 3068 unit tests were both blind to:

1. **Setup and invites heroes lost 14–18% of their height.** Both had replaced a square sprite's box with an unrelated `dvh` height. The sprite was 320×320 and sized by width with `object-contain`, so `aspect-square` restores the box it actually occupied.
2. **Everything else was 3–6% short**, because `MASCOT_ART_FILL` sat at the median of the old per-pose fills rather than the top of the range.

Both fixed in `38e751f`, then re-measured. Current state, comparing **rendered artwork height** (opaque-pixel bounds for the raster, `getBBox` for the vector — the drawing, not the box):

| screen | pose | before | after | delta |
|---|---|---|---|---|
| setup signup | thinking | 232.4 | 248.8 | **+7%** |
| landing hero | waving-chill | 227.6 | 239.3 | **+5%** |
| careers | too-cool | 234.3 | 239.7 | **+2%** |
| maintenance | worried | 230.5 | 235.6 | **+2%** |
| invalid invite | sad | 89.4 | 87.9 | −2% |
| setup finish | waving-chill | 260.7 | 225.6 | −13% |
| setup landing | waving-chill | 244.3 | 210.7 | −14% |

The two `waving-chill` rows are expected rather than a defect: that rig is genuinely wider than the whistling sprite it replaces (art aspect 1.21 against 1.01), so in the same square box it trades height for width. Visible in the pairs — it reads as the same visual weight.

### Not covered

**cheering, pointing, pointing-down and walking** sit behind auth or seeded state (claim success, payment success, physical card, marketing CTA), so no canary shot reaches them. They have component tests and appear in `/dev/components`, but nobody has seen them on their real screen. If someone with a local session can open those four, that closes the last gap.

0xkkonrad:
## Independent visual QA — verdict: safe to ship

An independent pass read all 24 before/after pairs plus 10 high-magnification crops. Summary of what came back.

### Clean

- **No missing or invisible mascot.** All 24 after-shots render a rig.
- **No clipping.** The host has `overflow-hidden`, so the two plausible failure modes were checked specifically — the top edge of the signup panel at 320/375/390 and the right edge of the hero at 390/430. Heads, gloves, shoes and glasses fully drawn; nearest approach to a container edge is ~45 css px.
- **No layout shift.** Whole-frame pixel diffs show the only changes outside the mascot's own box are the animated marquee phase and a loading spinner. The SIGN UP button and yellow ticker land on identical `y` in both builds.
- **Both forced pose remaps read correctly.** `waving-chill` keeps the whistling "O" muzzle and round eyes, so it's recognisably the same character. `worried` still cries actual tear drops with downturned brows — same emotional register as the pose it replaces.
- **careers and invalid-invite are effectively unchanged** — within 4 css px on every axis.

### The sharpness win is real, not just different

Three zoom pairs, all decisive:

- **careers face at 3.4×** — the WebP has stepped edges on the sunglasses frame, a grey halo along every black stroke, banding in the yellow fill. The Lottie is flat colour, clean curves, zero fringing.
- **signup face at 2.6×** — the WebP renders the lenses as grey mush with magenta fringing; the Lottie draws the glare as clean white strokes.
- **hero glove at 1.6×** — the WebP outline is soft with a pink halo bleeding into the background; the Lottie stroke is solid black on flat pink.

Most visible at 414/430, where the raster was being upscaled hardest.

### One judgement call — the hero, and it's yours not mine

👉 **[See it side by side at 430 / 375 / 320](https://claude.ai/code/artifact/a3740689-cd83-46a8-b8ca-53693ea4cc8b)**

The whistling sprite held its arms on a diagonal — one down-left, one up-right. The `waving-chill` rig holds both arms out **horizontally**, at roughly the height of the graffiti headline behind it. At 430 that puts the gloves and head across "LOCAL FEEL", which the old pose left more readable. At 375 and 320 the difference is marginal.

Worth being precise about the scope: the headline is decorative artwork, not copy. "GLOBAL CASH" and "TAP. SCAN. ANYWHERE." are untouched, and both builds already overlap it deliberately — the hero comment in `hero.tsx` says the overlap is intended. So this is a question of degree on the flagship marketing page.

**I have deliberately not changed it.** It's a composition call for whoever owns the hero, and the lever is that component's own box rather than anything in `PeanutMascot`. Happy to adjust if you want it to occlude less.

Two smaller notes, both nitpicks, neither a defect: the mascot sits ~10% larger and slightly right-of-centre on maintenance and setup-signup, and setup-finish is arguably improved — the old raster flung one arm toward the bottom-left corner, the rig is symmetrical.

0xkkonrad:
## Adversarial code review — 2 real bugs found and fixed (`00699aea5`)

An independent review went at the diff looking for user-facing breakage. Two findings were real and are fixed; I checked each myself rather than taking them on trust.

### 1. HIGH — a failed mascot chunk could reload the page

`PeanutMascot/index.tsx` fired the lazy loader with **no `.catch`**. Webpack rejects a failed dynamic import with `ChunkLoadError`, and this app inlines a global handler in `layout.tsx` that listens on `unhandledrejection` and calls `window.location.reload()` for exactly that error.

Verified: [`chunk-error-recovery.ts:99`](https://github.com/peanutprotocol/peanut-ui/blob/feat/mascot-lottie/src/utils/chunk-error-recovery.ts#L99) does call `window.location.reload()`, and [`layout.tsx:168`](https://github.com/peanutprotocol/peanut-ui/blob/feat/mascot-lottie/src/app/layout.tsx#L168) inlines it.

Trigger: a tab open across a deploy — the deploy-skew case that file's own header cites as a real production incident — or a flaky mobile connection. A **decorative** asset 404s and the page reloads under the user. That lands on terminal screens: payment success, claim success, the qr-pay loading state, the delete-account confirmation. Confetti and in-page state discarded.

Before this PR those screens had zero dynamic imports for their artwork — a failed `` was inert. Now caught and swallowed, so the box just stays empty, which is what the `` did.

### 2. MEDIUM — qr-pay mascot lost its centring

The old markup was `<Image layout="fill" objectFit="contain" className="absolute z-0 h-32 w-32 ...">`. next/image maps `layout="fill"` to inline `position:absolute; inset:0; width:100%; height:100%`, which **silently overrode** the `h-32 w-32` classes — so the mascot stretched across the wrapper and appeared centred.

The new host actually obeys `h-32 w-32`. With all four insets `auto`, an absolutely-positioned element lands at its static position: the wrapper's **top-left corner**. The mascot would have jumped left by roughly `(cardWidth − 128) / 2` on every qr-pay loading state. This was the one call site where the class list was inert before and load-bearing now. Now centred explicitly with `left-1/2 -translate-x-1/2`.

### Also added — a guard the review asked for

`MASCOT_CANVAS_WIDTH/HEIGHT` duplicated each comp's dimensions with nothing enforcing it; re-exporting one rig at a different canvas size would silently mis-centre that pose with all 3068 tests still green. Now asserted against the JSON.

Writing that test turned up something worth recording: **`walking` reaches 16.8 units past the bottom of its 1050×1000 comp mid-stride.** Harmless — the art box drives placement and carries the overflow with it, and the sprite it replaces came off the same comp — but it is now documented rather than latent.

### One finding I did not action

The review flagged `alt="Peanutman crying 😭"` and `t('cryingPeanutAlt')` as stale, since `PeanutCrying` is now `pose="worried"`. **The copy is still accurate** — the `worried` rig has literal `Tear 1` and `Tear 2` layers and the peanut visibly cries. Verified in the JSON and in the maintenance screenshots. Renaming the keys would make them less true, not more.

Two other notes accepted as tradeoffs rather than defects: mascots are now client-rendered so the `priority` preload is gone (an empty box on first paint, in exchange for 368 KB → 55 KB on the thinking screen), and a mascot scrolled out and back restarts its loop at frame 0.

**Gate after the fixes:** typecheck 0 errors, 238 suites / 3069 tests green, prettier clean.

0xkkonrad:
## Ready for human review

7 commits, all signed. CI green. This went through the full `peanut-pr` lifecycle: local gate → CI → automated review → smell/docs pass → re-verify. Stopping here, per the rule that the skill never merges.

### Gate

| | baseline (`origin/dev`) | this branch |
|---|---|---|
| `tsc --noEmit` | 0 errors | **0 errors** |
| `npm test` | 237 suites / 3060 | **238 suites / 3071** |
| `prettier --check` | clean | **clean** |
| `next build` | — | **passes** |
| CI | — | **18 pass, 0 fail** |

Nothing skipped or deleted to get green. The extra 11 tests are the component's own.

### What review caught, and what it cost

Four separate passes ran over this — two adversarial subagents, the `/code-review` skill, and a measured visual canary. Between them they found **six real defects**, none of which typecheck or 3060 unit tests would have caught:

1. **A leaked rAF loop.** The shared clock re-armed itself after its last subscriber left, spinning forever with zero subscribers. Fixed with a regression test verified to fail without the fix.
2. **A failed chunk could reload the page.** The lazy loader had no `.catch`, and this app reloads on `ChunkLoadError` via a global handler — so a decorative mascot failing to fetch during a deploy would reload a payment-success screen and discard its state.
3. **qr-pay lost its centring.** `<Image layout="fill">` had been emitting an inline `inset:0` that silently overrode the size classes; the new host obeys them and fell to the wrapper's top-left.
4. **Setup and invites heroes lost 14–18% of their height** — a square sprite's box replaced with an unrelated `dvh` height.
5. **Careers grew without bound** — sized by width inside a half-width column, ~1000px tall on desktop against a 320px sprite.
6. **The landing hero's feet stopped overlapping the headline** — it measured the host box, but the drawing fills 93% of it and is centred, so the feet floated by the leftover padding.

### Mascot sizing is now standardized

The original problem was that nothing was framed consistently — the WebP sprites were cropped tight, the Lottie comps carry per-pose padding, and square boxes fit the *wider* of a pose's two dimensions. Three things fix it:

- The host takes its **aspect from the pose**, so a call site gives a height and every pose renders at that height.
- **`MASCOT_HERO_CLASS`** for screens the mascot leads (setup steps, invite, waitlist).
- **`MASCOT_STATE_CLASS`** for screens where it sits above a headline (errors, card status, physical card).

Measured on the running app:

| group | 375 | 390 | 430 | spread |
|---|---|---|---|---|
| hero screens | 208–211 | 262–266 | 291–301 | **3–10px** |
| invalid invite | 173 | 178 | — | was 118, raster was 89 |

`MASCOT_ART_FILL` remains the single knob if you want the whole set bigger.

### Two things for the reviewer, not defects

1. **The hero pose occludes more of "LOCAL FEEL" at 430.** `waving-chill` holds its arms horizontally where the whistling sprite held them diagonally. The headline is decorative artwork and both builds already overlap it deliberately — this is a question of degree on the flagship page, and the lever is `hero.tsx`'s own box. Deliberately not changed.
2. **Mascots are client-rendered now**, so the landing hero's `priority` preload is gone: an empty box on first paint, and crawlers don't see it. In exchange the thinking screen goes 368 KB → 55 KB. Worth a call if landing LCP is tracked.

### Caveat on the canary

The [24 before/after screenshots](https://claude.ai/code/artifact/a3740689-cd83-46a8-b8ca-53693ea4cc8b) were captured **before** the last two commits, so they show the earlier sizing. The measured table above is current. Say the word and I'll re-shoot them.

### Not covered

**cheering, pointing, pointing-down and walking** sit behind auth or seeded state, so no canary shot reaches them. They have component tests and appear in `/dev/components`.

innolope-dev:
@0xkkonrad saw this PR.

Not sure this is the right move. In Capacitor the animation runs in the WebView as JS. lottie-web rasterizes every frame on the main JS thread (the SVG renderer is worse — DOM churn per frame). For a full character at 40+ frames in the Android WebView that is plausibly worse than the GIF we're currently having (we had to convert webp to gif for mobile apps as webp were losing frames). 

If the assets are ready I can add lottie-ios / lottie-android plugins to Capacitor to see how it will look. But most likely it will be a huge headache. 

The WebView renders all web content into one native surface. A LottieAnimationView is a sibling view added to the view hierarchy, so it will sit unconditionally above the entire WebView, with no way to interleave. 
1. Z-order will be gone
2. We'll have to hand-mirror DOM geometry
3. Scrolling will be almost impossible

0xkkonrad:
## Canary, reshot against the current branch

Replaces the earlier set, which predated the sizing work. Every pair below is the same route at the same viewport — **WebP on the left, Lottie on the right**.

before/after pairs — landing hero, careers, maintenance

before/after pairs — maintenance, setup landing

before/after pairs — setup landing, setup signup

### What changed since the last set

The mascot is now **sized by height, per kind of screen**, instead of by a square box that fit whichever of the artwork's two dimensions was wider:

| group | 375 | 390 | 430 | spread |
|---|---|---|---|---|
| hero screens (setup steps, invite, waitlist) | 208–211 | 262–266 | 291–301 | **3–10px** |
| invalid invite | 173 | 178 | — | was 118; the raster was 89 |

So the invalid-invite mascot is now **double** the raster it replaced and in the same family as the screens either side of it, and the setup steps are back above their pre-migration height.

### The one composition call left for a human

landing hero at 430, 375 and 320 — before and after

The whistling sprite held its arms diagonally; `waving-chill` holds them **horizontally**, at about the height of the graffiti headline. At 430 that covers more of "LOCAL FEEL" than before. The headline is decorative artwork and both builds already overlap it deliberately, so this is a question of degree on the flagship page — the lever is `hero.tsx`'s own box, not the component. Deliberately unchanged.

### Coverage and method, honestly

- **19 pairs across 7 screens** and 6 widths: landing hero, careers, maintenance, setup landing, setup signup, setup finish, invalid invite.
- The 3 `/invite` shots from the previous set are **dropped, not missing** — that route redirects to the landing page, so they duplicated `landing-hero`.
- **before** frames come from a production build of `origin/dev`; **after** frames from a dev build of this branch, with the dev overlay hidden. This sandbox kept running out of memory during `next build`, and a matched prod/prod capture lost to that. Same DOM and CSS either way, but worth stating rather than implying otherwise.
- **cheering, pointing, pointing-down and walking** still are not reachable without auth or seeded state. They have component tests and appear in `/dev/components`.

Images live on `canary/mascot-lottie-shots` — screenshots only, never merged, nothing imports it. That branch is how they render inline here.

0xkkonrad:
Thanks @innolope-dev  — the WebView main-thread point is the right thing to be worried about, and it's the one thing nobody had measured. I went and measured it. But first, the framing, because I think we're weighing different things.

## What this PR is actually for

Not the file type, and not the megabytes. It's **programmability**. A GIF is a sealed loop — it can't react to a tap, hold a pose, aim at anything, or morph into another state. A rig can. That's the payoff, and nothing else on the table buys it.

Three objectives are getting bundled together here, and only one of them actually needs runtime Lottie:

1. **One vector source of truth** for the mascot set — could be Lottie-as-source with generated stills in the FE.
2. **Programmatic/interactive mascots** — click-to-react, pose held on a state, cursor/tap-aimed. **This is the only one that needs a live rig at runtime.**
3. **Killing 6 MB of assets** — agreed, achievable several ways.

1 and 3 have alternatives. 2 doesn't. So the question isn't "Lottie vs GIF", it's narrower: **can lottie-web hold up in our Android WebView on our floor device?** If yes, we get all three. If no, we still get 1 and 3 by generating stills from the same source, and only 2 is lost on the weakest devices.

Worth saying plainly: **this PR doesn't ship any interaction yet.** It's the enabler. The interaction work is a separate task (3/3 in the Lottie project, Sprint 155) and the techniques are already prototyped and working in `mono/inbox/lottie-v3/interactive.html`. Judged only on "smaller files", this PR is a weak trade. Judged on what it unlocks, it isn't.

## The measurements

Chromium at 6× CPU throttle (rough low-end Android), 390×844 @3x, real rigs, this PR's own clock. Total main thread — script plus style, layout and paint:

| variant | main thread busy | rAF |
|---|---|---|
| idle page | 1.5% | 60 fps |
| **GIF (ships today)** | **1.1%** | 60 fps |
| WebP | 1.1% | 60 fps |
| **Lottie SVG** | **12.1%** | **60 fps** |
| Lottie SVG, heaviest rig (`walking`) | 11.1% | 60 fps |
| Lottie SVG × 3 at once | 21.9% | 60 fps |
| **Lottie canvas** | **99.3%** | **6.8 fps** |

**You're right about the mechanism.** A GIF costs the main thread nothing and a rig costs ~11× that. Real change in kind, correctly called.

**But the SVG/canvas ordering is the other way round.** Canvas is the renderer that melts — 99.3% busy, frame rate collapsed to 6.8. SVG holds 60 fps at 12%. Same rig, same clock. This PR is already on `lottie_light` SVG, which turns out to be the safe one.

Two more corrections, offered as data not as argument:

- **SVG doesn't rasterize on the JS thread** — it writes attributes and the compositor rasterizes. That said, only ~30% of that 12.1% is script; the rest is exactly the style/layout/paint churn you're pointing at. My first pass timed only the JS call and got 2.9%; the honest figure is 12.1%.
- **Frame count isn't the cost driver.** Frames are keyframe data, not pre-rendered images. Cost is layers and paths (16–26 / 70–125 here), and we only repaint ~19.5×/s because `MASCOT_HOLD_FRAMES = 2` skips the write whenever the held frame hasn't changed — the 1930s stutter is also a third fewer paints than a naive 60 fps rig. Offscreen mascots don't tick at all (IntersectionObserver), one shared rAF drives the page, and `prefers-reduced-motion` holds a static frame.

On WebKit the same rigs cost 0.29–0.36 ms/render, no worse than Chromium — though that run was unthrottled, so it says the engine isn't pathological, not that an iPhone 8 is fine.

Also — the webp→gif fallback was yours (`eb88b073b`), and the commit message pins the cause to iOS <17 / A11 devices, not Android. Those phones were choking on *raster decode*, which is a cost a rig doesn't have at all. Different failure mode, so I don't think it transfers directly. Android has been on WebP throughout.

## Native plugins: agreed, dead end

Everything you list is correct — the WebView is one native surface, a `LottieAnimationView` is a sibling that sits unconditionally above it, and you'd be hand-mirroring DOM geometry and fighting scroll forever. Not worth it for a decorative mascot. You talked yourself out of it in the same comment and I think you were right to.

## What would settle it

**The measurement I can't make: a real device.** Throttled desktop Chromium is a proxy — it doesn't model WebView's raster path or a real thermal budget. You own that pipeline and `webContentsDebuggingEnabled` is already true for non-prod, so an Android debug build with the profiler open on the setup flow would settle this properly. If you can do that, I'll take the result either way.

**And so merging isn't a one-way door:** the component already holds frame 0 for `prefers-reduced-motion`. Adding `isNativeBridge()` plus the low-end gate you wrote in `instrumentation-client.ts` (`deviceMemory >= 4 && hardwareConcurrency >= 6`) to that same branch means native — or just weak devices — gets a crisp static mascot, still at a fraction of today's bytes. Then a bad profiling result is a one-line constant, not a revert. Happy to add that here before merge if you'd like it in from day one.

0xkkonrad:
Cross-note from #2760 (LP homepage rebuild → `main`), which touches `hero.tsx` too — no action needed now, just a string to carry when this PR resolves its mascot conflict.

#2760 lands four class edits on `hero.tsx` for TASK-21626 (the hero SIGN UP sits below the fold on every common laptop and on short phones). One of them is on the mascot `<Image>`:

```
md:max-h-[min(40vh,calc(100svh-28rem))]
```

This PR replaces that `<Image>` (`h-auto max-h-[40vh]`) with a fixed-height host div (`h-[40vh]`) and renames `PeanutMascot` → `HeroMascot`. So when the two meet — on the `dev` side, against #2690, since #2760 targets `main` — the cap has to move onto the host div, and it has to change form:

```
absolute left-1/2 z-10 h-[40vh] w-[90%] -translate-x-1/2 md:h-[min(40vh,calc(100svh-28rem))]
```

`md:h-`, **not** `md:max-h-`. `max-h` was correct against `h-auto`; against an explicit `h-[40vh]` it does nothing. Your positioning JS reads `hostHeight * MASCOT_ART_FILL` at runtime, so a capped height is picked up with no JS change.

Also worth knowing: only the mascot line actually conflicts. The other three edits (artwork `max-h-[calc(100svh-23rem)]`, h2 `md:mt-12`, CTA container `mt-8`) sit on lines this PR does not touch and three-way merge in silently — but the h2 line falls inside your trailing hunk context, which is where a hand-resolved conflict tends to lose `md:mt-12` without anyone noticing. Worth confirming all four survive after any resolution.

Without the cap, the 40vh mascot covers "LOCAL FEEL" at 1366×657.

jjramirezn:
I agree with @innolope-dev on the concern, and I think the PR has merit. But one fact changes the weighing: **in about a month native is the only client.** No PWA.

That removes the merge path we were discussing. A static frame on native was the safety net; with native as everything, it becomes the product. Every user would download the player to draw one frame, and the programmability payoff never runs anywhere. So "merge now, gate on device later" is off the table. The real-device profile is the only gate, and it competes with native work for the next month.

**My call: do not merge as-is, do not close it either. Reframe it.**

Keep the parts that are done and good:
- the ten rigs as the single vector source of truth
- the pose-typed `PeanutMascot` API and the call-site migration
- the art-box sizing work

Swap the renderer:
1. **Build-time export** from each rig to a sprite sheet. The stutter helps here: `MASCOT_HOLD_FRAMES = 2` means the sheet needs half the frames.
2. `PeanutMascot` draws from the sheet with CSS `steps()`. Compositor only, zero JS per frame, same cost class as the GIF we ship today.
3. Same `pose` prop. Hold a pose, play once, switch pose on tap all stay possible by class changes. Cursor-aimed and morph do not, so the Sprint 155 interaction task gets scoped to what a sheet can do.
4. If the sheets come out too heavy, fall back to an **animated WebP from the same export**. That still gets the source of truth and the byte savings.

The export script is the durable piece either way. If Slava later gets a real-device number that says the WebView copes with lottie-web SVG, the runtime player comes back as a follow-up and nothing here is wasted.

@0xkkonrad before committing to sheets over WebP: export one pose (waving-chill, at the hero's size) and post the file size. Ten poses at hero resolution could land heavier than the rasters they replace, and I want that number before you build the CSS side.

PR #2702, opened 2026-08-17
Merge mobile-release into dev — unify the release line

innolope-dev:
## What

Merges `mobile-release` (tip `250e87b17`) into `dev` (retargeted from main at slavik's request), so the native release line, `main`'s web work, and `dev`'s feature line converge. Apps build from the unified line after this lands; `mobile-release` is frozen — no deletions yet.

## Contents

- Full native layer as shipped in the v1.0.52/1.0.53 store builds — byte-identical to the binaries on devices.
- The one-tap session-key spend job (dark `SESSION_KEY_SPEND` flag) from the Mobile release regressions session.
- All of `main` (badges, Crisp proxy hardening, FX fixes) and the 1.0.53 version bump.
- Merge of current `dev` (deferred handoff, referral funnel, pt-BR, SEO sweep, unread-support badge, windowed connectivity).

## Conflict resolutions of note

- **native-routes**: dev's null dead-end guard wrapped around the richer recipient mapping (profile view, request pots, charge context); `/invite` keeps its params and writes the invite cookie (belt-and-suspenders).
- **sentry.utils**: the `reportNonOkResponse` refactor + native-http fallback rebased onto dev's windowed-connectivity model (`reportNetworkOk` is gone; two tests updated).
- **SupportDrawer**: dev's unread-badge clearing inside the hardened (`.catch`-guarded) native open; both new test suites kept.
- **Drawer**: dev's `scrollAreaClassName` prop with the `pb-safe-bottom` token (Android <15 safe-area fix preserved).
- **capgo-deploy-ios.yml**: stays deleted — its `--version-exists-ok` tweak already lives in the consolidated `capgo-deploy.yml`.
- **src/content submodule**: kept at main's newer pointer (`9101046`), no published content rolls back.

## Verification

- 256/256 jest suites pass (3,239 tests) on the merge result.
- `prettier --check` clean on all resolved files.
- Native paths byte-identical to `mobile-release` (= the shipped binaries).

PR #2737, opened 2026-08-18
chore(native): retire the transport-migration probes

innolope-dev:
## Why

`environment:native` produced **970 Sentry events in 24h**. The level split:

| level | events |
|---|---|
| info | 691 |
| error | 258 |
| warning | 21 |

**71% is instrumentation, not faults.** Nearly all of it was written to watch the CapacitorHttp → direct-fetch transport switch (PEANUT-UI-R44). That switch shipped; the probes didn't get removed.

## What goes

**`native-canary` — deleted.** Five `captureMessage` calls per app launch (`get-healthz`, `get-users-me`, `post-healthz`, `get-healthz-nocors`, `get-healthz-native`), ~110 launches/day → **~550 events/day**. It also fires five extra API requests on every cold start. Its own header comment states the purpose: watching CORS + credential behaviour "before and after the transport switch".

**`legacy-cookie native transport engaged` / `native http fallback engaged` — deleted.** Once-per-session census notes (~25/day) for the same migration: how many users still sit on legacy cookie-jar auth, how often the WebView path gets rejected.

> Worth flagging: `native http fallback engaged` was the one of these with residual alarm value — a spike would mean the WebView transport regressed. It's a warning-level, once-per-session note, so it's cheap. If you'd rather keep that signal, say so and I'll re-add it as a PostHog event instead of dropping it. The transport *behaviour* is untouched either way.

**`onesignal subscription snapshot` — moved to PostHog** (~86/day). The data is genuinely useful (permission state, token presence, whether `login()` linked the external_id), but it's a per-launch state snapshot, not an error — Sentry is the wrong store. Now `ANALYTICS_EVENTS.NOTIFICATION_SUBSCRIPTION_SNAPSHOT`. The `snapshot failed` variant **stays in Sentry**: failing to read subscription state is a real fault.

## What does not change

The transport logic itself. `preferNativeTransport` still tries the OS HTTP client first for tokenless native sessions, and the WebView-rejection fallback still rescues requests. Only the reporting is gone — both tests now assert those paths stay silent rather than asserting a notice count.

## Verification

- `sentry-fallback.test.ts`, `sentry-prefer-native.test.ts`, `native.adapter.test.ts` — 11 passed, 3 suites.
- `tsc --noEmit`: 232 errors, all pre-existing `Cannot find module '@/assets/….svg'` asset-declaration noise from a fresh worktree with no generated `next-env.d.ts`. **Zero** in any touched file.
- `prettier --check`: clean.

## Base

Branched off `merge/mobile-release-into-main-20260817`, because that's the only branch carrying this code — canary v3, the OneSignal snapshot, and both transport notes are absent from `dev` and `main`. **Until #2702 lands, this PR's diff will also show #2702's changes.** Review the two commits on top, or merge #2702 first and this reduces to the removal alone.

## Expected effect

~660 events/day removed from `environment:native`, taking info-level volume to near zero and leaving the error tier as the actual signal.

## Related

- #2735 fixes the `beforeSend` chained-exception bug and suppresses ~95/day of Capgo updater noise.
- #2708 fixes the `disable_auto_update_under_native` cause (105/day) — deliberately *not* filtered, since it means OTA is dead for that binary.

## Update — the conflict banner is #2702's, not this change's

GitHub shows this PR as conflicting with `dev`. The four conflicting files are:

```
instrumentation-client.ts
src/components/Global/SupportDrawer/__tests__/SupportDrawer.test.tsx
src/utils/__tests__/native-routes.test.ts
src/utils/native-routes.ts
```

The eight files this PR's commit touches are:

```
src/config/peanut.config.tsx
src/constants/analytics.consts.ts
src/services/onesignal/native.adapter.ts
src/utils/native-canary.ts                      (deleted)
src/utils/__tests__/native-canary.test.ts       (deleted)
src/utils/__tests__/sentry-fallback.test.ts
src/utils/__tests__/sentry-prefer-native.test.ts
src/utils/sentry.utils.ts
```

**Zero overlap.** `git merge-tree origin/dev origin/merge/mobile-release-into-main-20260817` conflicts on its own, so the banner is inherited wholesale from #2702's base — this branch is #2702's head plus one commit.

I deliberately did not resolve them here: doing so would mean making reconciliation decisions for `native-routes` and `instrumentation-client` on #2702's behalf, in files this change doesn't touch, and #2702 has to make those same decisions anyway. Resolving them in two places invites divergence.

**Unblocking order:** land #2702 first, then this rebases onto `dev` as the removal alone. If #2702 is going to sit, say so and I'll re-cut this against `dev` once the probe code is there.


innolope-dev:
@coderabbitai review

PR #2739, opened 2026-08-18
fix(passkey): stop reporting a cancelled passkey prompt as an error

innolope-dev:
## The intent that was being undone

`useZeroDev.handleLogin` already does the right thing. It classifies the raw WebAuthn failure, captures it with full context, and throws a curated `PasskeyError` carrying user-facing copy. For a plain user cancel it captures **nothing** on web:

```ts
// Cancel saved no state; everything else clears stale state and reports the error to Sentry.
if (code !== 'LOGIN_CANCELED') {
    …
    captureException(err, { tags: { error_type: 'login_error' } })
} else if (isCapacitor()) {
    // keep visibility without alerting
    captureException(err, { level: 'warning', tags: { error_type: 'login_canceled_native' } })
}
```

Four call sites then reported the wrapper anyway, which is why `PEANUT-UI-R20` — *"Login was cancelled, or no passkey was found on this device"* — exists at **error** level despite that deliberate silence.

## What changed

| Site | Was | Now |
|---|---|---|
| `InvitesPage` | `void handleLoginClick()`, no catch → **unhandled rejection** on every cancel | caught; curated message surfaced like every other login entry point |
| `GuestLoginModal` | `console.error(e)` → `captureConsoleIntegration` event | dropped; toast unchanged |
| `Landing` | `Sentry.captureException(wrapper)` | skipped for `PasskeyError` |
| `JoinWaitlist` | `Sentry.captureException(wrapper)` | skipped for `PasskeyError` |

The guard is a named helper in `webauthn.utils.ts` rather than four inline checks, and it's narrow — `Landing` and `JoinWaitlist` still report anything that **isn't** a `PasskeyError`, so an unexpected failure in the login path is not silenced. PostHog's `SIGNUP_LOGIN_ERROR` is untouched: a failed login is still a funnel event, just not a crash report.

## Note on `InvitesPage`

That one is a real bug independent of Sentry — a floating promise with no rejection handler. It previously showed the user nothing at all on failure; it now shows the curated message, consistent with `Landing`, `SetupPasskey` and `JoinWaitlist`.

## Verification

- `src/components/Invites`: **61 passed, 1 failed** — identical to the `dev` baseline. The one failure is `badge-campaign-context.test.ts › source-qualifies every published content UTM`, which reads `@/content/generated/…` and fails in any worktree without the `src/content` submodule initialised. Confirmed pre-existing by re-running with this change reverted.
- `src/components/Setup`, `GuestLoginModal`, `webauthn`: 8 passed, 0 failed.
- `tsc --noEmit`: 231 errors, **zero in any touched file** — 229 are missing `@/assets/*` declarations in a fresh worktree, 2 are the same `src/content` submodule.
- `prettier --check`: clean.

`InvitesPage.test.tsx` gained a `useToast` mock alongside the existing `useLogin` one, since the suite renders `InvitePageContent` outside a `ToastProvider`.

## Related

#2735 adds `PasskeyError` to the `alreadyReported` beforeSend list as a backstop, so a future call site can't reintroduce this. This PR removes the reports at source; that one stops them arriving if it happens again.

## Correction to the commit message

The commit says *"PEANUT-UI-QRW and PEANUT-UI-R20: 19 events yesterday"*. The accurate figure is **13/day** — QRW 9 + R20 4.

The extra 6 was `PEANUT-UI-SFV` (*"No matching passkey was found"*), which this PR does **not** fix. SFV is the *raw* Android error captured at the throw site in `useZeroDev`, not the `PasskeyError` wrapper, so none of the four call-site changes touch it. It's arguably also an expected outcome (no passkey enrolled on this device) and worth reclassifying to `LOGIN_CANCELED`, but that changes user-facing copy and belongs in its own change.

peanut-ui blocks force-push, so the commit message stands as written.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved login failure handling across invitations and setup flows.
  * Login errors now display appropriate localized messages without causing unhandled promise rejections.
  * Prevented duplicate reporting of passkey and WebAuthn errors while preserving user notifications and tracking.

* **Tests**
  * Updated invitation flow tests to safely handle toast notifications.

### DIFF
--- a/src/components/Global/GuestLoginModal/index.tsx
+++ b/src/components/Global/GuestLoginModal/index.tsx
@@ -25,8 +25,9 @@ const GuestLoginModal = () => {
                     onClick={() => {
                         handleLogin()
                             .then(closeModal)
-                            .catch((e) => {
-                                console.error(e)
+                            .catch(() => {
+                                // useZeroDev already reported the underlying failure;
+                                // console.error here would capture the wrapper again.
                                 toast.error(t('guestLoginModal.loginError'))
                             })
                     }}
--- a/src/components/Invites/InvitesPage.test.tsx
+++ b/src/components/Invites/InvitesPage.test.tsx
@@ -53,6 +53,10 @@ jest.mock('@/redux/slices/setup-slice', () => ({
     },
 }))
 jest.mock('@/hooks/useLogin', () => ({ useLogin: () => ({ handleLoginClick: mockLogin, isLoggingIn: false }) }))
+jest.mock('@/components/0_Bruddle/Toast', () => ({
+    ...jest.requireActual('@/components/0_Bruddle/Toast'),
+    useToast: () => ({ error: jest.fn(), success: jest.fn() }),
+}))
 jest.mock('@/hooks/useGuestStoreHandoff', () => ({
     useGuestStoreHandoff: (opts: { trackImpressionWhenGuest?: boolean }) => {
         mockUseGuestStoreHandoff(opts)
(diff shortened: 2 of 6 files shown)


innolope-dev:
@coderabbitai review

PR #2750, opened 2026-08-19
fix(native): paint the status-bar strip in the app background, not black

innolope-dev:
Rescues the one commit that was stranded on `mobile-release-fixes` with no PR. As `mobile-release` is retired into `dev`, it would otherwise have been dropped on the floor.

**Stacked on #2702** — based on that branch so the diff shows only the 2 real files. Once #2702 merges into `dev`, this should retarget to `dev` (GitHub does it automatically if the #2702 branch is deleted on merge; otherwise `gh pr edit 2750 --base dev`).

The top strip was forced black in `bd9a1b377` so it would stop flipping between black and beige above the pink beta feedback ribbon. That ribbon is now hidden on iOS (`e56254ef0`), and the underlying inconsistency was really a sizing/source problem, fixed in `9b3384e11` by reading Capacitor's natively measured insets.

Recolor only — the safe zone keeps `h-safe-top`, so the natively measured inset on Android 15+ and the `env()` fallback everywhere else are unchanged. `Style.Light` pairs dark status-bar icons with the light strip.

### DIFF
--- a/src/app/(mobile-ui)/layout.tsx
+++ b/src/app/(mobile-ui)/layout.tsx
@@ -180,12 +180,11 @@ const Layout = ({ children }: { children: React.ReactNode }) => {
 
     return (
         <div className="flex min-h-[100dvh] w-full bg-background pt-safe-top">
-            {/* Status-bar safe zone. On Android 15+ edge-to-edge the webview draws
-                under the status bar, where bg-background would otherwise show beige.
-                Fill the inset strip (above the feedback ribbon) with black so the top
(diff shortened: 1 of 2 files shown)


innolope-dev:
Closing in favour of collecting native fixes on a single branch.

This commit now lives on `mobile-release-fix` (branched off `dev`), which will accumulate the remaining native bug fixes and go up as one PR to `dev` later. Nothing is lost — the branch `fix/native-status-bar-strip-dev` stays on origin as a backup of the commit.

PR #1370, opened 2025-10-28
[TASK-15149] feat: custom devconnect claim flow

Zishan-7:
image

PR #2765, opened 2026-08-20
fix(kyc): keep the Sumsub SDK open through the EEA uplift questionnaire

abalinda:
## Summary

Sumsub's `bridge-requirements` workflow now branches every EEA applicant to the `bridge-eea-uplift` questionnaire as soon as the documents go GREEN. The frontend closed the SDK on the first submit for every non-LATAM flow, so the questionnaire never showed in-session. About 3 minutes later the backend poller read the follow-up level's `init` state, and the KYC drawer fell back to "Start verification". The user could not reach the questionnaire at all.

This PR keeps the SDK open through the questionnaire and fixes the drawer copy.

**Task:** TASK-21510

## What changed

**1. `isMultiLevel` now uses the intent the SDK was opened with.**

`isMultiLevel` is the flag that holds the SDK open past the first submit (`SumsubKycWrapper.handleSubmitted` returns early on it, and `evaluateSumsubStatusEvent` suppresses the `completed + GREEN` auto-close). It was `regionIntent === 'LATAM' && !isActionFlow`, read from the `regionIntent` **hook prop**.

Almost every entry point withholds that prop on purpose. The bank, claim and add-money flows pass the intent to `handleInitiateKyc` at call time, so that mounting a page does not create a backend record. The flag was therefore false for all of them — including every EU flow this fix targets. Changing only the intent list would have shipped a no-op.

The flag now lives in `useSumsubKycFlow`, where the effective intent is known. `handleInitiateKyc` raises it for **LATAM and EU**, unless the response is an applicant action; both close handlers clear it, so a closed SDK always reads false. `useMultiPhaseKycFlow` re-exports the value instead of deriving a second copy.

**NA is excluded, which deviates from the task card's "EU/NA" wording.** NA shares the `bridge-requirements` workflow, but only EEA applicants branch to the uplift questionnaire, so a US applicant has no second level. Marking NA multi-level would park every US applicant in an open SDK on Sumsub's "documents submitted" screen until approval, and would suppress the submit handler that fires `KYC_SUBMITTED`.

Known remainder: a non-EEA **EU**-intent user (UK on Faster Payments, say) also does not branch, and will wait on the SDK screen until approval. Accepted for now — see follow-up 3.

Side effect worth naming: first-time **LATAM** users on the bank / claim / add-money entry points had the same bug, for the same reason. They get the Manteca questionnaire in-session now too.

**2. ACTION_REQUIRED no longer tears down an open multi-level session.**

The companion `peanut-api-ts` PR maps the follow-up level's `init` state to `ACTION_REQUIRED`, so that status now lands about 3 minutes after the documents are submitted — while the user is still filling the questionnaire. The status-transition effect closed the flow on any non-success terminal state. It now holds the state when the status is `ACTION_REQUIRED`, the SDK is open, and the session is multi-level. The questionnaire IS the required action.

The transition is **deferred, not consumed**. The guard returns before `prevStatusRef` is advanced, so closing the SDK re-runs the effect and the same transition is evaluated for real. Advancing the ref first would swallow it, and a user who abandoned mid-questionnaire would then sit on a stale "verifying" modal with nothing left to close it.

Every other terminal state (`REJECTED`, `FAILED`) still closes, the suppression ends the moment the SDK closes, and the SDK's own rejection and retry handlers still close explicitly. Four boundary tests pin those edges.

**3. Card copy.** `kyc.actionMessageActionRequired` said "resubmit your documents", which is wrong for a questionnaire. It is now neutral: "We need a bit more to verify your identity. Tap below to continue." Translated in `es-419`, `es-AR` (voseo) and `pt-BR`.

**4. CTA copy.** `KycActionRequired`'s button said "Re-submit verification" under the new neutral card copy. The generic branch now reads "Continue" (`tCommon('continue')`, as its sibling `KycActionRequiredModal` already does). The reject-labels branch keeps "Re-submit verification" — that one genuinely is a re-submission.

## Risk

Medium, and scoped to the KYC entry flow.

- **Deploy order:** this PR is safe on its own. Change 2 only matters once the companion `peanut-api-ts` PR ships. Without it, the guard never fires.
- **Only EU and LATAM are held open.** NA, ROW and STANDARD behave exactly as before.
- **A non-EEA EU-intent user** does not branch to the questionnaire, so they wait on Sumsub's own "in review" screen until approval instead of our verification-progress modal. LATAM users already have this behavior, and approval still closes the SDK through the `APPROVED` transition. Follow-up 3 is the durable fix.
- **Native is unaffected.** `SumsubNativeSdk` never reads `isMultiLevel`. The Cordova SDK walks all levels of the workflow before it dismisses, so it calls `onComplete` only once the whole session ends.

- `isMultiLevel` is raised only by the initiate open and cleared by **all three** paths that close the SDK — both close handlers and the `APPROVED` branch — so "closed implies single-level" holds by construction and no later action open can inherit a stale true. The `showWrapper` conjunct in the guard is kept as defence in depth for any future close path.
- The `ACTION_REQUIRED` guard reads `showWrapper` and `isMultiLevel` as committed state, with both in the status effect's dependency array, rather than through a render-written ref (CodeRabbit, [#discussion_r3821474016](https://github.com/peanutprotocol/peanut-ui/pull/2765#discussion_r3821474016)). The pre-existing render-time write to `showWrapperRef` is untouched — it is read by an async callback with its own documented race, and rewiring it belongs in its own PR.
- Deferring the transition means that if the status is still `ACTION_REQUIRED` when the user finishes the questionnaire, the deferred close fires as the SDK closes and the drawer shows the action state until the backend catches up. That is the honest state, and it is the safer of the two failures — the alternative strands the user on a modal that nothing can close.

## Known follow-ups

1. **Drawer resume passes no intent.** `KycStatusDrawer` calls `handleInitiateKyc()` bare, so a resumed session is single-level. Benign today: a fresh `bridge-requirements` token resumes AT the questionnaire, so the close-on-submit happens after the questionnaire is done (applicant `6a8189f7` on the task card). No code change.
2. **`handleResubmitted` closes unconditionally** in `SumsubKycWrapper`, a deliberate LATAM-era decision. An EEA user who retries a RED document in-SDK is closed before the questionnaire and re-enters through the drawer. Changing it needs sandbox verification of which Sumsub event fires first. Follow-up, not this PR.
4. **A consumed ACTION_REQUIRED cannot re-fire.** The close branch keys on a status *transition*. If `ACTION_REQUIRED` is consumed just before the SDK opens, `prevStatusRef` already holds it, so when `handleSdkComplete` opens the verification-progress modal after the questionnaire, no transition arrives to close it. The user sees "verifying" until the poll observes a different status or they dismiss the modal. Flagged by CodeRabbit as a moderate merge risk on `85e80fd2e`; **explicitly accepted here**. It predates this PR — the close branch has always been transition-keyed — but this PR makes `ACTION_REQUIRED` a routine mid-flow status, so it raises the odds. The modal stays dismissible and the poll still resolves it on any later status change. Fixing it means changing when `handleSdkComplete` opens the modal, which touches every KYC flow, so it does not belong in this diff.

3. **The frontend should stop hand-rolling workflow knowledge.** The intent list here will drift (`STANDARD` maps to `general` and already disagrees). The durable fix is for the initiate response to carry `levelName` / a multi-level flag from the backend. Cross-repo follow-up.

## QA

- `npm test` — 247 suites, 3188 tests green. `npm run typecheck` and `prettier --check .` clean. `eslint` clean on every changed file.
- `src/hooks/__tests__/useSumsubKycFlow.test.ts`:
  - Intent passed as a call-time override → `isMultiLevel` true for EU and LATAM, false for NA, ROW and STANDARD. The old prop-only derivation got this wrong for every override caller.
  - Falls back to the `regionIntent` prop when no override is passed.
  - A `manteca` applicant action stays single-level even for LATAM.
  - `ACTION_REQUIRED` with the multi-level SDK open holds the flow; **closing the SDK then re-evaluates the deferred transition** and runs the normal close; the ROW twin closes immediately; the same status closes once the SDK is closed; `REJECTED` still closes with the SDK open.
- `src/components/Kyc/__tests__/sumsubStatusEvent.utils.test.ts`: multi-level + RED keeps the SDK open for an in-session retry.
- `src/components/Kyc/states/__tests__/KycStates.test.tsx`: the generic branch renders the new card copy and a "Continue" CTA; the reject-labels branch keeps "Re-submit verification".

**Screenshots:** ⚠️ NONE. The two visible changes are the drawer card string and its button label, both inside the `action_required` KYC state — reaching it needs a live applicant the backend has flipped to `ACTION_REQUIRED`, which the local sandbox cannot produce. Both strings are asserted in `KycStates.test.tsx` instead. The questionnaire itself is Sumsub-rendered.

## Review round — Jota (`acde2e184`)

Four of the five findings were valid and are fixed; the fifth is a comment correction plus a declined scope change.

1. **The sibling effect in `useMultiPhaseKycFlow` fired `KYC_REJECTED` + `fetchUser` on every happy-path ACTION_REQUIRED.** Same defer guard as the transition effect. On the happy path the status is APPROVED by close time, so nothing fires.
2. **The deferred transition replayed on EVERY close**, so an in-session resubmit closed the progress modal it had just opened. The defer is now split by close type: a submission close (`handleSdkComplete`) **consumes** the held transition (the user just acted — it is stale); the manual close keeps the replay (abandoning really does leave the action required, and the SDK can sit on top of an open progress modal). This supersedes the "deferred, not consumed" wording above.
3. **Multi-level sessions emitted no `KYC_SUBMITTED` at all.** New optional `onSubmitted` on `SumsubSdkProps`: the wrapper reports the Level-1 submit it previously swallowed, `useMultiPhaseKycFlow` captures the event with the usual `region_intent`. Single-level and native paths unchanged.
4. **`handleRestartIdentity` reopened the SDK with `isMultiLevel` false**, closing a restarted LATAM session before the questionnaire. It re-derives the flag from `regionIntentRef` (best-effort — undefined ref keeps today's behavior).
5. **The "NA has no second level" comments were wrong** — the backend registry documents organic `source-of-funds` / `proof-of-address` branches. Comments corrected. NA stays out of the multi-level set: the hold applies to every applicant of the intent, and the rare branch cohort converges through the ACTION_REQUIRED drawer round-trip.

## Summary by CodeRabbit

* **New Features**
  * Improved multi-step verification flows so follow-up questionnaires can continue without closing the verification experience.
  * Added submission handling that keeps eligible verification sessions open for in-session retries.
  * Updated action-required prompts with clearer “Continue” guidance in English, Spanish, and Portuguese.

* **Bug Fixes**
  * Prevented premature rejection tracking and modal closure during multi-step verification.
  * Preserved document resubmission guidance for cases that specifically require it.

### DIFF
--- a/src/components/Kyc/sumsubStatusEvent.utils.ts
+++ b/src/components/Kyc/sumsubStatusEvent.utils.ts
@@ -44,8 +44,9 @@ export function evaluateSumsubStatusEvent({
     if (isMultiLevel) {
         // Level 1 fires completed+GREEN before Level 2 is shown — don't close.
         // Level 2 is the Manteca questionnaire (LATAM) or the EEA uplift
-        // questionnaire (EU). NA shares the `bridge-requirements` workflow but
-        // never branches to the uplift, so it is not multi-level.
+        // questionnaire (EU). NA shares the `bridge-requirements` workflow but is
+        // deliberately not marked multi-level — its second levels are rare
+        // organic branches (see isMultiLevelIntent in useSumsubKycFlow).
         return { markSubmitted: false, autoClose: false }
     }
     if (isCompletedGreen) {
--- a/src/components/Kyc/sumsubStatusEvent.utils.ts
+++ b/src/components/Kyc/sumsubStatusEvent.utils.ts
@@ -44,7 +44,8 @@ export function evaluateSumsubStatusEvent({
     if (isMultiLevel) {
         // Level 1 fires completed+GREEN before Level 2 is shown — don't close.
         // Level 2 is the Manteca questionnaire (LATAM) or the EEA uplift
-        // questionnaire (EU / NA, both on `bridge-requirements`).
+        // questionnaire (EU). NA shares the `bridge-requirements` workflow but
+        // never branches to the uplift, so it is not multi-level.
         return { markSubmitted: false, autoClose: false }
     }
     if (isCompletedGreen) {
--- a/src/components/Kyc/sumsubStatusEvent.utils.ts
+++ b/src/components/Kyc/sumsubStatusEvent.utils.ts
@@ -43,6 +43,8 @@ export function evaluateSumsubStatusEvent({
     }
     if (isMultiLevel) {
         // Level 1 fires completed+GREEN before Level 2 is shown — don't close.
+        // Level 2 is the Manteca questionnaire (LATAM) or the EEA uplift
+        // questionnaire (EU / NA, both on `bridge-requirements`).
         return { markSubmitted: false, autoClose: false }
     }
     if (isCompletedGreen) {
--- a/src/hooks/useMultiPhaseKycFlow.ts
+++ b/src/hooks/useMultiPhaseKycFlow.ts
@@ -248,6 +248,15 @@ export const useMultiPhaseKycFlow = ({
     // refresh user store when kyc status transitions to a non-success state
     // so the drawer/status item reads the updated verification record
     useEffect(() => {
+        // Same defer guard as the transition effect in useSumsubKycFlow: while a
+        // multi-level SDK session is open, ACTION_REQUIRED is the routine "the
+        // follow-up questionnaire is showing" status (the backend maps the
+        // follow-up level's `init` state to it ~3 min after the documents are
+        // submitted). Acting on it here would log a bogus KYC_REJECTED for every
+        // succeeding EEA applicant and flip the user store to action_required
+        // under the open SDK. The effect re-runs when the SDK closes, so an
+        // abandoned session still gets the capture and the store refresh.
+        if (liveKycStatus === 'ACTION_REQUIRED' && showWrapper && isMultiLevel) return
         if (liveKycStatus === 'ACTION_REQUIRED' || liveKycStatus === 'REJECTED') {
             posthog.capture(ANALYTICS_EVENTS.KYC_REJECTED, {
                 region_intent: lastIntentRef.current ?? regionIntent,
@@ -255,7 +264,15 @@ export const useMultiPhaseKycFlow = ({
             })
             fetchUser()
         }
-    }, [liveKycStatus, fetchUser, regionIntent])
+    }, [liveKycStatus, fetchUser, regionIntent, showWrapper, isMultiLevel])
+
+    // A multi-level session never reaches handleSdkComplete on the happy path:
+    // the SDK stays open through the follow-up level and the APPROVED transition
+    // closes it without onComplete. The wrapper reports the Level-1 submit
+    // through onSubmitted instead, so the funnel still gets its KYC_SUBMITTED.
+    const handleSdkSubmitted = useCallback(() => {
+        posthog.capture(ANALYTICS_EVENTS.KYC_SUBMITTED, { region_intent: lastIntentRef.current ?? regionIntent })
+    }, [regionIntent])
 
     // wrap handleSdkComplete to track real-time flow
     const handleSdkComplete = useCallback(() => {
@@ -465,6 +482,7 @@ export const useMultiPhaseKycFlow = ({
         accessToken,
         handleSdkClose: handleClose,
         handleSdkComplete,
+        handleSdkSubmitted,
         refreshToken,
         isMultiLevel,
 
--- a/src/hooks/useMultiPhaseKycFlow.ts
+++ b/src/hooks/useMultiPhaseKycFlow.ts
@@ -237,6 +237,7 @@ export const useMultiPhaseKycFlow = ({
         isVerificationProgressModalOpen,
         closeVerificationProgressModal,
         isActionFlow,
+        isMultiLevel,
     } = useSumsubKycFlow({ onKycSuccess: handleSumsubApproved, onManualClose, regionIntent })
 
     // keep ref in sync
@@ -441,10 +442,6 @@ export const useMultiPhaseKycFlow = ({
     }, [clearPreparingTimer, stopTracking])
 
     const isModalOpen = isVerificationProgressModalOpen || forceShowModal
-
-    // multi-level only for first-time LATAM (workflow with conditional questionnaire).
-    // cross-region LATAM uses an applicant action (single level, not multi-level).
-    const isMultiLevel = regionIntent === 'LATAM' && !isActionFlow
 
     // Derive preparing stage from elapsed time for progressive copy
     const preparingStage = useMemo<'initial' | 'configuring' | 'slow'>(() => {
--- a/src/hooks/useSumsubKycFlow.ts
+++ b/src/hooks/useSumsubKycFlow.ts
@@ -54,6 +54,18 @@ const ACTION_ERROR_KEYS = {
     unexpected: 'unexpectedError',
 } as const satisfies Record<SumsubActionErrorCode, string>
 
+/**
+ * A workflow is multi-level when Sumsub can show a SECOND level in the same
+ * session, after the first submit:
+ *   - LATAM → `general`, which branches AR/BR applicants to `manteca-requirements`
+ *   - EU / NA → `bridge-requirements`, which branches EEA applicants to the
+ *     `bridge-eea-uplift` questionnaire
+ * ROW and STANDARD are single-level. Applicant actions are always single-level,
+ * whatever the region — see the `isActionFlow` checks at each SDK open.
+ */
+const isMultiLevelIntent = (intent: KYCRegionIntent | undefined): boolean =>
+    intent === 'LATAM' || intent === 'EU' || intent === 'NA'
+
 const getKycPollDelayMs = (elapsedMs: number): number => {
     for (const { untilMs, delayMs } of KYC_POLL_SCHEDULE) {
         if (elapsedMs < untilMs) return delayMs
@@ -83,9 +95,17 @@ export const useSumsubKycFlow = ({ onKycSuccess, onManualClose, regionIntent }: 
     const [rejectLabels, setRejectLabels] = useState<string[] | undefined>(undefined)
     // true when the SDK is showing an applicant action (not a standard level)
     const [isActionFlow, setIsActionFlow] = useState(false)
+    // true when the open SDK session runs a multi-level workflow, so the SDK must
+    // stay open past the first submit and show the follow-up level. Set by each
+    // path that opens the SDK, because the answer depends on the intent that open
+    // actually used — most entry points pass it to handleInitiateKyc rather than
+    // as the `regionIntent` prop, so the prop alone is not the effective intent.
+    const [isMultiLevel, setIsMultiLevel] = useState(false)
     const prevStatusRef = useRef(liveKycStatus)
     const showWrapperRef = useRef(showWrapper)
     showWrapperRef.current = showWrapper
+    const isMultiLevelRef = useRef(isMultiLevel)
+    isMultiLevelRef.current = isMultiLevel
     // tracks the effective region intent across initiate + refresh so the correct template is always used
     const regionIntentRef = useRef<KYCRegionIntent | undefined>(regionIntent)
     // tracks the level name across initiate + refresh (e.g. 'peanut-additional-docs')
@@ -137,6 +157,13 @@ export const useSumsubKycFlow = ({ onKycSuccess, onManualClose, regionIntent }: 
             liveKycStatus !== 'PENDING' &&
             liveKycStatus !== 'REVERIFYING'
         ) {
+            // In a multi-level workflow the follow-up questionnaire IS the required
+            // action, and the backend flips to ACTION_REQUIRED while the user is
+            // still filling it in the open SDK. Tearing the flow down under them is
+            // the bug this branch would cause, so hold the state while the SDK is
+            // open. The SDK's own rejection / retry handlers still close explicitly,
+            // and every other terminal state (REJECTED, FAILED) still closes here.
+            if (liveKycStatus === 'ACTION_REQUIRED' && showWrapperRef.current && isMultiLevelRef.current) return
             // close modal for any non-success terminal state (REJECTED, ACTION_REQUIRED, FAILED, etc.)
             setIsVerificationProgressModalOpen(false)
         }
@@ -332,6 +359,7 @@ export const useSumsubKycFlow = ({ onKycSuccess, onManualClose, regionIntent }: 
                     // not just this one — reaches the right SDK.
                     setAccessToken(response.data.token)
                     setIsActionFlow(!!response.data.actionType)
+                    setIsMultiLevel(!response.data.actionType && isMultiLevelIntent(effectiveIntent))
                     setShowWrapper(true)
                     return true
                 } else {
@@ -431,6 +459,8 @@ export const useSumsubKycFlow = ({ onKycSuccess, onManualClose, regionIntent }: 
             }
             if (response.data?.token) {
                 setAccessToken(response.data.token)
+                // a restart re-opens a single level (the IDENTITY step only)
+                setIsMultiLevel(false)
                 setShowWrapper(true)
             } else {
                 userInitiatedRef.current = false
@@ -468,6 +498,8 @@ export const useSumsubKycFlow = ({ onKycSuccess, onManualClose, regionIntent }: 
 
                 if (response.data?.token) {
                     setAccessToken(response.data.token)
+                    // self-heal mints an action token — always a single level
+                    setIsMultiLevel(false)
                     setShowWrapper(true)
                 } else {
                     userInitiatedRef.current = false
@@ -508,6 +540,8 @@ export const useSumsubKycFlow = ({ onKycSuccess, onManualClose, regionIntent }: 
                 levelNameRef.current = response.data.levelName
                 setAccessToken(response.data.token)
                 setIsActionFlow(true)
+                // an RFI action level — always a single level
+                setIsMultiLevel(false)
                 setShowWrapper(true)
             } catch (e: unknown) {
                 userInitiatedRef.current = false
@@ -539,5 +573,6 @@ export const useSumsubKycFlow = ({ onKycSuccess, onManualClose, regionIntent }: 
         closeVerificationModalAndGoHome,
         resetError,
         isActionFlow,
+        isMultiLevel,
     }
 }
--- a/src/hooks/useSumsubKycFlow.ts
+++ b/src/hooks/useSumsubKycFlow.ts
@@ -162,10 +162,15 @@ export const useSumsubKycFlow = ({ onKycSuccess, onManualClose, regionIntent }: 
         prevStatusRef.current = liveKycStatus
 
         if (prevStatus !== 'APPROVED' && liveKycStatus === 'APPROVED') {
-            // if SDK is still open (LATAM multi-level), close it now —
+            // if SDK is still open (multi-level), close it now —
             // applicantWorkflowCompleted has fired, all levels are done.
             if (showWrapperRef.current) {
                 setShowWrapper(false)
+                // this is the third and last path that closes the SDK, so clearing
+                // here is what makes "closed ⇒ single-level" hold everywhere. A
+                // later action open (self-heal, restart-identity, start-action)
+                // would otherwise inherit a stale true and never close on submit.
+                setIsMultiLevel(false)
                 setIsVerificationProgressModalOpen(true)
                 userInitiatedRef.current = true
             }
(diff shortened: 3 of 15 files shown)


src/hooks/useSumsubKycFlow.ts
```diff
@@ -83,9 +95,17 @@ export const useSumsubKycFlow = ({ onKycSuccess, onManualClose, regionIntent }:
...
+    // stay open past the first submit and show the follow-up level. Set by each
+    // path that opens the SDK, because the answer depends on the intent that open
+    // actually used — most entry points pass it to handleInitiateKyc rather than
+    // as the `regionIntent` prop, so the prop alone is not the effective intent.
+    const [isMultiLevel, setIsMultiLevel] = useState(false)
     const prevStatusRef = useRef(liveKycStatus)
     const showWrapperRef = useRef(showWrapper)
     showWrapperRef.current = showWrapper
+    const isMultiLevelRef = useRef(isMultiLevel)
+    isMultiLevelRef.current = isMultiLevel
```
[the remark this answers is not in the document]
abalinda: Fixed in a598466d7 — thank you, the finding is valid.

I dropped `isMultiLevelRef` instead of moving it into an effect. The guard now reads `showWrapper` and `isMultiLevel` as committed state, with both added to the effect's dependency array, so there is no ref to leak an uncommitted value. A re-run with an unchanged status is a no-op: `prevStatusRef` already holds that status, so neither branch fires.

The render-time write to `showWrapperRef` on line 105 is pre-existing and is read by an async callback elsewhere in the hook (`fetchCurrentStatus`), which has its own documented race. Left alone to keep this diff scoped; noted as a follow-up.

abalinda:
@coderabbitai full review

abalinda:
@coderabbitai full review

abalinda:
@coderabbitai full review

src/hooks/useSumsubKycFlow.ts
```diff
@@ -83,6 +104,13 @@ export const useSumsubKycFlow = ({ onKycSuccess, onManualClose, regionIntent }:
     const [rejectLabels, setRejectLabels] = useState<string[] | undefined>(undefined)
     // true when the SDK is showing an applicant action (not a standard level)
     const [isActionFlow, setIsActionFlow] = useState(false)
+    // true when the open SDK session runs a multi-level workflow, so the SDK must
+    // stay open past the first submit and show the follow-up level. Raised by the
+    // initiate open (the only path that can start one) and cleared by both close
+    // handlers, so "closed" always means false. It depends on the intent that open
+    // actually used — most entry points pass it to handleInitiateKyc rather than
+    // as the `regionIntent` prop, so the prop alone is not the effective intent.
+    const [isMultiLevel, setIsMultiLevel] = useState(false)
```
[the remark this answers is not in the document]
abalinda: Fixed in 85e80fd2e — valid and important, thank you.

I cleared `isMultiLevel` in the APPROVED branch alongside `setShowWrapper(false)`. That was the third and last path that closes the SDK; the other two (`handleSdkComplete`, `handleClose`) already cleared it. With all three covered, "SDK closed implies single-level" now holds by construction.

On the second half of the suggestion — explicitly setting `false` in every non-regular token-open path — I did not reinstate those. They were deliberately removed earlier in review to make the default correct by construction rather than by repetition, and with every close path clearing the flag no open can inherit a stale true. `handleInitiateKyc` still sets it explicitly because it is the only path that can raise it.

Residual, accepted: an action open fired while the SDK is still open would keep the current value. That needs the drawer UI to be reachable from behind the SDK modal, which it is not.

New test `clears isMultiLevel when approval closes the SDK` fails without the fix — I checked by reverting it.

src/components/Kyc/sumsubStatusEvent.utils.ts
```diff
@@ -43,6 +43,8 @@ export function evaluateSumsubStatusEvent({
     }
     if (isMultiLevel) {
         // Level 1 fires completed+GREEN before Level 2 is shown — don't close.
+        // Level 2 is the Manteca questionnaire (LATAM) or the EEA uplift
+        // questionnaire (EU / NA, both on `bridge-requirements`).
```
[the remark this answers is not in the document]
abalinda: Fixed in 85e80fd2e. Both comments were left over from an earlier revision that treated NA as multi-level.

`sumsubStatusEvent.utils.ts` now says Level 2 is the Manteca questionnaire (LATAM) or the EEA uplift questionnaire (EU), and states that NA shares the `bridge-requirements` workflow but never branches to the uplift. The test comment now describes an EU applicant only.

abalinda:
@coderabbitai full review

abalinda:
@coderabbitai full review

src/hooks/useMultiPhaseKycFlow.ts
```diff
@@ -237,6 +237,7 @@ export const useMultiPhaseKycFlow = ({
         isVerificationProgressModalOpen,
         closeVerificationProgressModal,
         isActionFlow,
+        isMultiLevel,
```
jjramirezn, line 240: **[code-review] Rejection analytics + store flip fire on every happy-path ACTION_REQUIRED**

The ACTION_REQUIRED hold was added only to `useSumsubKycFlow`'s transition effect. The sibling effect below (~line 251) still runs `posthog.capture(KYC_REJECTED, …)` and `fetchUser()` on every `ACTION_REQUIRED`, with no `showWrapper`/`isMultiLevel` guard.

With api-ts #1370, ACTION_REQUIRED becomes a routine mid-flow state (~3 min after doc submit). Result: every succeeding EEA applicant logs a `KYC_REJECTED` event mid-questionnaire (corrupts the KYC funnel), and `fetchUser()` flips the user store to `action_required` under the open SDK, so store-reading surfaces show "Action needed" while the user is filling the questionnaire.

Suggested fix: apply the same `ACTION_REQUIRED && showWrapper && isMultiLevel` defer guard here.

src/hooks/useSumsubKycFlow.ts
```diff
@@ -116,14 +144,33 @@ export const useSumsubKycFlow = ({ onKycSuccess, onManualClose, regionIntent }:
...
+        //
+        // This returns BEFORE prevStatusRef is advanced, so the transition is
+        // deferred, not consumed. When the SDK closes, `showWrapper` changes, this
+        // effect re-runs and the branches below evaluate the same transition for
+        // real. Advancing the ref first would swallow it: a user who abandoned
+        // mid-questionnaire would then sit on a stale "verifying" modal forever.
+        //
+        // Both flags are committed state rather than refs, so an interrupted render
+        // can never leak a value this guard acts on.
+        if (liveKycStatus === 'ACTION_REQUIRED' && showWrapper && isMultiLevel) return
```
jjramirezn, line 168: **[code-review] Deferred stale transition closes the "verifying" modal right after a resubmit**

This guard defers the ACTION_REQUIRED transition without advancing `prevStatusRef`, so the stale transition re-evaluates on any later close path — including `handleSdkComplete`, which opens the verification-progress modal in the same batch.

Scenario: poller flips the row to ACTION_REQUIRED (deferred here). The user's docs then get a RETRY decline in-session (multi-level RED keeps the SDK open, per the new test), the user fixes and resubmits → `onApplicantResubmitted` → `handleSdkComplete` sets `showWrapper=false`, `isMultiLevel=false`, and opens the progress modal. The effect re-runs, the deferred PENDING→ACTION_REQUIRED finally evaluates, and `setIsVerificationProgressModalOpen(false)` closes the just-opened modal — the user lands on an ACTION_REQUIRED drawer that is stale with respect to the documents they just resubmitted.

Suggested fix: consume the deferred status on SDK close (e.g. advance `prevStatusRef` when the wrapper closes) instead of replaying it.

src/hooks/useSumsubKycFlow.ts
```diff
@@ -359,13 +407,15 @@ export const useSumsubKycFlow = ({ onKycSuccess, onManualClose, regionIntent }:
         selfHealProviderRef.current = null
         setShowWrapper(false)
         setIsActionFlow(false)
+        setIsMultiLevel(false)
```
jjramirezn, line 425: **[code-review] Multi-level sessions never emit KYC_SUBMITTED**

`handleSdkComplete` is the only emitter of `ANALYTICS_EVENTS.KYC_SUBMITTED`, and no multi-level path reaches it: `handleSubmitted`/`handleActionSubmitted` early-return on `isMultiLevelRef`, `evaluateSumsubStatusEvent` returns `autoClose: false` for multi-level, and the APPROVED websocket close calls `setShowWrapper(false)` without `onComplete`. Only the retry path (`handleResubmitted`) can still fire it.

Pre-PR this blackout existed only for prop-driven LATAM sessions; this PR extends it to every EU session and to the LATAM entry points it newly makes multi-level. The PR body names this mechanism only as the reason to exclude NA — the EU/LATAM event loss is not an accepted trade-off anywhere.

Suggested fix: emit KYC_SUBMITTED from the multi-level submit handlers (where they early-return) or from the APPROVED close path.

src/hooks/useSumsubKycFlow.ts
```diff
@@ -332,6 +379,7 @@ export const useSumsubKycFlow = ({ onKycSuccess, onManualClose, regionIntent }:
                     // not just this one — reaches the right SDK.
                     setAccessToken(response.data.token)
                     setIsActionFlow(!!response.data.actionType)
+                    setIsMultiLevel(!response.data.actionType && isMultiLevelIntent(effectiveIntent))
```
jjramirezn, line 391: **[code-review] `handleRestartIdentity` reopens the SDK without setting `isMultiLevel`**

The flag is now set only here in `handleInitiateKyc` and cleared on every close. `handleRestartIdentity` (~line 484, also `handleSelfHealResubmit`) reopens the SDK with only `setAccessToken` + `setShowWrapper(true)`, so the restarted session runs with `isMultiLevel=false`.

On the one surface that passes `regionIntent` (UnlockedRegions, LATAM), the deleted prop-derived flag was `true` for the whole component lifetime, covering restarts. Now: an AR/BR user who uploaded a wrong-country document and restarts gets the same multi-level `general` workflow, but the wrapper closes at first submit (if Sumsub emits `onApplicantSubmitted`) — before the manteca-requirements questionnaire — parking them at "verifying" while the backend holds IN_REVIEW. No backend rewrite resurfaces Manteca follow-ups (the ACTION_REQUIRED rewrite in #1370 is Bridge-levels only). Moot only if the restart flow exclusively emits `onApplicantResubmitted`, which always closed.

Suggested fix: set `isMultiLevel` (from the same `isMultiLevelIntent(effectiveIntent)` derivation) in `handleRestartIdentity` and `handleSelfHealResubmit` too.

src/hooks/useSumsubKycFlow.ts
```diff
@@ -54,6 +54,27 @@ const ACTION_ERROR_KEYS = {
...
+/**
+ * A workflow is multi-level when Sumsub can show a SECOND level in the same
+ * session, after the first submit:
+ *   - LATAM → `general`, which branches AR/BR applicants to `manteca-requirements`
+ *   - EU → `bridge-requirements`, which branches EEA applicants to the
+ *     `bridge-eea-uplift` questionnaire
+ *
+ * NA is deliberately NOT here, even though it shares the `bridge-requirements`
+ * workflow: only EEA applicants branch to the uplift questionnaire, so a US
+ * applicant has no second level. Marking NA multi-level would hold every US
```
jjramirezn: **[code-review] "NA has no second level" contradicts the backend level registry**

This comment (and the intent list it justifies) excludes NA on the claim that only EEA applicants branch. The backend registry (`peanut-api-ts/src/kyc/level-registry.ts:100`) documents `source-of-funds` (age ≥ 60) and `proof-of-address` (high-risk POI country) as organic `bridge-requirements` branches — NA applicants can hit both, and #1370's `isBridgeFollowupLevel` maps them to ACTION_REQUIRED.

Cost: a 60-year-old US applicant's SDK closes on first submit, and they get the ACTION_REQUIRED drawer round-trip (plus the spurious analytics/push per lap) instead of the in-session second level EU users get. Degraded but convergent — still, the JSDoc here predicts drift and the drift already exists at merge time.

Suggested fix: correct the comment, and consider including NA in the multi-level intent set. (The registry's own claim at level-registry.ts:99 that bridge-eea-uplift is reached only via the campaign moveToLevel is also stale versus both PR bodies — worth updating in #1370.)

src/hooks/useMultiPhaseKycFlow.ts
abalinda: Valid — fixed in acde2e184. The effect now has the same defer guard (`ACTION_REQUIRED && showWrapper && isMultiLevel`), with both flags in the dependency array. On the happy path the status is APPROVED by the time the SDK closes, so the deferred pass captures nothing — no funnel pollution, no store flip under the open SDK. An abandoned session still gets the capture and the `fetchUser` on close.

One known wart: an in-session resubmit also re-runs the effect on close and fires one late KYC_REJECTED. That rejection did happen (the RED decline that forced the resubmit), so the event is factual — just late. Distinguishing submit-close from abandon-close here needs more plumbing than it is worth.

src/hooks/useSumsubKycFlow.ts
abalinda: Valid — good catch on the same-batch replay. Fixed in acde2e184, with one refinement over "consume on any close": only a submission close consumes.

- `handleSdkComplete` now advances `prevStatusRef` when the live status is ACTION_REQUIRED. The user just acted, so the held transition is stale — the progress modal it opens stays open, and the poll (armed by the open modal) resolves the real status.
- `handleClose` keeps the replay. The SDK can sit on top of an already-open progress modal (submit once, then re-initiate from a second entry point — the both-open state `openSdkOverProgressModal` pins in the tests). Consuming there would strand that modal on "verifying": a poll that returns the same ACTION_REQUIRED value never re-runs the effect.

New regression test: "a submission close consumes the deferred transition — the progress modal stays open". The guard comment now documents both halves.

src/hooks/useSumsubKycFlow.ts
abalinda: Valid — fixed in acde2e184. New optional `onSubmitted` on `SumsubSdkProps`, fired by the wrapper's multi-level early-return branches on the first submit of a session (deduped via `hasSubmittedRef`, so the idCheck-prefixed twin event cannot double-fire it). `SumsubKycModals` wires it to a new `handleSdkSubmitted` in `useMultiPhaseKycFlow`, which captures KYC_SUBMITTED with the same `region_intent` resolution as `handleSdkComplete`.

Single-level paths are untouched — `onComplete` still carries their capture, and the wrapper never fires both for one submit. Native is unaffected: the Cordova SDK dismisses only after the last level, so its one `onComplete` already emits the event. Two new wrapper tests cover the multi-level and single-level branches.

src/hooks/useSumsubKycFlow.ts
abalinda: Fixed for `handleRestartIdentity` in acde2e184: it re-derives the flag with `setIsMultiLevel(isMultiLevelIntent(regionIntentRef.current))` before reopening. Best-effort: the ref is undefined when this hook instance never initiated (drawer-style mounts), which keeps today's single-level behavior; when the surface initiated in-session — the normal path to the Manteca country-ineligibility gate — it restores the pre-PR prop-derived coverage, and UnlockedRegions (intent prop) is covered either way.

Not applied to `handleSelfHealResubmit`: self-heal opens a provider resubmission action, single-level by construction, and "closed implies isMultiLevel false" already holds via all three close paths — so false is the correct value there. Same reasoning for `handleStartAction` (RFI level).

src/hooks/useSumsubKycFlow.ts
abalinda: You are right about the registry — comment corrected in acde2e184 (here, in `sumsubStatusEvent.utils.ts`, and in the test): NA second levels exist (`source-of-funds` at age >= 60, `proof-of-address` for high-risk POI countries) but are rare organic branches. The stale KYC_SUBMITTED half of the justification is also gone now that multi-level sessions emit it (see the onSubmitted thread).

NA stays out of the multi-level set, deliberately: the hold applies to every applicant of the intent, so marking NA multi-level parks the entire US mainstream funnel in an open SDK until approval to serve those rare branches. The branch cohort converges through the ACTION_REQUIRED drawer round-trip instead (#1370 maps the follow-up level to ACTION_REQUIRED, and a fresh bridge-requirements token resumes AT the follow-up level). Degraded, accepted, and now documented as such. The durable fix stays the backend-reported multi-level flag (follow-up 3).

Agreed the registry's own claim at level-registry.ts:99 is stale versus both PR bodies — that fix belongs in #1370, flagged for it.

jjramirezn (approved):

PR #2766, opened 2026-08-20
fix(qr-scanner): keep paste link above the My QR drawer on all locales and screen heights

abalinda:
## Summary

On the "Scan to pay" screen, the **"Click to paste" link and the clipboard chips were covered by the "My QR" drawer**. Portuguese users hit it on tall phones (reported on a Galaxy S24 Ultra), and on small screens it happened in **every** locale — on a 667px-tall viewport the collapsed drawer swallowed the paste link, the payment-method badges, and the bottom of the scan square.

**Task:** [TASK-21651](https://app.notion.com/p/3c183811757981f7bde0f6dd4a816554) — "PT locale: QR copy-paste option is hidden behind the Meu QR drawer on scan-to-pay"
**Field report:** PostHog session replay `01a01c66-369a-7b23-965d-6772d90e6de5`
(Notion also carries a duplicate of the same report: "PR bug: drawer covers 'Paste QR code' when app is in Portuguese".)

## Root cause

Two independent coordinate systems that were never made to agree.

- The paste link hung off the **scan square**, which is `fixed` and positioned from the **top of the viewport**. It landed at y≈520–541px on every device and in every locale — measured, identical in all 18 locale × viewport combinations.
- The drawer's collapsed peek grows from the **bottom**. `snapPoints = [0.75, 1]` are *fractional*, and vaul resolves a fractional snap as `windowHeight − snap × windowHeight` applied to the drawer's **natural (content) height**. On a content-sized drawer that makes the visible peek `contentHeight − 0.25 × windowHeight`.

So the peek depended on **both** inputs:

- **Locale** — pt-BR's `myQrText` ("Deixe outras pessoas escanearem para te pagar") wraps to two lines where EN's ("Let others scan this to pay you") is one. Measured drawer content: **445px in EN, 469px in pt-BR**.
- **Viewport height** — a smaller window subtracts less, so the peek grows.

Measured peek ranged from **212px to 309px**. EN on flagship phones fit by accident; the snap value had been tuned by eye on one device in one locale. The i18n extraction commits (`95e046d2b`, `fff17a424`) added the longer PT strings that tipped tall phones over.

## The fix

**1. `QRBottomDrawer` — deterministic geometry.** Switch to px snap points (`['150px', '520px']`, still module-scope so vaul's snap-sync effect doesn't refire) and add `mt-0 h-[100dvh]` to the content (twMerge drops the wrapper's `mt-24`). vaul resolves a snap point as `window.innerHeight - snapPoint`, so the drawer has to be exactly `innerHeight` tall for a px snap to **be** the visible height — identical on every device and in every locale.

`dvh` rather than `h-full` is load-bearing, and is the main thing the review caught (see "Review fixes" below).

Both constants are measured, not guessed:
- `150` = 87px to the bottom of the collapsed title, + 34px for the iOS home indicator that sits inside the peek, + slack for a title that wraps to two lines in some future translation.
- `520` = 448px to the bottom of the share button in the tallest locale (pt-BR), + 34px iOS safe-area pad, + slack.

**2. `QRScanner` — anchor the paste actions to that known peek.** Extract `PasteActions` and render it from a new bottom-anchored sibling at `bottom: QR_DRAWER_PEEK_PX + QR_DRAWER_PASTE_GAP_PX`. Both numbers come from a new `src/constants/qr-drawer.consts.ts` so the scanner and the drawer share one source of truth and neither imports the other. Badges stay where they were, with `[@media(max-height:729px)]:hidden` so they don't collide with the paste group on short screens.

The drawer stays at `z-[60]` above the link, so expanding it still slides over the link — that part was always correct.

## Before / after

| | Before | After |
|---|---|---|
| **pt-BR, 375×667** |  |  |
| **pt-BR, 384×832 (S24 Ultra) + chip** |  |  |
| **en, 375×667** |  |  |

<sub>Images are pinned to commit `f1583150b`, which is on this branch; the files themselves were removed in the following commit so they never reach `main`.</sub>

> **About the "Use copied address" chip in the screenshots** — it is **not introduced by this PR**. It is existing production behaviour, shipped in `9c78dc2bd` ("feat(qr-scanner): prompt-free iOS clipboard chip; restore Android pre-read"): on iOS/web it appears whenever the clipboard holds any text (a prompt-free `hasStrings` check), and on Android native it shows the detected address when the pre-read finds an EVM address. This PR only **repositions** it — it now rides with the paste link in the bottom-anchored group. Behaviour, copy and icon are untouched. It is shown here because it is the worst case for the overlap: it makes the paste group taller, which is what tipped the S24 Ultra from +30px to **-17px**.

## Measurements

Real `getBoundingClientRect()` geometry from a headless Chromium against the running app, one row per locale × viewport. "Link gap" is the distance from the bottom of the paste group to the top edge of the drawer — **negative means the drawer covers it**. Chip-present is measured with the chip's exact markup, because `clipboardHasStrings()` is iOS-native-only and can never be true on web.

| Locale | Viewport | Peek before | Peek after | Link gap before (no chip / chip) | Link gap after |
|---|---|---|---|---|---|
| en | 360x640 | 285px | 150px | **-186** ⚠️ / **-233** ⚠️ | +16 / +16 |
| en | 375x667 | 278px | 150px | **-152** ⚠️ / **-199** ⚠️ | +16 / +16 |
| en | 360x800 | 245px | 150px | 14 / **-33** ⚠️ | +16 / +16 |
| en | 384x832 | 237px | 150px | 54 / **7** ⚠️ | +16 / +16 |
| en | 390x844 | 234px | 150px | 69 / 22 | +16 / +16 |
| en | 430x932 | 212px | 150px | 179 / 132 | +16 / +16 |
| pt-BR | 360x640 | 309px | 150px | **-210** ⚠️ / **-257** ⚠️ | +16 / +16 |
| pt-BR | 375x667 | 302px | 150px | **-176** ⚠️ / **-223** ⚠️ | +16 / +16 |
| pt-BR | 360x800 | 269px | 150px | **-10** ⚠️ / **-57** ⚠️ | +16 / +16 |
| pt-BR | 384x832 | 261px | 150px | 30 / **-17** ⚠️ | +16 / +16 |
| pt-BR | 390x844 | 258px | 150px | 45 / **-2** ⚠️ | +16 / +16 |
| pt-BR | 430x932 | 212px | 150px | 179 / 132 | +16 / +16 |
| es-419 | 360x640 | 309px | 150px | **-210** ⚠️ / **-257** ⚠️ | +16 / +16 |
| es-419 | 375x667 | 278px | 150px | **-152** ⚠️ / **-199** ⚠️ | +16 / +16 |
| es-419 | 360x800 | 269px | 150px | **-10** ⚠️ / **-57** ⚠️ | +16 / +16 |
| es-419 | 384x832 | 237px | 150px | 54 / **7** ⚠️ | +16 / +16 |
| es-419 | 390x844 | 234px | 150px | 69 / 22 | +16 / +16 |
| es-419 | 430x932 | 212px | 150px | 179 / 132 | +16 / +16 |

Note the S24 Ultra row: pt-BR at 384×832 was **-17px with the clipboard chip showing** — the exact reported repro. After the change the peek is **150px in all 36 measured combinations** (zero variance, not merely ±2) and the gap is **+16px everywhere**, with or without the chip. The chip cannot reduce the gap by construction: the group is bottom-anchored, so a taller group grows upward.

## QA

Browser-verified against the running app (headless Chromium, fake camera device, `app-locale` cookie per run):

- Peek identical across en / pt-BR / es-419 / es-AR at every viewport; handle + collapsed title always fully visible.
- Paste link and chip clear the drawer by 16px at every viewport in the matrix, in both locales, chip present or not.
- Link is genuinely tappable — Playwright's click actionability asserts the element actually receives the pointer event, which is a direct proof it isn't covered.
- Expanded drawer: QR, body text and share button all visible, share button fully within the viewport even at 640px tall (no reliance on the inner scroll).
- Drag up collapsed→expanded, drag down back to 150px, and drag-down-at-collapsed still inert (`dismissible={false}`) — all pass on 360×640, 375×667, 384×832 and 430×932.
- Badge cutoff verified at the boundary: hidden at ≤729px, visible at ≥730px with a 16px gap to the paste group.

Local gate: prettier ✅ · typecheck ✅ · jest 247 suites / 3175 tests ✅ · build ✅ · eslint clean on the changed files ✅.

## Review fixes (second commit)

`/code-review` found two real problems in the first pass. Both are fixed, and the second is why the expanded snap is not simply "full height".

**1. `h-full` was wrong — it would have reproduced the original bug on the reported device.** A percentage height on a `position: fixed` element resolves against the *initial containing block*, which on a mobile browser with a retractable toolbar is the **large** viewport. vaul resolves snap points against `window.innerHeight` — the **small** viewport while the toolbar is showing. The two disagree by the toolbar's height (~56px on Chrome Android), so the collapsed peek would have come out at ~206px instead of 150px and covered the paste link again — on Chrome Android, which is exactly where the Galaxy S24 Ultra report came from. A headless browser has no toolbar, so **the 36-combination matrix structurally could not catch this**. Now `h-[100dvh]`, which tracks `innerHeight`. (`dvh` is already used elsewhere in this repo.)

**2. A fixed expanded height can clip the share button with nothing to scroll.** The drawer no longer grows with its content, so once a longer translation or the reader's font-size setting pushes the content past the 520px window, the overflow is simply cut off — the shared `max-h-[80vh]` cap is never reached inside a 520px window, so nothing scrolls. Reproduced by scaling the root font size: at 22px the share button sat at y=703 on a 667px screen, unreachable.

Fixed by capping the drawer's scroll area to the expanded window itself, less the drag-handle block above it. That block (`p-5` top + `my-4` + the handle) is rem-based, so the cap is written as `calc(520px - 3.625rem)` and tracks the font size:

| root font | scroll max-height | share button | reachable |
|---|---|---|---|
| 16px (default) | 462px | y=595 | ✅ (no scroll needed) |
| 20px | 447.5px | y=667 | ✅ |
| 22px | 440.25px | y=668 after scroll | ✅ (now scrolls) |
| 26px | 425.75px | y=668 after scroll | ✅ (now scrolls) |

The scroll region now lands on the bottom of the viewport at every font size. Making the expanded snap full-height (`1`) would also have fixed it, but it left ~40% of the sheet as empty background on a 932px phone — a worse regression than the one being fixed.

**Also applied from the review:** the shared geometry moved into `src/constants/qr-drawer.consts.ts`, so `QRScanner` no longer imports the `QRBottomDrawer` component (removing a component→component module edge and matching the repo's constants convention), and `min-h-[200px]` was dropped now that `h-[100dvh]` makes it dead.

**Declined, with reason:** the review also flagged that `[@media(max-height:729px)]:hidden` keys off viewport height alone, so the badges also hide in landscape and in short desktop/PWA windows. That is accepted — in those cases there genuinely is no room for the badge row either, so hiding it is the behaviour we want, and a height threshold is the cheapest honest proxy.

- **`mt-10` on the paste button is kept verbatim.** In the old in-square column it spaced the link from the badge row; in the bottom-anchored container it is inert margin (the container is bottom-anchored, so the link's position comes from the `bottom` offset, not from this margin). It is retained so the `dev` back-merge stays mechanical. It is transparent and sits over a non-interactive area, so it changes nothing visually or functionally.
- **Landscape.** The app asks for a portrait lock (`ScreenOrientationLocker`), which holds on Android and installed PWAs but not on iOS Safari or desktop. In a landscape window shorter than the 520px expanded snap the drawer's handle sits above the top edge when expanded. `main` had the same failure mode (its content-height drawer overshot a short landscape window too), just ~50px milder. Not introduced here and not worth a dynamic snap for a portrait-locked screen, but worth knowing.
- **The regression test pins the coupling, not the pixels.** jsdom has no layout engine, so the geometry itself is verified in a browser. The test asserts that the anchor offset is *derived from* `QR_DRAWER_PEEK_PX` — which is the failure mode that would actually recur (someone retunes the peek and forgets the scanner).
- **Hard-fling re-test (vaul 1.1.2).** The known out-of-bounds snap index still exists. It is reachable **only** with a physically impossible synthetic gesture — a 500px pointer teleport in a single frame (Playwright `steps: 3`, 0ms) — and only at viewport heights 660–700px, where the drawer rests at ~317px instead of 520px. Every realistically-timed swipe, down to an 80ms flick, snaps correctly; the state is fully recoverable with one normal drag and logs no errors. Not introduced by this PR in any user-reachable sense, and left alone rather than worked around in our code.

## Pre-existing issues noticed (NOT fixed here — out of scope)

- `QRBottomDrawer` renders `<h2>` inside `DrawerTitle`, which is itself an `<h2>`, so React logs a `<h2> cannot be a child of <h2>` hydration warning. Present on `main`, unrelated to this change.
- `QRScanner` seeds `isScanning` from `isOpen`, so force-mounting it open would call `createPortal(document.body)` during SSR. The app never does this (the overlay mounts it only after a client-side state flip), so it is latent, not live.

## Risk

Low, and contained to the QR scanner screen. Pure web-layer CSS/DOM — **no platform code**. Two files plus one test; no i18n strings changed, no changes to the shared `Drawer` wrapper (all classes are passed through `QRBottomDrawer`'s own `className`), no `NEXT_PUBLIC_API_VERSION` bump.

The one behaviour change a reviewer should look at deliberately: **the collapsed drawer is now shorter** (150px instead of 212–309px), which is what frees the space for the paste link. Less of the QR code peeks out while collapsed.

**Native:** ships to the PWA on deploy and to iOS/Android through the normal app release or OTA. No native changes needed.

## ⚠️ Back-merge note (main → dev)

This is a hotfix onto `main`, and `dev` has already diverged in `QRScanner/index.tsx`. `dev` **already has a `PasteActions` component** extracted from the same code, plus behavioural changes this PR deliberately does **not** import (it uses the `paste` icon and a `useCopiedCode` key, and hands raw clipboard text to `scanValue` instead of extracting an EVM address; it also renders `PasteActions` inside `ErrorView` and passes `onPaste` to `CameraPermissionModal`).

The back-merge therefore needs a small manual resolution:

1. **Keep `dev`'s `PasteActions`** — its body, its icon, its i18n key, its call sites. Do not take this PR's copy.
2. **Take this PR's bottom-anchored container** and render `dev`'s existing `PasteActions` from it, instead of from inside the scan-square column.
3. **Take `QRBottomDrawer` verbatim** from this PR — `dev` has no competing changes there.
4. Keep the `[@media(max-height:729px)]:hidden` on the badge row.


abalinda:
@coderabbitai review

abalinda:
@coderabbitai review

innolope-dev (approved):
Reviewed the diff plus the load-bearing claims against the code (vaul px-snap semantics, twMerge overrides, dvh vs h-full). Approach and measurements check out.

PR #2767, opened 2026-08-20
chore(marketing): consolidate the stray pages, and give /careers a real one

0xkkonrad:
Cleans up the stray marketing pages on peanut.me, found in a routing + copy audit of `origin/dev`. Six commits, each independently reviewable.

Two items from that audit are deliberately **not** here, per Konrad: the public `/dev/*` pages (filed as a Notion task for Jota instead) and the `/pricing` fee wording.

## What changes

**`/careers` becomes a real page.** It was a 30-word interstitial — a mascot, one sentence, a link out to Notion. No `<h1>`, no footer, no Crisp, and metadata still selling "Instant Global P2P Payments". The footer's Jobs link bypassed it entirely for a *second* URL pointing at the same Notion page. It now has its own content (what Peanut is, how the work works, open roles), sourced from `content/press/en.md` so nothing is invented, and Notion is reached only by **Apply on Notion**. The footer now links `/careers`.

**`/shhhhh` drops the card reward claims.** The Rain resubmission note records the "$10 after $100 card spend" activation reward and the "$10 / $10 referral" as *removed everywhere they appeared*. Both were still live in all four locales — and the same claim was live in the app home carousel. Also moves the "where can I get the card?" answer from issuance framing to acceptance framing, per the rule that public card copy describes acceptance only. Peanut's general referral share is untouched; the ruling keeps it, because it isn't tied to the card.

**Four routes retired**, all permanent redirects:

| Route | → | Why |
|---|---|---|
| `/lp/card` | `/shhhhh` | Dead behind `disableCardPioneers`: 200 with full metadata, zero body copy, sitemap priority 0.9 |
| `/exchange` | `/` | Renders the same `NoFees` component already on the homepage; ships an empty client shell |
| `/:locale/blog` | `/:locale/content` | Byte-identical to `/content` with `fixedType="blog"`. Index only — all 21 posts survive |
| `/:locale/team` | `/:locale/press` | `/press` already renders the same two members from the same source, with real photos |

**The content hub is crawlable again.** `/en/content` and `/en/blog` emitted **zero** article links in server HTML while the sitemap listed 21 posts — `ContentLanding` is a client component read through nuqs, so Next prerenders the Suspense fallback, and the fallback was a pulsing skeleton. The list rendering is now an exported `ContentLinkList` that touches no URL state, and the fallback renders it.

**Plus:** `/en/stories/index` (a live duplicate of its own hub, linked from that hub) is gone; `/press` section headings are localized; three client routes stop declaring the homepage as their canonical; the root title and description drop the protocol-era framing; `@PeanutProtocol` → `@joinpeanut` in all six places; and the five card legal pages join the sitemap.

**Ragdoll.** It had no `prefers-reduced-motion` handling at all — it now follows `confetti.ts` and degrades to a single settled frame rather than vanishing. Two new homes, both non-scrolling panes with a definite height: `/maintenance` (whose copy was the worst voice on the site) and the waitlist jail pane, gated on `step === 'jail'` so plain `/invite` doesn't get a draggable toy beside its CTA.

## One real bug fixed on the way

The `(mobile-ui)` layout registers pull-to-refresh on `document` before the waitlist early-return, and its `#scrollable-content` guard is absent on that screen, so the callback fell through to `true`. Dragging the peanut down ~160px would have fired `window.location.reload()` and wiped the user's typed invite code. The ragdoll host now swallows `touchstart` — a native listener, because React hydrates on `document` and a synthetic `stopPropagation` wouldn't stop a same-target sibling.

## Verification

- `npx jest` — **261 suites, 3344 tests, 0 failures**
- `npx tsc --noEmit` — **0 errors**
- `npx eslint` on all 37 changed TS/TSX files — 0 errors; `prettier --check` clean
- `verify-content.ts` — Pass 11 green, all 1165 expected URLs have routes
- `generateSitemap()` executed: 742 URLs, 21 blog posts intact, 0 blog-index, 5 card-legal, `/careers` still present
- Redirects checked statically: no duplicate sources, every source route removed, every destination resolves. The two pre-existing `/:path*` catch-alls are host-scoped to `docs.peanut.*` and can't shadow these
- `ContentLinkList` proven to server-render real anchors (`/en/blog/rewards`, `/en/stories/lynn`) with the `| Peanut` suffix stripped

**Not verified locally:** `next build` could not complete — earlyoom SIGTERMs the build worker at ~2.7 GB on this box, with zero compile or type errors in the log before the kill. CI is the real gate for the build. Runtime redirect behaviour also can't be exercised locally: [next.config.js:363](https://github.com/peanutprotocol/peanut-ui/blob/dev/next.config.js#L363) deliberately applies `redirects.json` only outside development.

## Coordination

- **Overlaps open PR #2690**, which also retires `/lp/card`. Konrad asked for it here too. This branch copies #2690's exact destination, sitemap comment and script edits, so whichever lands second should conflict trivially on `redirects.json`, `sitemap.ts` and `underMaintenance.config.ts`.
- `CardPioneers.tsx` is left in place (now pushing to `/shhhhh` rather than into a 308) because #2690 deletes it. Note for that PR: deleting it also orphans `PioneerCard3D.tsx` and six `landingCard*` keys.

## Follow-ups, deliberately not in this PR

- `SEOFooter` still links Jobs → Notion from `footer-manifest.json` in the peanut-content submodule. That's a content change, and content and code don't travel in the same PR.
- The `/shhhhh` "20 a week" strings predate the ruling recorded in `product/card.md` that the pace is now ~50/week — 7 catalog strings, 1 metadata line, and 3 hardcoded literals in `ScarcityCounter`.
- `/en/help` looks to have the same client-component crawlability gap: its prod HTML carries only 3 `/en/help/` links.
- `/maintenance` and the 404 are still hardcoded English.


0xkkonrad:
Superseded by #2782, which is rebased onto current `dev`.

Two parts of this PR were overtaken while it sat open:

- **`/lp/card`** was retired by #2760 (the homepage rebuild that succeeded #2690/#2694) — page deleted, 308 to `/shhhhh`, sitemap updated. My duplicate is dropped.
- **The `$10` card reward claims** were removed on `dev` separately, by deleting `shhhhh.faq.q4` outright rather than replacing it. That version is kept. Only `q6` — the availability/issuance framing — was still unfixed, and it carries over.

Everything else still applies and is in #2782: the real `/careers` page, the `/exchange`, blog-index and `/team` retirements, the content-hub crawlability fix, the `stories/index` duplicate, the canonical/identity work, and the ragdoll reduced-motion guard plus its two placements. Plus the footer label, which now reads Careers.

One correction from this PR's description: the claim that `SEOFooter` still links Jobs to Notion was wrong — `RESOURCES_MOVED_ELSEWHERE` already suppresses that manifest entry (#2755), so no content-submodule change is needed.

PR #2768, opened 2026-08-20
fix(qr-scanner): anchor paste actions to a deterministic drawer peek (dev port of #2766)

innolope-dev:
## Summary

Port of #2766's geometry fix onto `dev`, composed with `dev`'s newer paste behaviour. This is the planned manual back-merge done as its own PR instead of a `main → dev` merge commit, because the resolution isn't mechanical: `dev` already reworked `PasteActions` (raw-clipboard pass-through, `useCopiedCode` chip, error/permission-denied entry points) and none of that should be overwritten by the hotfix's behaviour-frozen copy.

**Problem being ported:** the My QR drawer's fractional snap points made the collapsed peek a function of both locale text length and viewport height (212–309px measured on `main`), while the paste link hung off the top-anchored scan square — so the drawer covered the link on short screens in every locale, and on tall phones in locales whose drawer text wraps (pt-BR on a Galaxy S24 Ultra, TASK-21651). `dev` has the identical geometry today; full root-cause analysis and the 36-combination locale × viewport measurement matrix are in #2766.

## Composition

**Taken from #2766, verbatim:**

- `QRBottomDrawer`: px snap points (`150px` / `520px`) on a viewport-height drawer — the peek becomes the same number on every device in every locale. Includes the scroll-area cap (`calc(520px - 3.625rem)`, rem-based so it tracks font-size settings) that keeps the share button reachable when content outgrows the fixed window.
- `src/constants/qr-drawer.consts.ts`: shared geometry constants, so the scanner and drawer agree without importing each other.
- `QRScanner`: the bottom-anchored container at `peek + 16px`, and the payment-method badge row hidden below 730px viewports (measured collision threshold).
- The regression test pinning that the anchor offset is *derived from* `QR_DRAWER_PEEK_PX` — the failure mode that would actually recur.

**Kept from `dev`, untouched:** the `PasteActions` component body and all call sites — the `useCopiedCode` chip and `paste` icon, raw clipboard text handed to `scanValue` (so Pix copia-e-cola works, not just EVM addresses), and the `ErrorView` / `CameraPermissionModal` paste entry points. Only the happy-path render location moved (into the anchored container).

**Two deliberate additions over #2766** (dev-only for now; worth folding back into `main` if it gets another hotfix):

1. **`h-screen` fallback under the `dvh` sizing** — `supports-[height:100dvh]:h-[100dvh]` with `h-screen` as base. `dvh` needs iOS 15.4+ / WebView Chromium 108+, and the app's deployment target is iOS 15.0; without a valid height the drawer would translate entirely off-screen on a 15.0–15.3 WebView (worse than the bug being fixed). Inside a WebView there is no retractable toolbar, so `100vh == 100dvh` exactly where the fallback applies. Verified against the compiled project CSS that the `@supports` rule lands after `.h-screen` in the cascade, so `dvh` still wins everywhere it exists.
2. **`pointer-events-none` on the anchor strip** (re-enabled on the actions wrapper) — the strip is full-width and transparent; without this it would swallow taps landing beside the buttons.

## Not changed / follow-ups

- `ErrorView`'s `PasteActions` instance stays in-flow near the top of the screen (dev's placement). It renders far from the drawer, but its geometry has never been measured the way the happy path now is — worth a look if that state ever gets layout complaints.
- Same accepted trade-offs as #2766: shorter collapsed peek (less QR visible), badge row also hides in short landscape/desktop windows, and the vaul 1.1.2 hard-fling note.

## QA

- `tsc --noEmit` clean; `jest` QRScanner suites 26/26 (includes the new anchoring pin); `eslint` and `prettier` clean on the touched files.
- Real-browser checks against the compiled project CSS: `@supports` cascade order for the dvh fallback; hit-testing on the anchor strip (paste link and chip receive the pointer, the empty band beside them clicks through); strip anchored exactly 166px above the viewport bottom, full width.
- The geometry itself (peek = 150px in all 36 locale × viewport combinations, +16px link gap) was measured in #2766 against the running app; the constants and layout are ported unchanged, so those measurements carry over. A device pass on iOS (notch) and Android 15 (gesture nav) before the next native/OTA release is still the recommended final check for both PRs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved QR scanner drawer positioning across screen sizes and mobile WebViews.
  * Kept “Click to paste” controls consistently visible above the collapsed drawer.
  * Prevented content from overflowing when the drawer is expanded.
  * Improved handling of short screens, safe-area spacing, and longer localized text.
* **Tests**
  * Added coverage to verify QR paste controls remain correctly positioned relative to the drawer.

### DIFF
--- a/src/components/Global/QRBottomDrawer/index.tsx
+++ b/src/components/Global/QRBottomDrawer/index.tsx
@@ -4,6 +4,7 @@ import ShareButton from '@/components/Global/ShareButton'
 import { useTranslations } from 'next-intl'
 import { useState } from 'react'
 import { Drawer, DrawerContent, DrawerTitle } from '../Drawer'
+import { QR_DRAWER_EXPANDED_PX, QR_DRAWER_PEEK_PX } from '@/constants/qr-drawer.consts'
 
 interface QRBottomDrawerProps {
     url: string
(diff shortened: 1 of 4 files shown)


innolope-dev:
@coderabbitai review

PR #2771, opened 2026-08-20
feat(rewards): restore referral surfaces on iOS, reframe as cashback

innolope-dev:
Restores the referral surfaces that were hidden on iOS in `4564b401` and reframes the programme as **cashback** throughout, in all four locales.

The incentive stays visible and legible — this is a deliberate product decision, not a compliance retreat. What changes is the vocabulary, what the money is attributed to, and three items that carried App Store exposure without buying anything.

## Why the gate comes off

`isReferralRewardsHidden()` hid `/rewards`, `/rewards/invites`, the home pill, the profile row, both invite carousel CTAs, the surprise-claim treatment and the receipt points row. Hiding the programme while the backend kept accruing — and while `referral.reward.earned` pushes kept advertising it — is worse than showing it: guideline **2.3.1** treats hidden-but-discoverable features as grounds for removal, and escalates to account level for repeat behaviour.

The programme will instead be **disclosed in Notes for Review**. That is the load-bearing mitigation: disclosed, the worst case is a rejection cycle; undisclosed and discoverable, it is not.

Note the old citation was wrong. `appStoreCompliance.ts` and the original commit both cited **3.1.5 (ii)**, which is Mining. The relevant clause is **3.1.5 (v)** — "Cryptocurrency apps may not offer currency for completing tasks, such as downloading other apps, encouraging other users to download…".

## What changed

**Gate removed** — `src/config/appStoreCompliance.ts` deleted along with its seven call sites. `src/config/__tests__/ios-store-gating.test.ts` keeps its cross-chain-withdraw coverage and drops the referral block.

**The cross-chain withdraw gate stays.** It rides in `underMaintenance.config.ts`, rests on a different clause — **3.1.5 (iii)**, Exchanges, which requires per-region licensing and is enforced far more consistently — and multi-chain swaps are the strongest "this is a crypto app" signal we ship. Untouched by this PR.

**Vocabulary: rewards → cashback.** 45 strings per full locale (en, es-419, pt-BR) plus 27 es-AR voseo overrides. `Recompensas` → `Cashback` as a loanword, which is the standard term in both LatAm markets.

**Attribution: signups → transactions.** `used Peanut` → `paid with Peanut`, `The more they use` → `The more they pay`. Same person, same money, same notification — but the cashback is tied to a transaction rather than to a download, and downloading is what clause (v) names.

**New "How cashback works" card** on `/rewards`. The earn instruction used to be a half-sentence appended to the lifetime total (`"Lifetime rewards: $X. To earn more, invite friends."`). It is now a four-step explainer — invite → they pay → cashback builds → make a payment to claim — and the lifetime line is just the number. Clearer, and it states the process without pinning a specific dollar to a specific person.

**Two multi-level admissions removed.** `qrPay.claim.inviteQrDescription` said "Points keep coming from the activity of friends **& their friends**" — the only place the UI stated the transitive structure, on a shareable QR-sticker screen. `claimSuccess.stickerDescription` lost "contribute towards your points forever". Neither bought anything.

**`profile.menu.points` → `profile.menu.cashback`**, labelled "Cashback" in all locales, matching the home pill.

## The push deep link fixes itself

`referral.reward.earned` deep-links to `/rewards` — a route the current iOS build guards and redirects away from. Today an iOS user is told they earned money, taps, and lands on `/home`. Removing the route guard fixes that. Copy changes are in the companion peanut-api-ts PR.

## Kept deliberately

`inviteFriendsModal.descriptionNoRewards` ("Share your link so friends can join you on Peanut.") is now unreferenced but stays in all three catalogs. It is the reward-free fallback copy — if we are rejected under (v), that string plus the carousel equivalents make the response a resubmit rather than a redesign.

## Verification

- `messages.test.ts` — catalog key parity and duplicate-value drift across all four locales
- `glossary.test.ts` — voseo/tuteo separation holds for the new and edited strings
- `ios-store-gating.test.ts` — cross-chain withdraw gate still on for iOS, off for web/Android
- prettier

## Follow-ups (not in this PR)

- **Notes for Review** text describing the cashback programme — required before submission
- T&Cs, Crisp support macros, App Store listing copy and screenshots, and the `earn-with-peanut-*` blog set still describe "rewards" and the referral revenue share
- mono `product/rewards.md` pins `never_use: [cashback, …]`; that nomenclature rule needs updating to match. It is a documentation rule, not a CI check — the i18n glossary suite enforces locale register, not reward nomenclature, so nothing here goes red on it.


innolope-dev:
Superseded — the cashback copy needs to be iOS-native only, not global. Replacing with a branch that layers the copy behind a platform check so web and Android are byte-for-byte unchanged.

PR #2773, opened 2026-08-20
feat(rewards): iOS-only cashback copy, referral surfaces restored

innolope-dev:
Restores the referral surfaces hidden on iOS in `4564b401`, and presents the programme as **cashback in the native iOS build only**. Web and Android are unchanged.

Supersedes #2771, which made the copy change globally.

## Why the gate comes off

`isReferralRewardsHidden()` hid `/rewards`, `/rewards/invites`, the home pill, the profile row, both invite carousel CTAs, the surprise-claim treatment and the receipt points row — while the backend kept accruing and `referral.reward.earned` kept pushing *"You earned $12.40! Bruno joined Peanut with your invite"* to the same devices, deep-linking to the guarded route.

Guideline **2.3.1** treats hidden-but-discoverable features as grounds for removal and escalates to account level for repeat behaviour. Disclosed in Notes for Review, the worst case is a rejection cycle. That disclosure is the load-bearing mitigation and is a release-checklist item, not a code change.

The old citation was wrong: `appStoreCompliance.ts` cited **3.1.5 (ii)** — Mining. The clause about offering currency for encouraging downloads is **3.1.5 (v)**.

## Web and Android are untouched — how that's enforced

Every pre-existing catalog string is byte-identical to `dev` across all four locales. Overrides live in a separate `iosCopy` block; the base keys are not edited. Verified mechanically:

```
en:     changed=none  removed=none
es-419: changed=none  removed=none
es-AR:  changed=none  removed=none
pt-BR:  changed=none  removed=none
```

The two new UI elements — the "How cashback works" card and the lifetime caption on the invites page — are behind `isIOSNative()`.

Three call-site rewrites are no-ops off iOS by construction: the invite-modal ternary already resolved to `description` on web; the profile menu position already resolved to `middle`; and the invites page's new `invitesTitle` key holds the same string as the `title` it replaced.

## The copy layer

`useAppTranslations` wraps `useTranslations` and prefers `iosCopy.<key>` when the namespace has one, falling back to the base string otherwise. Call sites keep their existing keys — only the hook name changes.

```ts
const t = useAppTranslations('rewards')
t('title')     // iOS: rewards.iosCopy.title → "Cashback"
               // web: rewards.title         → "Rewards"
t('inviteNow') // both: rewards.inviteNow    → "Invite Now"
```

Off iOS the hook returns the original translator unchanged, so there is no wrapper in the non-iOS path at all.

**The block is `iosCopy`, not `ios`, on purpose.** `profile.backup.steps.ios` is already content — a namespace whose own content sat under `ios` would have had every one of its keys silently redirected on iOS. A test pins that distinction.

Platform is read at render time, never at module scope: the Capacitor bridge only lands on `window` after module eval and is absent during prerender.

## What changes on iOS

45 overrides per full locale, plus voseo deltas for es-AR.

- **rewards → cashback.** "Cashback" as a loanword is the standard term in both LatAm markets.
- **Attribution moves from signups to transactions.** `used Peanut` → `paid with Peanut`; `The more they use` → `The more they pay`. Same person, same money — but the cashback is tied to a transaction rather than a download, and downloading is what clause (v) names.
- **New four-step "How cashback works" card** on `/rewards`. The earn instruction used to be a half-sentence appended to the lifetime total (`"Lifetime rewards: $X. To earn more, invite friends."`). As its own card it reads better and states the process without pinning a specific dollar to a specific person.
- **Two multi-level admissions removed.** `qrPay.claim.inviteQrDescription` said "Points keep coming from the activity of friends **& their friends**" — the only place the UI stated the transitive structure, on a shareable QR-sticker screen. `claimSuccess.stickerDescription` lost "contribute towards your points forever".

Button labels stay Title Case (`Unwrap Cashback`); mid-sentence uses stay lowercase (`Invite friends. Earn cashback`).

## Deliberately not changed

**Push notification copy.** `PushChannel.send` targets by `userId`, and OneSignal fans one notification out to every subscription that user has. There is no per-platform copy path, and a single notification cannot say two different things to the same person's devices — so the referral push keeps its current wording on every platform, iOS included. Splitting it would need per-subscription platform storage and a two-send fan-out.

**The cross-chain withdraw gate.** It lives in `underMaintenance.config.ts`, rests on a different clause — **3.1.5 (iii)**, Exchanges, which requires per-region licensing and is enforced far more consistently — and multi-chain swaps are the strongest "this is a crypto app" signal we ship. `ios-store-gating.test.ts` keeps its coverage.

**`inviteFriendsModal.descriptionNoRewards`** stays in the catalogs though it is now unreferenced. It is the reward-free fallback copy — if we are rejected under (v), that string makes the response a resubmit rather than a redesign.

## Verification

- `pnpm jest src/i18n src/config src/components/Home src/components/Profile src/components/TransactionDetails src/app` — 764 passed, 3 skipped, 55 suites
- New `useAppTranslations.test.tsx` — both platforms, fall-through, all four locales, `has()` across both layers, and two catalog invariants: every override shadows a real base key (a typo would otherwise silently never resolve), and `iosCopy` stays distinct from the existing `ios` content key
- `messages.test.ts` key parity and duplicate-value drift, `glossary.test.ts` voseo/tuteo separation
- `tsc --noEmit` clean; the `@/assets/*.svg` TS2307s are pre-existing in any fresh worktree (`next-env.d.ts` is gitignored)
- prettier

## Follow-ups

- **Notes for Review** text describing the cashback programme — required before submission
- T&Cs, Crisp support macros, App Store listing copy and screenshots, and the `earn-with-peanut-*` blog set still describe rewards and the referral revenue share
- mono `product/rewards.md` pins `never_use: [cashback, …]`. That is a documentation rule, not a CI check — the i18n glossary suite enforces locale register, not reward nomenclature, so nothing here goes red on it.

## Summary by CodeRabbit

* **New Features**
  * Added iOS-specific cashback terminology and localized messaging across rewards, payments, profiles, transactions, and invitations.
  * Added an iOS-only “How it works” guide explaining cashback steps.
  * Rewards and points-related links and details are now available for activated users across supported screens.
  * Updated invite, claim, QR payment, and transaction messaging for clearer platform-specific experiences.

* **Bug Fixes**
  * Improved translation fallback behavior across supported languages and platforms.


innolope-dev:
@coderabbitai review

PR #2776, opened 2026-08-20
fix(claim): stop blaming the sender, and offer a retry when one would work

innolope-dev:
Two user reports on the send-link claim flow: the recipient couldn't claim, and the modal blamed the sender for not being "verified for that method". Those turned out to be two independent bugs that happen to land on the same screen.

> **Base note:** this is the `main` PR, targeting the branch the web app ships from. The `dev` twin ([#2777](https://github.com/peanutprotocol/peanut-ui/pull/2777)) is already merged.

## 1. The guest-verification modal blamed the sender for things that aren't the sender

`Initial.view.tsx` rendered one hardcoded line for every trigger:

> "The sender isn't verified for this method. You'll have to create an account, verify your identity, and then your funds will be deposited to your bank."

Three paths raise that modal, and it's only right on one of them:

| Trigger | Actual blocker |
|---|---|
| `SendLinkActionList` → bank, `GuestKycNeeded` | *sometimes* the sender |
| `SendLinkActionList` → mercadopago/pix, `!user` | the claimer has no account |
| `Initial.view` effect → `claimToMercadoPago && !user` | the claimer has no account |

MercadoPago and Pix never consult the sender's rails at all. And `GuestKycNeeded` is *also* reached when the link carries no `senderUserId`, or when `getUserById` 404s or throws — cases where nothing whatsoever was established about the sender. In every one of these the claimer is handed a fact they can't act on, and usually an untrue one.

**Fix:** callers pass a reason. `useDetermineBankClaimType` now exposes the sender's capability as a tri-state (`true` / `false` / `null`), because "we asked and the answer was no" and "we never got an answer" are different facts and only the first may be shown as the sender's fault.

Copy updated in `en`, `es-419` and `pt-BR`, with the voseo delta in `es-AR` (`es-AR` is deltas-over-`es-419`, so the shared keys stay in one place).

## 2. A retryable claim failure told users to contact support

When ZeroDev's paymaster failed on `zd_sponsorUserOperation`, the claim never happened and the API had already rolled the link back — the next attempt would most likely have worked. Users were told to contact support instead.

- **[PEANUT-UI-SJ5](https://peanut-c34d84c05.sentry.io/issues/PEANUT-UI-SJ5)** ← **[PEANUT-API-3M](https://peanut-c34d84c05.sentry.io/issues/PEANUT-API-3M)** — six users inside 30 minutes on 2026-08-19. Two of them hit it directly on `peanut.me/claim` within 15 minutes of each other, which is almost certainly the pair who reported this.
- Note the UI issue is sampled at 10% (`client_sample_rate: 0.1`), so real impact is higher than its 24 events suggest.

The API sanitizes 5xx prose (rightly — that path has leaked the vault address before), and `postJson` was **dropping the `code`** from the response body. So `friendlyError` had nothing left to classify and fell through to the support fallback.

**Fix:** `postJson` carries the wire code through, and `CHAIN_INFRA_UNAVAILABLE` maps to the existing localized retryable copy.

It stays a plain `Error` rather than `ApiError` on purpose: `ApiError` would also opt this endpoint into the blanket "any 5xx is retryable" rule, and a claim can fail 5xx for reasons retrying never fixes (an unsupported vault, say). Only codes the API states explicitly change the advice.

## Deploy ordering — safe in either direction

The backend half is [peanut-api-ts#1382](https://github.com/peanutprotocol/peanut-api-ts/pull/1382). **There is no ordering hazard**, because every change on both sides is additive:

- **This frontend first, backend later.** Fix 1 is entirely frontend and works immediately — it reads `canReceiveBankOfframp`, which the API already returns. Fix 2 becomes a no-op until the API ships `CHAIN_INFRA_UNAVAILABLE`: no code arrives, `friendlyError` falls through exactly as it does today, and users keep seeing the current "contact support" copy. There's a test pinning that (`the same prose WITHOUT the code keeps the support fallback`).
- **Backend first, frontend later.** The new code reaches an old frontend that drops it in `postJson` and falls through to the same copy as today. Rails parked in `REQUIRES_SUPPORT` render through a capability-resolver path the current frontend already handles.

So the only cost of splitting them is that each half's improvement waits for the other. Nothing breaks either way.

## Testing

CI green: unit **3384 ran, 0 failed**, plus typecheck, eslint, format, e2e, CodeQL and analyze.


abalinda (approved):

abalinda:
Code review(high · 7 findings)
src/components/Claim/useClaimLink.tsx
  ●   39 [correctness]    postJson dies on non-JSON or JSON-null error bodies, so the new wire-code contract degrades exactly during the infra outages it was built for (pre-existing lines in the touched function).
  ●   63 [reuse]          withWireCode is the codebase's third private copy of 'attach body.code to a thrown Error' and belongs in api-error.ts next to wireErrorCode; the code-dropping bug class stays live at sibling fetch boundaries.
src/components/Claim/Link/SendLinkActionList.tsx
  ●  152 [correctness]    A guest tapping Bank before the async sender lookup resolves bypasses the verification modal entirely (pre-existing race inherited by the touched handler).
src/components/Claim/Link/Initial.view.tsx
  ● 1132 [correctness]    The modal's onClose clears the show flag but not verificationPromptReason, so 'sender-unverified' stays sticky in context until resetFlow.
src/hooks/useDetermineBankClaimType.ts
  ●   27 [test-coverage]  The new tri-state gating logic ships with zero test coverage of the hook — the only test 'coverage' mocks the hook wholesale (CLAUDE.md: 'Every custom hook that fetches data, gates a flow, or holds persistent state needs a test').
src/components/Claim/Link/__tests__/SendLinkActionList.verificationReason.test.tsx
  ●   15 [simplification] ~75 lines of mock preamble duplicated near-verbatim from the existing SendLinkActionList.test.tsx; the 3 new tests could live there using its established mutable-mock pattern.
src/components/Claim/__tests__/useClaimLink.wireCode.test.ts
  ●   53 [simplification] Tests 1 and 2 re-assert the friendlyError mapping (networkBusyTimeout/genericSupport) that friendly-error.utils.test.tsx:275-289 in this same PR already pins — minor.

Review of peanut-ui#2776 is done — the PR is sound and nothing blocks merge. Both headline fixes verify end-to-end: the wire code survives the full path to friendlyError (allow-list gated, no TanStack re-wrap, retry: false), the deleted locale key has no orphaned consumers, es-AR merges cleanly, and every modal-open path sets a reason today.

7 findings reported above (2 other candidates were refuted during verification). The top three are hardening: an unguarded response.json() in postJson that loses the wire code exactly during infra outages, a race where a guest tapping Bank before the sender lookup resolves skips the verification modal, and a sticky verificationPromptReason that onClose never resets. The rest are reuse/test-coverage cleanups (third private copy of the attach-code helper, untested tri-state hook, duplicated test preamble/assertions).


Merging as is

PR #2778, opened 2026-08-20
feat(kyc): show a region screen instead of a support dead end

innolope-dev:
Same branch as peanutprotocol/peanut-ui#2775, targeting `dev` instead of `main`. **Merge one, close the other** — they are the identical three commits.

Backend: peanutprotocol/peanut-api-ts#1381.

## What this does

A user whose ID is rejected for its **jurisdiction** currently hits one of four endings, none of them honest:

| Surface | Today | Ending |
|---|---|---|
| Home activation card | "Verification issue — Contact support" | Ticket nobody can resolve |
| Identity row → drawer | "Retry verification" | Retry that can never pass |
| `/profile/identity-verification` | "Let's try that again" | Retry that can never pass |
| Add-money / withdraw bank gate | "We couldn't unlock this — contact support" | Ticket nobody can resolve |

All four are replaced by one screen: explains the block, **names no country**, offers no retry, offers no support link, and hands the user the part of the app that still works.

> **We can't verify IDs from your country**
>
> Our verification partner doesn't accept documents issued in your country. Re-uploading won't change the result.
>
> Your funds are safe, and you can still use Peanut to send and receive money with other people.
>
> **[ Send or request money ]**

## Two different endings, deliberately

The third commit fixes a **separate pre-existing bug** found while building this, and the two cases are deliberately handled differently:

| | Region-restricted | Fraud / sanctions / age / forgery |
|---|---|---|
| Explain the cause? | Yes, fully | **No** — compliance exposure, and it tips off actual fraudsters |
| Can support help? | No — it cannot lift a jurisdictional block | **Yes** — a human can review a misclassification |
| Retry offered? | No | No |
| Ending | Explanation + "Send or request money" | Generic + **Contact support** |

**The pre-existing bug:** `UnlockedRegions.view.tsx` hardcoded all three of `isTerminalRejection`'s inputs to `null`/`undefined` during the capabilities migration, so the check *always* returned "retryable". Every terminal rejection — fraud, sanctions, age — has been showing "Let's try that again" on that page. The drawer never had a terminal branch at all. Both now read the backend's verdict instead of reconstructing it from raw labels and attempt counts.

## Enforced at the choke point

Six gates open `InitiateKycModal`, each computing its variant from a rail gate that **cannot see why identity failed** — a region-restricted user reads as `needs-identity` and would be offered "Unlock now" straight back into the Sumsub SDK. The short-circuit lives inside the shared component, so a future call site cannot miss it. Tested against all five variants.

## Deploy-order safety

`isRegionRestricted` requires `identityVerification.reason`, which only the backend PR emits. **If this ships first, the region screen simply never renders** — every region path stays exactly as it is today. Nothing breaks.

The terminal fix is different and **does** take effect immediately, by design: `isTerminalFailure` treats a missing `canRetry` as terminal (fail-closed), on the grounds that offering a retry which cannot pass is worse than a support link that wasn't strictly needed. So the fraud/sanctions retry loop is fixed on FE deploy regardless of backend timing. The only cost in that window is the small `reviewAnswer: ERROR` cohort losing their (legitimate) retry until the backend ships. Say the word if you'd rather default the other way — it's one operator in `useIdentityVerification`.

## Verification

- **31 new tests** — 20 on the region screens (contract assertions run against every surface via `describe.each`, so drawer and modal can't drift), 8 on the hook, 3 on the home card.
- **`src/components`, `src/hooks`, `src/utils`, `src/i18n`: 2835 passed.** One pre-existing failure in `Invites/badge-campaign-context.test.ts` — fails identically on `main`, and passes in CI.
- CI on the `main` PR: **3404 unit tests, 0 failed**; typecheck, eslint, format, e2e all green.
- Copy added to all four locales, respecting each register (es-AR voseo, es-419 tuteo). No country named in any of them.

## Review focus

1. **Is peer-to-peer actually open for a Sumsub-rejected user?** The copy promises it. This PR doesn't gate anything, but it depends on that being true.
2. Is `/send` the right destination, or would `/home` be safer for a cohort likely holding no balance?
3. The fail-closed `canRetry` default described above.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EEmE861AXBaoE1y31ChW6e)_

### DIFF
--- a/src/components/Home/ActivationCTAs.tsx
+++ b/src/components/Home/ActivationCTAs.tsx
@@ -16,6 +16,7 @@ import { ANALYTICS_EVENTS } from '@/constants/analytics.consts'
 import { useCapabilities } from '@/hooks/useCapabilities'
 import { useCardInfo } from '@/hooks/useCardInfo'
 import { useIdentityVerification } from '@/hooks/useIdentityVerification'
+import { REGION_RESTRICTED_CTA_HREF } from '@/components/Kyc/KycRegionRestrictedContent'
 import ActionModal from '@/components/Global/ActionModal'
 import { useAuth } from '@/context/authContext'
 import { buildContactSupportMessage } from '@/utils/contact-support.utils'
@@ -49,6 +50,7 @@ export default function ActivationCTAs({ activationStep, onDismissCard }: Activa
     const t = useTranslations('home.activation')
     const tCommon = useTranslations('common')
     const tIdentity = useTranslations('identity')
+    const tRegion = useTranslations('kyc.regionRestricted')
     const router = useRouter()
     const { setIsQRScannerOpen, openSupportWithMessage } = useModalsContext()
     const { rails, channelOf, nextActions } = useCapabilities()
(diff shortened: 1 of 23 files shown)


innolope-dev:
Closing: all three commits from this PR are merged unchanged into #2790, the joint onboarding PR targeting `dev` (the region-restricted interrupt was semantically merged with the new home checklist — it outranks it, pinned by test).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2786, opened 2026-08-21
fix(send): stop over-promising account-less claims and surface the $5 fiat minimum at send time

innolope-dev:
Closes out the send-flow half of TASK-21724 (Michael Koh's report). The claim side is behaving as designed — Pix/Mercado Pago claims require a Peanut account, and the fiat rails have a $5 minimum — but the send flow set the wrong expectation on both counts.

## 1. The "no account needed" over-promise

The send-via-link card said **"No account needed to receive."**, which is only true for bank and external-wallet claims. Pix and Mercado Pago claims require a Peanut account (claim to balance first), so Michael's recipient hit "requires verification" on Mercado Pago and concluded the link was broken.

The copy now scopes the promise and names the caveat:

> No account needed for bank or wallet claims. Pix and Mercado Pago need a Peanut account.

Updated in `en`, `es-419`, and `pt-BR`. `es-AR` inherits the `es-419` string through the deltas-only overlay — the new copy has no second-person forms, so no voseo override is needed (the resolved-catalog glossary test confirms).

## 2. The invisible $5 fiat-claim minimum

`MIN_BANK_TRANSFER_AMOUNT`, `MIN_MERCADOPAGO_AMOUNT` and `MIN_PIX_AMOUNT` are all $5, but nothing surfaced this at send time — a sender could create a $3 link and strand the recipient with no fiat claim route (the "$5 minimum … but it didn't work" half of the report). The claim screen is too late: the money is already locked in the link.

`Initial.link.send.view.tsx` now shows a **non-blocking** warning `InfoCard` when the entered amount is below the smallest fiat minimum, telling the sender the recipient will only be able to claim to a Peanut account or crypto wallet. Small links stay sendable — account and wallet claims have no minimum — so Create link is never disabled by this. The threshold is `Math.min` of the three constants, so it tracks any future value change.

## Not in this PR

- Juan's peer-send failure (6cfc9e) is a different surface with no reproduction yet — split into its own task.
- The backend half of the claim-failure cluster (retryable paymaster failures, parked Bridge profiles) is peanut-api-ts#1382.

## Testing

- Two new tests in the existing view suite: the warning shows at $3 without disabling Create link; absent at $5.
- All 202 tests across the 15 `src/i18n` + `src/components/Send` suites pass on this branch, including catalog key parity, glossary compliance, and ICU compilation for the new keys.
- ESLint and Prettier clean on changed files; typecheck error set verified byte-identical before/after the change (pre-existing failures are missing generated assets in the sandbox).

Same change is also on `claude/send-link-claim-failures-ocfjvf` (main-based) if a production hotfix PR is wanted, mirroring the #2776/#2777 split.

https://claude.ai/code/session_01BLNUBFAcvLkHKdMMawQ1Zv

---
_Generated by [Claude Code](https://claude.ai/code/session_01BLNUBFAcvLkHKdMMawQ1Zv)_

### DIFF
--- a/src/components/Send/link/views/Initial.link.send.view.tsx
+++ b/src/components/Send/link/views/Initial.link.send.view.tsx
@@ -2,7 +2,9 @@
 
 import { useCreateLink } from '@/components/Create/useCreateLink'
 import ErrorAlert from '@/components/Global/ErrorAlert'
+import InfoCard from '@/components/Global/InfoCard'
 import PeanutActionCard from '@/components/Global/PeanutActionCard'
+import { MIN_BANK_TRANSFER_AMOUNT, MIN_MERCADOPAGO_AMOUNT, MIN_PIX_AMOUNT } from '@/constants/payment.consts'
 import { PEANUT_WALLET_TOKEN_DECIMALS } from '@/constants/zerodev.consts'
 import { TRANSACTIONS } from '@/constants/query.consts'
 import { loadingStateContext } from '@/context/loadingStates.context'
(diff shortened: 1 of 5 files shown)


innolope-dev:
Superseded by #2812 — identical diff (tree-identical head), reopened from a new branch to fix commit authorship; force-pushes are blocked repo-wide so the history could not be rewritten in place.

PR #2794, opened 2026-08-22
fix: block account deletion while the wallet still holds funds

innolope-dev:
## Problem

Settings offered **Delete My Account** identically to a user with $500 and a user with $0 — same modal, same copy, no mention of money. Deletion is irreversible (login is blocked forever server-side, no reactivation path), so any balance left behind is out of reach afterwards. One tap sequence, real money, no warning.

## Change

Gate the flow on the wallet's spendable balance (smart account + Rain collateral, via `useWallet`):

- **Funds present** → the button opens a *"Move your money first"* step showing the amount, with a CTA into `/withdraw` and a dismiss. The delete confirmation is unreachable until the balance is down to dust. Emits `delete_account_blocked_balance`.
- **Balance still loading** → falls through to the confirm step and relies on the backend's live read. `usersApi.requestDeletion` now surfaces the server's `409` as a typed `AccountHasBalanceError`, which flips the modal to the same blocked step using the server's figure — rather than the generic "try again" toast.
- **Dust / zero** → unchanged flow.

The `$0.01` floor mirrors the backend's authoritative threshold: sub-cent dust is neither displayable nor withdrawable, and trapping someone in an account they asked to delete over rounding is worse than losing it.

## Depends on

peanut-api-ts https://github.com/peanutprotocol/peanut-api-ts/pull/1394 — the server-side guard, which is what actually stops an old or patched client. This PR is the humane half; it degrades to the existing error toast if it ships first.

## Copy

New keys in `settings.deleteAccount` across en / es-419 / pt-BR, plus voseo deltas for es-AR (the new copy is imperative-heavy, so the tuteo inherited from es-419 reads wrong in AR).

## Testing

- 5 new `DeleteAccountButton` tests covering the blocked step, the `/withdraw` handoff, dust passing through, the unknown-balance fallthrough, and the server-refusal path — 10/10 green
- 186 i18n tests (catalog parity, duplicate-value drift, ICU compilation) green
- `tsc --noEmit` clean, `prettier --check` clean, `eslint` clean on touched files

## Summary by CodeRabbit

* **New Features**
  * Account deletion is now blocked when the wallet contains funds above the dust threshold.
  * A withdrawal option guides users to move remaining funds before deleting their account.
  * Balance-related deletion errors now display the reported amount when available.
  * Added localized messaging in English, Spanish, and Brazilian Portuguese.

* **Accessibility**
  * Added alternative text for the account-deletion illustration.

* **Analytics**
  * Added tracking for deletion attempts blocked by a remaining balance.

### DIFF
--- a/src/components/Settings/DeleteAccountButton.tsx
+++ b/src/components/Settings/DeleteAccountButton.tsx
@@ -16,6 +16,14 @@ import { DELETION_BALANCE_DUST_UNITS } from '@/utils/balance.utils'
 
 type ModalState = 'closed' | 'blocked' | 'confirm' | 'done'
 
+type Step = {
+    mascot: string
+    mascotAlt: string
+    title: string
+    description: string
+    ctas: ActionModalButtonProps[]
+}
+
 // A big animated mascot at the top of the modal instead of the tiny alert icon.
 // `unoptimized` keeps the animated WebP playing (Next's optimizer flattens it).
 const Mascot: FC<{ src: string; alt: string }> = ({ src, alt }) => (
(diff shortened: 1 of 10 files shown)


innolope-dev:
@coderabbitai review

innolope-dev:
@coderabbitai review

innolope-dev:
@coderabbitai review

PR #2795, opened 2026-08-22
fix(capgo): give every OTA upload a unique bundle version

### LINKED ISSUE #2638: Capgo OTA: --version-exists-ok greens the check but ships nothing (version never bumped)

innolope-dev:
Closes #2638. Replaces #2708 — same change, rebuilt on current `dev` (see *Why a new PR* below).

## Problem

`package.json`'s version only moves on a native release, so every OTA upload after the first reused the same bundle version. Capgo failed those uploads loudly until #2636 added `--version-exists-ok`, which turned them into silent no-ops: green check, nothing shipped.

## The fix that must not be used — and is currently in `dev`

Appending the short SHA looks equivalent and is not. `1.0.53-abc1234` is a semver **prerelease** and sorts **below** plain `1.0.53`. Capgo's update endpoint rejects any bundle that sorts under the device's native version (`disable_auto_update_under_native`, on by default — [update.ts](https://github.com/Cap-go/capgo/blob/main/supabase/functions/_backend/plugin_runtime/utils/update.ts)). So the upload succeeds, the dashboard shows a new bundle, and every device on the shipped binary refuses it.

`capgo-deploy.yml` on `dev` does exactly this for every non-production lane today.

## Changes

- **`scripts/capgo-bundle-version.mjs`** — prints `<major>.<minor>` from `package.json` + the git commit count (today: `1.0.10175`). Anchoring major/minor to `package.json` keeps the version above the native floor when a native release bumps it; the commit count makes it unique per commit. It refuses a shallow clone (would emit `1.0.1`), a prerelease override, and any `CAPGO_BUNDLE_VERSION` override below the native version.
- **`capgo-deploy.yml`** — `fetch-depth: 0`, a "Resolve bundle version" step, `--bundle <version>` on the upload, and a post-upload assertion that `channel currentBundle` actually serves the version the run just built. `--version-exists-ok` stays, but now only covers a re-run of the same commit; any other no-op leaves the channel stale and turns the job red. That check is #2638's "verify when fixed" criterion, enforced in CI.
- **Production is deliberately left alone.** `ota-1.0.48` is already plain and above the native floor, and `ota-1.0.48-hotfix1` is an intentional prerelease that still sorts above it. Production keeps its tag-derived version and does not go through the script, which refuses prereleases by design.
- **`capgo:upload:dev|staging`** package scripts get the same derived version — they had the identical collision.

## Verification

- 6 new jest tests (`scripts/__tests__/capgo-bundle-version.test.js`), including a real shallow-clone repo.
- The verification step was simulated under `bash -e` for match / stale-channel / CLI-error; it is green only on a real match.
- `prettier --check` clean on the touched files; the workflow parses.

## Why a new PR

#2708 was cut on 2026-08-17. Since then `dev` deleted `capgo-deploy-ios.yml` (folded into `capgo-deploy.yml`), moved production onto `ota-*` tags, and added the SHA-suffix scheme above — so #2708 conflicted. It could not be fixed in place: merging `dev` into that branch drags in 7 pre-2026-08-18 unsigned `src/content` bumps and the "All branches" ruleset rejects the push (`GH013`), while `non_fast_forward` blocks the rebase escape hatch. Same content, rebuilt on `dev`, reconciled with dev's consolidated lane. #2708 is closed in favour of this.

## Before the first release

Check on the Capgo dashboard that the channels' update policy is not `patch` — that setting blocks any patch-differing bundle, i.e. all OTA, regardless of the version scheme. Also note a native release that overrides `versionName` without bumping `package.json` would strand OTA bundles under the shipped binary.

### DIFF
--- a/.github/workflows/capgo-deploy.yml
+++ b/.github/workflows/capgo-deploy.yml
@@ -68,6 +68,8 @@ jobs:
               with:
                   submodules: true
                   token: ${{ secrets.SUBMODULE_TOKEN }}
+                  # Full history: the bundle version's patch is the commit count.
+                  fetch-depth: 0
 
             - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6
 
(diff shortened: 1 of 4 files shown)


innolope-dev:
@coderabbitai review

PR #2796, opened 2026-08-24
fix(seo): link localized pages to content owners

0xkkonrad:
## Summary

- point localized internal content links at the locale that owns the rendered prose, instead of emitting crawlable fallback URLs that canonicalize elsewhere
- keep genuinely owned `es-ar` pages local while resolving fallback-only pages through `es-419` and then English; preserve query strings, fragments, external URLs, and anchors
- resolve exact-origin `https://peanut.me` content URLs too, while leaving app/hub URLs and lookalike or insecure origins untouched
- keep filesystem-backed resolution server-side by splitting the visual footer chrome and serializing a resolved homepage href map across the server/client boundary

## Task

[TASK-21781](https://app.notion.com/p/3c68381175798101b16ec08033342716?pvs=204)

## Risks / breaking changes

- changes link destinations across marketing homepages, content directories, help/stories listings, destination grids, related pages, and MDX links
- no copy, layout, API, or user-flow change; locale ownership follows the checked-in content mirror, so future content additions automatically retain the requested locale
- production hotfix targets `main`; back-merge to `dev` is required after merge

- encoded content-path segments are currently looked up in their emitted form; the checked 763-file content inventory has no encoded, non-ASCII, or space-bearing content directories/links, so decoding and malformed-percent semantics are deferred to a follow-up rather than expanded into this hotfix

## QA

- Prettier: full repository check ✅
- TypeScript: detached `origin/main` baseline and hotfix head ✅
- Jest: 8 deterministic in-band shards, 274 suites / 3,427 passed / 3 skipped ✅
- focused locale-owner coverage: 46/46 tests ✅
- sitemap generator: exact 35 `es-ar` URLs enumerated ✅
- client dependency graph: all 489 `use client` roots traversed; 1,079 reachable modules / 5,785 runtime edges / zero `fs`, `path`, or `server-only` leaks ✅
- CI/Vercel: format, ESLint, typecheck, 3,430 unit tests, e2e, analysis, CodeQL, reports, aggregate CI, and hosted preview all green on `fa42b5bf` ✅
- hosted Vercel crawl: all 35 `es-ar` sitemap pages hydrated; 200 unique rendered internal targets checked with zero bad responses and zero fallback-canonical owner targets ✅
- hosted visual comparison at a 390×844 mobile viewport: 25/35 pages were raw pixel-identical; nine differed only in the live USD/ARS quote (0.0027–0.0032% pixels), while the homepage's decorative runtime animation differed by 0.5130%, below a production-vs-production self-noise control of 0.5209%; dimensions and non-dynamic copy/styles matched ✅
- local production build exception: three local webpack variants reached compilation without a source diagnostic, but the 11 GiB shared host killed them for memory (final documented `webpackMemoryOptimizations` run was kernel-OOM-killed at 5.36 GiB anonymous RSS); the authoritative CI/Vercel builds are green

Screenshots: quantified hosted comparison recorded above; no static artifact retained because the only observed deltas were live-rate and decorative animation noise (this PR changes href destinations, not rendered copy/layout/style)

## Summary by CodeRabbit

* **New Features**
  * Added consistent locale-aware links across landing pages, help articles, stories, destinations, and related content.
  * Updated landing-page navigation, comparison links, security resources, city links, and footer links to open the appropriate localized content.
  * Introduced a refreshed responsive footer with social links, legal resources, navigation, attribution, and locale switching.

* **Bug Fixes**
  * Corrected localized links to use the locale owning available content, with appropriate fallbacks.
  * Preserved query parameters and anchors when resolving localized URLs.
  * Added footer coverage for the quests page.

### DIFF
--- a/src/app/[locale]/(marketing)/help/page.tsx
+++ b/src/app/[locale]/(marketing)/help/page.tsx
@@ -3,7 +3,7 @@ import { type Metadata } from 'next'
 import { generateMetadata as metadataHelper } from '@/app/metadata'
 import { SUPPORTED_LOCALES, isValidLocale, getAlternates } from '@/i18n/config'
 import { getTranslations } from '@/i18n'
-import { readPageContentLocalized, listContentSlugs, resolveContentHref } from '@/lib/content'
+import { readPageContentLocalizedResolved, listContentSlugs } from '@/lib/content'
 import { notFound } from 'next/navigation'
 import { ContentPage } from '@/components/Marketing/ContentPage'
 import { Hero } from '@/components/Marketing/mdx/Hero'
@@ -91,11 +91,13 @@ export default async function HelpPage({ params }: PageProps) {
     const slugs = listContentSlugs('help')
     const articles = slugs
         .map((slug) => {
-            const content = readPageContentLocalized<HelpFrontmatter>('help', slug, locale)
-            if (!content || content.frontmatter.published === false) return null
+            const resolved = readPageContentLocalizedResolved<HelpFrontmatter>('help', slug, locale)
+            if (!resolved || resolved.content.frontmatter.published === false) return null
+            const { content, lang } = resolved
             return {
                 slug,
-                href: resolveContentHref(`/en/help/${encodeURIComponent(slug)}`, locale),
+                // The serving locale owns the prose, so link it directly.
+                href: `/${lang}/help/${encodeURIComponent(slug)}`,
                 title: content.frontmatter.title.replace(/\s*\|\s*Peanut Help$/, ''),
                 description: content.frontmatter.description,
                 category: content.frontmatter.category ?? 'General',
--- a/src/app/[locale]/(marketing)/help/page.tsx
+++ b/src/app/[locale]/(marketing)/help/page.tsx
@@ -3,7 +3,7 @@ import { type Metadata } from 'next'
 import { generateMetadata as metadataHelper } from '@/app/metadata'
 import { SUPPORTED_LOCALES, isValidLocale, getAlternates } from '@/i18n/config'
 import { getTranslations } from '@/i18n'
-import { readPageContentLocalized, listContentSlugs } from '@/lib/content'
+import { readPageContentLocalized, listContentSlugs, resolveContentHref } from '@/lib/content'
 import { notFound } from 'next/navigation'
 import { ContentPage } from '@/components/Marketing/ContentPage'
 import { Hero } from '@/components/Marketing/mdx/Hero'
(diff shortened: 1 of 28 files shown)


0xkkonrad:
@coderabbitai review

0xkkonrad:
@coderabbitai review

0xkkonrad:
@coderabbitai review

innolope-dev:
Pushed 3709ae983 on top of fa42b5bf5 with a set of review hardenings to the resolver. No behavior change for any link the PR already emits correctly — all existing tests pass unchanged; the changes close latent gaps:

**Correctness**
- **Ownership now requires published prose.** `existingPageLocale`/`existingCorridorLocale`/`existingSingletonLocale` walk the fallback chain once and skip files with `published: false`. Before, the first *existing* file won, so a draft translation (e.g. an unpublished `es-419` file over a live `en` one) would have made the SEO footer link a URL whose route `notFound()`s — sitewide. Nothing in the current mirror triggers this, which is exactly why it would have been invisible when it did.
- **Percent-encoded segments resolve.** Segments are decoded (with a malformed-percent fallback) before the content lookup, so the first non-ASCII slug won't silently lose its owner and revert to a fallback-only URL. This closes the trade-off deferred in the PR description; emitted hrefs keep their original encoding.
- **Absolute/relative consistency.** `https://peanut.me/en/help` now localizes to the requested locale like `/en/help` does, instead of staying English when no owner is found. Non-en absolute URLs (e.g. `https://peanut.me/es-ar/help`) are still treated as deliberate locale pins and pass through, per the existing tests.

**Structure**
- Legal slugs are derived from `listContentSlugs('legal')` instead of a hand-kept set, so the next legal doc can't be misclassified as a countries page.
- `RegulatedRails`/`YourMoney`/`SecurityBuiltIn` resolve their own hrefs via a shared memoized `contentHrefsFor` (new `landingContentHrefs.server.ts`); where the map remains a prop (`LandingPageClient`, `NoFees`, `FooterChrome`'s security href) it is now **required**, so a call site that forgets it fails to compile instead of shipping localized copy with `/en/` links. `NoFees` also lost its unused `locale` prop.
- The help and stories hubs build each link directly from the locale that served the prose (`readPageContentLocalizedResolved`), removing the second fallback-chain walk per slug and keeping the href consistent with the file the listing's published-gate saw.
- `localizeContentHref` (now caller-less and divergent from the resolver) is deleted with its tests; `LandingContentHrefs` is derived from the map instead of a hand-written 21-field interface; the test imports the production `contentHrefsFor` instead of re-implementing it.

**Verification**: `tsc --noEmit` clean; the 7 affected suites pass (63 tests, incl. new cases for absolute-hub localization and encoded-segment lookup); prettier clean. Full suite left to CI.

abalinda (dismissed):

innolope-dev:
48aa7ffa9 fixes the Vercel failure from 3709ae983: the three landing sections that now resolve hrefs from the filesystem were still re-exported from the `LandingPage` barrel, which client components import — pulling `fs` into the client bundle (`Module not found: Can't resolve 'fs'`). The barrel is client-safe again (server sections dropped from it; `/quests` imports them directly). Verified with a full local `next build` before pushing.

abalinda (approved):

PR #2800, opened 2026-08-24
fix(i18n): deposit note claimed a 0.1% bridging fee that does not exist

abalinda:
## Summary

The crypto deposit screen told every user "A small bridging fee (~0.1%) applies — you'll receive slightly less than you send." No such fee exists: deposits ledger 1:1 and the Rhino fee is booked on withdraw only. The copy drifted in during the i18n string extraction (0cffa3e1e) and contradicted both the same screen's tooltips (which correctly say ±0.1% *variance*) and the fees help page (deposits free). A user caught the contradiction and disputed it in support.

Fix: rename `bridgingFeeNote` → two variance notes.
- EVM: "USDC on Arbitrum arrives exactly. Deposits from other chains or tokens are bridged and can vary slightly (±0.1%)."
- SOL/TRON (always bridged): "Deposits are bridged to Arbitrum and can vary slightly (±0.1%)."

All four locales updated (en, es-419, es-AR, pt-BR).

## Task

TASK-21792 — https://app.notion.com/p/3c68381175798144a30eebe47bcb1847

## Risks / breaking changes

Copy-only. No backend, no logic beyond an `isEvm` ternary on an existing prop. No cross-repo impact.

## QA

Open Add money → Crypto on each network type: EVM shows the "arrives exactly / other chains vary" note, SOL/TRON show the bridged-variance note, the offramp-migration variant is unchanged (`multipleTransfersNote`).

Screenshots: ⚠️ NONE — copy-only swap inside an existing `<p>` (same element, same classes, layout untouched). The exact new strings are quoted above; capturing all 3 network states would need a full sandbox boot for a text diff already visible in this body.

- Two keys instead of one parameterized string: the EVM and SOL/TRON messages differ structurally (EVM needs the "arrives exactly" clause), so a ternary on the existing `isEvm` beats string interpolation.
- `CryptoDeposit.view.tsx` complexity (CC 49) flagged by code-analysis is pre-existing; this PR adds one ternary (+0.19 painscore, 0 net findings). Not touched — out of blast radius.


abalinda:
@coderabbitai review

kushagrasarathe (approved):

PR #2806, opened 2026-08-24
ci(release): couple native releases to the production OTA lane

innolope-dev:
TASK-21793. Follow-up to #2795 (which fixed the silent-no-op bug class for the dev→staging lane and predicted this incident in its own PR body).

## The incident

Internal store builds shipped with versionName **1.0.53** while the newest production OTA bundle was **1.0.51**. Capgo's `disable_auto_update_under_native` rule rejects any bundle sorting below the installed native version, so every device on the new binaries refused all existing bundles — OTA silently dead for the whole beta fleet (102 devices / 1,002 refusal events in 30 days; 26 devices stuck on July-29 JS), with green CI throughout. The immediate remediation (`ota-1.0.54` on the 08-21 build commit) already shipped; this PR makes the failure class impossible to repeat silently.

## Changes

- **`android-release.yml` / `ios-release.yml`** — three new pieces, kept in lockstep across the two platforms:
    - *Resolve release versionName* up front (input > tag > package.json, the same precedence both workflows already used inline), so every later step sees the exact version the binary ships with.
    - *Check production OTA floor* before the long build: read `channel currentBundle production` from Capgo and compare against the versionName with `scripts/semver-newer.mjs`. Unreadable channel or unparseable version fails the job — a binary that may strand the fleet does not ship on a guess.
    - *Publish matching production OTA bundle* after the store upload, only when the binary would leapfrog the newest bundle: upload the release's own `out/` (the exact JS baked into the binary) under the binary's versionName, then assert the channel actually serves it. The lanes can no longer drift apart; a failed publish turns the job red instead of the fleet going dark. Both platforms publishing the same version from the same tag is a verified no-op via `--version-exists-ok`.
- **`scripts/semver-newer.mjs`** — prerelease-aware "sorts strictly above" comparison. `sort -V` gets the case that matters here backwards: `1.0.54-hotfix1` sorts *below* `1.0.54` in semver (and in Capgo's on-device check), *above* it in GNU version sort. Fails loudly on unusable input so a `bash -e` step dies rather than misreading an error as "false".
- **`capgo-deploy.yml` / `docs/NATIVE-RELEASE.md`** — the "production OTA fires on an `ota-*` tag and nothing else" design notes now document this one deliberate exception.

## Task items covered elsewhere

- *Post-upload assertion on the production `ota-*` lane*: already live — #2795's "Verify channel serves the new bundle" step merged unconditional (no channel guard), so it ran on the `ota-1.0.54` remediation deploy. No change needed.
- *Sentry alert on `disable_auto_update_under_native`* and the *Capgo channel update-policy check* are ops-side, tracked in TASK-21793.

## Verification

- 16 new jest tests for `semver-newer.mjs` (plain/prerelease ordering incl. numeric identifiers, loud failure on unusable input); full `scripts/__tests__` suite green (35 tests).
- The floor-check step simulated under `bash -e` for all four paths: ahead → publish, covered → skip, unreadable channel → red, unparseable version → red.
- All three workflows parse; `prettier --check` clean on every touched file.

### DIFF
--- a/.github/workflows/android-release.yml
+++ b/.github/workflows/android-release.yml
@@ -113,22 +113,61 @@ jobs:
                       echo "NEXT_PUBLIC_ONESIGNAL_DEBUG=true" >> .env.production.local
                   fi
 
+            - name: Resolve release versionName
+              id: version
+              env:
+                  INPUT_VERSION: ${{ github.event.inputs.versionName || '' }}
+              run: |
+                  # versionName: manual dispatch input wins; else the tag name minus
+                  # its leading 'v' (v1.0.10 -> 1.0.10); else package.json (the same
+                  # fallback build.gradle uses). Resolved once, up front, so the OTA
+                  # floor check and the bundle publish below see the exact version
+                  # the binary ships with. Keep in lockstep with ios-release.yml.
+                  VERSION_NAME="$INPUT_VERSION"
+                  if [ -z "$VERSION_NAME" ] && [ "$GITHUB_REF_TYPE" = "tag" ]; then
+                      VERSION_NAME="${GITHUB_REF_NAME#v}"
+                  fi
+                  if [ -z "$VERSION_NAME" ]; then
+                      VERSION_NAME="$(node -p "require('./package.json').version")"
+                  fi
+                  echo "name=$VERSION_NAME" >> "$GITHUB_OUTPUT"
+                  echo "Release versionName: $VERSION_NAME"
+
+            # A binary whose versionName is ahead of the newest production OTA bundle
+            # makes Capgo's disable_auto_update_under_native rule reject every bundle
+            # on those devices — OTA goes silently dead for the whole fleet while CI
+            # stays green (TASK-21793: 102 devices, a month of refused updates). Read
+            # the floor before the long build; the publish step after the store
+            # upload closes the gap this detects. Keep in lockstep with
+            # ios-release.yml.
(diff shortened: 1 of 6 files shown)


innolope-dev:
@coderabbitai review

abalinda:
Decision: **merge this PR.** The mechanism is right — a native release that would leapfrog the production bundle now publishes its own `out/` under the same version and asserts the channel serves it, so the incident class cannot repeat silently on releases cut from this workflow.

One thing this PR does not cover that we need **before the next native build**: **bump `package.json` to ≥ 1.0.55** (on `dev`, `main`, and `mobile-release`).

Why: `package.json` is 1.0.53 and the production channel now serves `ota-1.0.54` (the 08-24 remediation, JS from the 08-21 commit). A native build cut today without a bump gets versionName 1.0.53, the floor check reports it as "covered", no bundle is published — and on first launch those devices download bundle 1.0.54, which is **older** JS than what the binary bakes in. Silent downgrade. The "equal-or-below the current bundle = error" rule that closes this properly can be a follow-up; the bump is what prevents it right now.

Follow-ups, not merge gates: gate the auto-publish on `track == production` + a `v*` tag (a manual dispatch on a feature branch currently publishes that branch's JS to every device), re-read `currentBundle` right before publishing, anchor the version parse to the channel line, and pin `@capgo/cli`. Also note this guard only runs on releases cut from commits that contain it — it needs to reach `main` and `mobile-release` too.

PR #2810, opened 2026-08-24
fix(sentry): give stackless console events a stack and stop issue explosion

innolope-dev:
## What

Started from [this PostHog view](https://eu.posthog.com/project/138913/activity/explore) — `$exception` events where `$cymbal_errors` is set. That property is PostHog's flag for *"empty exception list"*: the event arrived with a message but no `exception.values`, so no type and no stack.

**Those events were reaching Sentry fine.** The problem is the shape they arrived in, and how they grouped. 433 events / 255 people over 7d (~3.4% of exception volume).

## Why they were stackless

`captureConsoleIntegration` only calls `captureException` when one of the console args **is an `Error` instance**; otherwise it calls `captureMessage`, which carries no stack at all ([`captureconsole.js`](https://github.com/getsentry/sentry-javascript/blob/8.55.0/packages/core/src/integrations/captureconsole.ts)). Every event on that page was either a `console.error` whose args held no `Error`, or one of our deliberate `captureMessage` calls.

## Changes

**1. `attachStacktrace: true`** — synthesizes a stack for every message event, including `fetchWithSentry` / `native-auth-capture`, and third-party console noise where we can't touch the call site. One line, covers all 433.

**2. Call sites that destroyed their payload** — a stack can't recover an error the call site already threw away:

| Site | Was |
|---|---|
| `SetupPasskey.tsx` | passed `err.name, err.message` instead of `err` — largest family, ~126/wk |
| `useQRScanner.ts` ×2 | qr-scanner rejects with bare strings (`'Camera not found.'`) |
| `SumsubKycWrapper.tsx` | no `Error` object at all |
| `withdraw/crypto/page.tsx` | logged `[object Object]` — money-moving flow |
| `useSendMoney.ts` | logged `[object Object]` — money-moving flow |

Adds `utils/to-error.ts`, replacing what were already three ad-hoc copies of the same normalization. It JSON-serializes plain objects so the shape survives into the title instead of collapsing.

Passkey filtering is unchanged — the `AbortError` / `not allowed by the user` variants still match, via exception type rather than message.

**3. `sanitizeUrl` fingerprint gap** — it normalized numeric query values but not the rest, so one upstream 429 was three separate issues:

| Issue | Title | Events (7d) |
|---|---|---|
| PEANUT-UI-SGN | `exchange-rate?accountType=iban` | 273 |
| PEANUT-UI-SQ4 | `exchange-rate?accountType=clabe` | 151 |
| PEANUT-UI-SSA | `exchange-rate?accountType=gb` | 131 |

Now normalizes every value. A lookahead skips values the numeric pass already replaced, so fingerprints that were *already* correct keep their existing issue rather than forking a new one on deploy.

While adding tests for this I found a pre-existing bug: the numeric rule was unanchored, so `?address=0xaf88d065` had its leading `0` replaced and became `{id}xaf88d065` — every wallet still got its own fingerprint *and* the value was unreadable. Anchored with `(?=&|$)`.

**4. OneSignal issue explosion** — its op-failure message inlines a per-device `onesignalId`, so every device minted its own Sentry issue. These are **fingerprinted, not dropped**: push-subscription writes failing is exactly the signal that was missing while iOS push was dead on a mis-provisioned APNs key. Collapsed onto one issue per op name. The worker messenger's no-service-worker warning is pure noise and is ignored.

## Note

`attachStacktrace` lands on `threads`, not `exception.values`, so PostHog's mirror will still flag the remaining third-party ones as an empty exception list. That flag only clears where we now pass a real `Error`. Sentry is fully fixed; the PostHog view shrinks but won't empty out.

## Testing

- New: `to-error.test.ts` (8 cases), `sentry-fingerprint-url.test.ts` (7 cases, incl. the no-refingerprint regression guard)
- Extended `sentry.utils.test.ts` with the OneSignal fingerprint + ignore cases
- 94 sentry/`toError` tests and 272 tests across the touched components pass; prettier and eslint clean

## Overlap

Touches `sentry.utils.ts`, as does #2735 (chained-error matching + Capgo noise). No textual conflict — different regions, and `collapseNoisyFingerprint` composes with its `shouldIgnoreError` changes — but whichever lands second deserves a re-read.

### DIFF
--- a/sentry.utils.test.ts
+++ b/sentry.utils.test.ts
@@ -1,5 +1,5 @@
 import type { ErrorEvent } from '@sentry/nextjs'
-import { shouldIgnoreError } from './sentry.utils'
+import { beforeSendHandler, shouldIgnoreError } from './sentry.utils'
 import { criticalFlowTags } from '@/utils/sentry-critical-flow'
 
 function eventWith(partial: {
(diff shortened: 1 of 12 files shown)


innolope-dev:
@coderabbitai review

PR #2812, opened 2026-08-25
fix(send): stop over-promising account-less claims and surface the $5 fiat minimum at send time

innolope-dev:
Closes out the send-flow half of TASK-21724 (Michael Koh's report). The claim side is behaving as designed — Pix/Mercado Pago claims require a Peanut account, and the fiat rails have a $5 minimum — but the send flow set the wrong expectation on both counts.

> **Supersedes #2786** — identical diff, reopened from a new branch to fix commit authorship (force-pushes are blocked repo-wide, so the history could not be rewritten in place).

## 1. The "no account needed" over-promise

The send-via-link card said **"No account needed to receive."**, which is only true for bank and external-wallet claims. Pix and Mercado Pago claims require a Peanut account (claim to balance first), so Michael's recipient hit "requires verification" on Mercado Pago and concluded the link was broken.

The copy now scopes the promise and names the caveat:

> No account needed for bank or crypto-wallet claims. Pix and Mercado Pago need a Peanut account.

Updated in `en`, `es-419`, and `pt-BR`. `es-AR` inherits the `es-419` string through the deltas-only overlay — the new copy has no second-person forms, so no voseo override is needed (the resolved-catalog glossary test confirms).

## 2. The invisible $5 fiat-claim minimum

`MIN_BANK_TRANSFER_AMOUNT`, `MIN_MERCADOPAGO_AMOUNT` and `MIN_PIX_AMOUNT` are all $5, but nothing surfaced this at send time — a sender could create a $3 link and strand the recipient with no fiat claim route (the "$5 minimum … but it didn't work" half of the report). The claim screen is too late: the money is already locked in the link.

`Initial.link.send.view.tsx` now shows a **non-blocking** warning `InfoCard` when the entered amount is below the smallest fiat minimum, telling the sender the recipient will only be able to claim to a Peanut account or crypto wallet. Small links stay sendable — account and wallet claims have no minimum — so Create link is never disabled by this. The threshold is `Math.min` of the three constants, so it tracks any future value change.

## Not in this PR

- Juan's peer-send failure (6cfc9e) is a different surface with no reproduction yet — split into its own task.
- The backend half of the claim-failure cluster (retryable paymaster failures, parked Bridge profiles) is peanut-api-ts#1408.

## Testing

- Two new tests in the existing view suite: the warning shows at $3 without disabling Create link; absent at $5.
- All 202 tests across the 15 `src/i18n` + `src/components/Send` suites pass on this branch, including catalog key parity, glossary compliance, and ICU compilation for the new keys.
- ESLint and Prettier clean on changed files; typecheck error set verified byte-identical before/after the change (pre-existing failures are missing generated assets in the sandbox).

A main-based mirror can be cut for a production hotfix if wanted, mirroring the #2776/#2777 split.

### DIFF
--- a/src/components/Send/link/views/Initial.link.send.view.tsx
+++ b/src/components/Send/link/views/Initial.link.send.view.tsx
@@ -4,7 +4,7 @@ import { useCreateLink } from '@/components/Create/useCreateLink'
 import ErrorAlert from '@/components/Global/ErrorAlert'
 import InfoCard from '@/components/Global/InfoCard'
 import PeanutActionCard from '@/components/Global/PeanutActionCard'
-import { MIN_BANK_TRANSFER_AMOUNT, MIN_MERCADOPAGO_AMOUNT, MIN_PIX_AMOUNT } from '@/constants/payment.consts'
+import { CLAIM_RAIL_MINIMUMS } from '@/constants/payment.consts'
 import { PEANUT_WALLET_TOKEN_DECIMALS } from '@/constants/zerodev.consts'
 import { TRANSACTIONS } from '@/constants/query.consts'
 import { loadingStateContext } from '@/context/loadingStates.context'
@@ -28,8 +28,10 @@ import { ANALYTICS_EVENTS } from '@/constants/analytics.consts'
 // Below the smallest fiat minimum the recipient loses every fiat claim rail
 // (bank / Pix / Mercado Pago all reject at claim time) and is left with only
 // Peanut-account or wallet claims. Warn the sender here — the claim screen is
-// too late, the money is already locked in the link.
-const MIN_FIAT_CLAIM_AMOUNT = Math.min(MIN_BANK_TRANSFER_AMOUNT, MIN_MERCADOPAGO_AMOUNT, MIN_PIX_AMOUNT)
+// too late, the money is already locked in the link. The warning copy asserts
+// that conjunction, which only holds while the per-rail minimums agree — a
+// test pins them equal so a divergence forces the copy question.
+const MIN_FIAT_CLAIM_AMOUNT = Math.min(...Object.values(CLAIM_RAIL_MINIMUMS))
 
 const LinkSendInitialView = () => {
     const t = useTranslations('send')
@@ -64,10 +66,8 @@ const LinkSendInitialView = () => {
 
     // Informational only — small links are legitimate (Peanut-account / wallet
     // claims have no minimum), so this never blocks Create link.
-    const isBelowFiatClaimMinimum = useMemo(() => {
-        const amount = parseFloat(tokenValue ?? '')
-        return amount > 0 && amount < MIN_FIAT_CLAIM_AMOUNT
-    }, [tokenValue])
+    const enteredAmount = parseFloat(tokenValue ?? '')
+    const isBelowFiatClaimMinimum = enteredAmount > 0 && enteredAmount < MIN_FIAT_CLAIM_AMOUNT
 
     const handleOnNext = useCallback(async () => {
         try {
--- a/src/components/Send/link/views/Initial.link.send.view.tsx
+++ b/src/components/Send/link/views/Initial.link.send.view.tsx
@@ -2,7 +2,9 @@
 
 import { useCreateLink } from '@/components/Create/useCreateLink'
 import ErrorAlert from '@/components/Global/ErrorAlert'
+import InfoCard from '@/components/Global/InfoCard'
 import PeanutActionCard from '@/components/Global/PeanutActionCard'
+import { MIN_BANK_TRANSFER_AMOUNT, MIN_MERCADOPAGO_AMOUNT, MIN_PIX_AMOUNT } from '@/constants/payment.consts'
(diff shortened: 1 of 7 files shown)


innolope-dev:
@coderabbitai review

innolope-dev:
Applied the CodeRabbit copy finding in 0574c9b: the caveat now says crypto wallet explicitly in all three locales (en "bank or crypto-wallet claims", es-419 "billetera cripto", pt-BR "carteira cripto"). es-AR still inherits es-419. All 202 tests across the i18n + Send suites pass; prettier unchanged.

src/i18n/app/messages/en.json
```diff
@@ -744,7 +744,7 @@
     "send": {
         "linkCard": {
             "title": "Send money with a link",
-            "description": "No account needed to receive.",
+            "description": "No account needed for bank or crypto-wallet claims. Pix and Mercado Pago need a Peanut account.",
```
jjramirezn: **"No account needed for bank ... claims" is false on this branch** (verified)

Every account-less path to a bank claim is currently a dead end: `useDetermineBankClaimType` resolves a guest to `GuestBankClaim` — hard-disabled, `SendLinkActionList.tsx:119-129` greys the bank card as "Soon!" because the BE 503s `POST /bridge/offramp/create-for-guest` — or `GuestKycNeeded`, which is a signup+KYC wall, i.e. an account requirement.

So the rewritten card re-creates, for the bank rail specifically, the exact over-promise this PR ("stop over-promising account-less claims") sets out to remove: a sender reads "no account needed for bank claims", sends to a recipient without a Peanut account, and the recipient finds bank greyed out or walled. Same copy in `es-419.json:747` and `pt-BR.json:747` (and es-AR inherits via deep-merge).

Suggestion: scope the sentence to crypto-wallet claims only, or reintroduce the bank mention when guest bank claims ship.

src/components/Send/link/views/Initial.link.send.view.tsx
```diff
@@ -23,6 +25,12 @@ import { usePendingTransactions } from '@/hooks/wallet/usePendingTransactions'
 import posthog from 'posthog-js'
 import { ANALYTICS_EVENTS } from '@/constants/analytics.consts'
 
+// Below the smallest fiat minimum the recipient loses every fiat claim rail
+// (bank / Pix / Mercado Pago all reject at claim time) and is left with only
+// Peanut-account or wallet claims. Warn the sender here — the claim screen is
+// too late, the money is already locked in the link.
+const MIN_FIAT_CLAIM_AMOUNT = Math.min(MIN_BANK_TRANSFER_AMOUNT, MIN_MERCADOPAGO_AMOUNT, MIN_PIX_AMOUNT)
```
jjramirezn: **`Math.min` of the three minimums is only correct while they happen to be equal** (verified; latent drift)

The copy gated by this threshold asserts a conjunction — "can't be claimed to a bank, Pix or Mercado Pago" — that holds only because all three constants are $5 today. `payment.consts.ts` already carries a diverged per-rail value (`MIN_PIX_AMOUNT_BRL = 1`), and no test pins the three equal.

If any one minimum drops (say Pix to $1), `Math.min` silently follows: a $3 link shows **no warning** although bank and Mercado Pago still reject at claim time — the exact TASK-21724 blind spot this PR closes — and a $0.50 link shows copy naming Pix as unavailable when it is claimable.

Fiat-minimum knowledge now lives in three uncoordinated places: `validateMinimumAmount`'s map (payment.consts.ts), the per-rail ternary in `SendLinkActionList.tsx:140-144`, and this `Math.min` + rail names hardcoded in three locale strings. Suggest exporting a shared `CLAIM_RAIL_MINIMUMS` map from `payment.consts.ts` that all three derive from, so a divergence forces the copy question instead of silently changing the gate.

jjramirezn:
**Optional nit**

`useMemo` around `isBelowFiatClaimMinimum` memoizes a `parseFloat` plus two comparisons producing a primitive — the memo machinery costs as much as the computation, and nothing depends on referential stability. A plain `const amount = parseFloat(tokenValue ?? ''); const isBelowFiatClaimMinimum = amount > 0 && amount < MIN_FIAT_CLAIM_AMOUNT` is less code with identical behavior. (Matches the local style of neighboring memos, so feel free to ignore.)

src/i18n/app/messages/en.json
innolope-dev: Fixed in 24f8c58 — bank is out of the promise: "No account needed for crypto-wallet claims. Bank, Pix and Mercado Pago need a Peanut account." (es-419/pt-BR match, es-AR still inherits). Reintroducing bank when guest bank claims ship is a one-string change.

src/components/Send/link/views/Initial.link.send.view.tsx
innolope-dev: Fixed in 24f8c58 — added CLAIM_RAIL_MINIMUMS to payment.consts.ts; validateMinimumAmount, the SendLinkActionList per-rail lookup and the send-time gate all derive from it, and a test pins the three values equal with a comment saying exactly what to revisit (the warning copy + Math.min gate) when it fails. The useMemo nit is in the same commit.

jjramirezn (approved):

PR #2817, opened 2026-08-25
ci: stop double-running Tests on every PR commit

innolope-dev:
## What

Restrict the `Tests` workflow's `push` trigger from `branches: ['**']` to `branches: [main, dev]` (post-merge signal only). Also updates the now-stale comment in `content-publish-automerge.yml` that described the two-runs-per-SHA behavior.

## Why

Every PR commit currently runs the **entire suite twice**: the `push` and `pull_request` events fire on the same SHA but land in different concurrency groups (`pull_request.number` vs `ref`), so neither cancels the other. The run list shows push/PR pairs seconds apart, each ~2 min across 7 jobs — roughly half of this repo's Tests compute is duplicate work. The `pull_request` run is the better of the two: it tests the merge commit against the base.

## Safety check on content publishes

`content-publish-automerge`'s `merge-on-green` path triggers on **any** completed `Tests` run (including `pull_request`-event runs) and its gate requires *"at least one and ALL green"* `ci-success` check runs per SHA — one run per SHA satisfies it. No other workflow consumes push-triggered Tests runs (checked `workflow_run` references across `.github/workflows/`).

### DIFF
--- a/.github/workflows/content-publish-automerge.yml
+++ b/.github/workflows/content-publish-automerge.yml
@@ -134,9 +134,10 @@ jobs:
                   # on; the run conclusion also folds in advisory jobs (e2e). Gate on
                   # exactly what the ruleset gates on.
                   #
-                  # `Tests` runs on both `push` and `pull_request`, so a commit usually
-                  # carries TWO ci-success check runs. Require at least one and ALL of
-                  # them green — a single red run must block the merge.
+                  # `Tests` runs on `pull_request` (and `push` for dev/main), so a
+                  # commit can carry more than one ci-success check run. Require at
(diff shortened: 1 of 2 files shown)


innolope-dev:
@coderabbitai review

innolope-dev:
TL;DR: every PR commit was running the full test suite **twice** — the `push` and `pull_request` triggers both fired on the same commit and neither cancelled the other. Now it runs once per commit.

**Key speed results:** wall time per commit is unchanged (~2.2 min, the runs were parallel), but CI compute is **halved** — 7 duplicate jobs eliminated per commit (~15 runner-minutes saved each push), with zero change to what's tested.

PR #2819, opened 2026-08-26
qa: fixtures, screenshot diffing, and a CI job that can actually fail

kushagrasarathe:
## Summary

Verifying a design-system change meant building the base branch, screenshotting by hand, building the head branch, screenshotting again, and pairing the results into a table. PR #2758 shipped 25 of those, at one width, and the table could not say which screens had actually moved.

This makes that automatic, and removes a CI job that had reported success without running a test since April.

**Base: `feat/design-system`** (PR #2813). Ships with the design-system release.

## What lands

**1. Named app states — `?__fixture=<name>`**

30 fixtures. `<route>?__fixture=<name>` answers every API call from a named state, so any screen renders with **no database, no API, no provider keys and no harness secret**. The same URL works in a browser, in `/dev/devices` at six widths, on a Vercel preview, and under Playwright.

Hugo's question — "I want to see the profile screen with username `bh12ui2buibui52bi` and 1.2 billion points" — is now a URL. `/dev/fixtures` lists them all.

Built on `demo-api`, which already answers every route the app calls, so a fixture states only what differs. Most are one to three lines.

**2. Screenshots — `npm run test:visual`**

30 fixtures × 4 widths (320/375/393/430) = 120 PNGs in about 60 seconds. 320 is an iPhone on Larger Text and finds the most bugs; 430 is a Pro Max.

`npm run test:visual:diff a b` prints which screens moved, worst first.

A diff tool that cries wolf gets ignored in a week, so the run pins everything that moves: frozen clock, no cross-origin requests, no service worker, animations and transitions off, fonts and images awaited, and a wait for two equal text samples because the points total counts up from JavaScript. **Two full captures of one build differ in zero pixels.** A deliberate margin change on one heading surfaced exactly the three affected fixtures at all four widths, nothing else.

The run **fails rather than shoots** when a loader is on camera. A green run of 120 spinners looks identical to a green run of 120 screens.

**3. CI — `e2e` out, `ds-shots` in**

The `e2e` job has reported success without running a test since `271750934` (2026-04-27), which demoted it with `continue-on-error` and the note "TEST_HARNESS_SECRET not in CI env yet". It dies in `globalSetup` before a spec loads.

Adding the secret cannot fix it: `requireTestMode` only accepts a database on localhost, so no hosted API can ever serve `/dev/test-session`.

`ds-shots` replaces it — captures the head, diffs against a baseline captured on pushes to `dev`, and keeps one PR comment up to date with the screens that moved and the images inline. **Advisory: not in `ci-success.needs`, so it never blocks a merge.** The gate stays at about 110 seconds.

**4. The old spec suite**

18 specs, 2,154 lines, running nowhere. Eight contained no assertion at all. One asserted that the URL contained the step it had just typed into the URL.

Four earned their place and are kept. They did not actually run either — with no API the layout holds every screen on the loading mascot, so two failed and three passed against a page that had rendered nothing. `?__fixture=home` gives them a session; they now assert against real content in 11 seconds.

## Coverage this gives up

Stated plainly. No fixture covers the add-money or withdraw country forms, `/claim` link states, `/qr-pay`, `/send/success`, or the public routes outside `(mobile-ui)`. The claim page runs a real crypto check against the link's pubKey, which static fixture responses cannot satisfy — that belongs in the nightly sandbox tier. None of it was running before this PR either.

## Verified

- `typecheck` clean · `npx jest` 302 suites, 3632 tests · `prettier` clean · `ds-lint --check` no metric increased
- Kept specs: 7 tests, 11.3s, green against a production build
- Capture stability re-run after every change: `0 changed, 120 unchanged`
- Screenshots read back by eye, not trusted from a green run — eight of the original 38 fixtures rendered something other than what they claimed and were fixed or deleted

## Follow-ups, not in this PR

- `error-points` and `error-user` fixtures land after the production-fix PR, or they would capture the bugs those fixes remove
- Nothing runs the four kept specs in CI. `test:e2e:regression` takes 8 seconds and needs no secret — cheap to add
- The `/dev` chunks still ship in the production bundle; the routes are blocked, the JavaScript is not


kushagrasarathe:
## /code-review pass — 2026-08-26

Fixed in b70943f47 (typecheck + 3,632 unit tests + all 7 regression tests + a live capture smoke all verified green):
- The fixture cookie writes were origin-wide and unguarded: activate() overwrote a real jwt-token session, and while ?__fixture=off sat in the URL, every API call re-deleted a freshly created real login. Both writes now touch the cookie only when its value is the fixture's own.
- The four kept behaviour specs (setup reload-loop guard, icon-regression, send-link hydration, share-asset capture) ran in NO CI job. They now run inside ds-shots, which already has the preview build and Chromium — 11.5s.
- FROZEN_NOW (2026-06-15) predated the registry's newest timestamp (2026-08-01), so relative times rendered as future dates in every baseline. Now 2026-08-15; baselines recapture on the next dev push.

Found but NOT fixed (advisory, for the author to weigh):
- Server actions bypass the fixture layer (interception is client-only), so a fixture screen whose data flows through serverFetch can hit the real API from CI. Worth checking which of the 30 fixtures depend on server actions.
- Baselines save only on dev/main pushes; a PR based elsewhere falls back to the newest dev/main baseline and diffs against the wrong code line. Fine for the intended dev-based flow, misleading for feature-train PRs (like this one).
- ds-shots failure is escalated by nothing (not in ci-success.needs) — the 'advisory job silently dead for weeks' failure mode this PR describes for the old harness.
- Cleanups suggested, not applied: shared isBackendlessSession() helper for the isDemoMode()||fixture disjunction (3 sites today, ~30 potential), deepMerge reuse from i18n, dedup of the two Playwright configs and the three PR-comment builders.

jjramirezn:
Review notes. Full digest for the whole batch: https://claude.ai/code/artifact/9d565cfc-146b-44d9-98a6-2d76cd081bad

The fixture idea is good and the dev story holds up. `?__fixture=<name>` with no database, no API and no keys is the right primitive. Findings below went through a refute pass; two candidates died and are listed at the end so nobody chases them.

This also conflicts with its base right now, so it needs a rebase before it can merge.

### 1. The visual diff has never run on this PR

`tests.yml` limits the `pull_request` trigger to `main`, `dev` and `develop`. This PR is based on `feat/design-system`, so no `pull_request` run fires. On the `push` run, `ds-shots-filter` and `ds-shots` both show as skipped.

The baseline problem stacks on top. `Save this commit as the baseline` is gated on `github.event_name == 'push'` with the branch allowlist `["dev","main"]`, so nothing ever writes a `feat/design-system` baseline. Actions cache scoping lets a PR read caches from its own ref, its base branch and the default branch only — `main` is the default here, so a `dev`-written baseline is not even reachable from a DS-based PR. The `restore-keys: ds-shots-` fallback then serves whatever `main` entry exists, and every DS change reads as a visual regression.

So the headline feature cannot work on the branch it ships on, and it has not been seen to work in CI. Worth solving before merge — either compute the baseline in-job from the merge-base, or widen the trigger and the allowlist.

### 2. The capture spec never checks that fixture mode is on

`e2e/shots/fixtures.spec.ts:112` is `goto` then `settle(page)` then `screenshot`. Nothing asserts the URL after navigation, and nothing reads the fixture back. The only failure it can produce is a `settle()` timeout.

In CI this is safe — the `Build` step sets `NEXT_PUBLIC_VERCEL_ENV: preview`, and `test:visual` sets it too. Locally it is not. `pnpm test:visual:capture` against a normal build gives `DEV_TOOLS_ENABLED === false`, every protected route redirects to `/setup`, and `/setup` settles (it drops its mascot loader once `sessionChecked` flips). You get 120 screenshots of `/setup` under the fixture filenames and a `no screen moved` comment.

`reuseExistingServer: true` at `playwright.shots.config.ts:79` is unconditional, so a stale `next start` on 3080 gets adopted the same silent way. Gate it on `!process.env.CI`, and assert the URL or read back `peanut_fixture` after `goto`.

Also `FIXTURE_PARAM` is exported from `active.ts` but the spec hardcodes `?__fixture=`. Rename the param and the spec still passes while capturing nothing real.

### 3. Path filter blind spot

The `ds-shots-filter` KEEP regex covers `src/`, `e2e/shots/`, the visual scripts, the shots config and a few build files. It does not list `e2e/flows/` or `playwright.regression.config.ts`. A PR touching only those gets `render=false`, so the four behaviour specs never execute.

To be clear this is **not** a regression — the old `e2e` job had `continue-on-error: true` and `playwright.config.ts` has no `webServer` block, so it pointed at port 3000 with nothing running and could never pass or fail a test. Advisory before, advisory now. But the new filter adds a way to skip the specs silently, so please add the two paths.

### 4. Render-phase side effect in `useWallet`

`src/hooks/wallet/useWallet.ts:214` calls `ensureActiveFixture()` in the render body, not in an effect — the sibling `useDemoBalanceUnits` on the next line is a `useSyncExternalStore`, so the contrast looks unintended. On the first render of a `?__fixture=` URL that call writes `sessionStorage` (`active.ts:25`) and can write `document.cookie` (`active.ts:31`). React does not promise to run render once.

### 5. A real session gets faked for the rest of the tab

Not the direction I first suspected. A user who is **already logged in** and then opens a `?__fixture=` URL keeps their real session — `activate()` skips the cookie write because a cookie is present — while `sessionStorage` pins the fixture and `api-fetch` fakes every response until `?__fixture=off`. Worth a guard or at least a visible banner, given `/dev` tooling is reachable on previews.

### Checked and cleared — please ignore these

- **No hydration mismatch.** `(mobile-ui)/layout.tsx` holds `children` behind `isReady` and returns the mascot loader on both branches until the effect fires, so `useWallet` never runs during SSR for any of the 30 fixture routes. There is no server render of the balance to diverge from.
- **A fixture cannot survive into a real login.** Both cookie writes are guarded, and while a fixture is stored `api-fetch` intercepts every call, so the login response that would mint a real JWT is itself faked. The guards in `b70943f4` are correct.

Minor: the workflow comment says 33/132 screenshots. 30 fixtures at 4 widths is 120.

kushagrasarathe:
Re: [your review](https://github.com/peanutprotocol/peanut-ui/pull/2819#issuecomment-5431114065) — rebased on feat/design-system (merge 404bc6aaa) and all findings addressed in 64bd8dbe4. Every gate re-verified after the merge: typecheck clean, 3,664 unit tests, preview build + 7 regression tests, capture smoke.

1. **Diff never ran on this PR** — took your 'widen the trigger and the allowlist' option: `feat/design-system` added to the pull_request branches, the push branches, and the baseline-save allowlist (commented as the DS release train, remove after it merges to dev). Cache scoping then works: DS-based PRs read the baseline their base branch pushes.
2. **No fixture-mode assertion** — after settle the spec now `expect.poll`s `sessionStorage[FIXTURE_STORAGE_KEY]` to equal the fixture name (key exported from active.ts), and the URL is built from the imported `FIXTURE_PARAM` — rename the param and the spec breaks loudly instead of shooting /setup 120 times. Proven: mutating the expected name fails both widths with the 'fixture mode never engaged' message. `reuseExistingServer` is now `!process.env.CI` in both the shots and regression configs.
3. **Path filter** — `e2e/flows/` and `playwright.regression.config.ts` added to the KEEP regex.
4. **Render-phase write** — added a read-only `peekActiveFixture()` twin (URL→session, zero writes, 'off' reads as null); the useWallet render body uses it. `ensureActiveFixture` (the promoting write) stays for the non-render sites: api-fetch, respond.ts, the kernelClient effect, the websocket callback path.
5. **Invisible fake session** — new `FixtureBanner`: fixed bottom strip 'fixture: <name> — API faked · tap to exit' (plain `<a href="?__fixture=off">`), DEV_TOOLS_ENABLED-gated, mounted once in the (mobile-ui) layout. Side effect worth knowing: the banner is on camera, so all 120 baselines shift once — and every capture now carries visible proof of fixture mode.
6. Minor: 33/132 → 30/120 fixed (plus the same stale 152 in the spec comment).

Thanks for the refute-pass section — the two dead candidates stayed dead.

jjramirezn (approved):

PR #2820, opened 2026-08-26
fix: three production bugs the fixture work surfaced

kushagrasarathe:
## Summary

Three production bugs, found while building the fixture harness in #2819. None was caught by a test, by Sentry, or by review — each surfaced from looking at a real screen or a real log.

Kept apart from #2819 on purpose: that PR is tooling with no production behaviour, this one is ~130 lines that change what users see. If one of these is wrong during launch week, reverting it should not mean reverting a harness.

**Base: `feat/design-system`** (PR #2813).

## 1. Internal dev tooling answered 200 on peanut.me

The intent to block `/dev` on production has been in the tree since February (`6830d7f3e`). It was written in the wrong place, so it never worked.

`(mobile-ui)/dev/layout.tsx` called `notFound()` for everything outside a small allowlist. **Inside a route group that renders the not-found UI and still answers 200**, so peanut.me served the page and its chunks to anyone who asked.

Worse, the pages under `src/app/dev/` sit outside that group and had **no gate at all**. `/dev/kyc-flows` rendered internal KYC flow diagrams to anonymous visitors on production.

The check now runs in the proxy, which sees both route groups, returns a real status, and covers routes nobody has written yet. Measured on a production build: **17 dev routes went 200 → 404**; `full-graph`, `payment-graph` and `safe-area` stayed 200. The allowlist is the one the layout already had — nothing added, nothing removed.

The host decides, not the build. Blocking on the dev-and-preview flag alone would have taken `/dev` off **staging**, which `6830d7f3e` enabled deliberately. peanut.me is the only host that blocks, and a build with no base URL reads as peanut.me, so the gate fails closed. The layout keeps its check because the native build is a static export with no proxy.

A test walks `src/app` for every `page.tsx`, so a dev page added tomorrow is covered without anyone remembering.

## 2. A backend outage sent signed-in users to the signup flow

`(mobile-ui)/layout.tsx` renders `BackendErrorScreen` when the user fetch fails. An effect earlier in the same file matched the same state with no error guard, called `router.replace('/setup')`, and armed a 3-second hard-nav fallback behind it.

Instrumented first paints, before:

```
+3128ms   error screen paints
+7054ms   → /setup        ← the fallback fires
```

During an outage that reads as "you have been logged out", which is the worst thing to tell someone whose money is behind the login.

"No user" has two causes and the data already told them apart: the user query **returns null** for a 401 and **throws** for a 5xx. One condition. After: the screen holds past 11 seconds and the URL never moves. A logged-out visitor still reaches `/setup` on identical timing.

Jest could not import the layout at all — no CSS mapper — so this had never been testable. One line added; `jest-transform-stub` was already a dependency.

## 3. A failed points fetch spun forever

`/rewards` sat on the mascot loader with no way out when `GET /points` failed. The error branch below the loader was unreachable for that path.

`pointsApi.getTierInfo` catches its own failure and resolves with `data: null`, so react-query never reports an error and never retries. The loading guard read `!tierInfo?.data`, which stays true forever once the call has failed. The invites call throws properly, which is why the error branch looked live in review.

The guards now split three states instead of two. A disabled query reports `isLoading` false while both queries wait on `user`, so `isLoading` cannot tell "not started" from "settled and empty". `isPending` can.

The screen was only correct before because a parent covers for it, using a different predicate than the queries use. It is now correct on its own terms.

## Verified

- `typecheck` clean · `npx jest` **304 suites, 3639 tests** · `prettier` clean · `ds-lint --check` no metric increased
- Each fix has one focused spec, and each spec was **proven to fail against the unfixed code** before it passed — including against the original buggy version, not only against a strawman
- Production builds served and probed for 1 and 2; first-paint instrumentation for 2, because a screenshot after settle cannot see a redirect that fires at 7 seconds

## Found and deliberately not fixed

- **Points failures never reach Sentry.** History's error branch calls `captureException`; rewards only `console.error`s, and `getTierInfo` swallows before even that. This is why bug 3 survived
- A failed invites call still blanks the whole rewards screen instead of dropping one section
- `isRedirecting.current` never resets, so a later legitimate redirect in the same mount is suppressed
- The `/dev` chunks still reach production bundles. The proxy blocks the routes, not the JavaScript — that needs a build rule over every `/dev` route

### DIFF
--- a/package.json
+++ b/package.json
@@ -201,6 +201,7 @@
         ],
         "moduleNameMapper": {
             "\\.(svg|png|jpg|jpeg|gif|webp)$": "<rootDir>/src/utils/__mocks__/static-image.ts",
+            "\\.css$": "jest-transform-stub",
             "^@/config/wagmi\\.config$": "<rootDir>/src/utils/__mocks__/wagmi-config.ts",
             "^wagmi/chains$": "<rootDir>/src/utils/__mocks__/wagmi.ts",
             "^next/cache$": "<rootDir>/src/utils/__mocks__/next-cache.ts",
--- a/src/__tests__/dev-routes-gate.test.ts
+++ b/src/__tests__/dev-routes-gate.test.ts
@@ -31,11 +31,8 @@ const ORIGINAL_BASE_URL = process.env.NEXT_PUBLIC_BASE_URL
 async function loadProdBuild() {
     process.env.NEXT_PUBLIC_BASE_URL = 'https://peanut.me'
     jest.resetModules()
-    const [{ proxy, config }, { isBlockedDevRoute }] = await Promise.all([
-        import('@/proxy'),
-        import('@/constants/dev-tools.consts'),
-    ])
-    return { proxy, config, isBlockedDevRoute }
+    const { proxy, config } = await import('@/proxy')
+    return { proxy, config }
 }
 
 afterAll(() => {
@@ -51,14 +48,13 @@ describe('dev routes on peanut.me', () => {
     })
 
     it('404s every dev page except the three allowed ones', async () => {
-        const { proxy, isBlockedDevRoute } = await loadProdBuild()
+        const { proxy } = await loadProdBuild()
 
-        const answering = DEV_ROUTES.filter((route) => {
-            if (isBlockedDevRoute(route)) {
-                return proxy(new NextRequest(`https://peanut.me${route}`))?.status !== 404
-            }
-            return true
-        })
+        // every route goes through the proxy — so the allowlist is asserted
+        // to actually answer, not just skipped by the filter
+        const answering = DEV_ROUTES.filter(
+            (route) => proxy(new NextRequest(`https://peanut.me${route}`))?.status !== 404
+        )
 
         expect(answering.sort()).toEqual(ALLOWED_ON_PROD)
     })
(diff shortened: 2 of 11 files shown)


kushagrasarathe:
## /code-review pass — 2026-08-26

Fixed in 213588f55 (typecheck + 3,639 unit tests green):
- The one finding that mattered: the dev-route lockdown missed src/app/dev/* on the production NATIVE build — a static export where proxy.ts never runs, so /dev/kyc-flows and /dev/loading-words answered in the app. Both route groups now share one gate via shouldBlockDevRoute(), which keeps the flag and predicate together so a future call site cannot use half of it.
- The gate test now routes ALLOWED pages through the proxy too — before, a change that over-blocked /dev/full-graph on prod would have passed the suite.
- The rewards error branch logged a contentless null in exactly the swallowed-error path it was built for; it now logs the settled response.
- routes.ts comment described the pre-PR mechanism; updated.

Declined with reason: DEV_ROUTES_ENABLED inlines its own prod-domain literal (a 5th copy in the repo) because general.consts pulls viem into the edge bundle — consolidating means moving BASE_URL out of general.consts, out of scope here. The review's encoded-path bypass note (%64ev) rests on Next normalizing before matching, which current releases do; an integration probe on a prod build would pin it, also out of scope.

The three original fixes themselves verified clean — no regression found in their happy/error paths.

jjramirezn (approved):

PR #2821, opened 2026-08-26
fix(qr-scanner): recover the camera when the app returns to foreground

innolope-dev:
Closes TASK-21862 — https://app.notion.com/p/3c78381175798110b8bfd9529ac52418

## Problem

An iPhone user opened the QR scanner, backgrounded the app for 94 seconds, and came back to a permanent "Starting camera…" spinner. iOS had reclaimed the capture device, so the stopped scanner's own `start()` rejected with `Camera not found.` — even though the same session had started that camera successfully a minute earlier.

`useQRScanner`'s visibilitychange handler called `scanner.start()` raw and its `.catch` only did a `console.error`. `isCameraReady` stayed `false` with no `error` set, which is exactly the state `index.tsx` renders as the indefinite spinner. The only escape was closing the drawer.

Sentry: [PEANUT-UI-SV1](https://peanut-c34d84c05.sentry.io/issues/7692395912/) — 1 event / 1 user, iOS 18.7, production release `043991d`. No money moved, failed, or got stuck.

## Fix

- Foreground resume now goes through `startCamera()` rather than `scanner.start()`. That path rebuilds the scanner after the iOS hardware settle delay and, when the restart still fails, sets the existing error state instead of leaving a bare spinner.
- New `isStartingCameraRef` guard in `startCamera` so overlapping visibility events — or a retry racing a resume — can't start two cameras at once. (This is the whole diff in `useQRScanner.ts`: the body is wrapped in `try/finally`, so review with `?w=1`.)
- The visibility listener now reads `startCamera` through a ref and subscribes once, instead of resubscribing per `isScanning` change.
- `ErrorView` gained a Retry button next to Close, so a failed restart reaches retry / paste / close rather than close-only. Uses the existing `common.retry` key — no new strings.

## Tests

`useQRScanner.test.ts` gains a resume suite: successful restart returns to the live camera; a rejected restart lands on `cameraNotFound` instead of spinning; a retry after that failure recovers; three synchronous resume events produce one `start()`. Three of the four fail against the pre-fix hook.

## Verification left for QA

Definition of done includes a manual pass on iOS Safari/PWA: open the scanner, background the app for at least a minute, return, and scan a QR.

### DIFF
--- a/src/components/Global/QRScanner/__tests__/useQRScanner.test.ts
+++ b/src/components/Global/QRScanner/__tests__/useQRScanner.test.ts
@@ -8,6 +8,8 @@ import { captureException } from '@sentry/nextjs'
 import { useQRScanner } from '../useQRScanner'
 
 let onDecode: (result: { data: string }) => void
+let startBehaviour: () => Promise<void> = () => Promise.resolve()
+let startCalls = 0
 
 jest.mock('qr-scanner', () => ({
     __esModule: true,
@@ -15,7 +17,10 @@ jest.mock('qr-scanner', () => ({
         constructor(_video: HTMLVideoElement, cb: (result: { data: string }) => void) {
             onDecode = cb
         }
-        start = jest.fn().mockResolvedValue(undefined)
+        start = jest.fn(() => {
+            startCalls++
+            return startBehaviour()
+        })
         stop = jest.fn()
         destroy = jest.fn()
         setCamera = jest.fn()
@@ -64,4 +69,85 @@ it('camera scan: onScan failure is captured with the qr_scan_processing tag', as
             extra: { qrLength: 100, qrKind: 'pix', qrPayload: PIX_PAYLOAD },
         })
     )
+})
+
+describe('resume after backgrounding (TASK-21862)', () => {
+    function setHidden(hidden: boolean): void {
+        Object.defineProperty(document, 'hidden', { value: hidden, configurable: true })
+        document.dispatchEvent(new Event('visibilitychange'))
+    }
+
+    async function mountScanning() {
+        const { result } = renderHook(() => useQRScanner(jest.fn(), undefined, true))
+        const videoRef = result.current.videoRef as React.MutableRefObject<HTMLVideoElement | null>
+        videoRef.current = document.createElement('video')
+        await act(async () => {
+            await result.current.retryCamera()
+        })
+        // the mount-time start ran before the <video> existed and left a
+        // 100ms element-retry timer behind; let it fire before asserting
+        await act(async () => {
+            await new Promise((resolve) => setTimeout(resolve, 150))
+        })
+        return result
+    }
+
+    afterEach(() => {
+        Object.defineProperty(document, 'hidden', { value: false, configurable: true })
+        startBehaviour = () => Promise.resolve()
+    })
+
+    it('a successful restart returns to the live camera', async () => {
+        const result = await mountScanning()
+        expect(result.current.isCameraReady).toBe(true)
+
+        await act(async () => setHidden(true))
+        expect(result.current.isCameraReady).toBe(false)
+
+        await act(async () => setHidden(false))
+        expect(result.current.isCameraReady).toBe(true)
+        expect(result.current.error).toBeNull()
+    })
+
+    it('a rejected restart shows the error view instead of an endless spinner', async () => {
+        const result = await mountScanning()
+
+        await act(async () => setHidden(true))
+        startBehaviour = () => Promise.reject(new DOMException('Camera not found.', 'NotFoundError'))
+        await act(async () => setHidden(false))
+
+        expect(result.current.isCameraReady).toBe(false)
+        expect(result.current.error).toBe('qrScanner.cameraNotFound')
+    })
+
+    it('a retry after a failed restart recovers the camera', async () => {
+        const result = await mountScanning()
+
+        await act(async () => setHidden(true))
+        startBehaviour = () => Promise.reject(new DOMException('Camera not found.', 'NotFoundError'))
+        await act(async () => setHidden(false))
+        expect(result.current.error).toBe('qrScanner.cameraNotFound')
+
+        startBehaviour = () => Promise.resolve()
+        await act(async () => {
+            await result.current.retryCamera()
+        })
+        expect(result.current.error).toBeNull()
+        expect(result.current.isCameraReady).toBe(true)
+    })
+
+    it('overlapping resumes do not start the camera twice', async () => {
+        const result = await mountScanning()
+        const before = startCalls
+
+        await act(async () => setHidden(true))
+        await act(async () => {
+            setHidden(false)
+            setHidden(false)
+            setHidden(false)
+        })
+
+        expect(startCalls - before).toBe(1)
+        expect(result.current.isCameraReady).toBe(true)
+    })
 })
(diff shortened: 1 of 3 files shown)


innolope-dev:
@coderabbitai review

abalinda:
@innolope-dev — ran `/code-review` on this branch (head `8cdb561`). The ticket's symptom is fixed on the happy path, but the new `isStartingCameraRef` guard reintroduces the black-video / silent-spinner state on a realistic iOS gesture. Findings, most severe first.

### Blocking

**1. `useQRScanner.ts:229` — the guard drops a resume that lands mid-start, and the in-flight start then "succeeds" with no stream.**
Scanner opens; user backgrounds within the first ~1s (iOS 200ms settle + getUserMedia, or the 5s PWA race). `hidden` → `scannerRef.stop()` sets qr-scanner `_active=false`. `visible` → `startCamera()` returns at 229 because the flag is still set. The in-flight `start()` then resolves normally, but qr-scanner discards the stream when `!_active` (`_stopVideoStream`) and skips acquisition if `document.hidden`. Line 329 still runs `setIsCameraReady(true)` → spinner gone, black `<video>`, no decode, no error view, no retry. qr-scanner's own visibilitychange handler (`_active && this.start()`) is disarmed too. Pre-PR, the unconditional `scanner.start()` re-armed `_active` and recovered.

**2. `useQRScanner.ts:414` — resume now fires in the permission-denied / error state.**
The old guard needed `scannerRef.current`, which `cleanup()` nulls in the catch. Now every foreground while the CameraPermissionModal is up runs `setError(null)` / `setIsPermissionDenied(false)` (233-234) → modal unmounts, spinner shows, `<video>` remounts, getUserMedia hangs on the WKWebView denial for `CAMERA_START_TIMEOUT_MS` (5s), then the modal comes back. The paste-Pix escape hatch is unreachable for 5s per foreground, and each cycle emits `console.error('Error accessing camera')` → Sentry via captureConsoleIntegration. On Capacitor, `ensureNativeCameraPermission` re-runs on every foreground while denied.

**3. `useQRScanner.ts:297` — `retryCamera` is the guarded `startCamera`, so Retry is a silent no-op while a start is in flight.**
On Capacitor the native `scanner.start()` has no timeout race, so a hung getUserMedia pins the flag until unmount — `cleanup()`, `close()`, the hidden branch and the unmount effect never clear it. Pre-PR a Retry tap reached `cleanup()` and rebuilt the scanner. On web the dead-button window is the 5s race.

### Should fix

**4. `:422` — resume and `toggleCamera` don't guard each other.** A foreground during the `setCamera` handover (hundreds of ms on Android WebView per the comment at 373) rebuilds with the stale `facingMode` closure; `setCamera` resolves silently (`_restartVideoStream` skips when `!_active`) and 383-384 commit the *new* facingMode. State says new camera, stream is the old one, mirror/toggle inverted from then on. Cheapest fix: capture `const scanner = scannerRef.current` before the await in `toggleCamera` and bail if it changed.

**5. `index.tsx:360` — CameraPermissionModal still passes `retryCamera` directly**, so the MouseEvent becomes `preferredCamera` and qr-scanner builds `deviceId: {exact: MouseEvent}` — three failing getUserMedia attempts, then the browser default camera (often front-facing). You fixed the identical wiring on ErrorView at 362 but not here. Pre-existing, but the root-cause fix is one line in the hook: `retryCamera: () => startCamera()` at :462 — covers both callers and lets 362 go back to `onRetry={retryCamera}`.

**6. `:283` — `startCamera` never re-checks `isScanningRef` after its awaits.** A foreground inside the 300ms deferred `close()` window runs the rebuild on a closing component: `new QrScannerLib(null, …)` throws into the catch at 331 → Sentry noise + `setError` on a closed scanner. Pre-existing hole, but resume now routes through it.

**7. `:185`, `:195` — `handleQRScan`'s two resume sites keep the bare `scannerRef.current?.start()`** with no catch and no fallback. Same iOS device-reclaim mechanism this PR names, still unhandled at 2 of 3 call sites (slow `onScan` + phone call / Siri / background cycle → frozen last frame, no error view).

### Simplification

**8. `:425` — every foreground now does full `cleanup()` + new QrScannerLib (new decode worker) + Capacitor permission check + iOS 200ms delay + timeout race, on all platforms.** The old `scanner.start()` was one call and worked everywhere except the iOS device-reclaim case. Cheaper shape that still fixes the ticket and keeps the error view on failure:

```ts
scannerRef.current.start()
  .then(() => isScanningRef.current && setIsCameraReady(true))
  .catch(() => startCameraRef.current())
```

The rebuild runs only when `start()` actually rejects, and the common path never enters `startCamera` — which also removes most of the surface of 1, 2, 4 and 6.

**9. `:404` — `startCameraRef` + its sync effect exist only so the visibility effect can have `[]` deps.** Keeping `[startCamera]` as the dep re-subscribes a document listener only when facingMode/deviceType/t change, which is free. And the outer try/finally re-indents ~120 unchanged lines; a 6-line `busy` wrapper around an untouched inner function gives the same guarantee with an ~8-line diff instead of 247.

**10. `__tests__/useQRScanner.test.ts:82` — `mountScanning` relies on a real 150ms sleep vs the 100ms video-element retry timer.** 50ms margin on a loaded CI runner → the 'overlapping resumes' test reads a delta of 2. `startCalls` is module-level and never reset. Use `jest.useFakeTimers()` + `advanceTimersByTime(CONFIG.VIDEO_ELEMENT_RETRY_DELAY_MS)`, or set `videoRef` before the first act so no retry timer schedules.

### Suggested path

Take 8's shape (plain `start()`, rebuild only on rejection) plus a generation counter re-checked after each await in `startCamera` (`if (gen !== startGenRef.current || !isScanningRef.current) { cleanup(); return }`) — that closes 1, 2, 4 and 6 in one change. Add the one-line `retryCamera: () => startCamera()` for 5, then re-run the resume suite with fake timers.

innolope-dev:
Thanks @abalinda — findings 1, 2, 3 and 5 reproduce, and 1 is worse than the spinner: `start()` *resolves* in that path. From qr-scanner 1.4.2's own source:

```js
const { stream, facingMode } = await this._getCameraStream();
if (!this._active || this._paused) {
    QrScanner._stopVideoStream(stream);
    return;          // resolves, no stream attached
}
```

So the guard-dropped resume left `_active` false, the in-flight start returned normally, and line 329 reported ready over a dead `<video>`.

Pushed `402ca25` — took your suggested path.

**Guard removed.** Resume is back to `scannerRef.current.start()` and only falls back to a full `startCamera()` rebuild when that rejects. That re-arms the instance the hidden branch stopped (closing 1), keeps the common foreground off the decode-worker rebuild (8), and un-guards `retryCamera` (3). Also drops the 120-line re-indent (9) — `useQRScanner.ts` is now +77/-19.

**Generation counter** in `startCamera`, re-checked after the permission call and the iOS settle delay, plus an `isScanningRef`/`videoRef` check at the same point — covers 4 and 6.

**2** — the resume branch requires a live `scannerRef.current` again, which `cleanup()` nulls in the catch, so a foreground no longer re-enters the flow behind the permission modal.

**5** — fixed at the root as you suggested: `retryCamera: () => startCamera()`. ErrorView goes back to `onRetry={retryCamera}` and CameraPermissionModal is fixed by the same line.

**7** — both `handleQRScan` restarts get `.catch(() => startCameraRef.current?.())`.

One addition beyond the list: a start that resolves *without* a stream now fails closed to the error view (`cameraStartFailed`, already translated in all four catalogs) instead of dismissing the spinner. That makes the finding-1 class unreachable regardless of ordering, rather than relying on the re-arm.

**10** — fake timers, `startCalls` reset per test, and the mock now models the `_active` / `srcObject` contract the hook depends on. Suite went 160ms → ~10ms per test. Three new tests cover 1, 2 and 5; against `origin/dev` 4 of 8 fail, and against the previous head the three findings-driven ones fail.

Worth noting `overlapping resumes do not start the camera twice` now passes on `origin/dev` too — with a faithful mock, `if (this._active && !this._paused) return` already dedupes, which is your point that the guard was never load-bearing there. Kept it as a contract test.

Manual iOS pass from the DoD is still outstanding.

PR #2822, opened 2026-08-26
fix(startup): survive a null or throwing localStorage

innolope-dev:
Closes TASK-21835 — https://app.notion.com/p/3c68381175798174b75effb4b74fb0fd

## Problem

Restricted documents — Android in-app browsers (Twitter's, Samsung), Chrome with site data blocked — expose `window.localStorage` as `null`, or throw `SecurityError` from the property getter itself. Two startup paths reached for it unguarded:

**Locale detection ([PEANUT-UI-STC](https://peanut-c34d84c05.sentry.io/issues/7690378127/))** — `resolveStartupLocale` checked `typeof localStorage !== 'undefined'`, and `typeof null === 'object'`, so a null storage sailed through the guard and `getItem` threw. `localeReady()` then memoized the rejected promise, and `AppIntlProvider` awaits it with no rejection handler — so a supported-language visitor stayed in English for the whole session and `app_locale` analytics never fired.

**Wagmi config ([PEANUT-UI-STF](https://peanut-c34d84c05.sentry.io/issues/7690983802/))** — `@wagmi/core@2.19.0`'s `getDefaultStorage()` reads `window.localStorage` with no try/catch, and only wraps `setItem`. Our `createConfig` runs at module scope, so a throwing getter took the app shell down before first render.

Both are 1 anonymous production actor each, 0 transaction attempts, $0 exposure.

## Fix

New `src/utils/safe-storage.ts` guards the property access *and* every storage method, degrading to "no stored value":

- `locale-store` reads and writes through it, so a hostile storage falls through to the browser-language default. `localeReady()` also catches now — memoizing a rejection is what turned a one-off read failure into a session-long wedge.
- `wagmi.config` passes `createStorage({ storage: resilientWebStorage })` instead of relying on wagmi's default. Key prefix is unchanged (`wagmi`), so existing persisted connector state still loads.

## Tests

`safe-storage.test.ts` covers the healthy round-trip, a null `localStorage`, a getter that throws `SecurityError`, a host with no `localStorage` at all (SSR), and storage methods that throw once reached. `locale-store.test.ts` gains a `localeReady` suite: browser-language fallback under a null storage, stored-locale precedence, and the memoization now caching a usable locale rather than a rejection. The null-storage and memoization cases fail against the pre-fix module.

One note on the `webStorage(host)` seam: jest's jsdom global swallows throwing property getters and yields `undefined`, so the `SecurityError` branch is unreachable through `window` and needs an explicit host object to exercise.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved app startup reliability when browser storage is unavailable, blocked, or throws errors.
  - Added graceful fallback to the browser’s language when saved locale data cannot be accessed.
  - Preserved locale preferences more reliably without allowing storage failures to interrupt the app.
- **Tests**
  - Added coverage for unavailable, restricted, and error-prone browser storage scenarios.
  - Added tests for locale fallback, saved-locale precedence, and startup resolution behavior.

### DIFF
--- a/src/config/wagmi.config.tsx
+++ b/src/config/wagmi.config.tsx
@@ -1,6 +1,7 @@
 'use client'
-import { WagmiProvider, cookieToInitialState, createConfig, http, type Config } from 'wagmi'
+import { WagmiProvider, cookieToInitialState, createConfig, createStorage, http, type Config } from 'wagmi'
 import { arbitrum, bsc, celo, gnosis, linea, mainnet, optimism, polygon, scroll, worldchain } from 'wagmi/chains'
+import { resilientWebStorage } from '@/utils/safe-storage'
 
 // Base intentionally absent: WAGMI's `http()` (no URL) defaults to mainnet.base.org,
 // which IP-rate-limits us with 403s and pollutes the console. We don't use Base for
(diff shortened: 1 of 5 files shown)


innolope-dev:
@coderabbitai review

PR #2825, opened 2026-08-26
ci: resolve release versions as <major>.<build>.<ota> instead of typing them

innolope-dev:
## What

Versions become `<major>.<build>.<ota>`:

| component | means | moves when |
|---|---|---|
| `major` | app generation | a deliberate major upgrade — `package.json` `version`, the only place it is read from |
| `build` | native build counter | every store build; resets `ota` to 0 |
| `ota` | OTA counter within that build | every OTA on that binary; `.0` is the JS baked into it |

So `1.7.0` is the seventh store build and `1.7.3` is its third OTA. Two properties that used to be maintained by hand fall out of the shape: an OTA always sorts strictly above the binary it targets, and the version alone says which binary a bundle belongs to. A hand-picked number getting that ordering wrong is what left 102 devices refusing every update for a month (TASK-21793).

Nobody types a number any more. `scripts/release-version.mjs` resolves them from git tags plus the Capgo channel.

## Why the registry is tags and not a file

`dev` and `main` both carry a `pull_request` ruleset with **zero bypass actors**, and all branches require signed commits. No actor — PAT, `GITHUB_TOKEN` or admin — can push a version-bump commit to either, so `package.json` cannot be the auto-bumped source of truth. Tags have no ruleset, so CI can write those.

`package.json` `version` is now only the major anchor and the local-build fallback; it is fine for it to lag behind what ships.

## How a release is cut

Two `workflow_dispatch` buttons, both refusing any ref but `dev`:

| button | does |
|---|---|
| **Release Native** | resolves `<build+1>.0` → builds iOS + Android from that one number → TestFlight + Play `internal` → tags `v<version>` |
| **Release OTA** | resolves `<ota+1>` off the production channel → uploads the bundle → tags `ota-<version>` |

Neither is reachable from a push — merging to `dev` still only ships to the staging channel. They are the same deliberate act as the `git tag … && git push` they replace, minus the hand-picked number.

Driving both platforms from one resolution is deliberate: dispatching the two build workflows separately would let them land on different numbers for the same release.

The tag is written **after** the build, as the record of what shipped, so a failed run leaves no tag and a re-run resolves the same number instead of burning one. CI tags with `GITHUB_TOKEN` on purpose — GitHub does not start workflows from a push made with it, so the tag records the release without re-triggering the build it just finished. A PAT there would double-build.

## Verified

Simulated against the repo's real tag list:

```
Release Native      -> 1.1.0        (no seeding: v1.0.0 is the only v1.x.0 tag)
Release OTA         -> 1.1.1
Release OTA again   -> 1.1.2
staging on dev push -> 1.1.<commit count>
Release Native      -> 1.2.0
Release OTA         -> 1.2.1
```

Every ordering constraint holds through the migration, including `1.1.0 > 1.0.54` (today's newest production bundle) and `1.1.0 > 1.0.53` for Apple's marketing-version comparison. 15 new tests; all 45 script tests green.

## Behaviour changes to know about

- **The build workflows no longer fall back to `package.json`.** Dispatching one with no `versionName` now errors instead of quietly shipping whatever the file says. iOS gets `IOS_MARKETING_VERSION` passed into postsync so the Xcode project is stamped from the resolved version.
- **A bare manual dispatch of `capgo-deploy` to `production` now errors.** It would have uploaded a staging-shaped version and dragged production's OTA counter into the commit-count band, where the next resolved version collides with a bundle that already exists. Use **Release OTA**.
- **Tag pushes stay as break-glass** and now validate against the scheme. That check is what a glob cannot do: `v*` cannot reject `v2026.02.26` (a real tag on `main`) because it is `X.Y.Z` shaped, so only the major comparison catches it.
- **Staging keeps the commit count as its `ota` component** but now rides the current build number. Anchored to `package.json` it would have sorted below the first `1.1.0` binary and gone dead. The count itself is what keeps the two lanes from colliding in the shared bundle namespace, where a collision no-ops under `--version-exists-ok` and ships nothing.

## Note on `main`

`on: push: tags:` is not branch-scoped, so a `v*` tag on `main` starts both native builds today exactly as one on `dev` does. The dispatch path closes that for normal releases by pinning the ref to `dev`; the break-glass tag path keeps it, now with the version validated.

## Summary by CodeRabbit

* **New Features**
  * Added manual workflows for native and over-the-air releases.
  * Added reusable release workflows with configurable versions, channels, tracks, and debug options.
  * Added automated version resolution, validation, and post-release tagging.
* **Bug Fixes**
  * Improved release version handling by rejecting missing, invalid, or inconsistent versions.
  * Ensured iOS marketing versions use the CI-resolved release version.
* **Documentation**
  * Updated release procedures, versioning rules, workflow triggers, approvals, and tagging guidance.
* **Tests**
  * Added end-to-end coverage for native, OTA, staging, and validation scenarios.


innolope-dev:
@coderabbitai review

innolope-dev:
Addressed all three findings in 2487dcee2.

**Major — one production mutex.** Real, and the failure mode is exactly the one this PR exists to prevent. `release-native` and `release-ota` now both hold `production-release` across the whole resolve-through-publish span; scoping it to the publish alone would still let a version resolved against a stale channel win the race.

Break-glass paths join the mutex themselves, keyed off `github.event_name` — `push` only ever means a tag on these workflows, and a *called* run reports the caller's `workflow_dispatch`, so a callee never queues behind the parent already holding the mutex. On the `v*` tag path that serialises iOS and Android, which is the price of not racing a concurrent OTA. A direct dispatch of a leaf workflow is indistinguishable from a called one and stays outside the mutex — noted in the file as the deepest break-glass layer.

While in that block I fixed a second, unrelated hazard: `capgo-deploy`'s concurrency group ignored the channel and cancelled in progress, so a routine merge to `dev` could cancel a production write mid-flight. The channel is now part of the group and a production run never cancels.

**Minor — manual-dispatch contract** and **Minor — production dispatch docs.** Both correct; the input descriptions and the OTA table row now match the enforced behaviour.

PR #2828, opened 2026-08-26
feat(invites): emit invited_by in invite links, keep code as an alias

0xkkonrad:
## What

Invite links now emit `/invite?invited_by=<username>` instead of `/invite?code=<username>`. The `code` param stays supported forever as a read-side alias — every previously shared link, QR code, X post, and campaign URL keeps its exact behavior.

## Why

The value in the param *is* the inviter's username, so the param should say so. `code` stays because years of shared links carry it, and because the code slot also transports non-username values (`offramp`, legacy `INVITESYOU` suffixes, campaign slugs).

## How

- New `inviteCodeFromParams()` in `invite-code.utils.ts` is the single reader: `params.get('code') ?? params.get('invited_by')`. **Legacy `code` wins when both are present**, so a pre-existing link keeps its behavior even with the new param appended — the alias is purely additive.
- Readers switched to it: `InvitesPage`, `/setup` deep-link cookie persistence, native `openDeepLink` (`useNativeAppLinks`).
- Writers now emit `invited_by`: `generateInviteCodeLink()`, `inviteFlowUrl()`, and the PublicProfile guest-door navigation.
- No API change: `/invites/validate` and `/invites/accept` still receive the same bare-username code in the body. `INVITE_CODE_TO_CAMPAIGN_MAP` / badge-campaign resolution are untouched — they key off the resolved value, not the param name.
- Analytics: `link_type: 'invite_code'` wire value unchanged (registered constant, PostHog continuity).

## Tests

- New unit suite for `inviteCodeFromParams` (both params, code-wins precedence, absent case).
- Emitted-shape assertions updated (`general.utils`, `invite-flow-url`, `PublicProfile`, `BadgeDetailModal`).
- Legacy `?code=` deep-link tests kept as-is and now prove the alias; added `invited_by` variants for native route mapping and `InvitesPage`.
- 8 suites / 205 tests green; `tsc --noEmit` clean.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -348,13 +348,14 @@ let nextConfig = {
     },
     async redirects() {
         // Campaign/invite links on the root domain (peanut.me/?campaign=skip,
-        // peanut.me/?code=alice) hand off to /invite, which owns the claim flow.
+        // peanut.me/?invited_by=alice, legacy peanut.me/?code=alice) hand off to
+        // /invite, which owns the claim flow.
         // Query params carry over automatically. Deliberately NOT keyed on
         // utm_campaign — that would hijack ordinary marketing links to the
         // landing page; utm_campaign only resolves to a badge once on /invite.
         // Active in development too (unlike the locale redirects) so the flow is
         // locally testable.
-        const campaignRedirects = ['campaign', 'campaignTag', 'code'].map((key) => ({
+        const campaignRedirects = ['campaign', 'campaignTag', 'invited_by', 'code'].map((key) => ({
             source: '/',
             has: [{ type: 'query', key }],
             destination: '/invite',
--- a/src/__tests__/next-config-invite-redirects.test.ts
+++ b/src/__tests__/next-config-invite-redirects.test.ts
@@ -1,0 +1,32 @@
+// Root-domain invite links (peanut.me/?invited_by=alice) must hand off to
+// /invite like the legacy peanut.me/?code=alice shape does — the landing page
+// ignores the param, so a missed key silently drops attribution.
+//
+// Loaded under NODE_ENV=development: that branch exports the plain config
+// (no Serwist/Sentry wrappers) and redirects() returns exactly the campaign
+// hand-offs, which is all this test is about.
+
+async function campaignRedirects(): Promise<Array<Record<string, unknown>>> {
+    const previousEnv = process.env.NODE_ENV
+    process.env.NODE_ENV = 'development'
+    try {
+        let config: { redirects(): Promise<Array<Record<string, unknown>>> } | undefined
+        jest.isolateModules(() => {
+            config = require('../../next.config.js')
+        })
+        return await config!.redirects()
+    } finally {
+        process.env.NODE_ENV = previousEnv
+    }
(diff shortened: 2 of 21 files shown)


0xkkonrad:
## Adversarial review — 26 Aug

Verified on this PR's Vercel preview + prod curls; full grep for every other `code` reader/writer across peanut-ui (dev + main), peanut-api-ts, mono content/skills.

**Fixed in de3cfc1e1 (this branch)**
1. **P0** `src/app/invite/page.tsx` `generateMetadata` read only `searchParams.code` → every new `?invited_by=` link unfurled as "Invites | Peanut / Invalid invite link" with the generic image (preview: `/invite?invited_by=squirrel` vs `?code=squirrel`). Now uses `inviteCodeFromParams`; test added.
2. **P1** `next.config.js` root-domain hand-off keyed on `campaign|campaignTag|code` → `peanut.me/?invited_by=alice` was a 200 landing page with the param dropped. Added `invited_by`; test added.
3. Nits: empty `?code=` falls through to `invited_by` (`||` instead of `??`); stale `?code=` comments.

**Fixed elsewhere**
- **P2** `seo-route-policy.js` (main-only) crawl-allowed only `/invite?code=` → new shape crawl-blocked, Google can't see the noindex header → #2835 (hotfix → main).
- **P1** peanut-api-ts `POST /redirects/claim` extracted the invite code from the client `targetUrl` via `?code=` only; `qr/[code]/page.tsx` now sends `?invited_by=` → every new QR claim would have fallen back to `<USER>INVITESYOU` → peanutprotocol/peanut-api-ts#1431 (→ dev). **Merge order: api#1431 before this.**
- telly-crm creator link template → `invited_by` (mono 6ee64a08).

**Not changed, noted**
- Old native bundles read only `code` (universal links claim `/invite`, static export). Capgo OTA on push to dev covers it; neither store listing is public yet, so exposure is testers-only until their next OTA pull.
- mono `content/*` CTAs still use `invite?code=SQUIRRELINVITESYOU` — works via the alias, left alone.
- PostHog insights filtering `$current_url` on `code=` (if any) go blind for new links — not checked.

innolope-dev:
@0xkkonrad — this one is `CONFLICTING` against `dev` (single conflict, `src/components/Profile/components/PublicProfile.tsx`) so it needs a `dev` merge before it can go in. CI is otherwise green and there are no open review threads.

Worth doing soon rather than later: [#2876](https://github.com/peanutprotocol/peanut-ui/pull/2876) brings the design system to `dev`, and it rebuilds `PublicProfile.tsx` — so resolving against today's `dev` is a much smaller job than resolving against it afterwards.

No migration work needed on this branch either way: it's the only feature PR in the current queue with zero DS debt — no legacy palette classes, no stock text sizes. Once the conflict is resolved it can merge straight away, and the integration branch absorbs it.

0xkkonrad:
Closing per TASK-22850. Decision: keep `?code=` as the invite param. The rename is cosmetic and this branch now conflicts with the setup stepper and invites refactors. Any remaining crawlability work is tracked in a new task.

PR #2834, opened 2026-08-26
feat: reimagined registration and onboarding (residence step, Unlock payments, home checklist, region-restricted screens)

innolope-dev:
> **Replaces #2790.** Identical content — same tree hash `4ec38cf`, all 42 commits (both merges preserved), dates and messages unchanged — re-authored under a human account so the `human-authors` check passes. This had to be a new branch because `non_fast_forward` is enforced repo-wide with no bypass actors, so #2790's branch could not be rewritten in place.
>
> The `unit` failure on #2790 is already fixed by the head commit (it regenerated the four marketing subsets that were missing `common.userAvatarAlt`); that run predates the fix.

---

Frontend half of the registration/onboarding rework. BE: peanutprotocol/peanut-api-ts#1429 (**merge that first** — every dependency here degrades gracefully without it, but the full experience needs the residence endpoints). **Supersedes #2778 and #2775** (their three commits are merged here unchanged — both can be closed when this merges).

## Signup flow
- **`signup_step_viewed` analytics + URL step mirroring** — every setup step fires a funnel event (`screen_id`, `step_index`, `nav_type`), and the active step mirrors to `?screen=` via shallow History API so browser/hardware Back walks the steps instead of ejecting the user from `/setup`. The URL is a mirror, never a source of truth: reload still routes through `determineInitialStep`, and points of no return (post-passkey) neutralize popstate.
- **New residence step** between username and passkey: legal-residence question ("the country your ID shows, not where you're travelling"), geo prefill as a suggestion only, "Have documents from more than one country?" second selector, and honest pre-account heads-ups — full restriction (CN/IR/RU/BY/KP/SY/CU/HK/GB: generic copy, "Continue anyway" or email capture) and partial restriction (card-only per Rain's list, banking-only per Bridge's). The selector supplements the add-money country list so sanctioned-country residents can answer truthfully. Answer persists to the API post-registration.
- Restriction tiers are consumed from `GET /config/residence-restrictions` with the bundled mirror as instant fallback.

## Unlock payments (replaces Unlocked Regions)
- Regions survive as presentational groups; every row is a concrete product with a live status chip (Active / Always on / Unlock / Processing / Not available). Everywhere leads (Peanut-to-Peanut payments, the card as its own row), the user's region floats up, LATAM never appears (it stays an internal KYC bucket), and the rest-of-world dead end is gone.
- **The Europe-tap → /card hijack is dead by construction** — a bank-method tap can only start bank KYC (pinned by test). All existing KYC modal machinery carried over unchanged; `UnlockedRegions.view.tsx` left untouched on disk.
- Residence anchor row (declared/verified) with the **residence-change flow**: declared saves immediately; a pick differing from the verified residence offers explicit "Save &amp; re-verify now" over the existing restart-identity primitive. Method-worded unlock sheet (`UnlockMethodModal`).

## Home
- **3-item getting-started checklist** replaces the happy-path funnel card: Create your account (pre-checked) · Add money (label follows residence — PIX/SEPA/SPEI/bank — with the honest KYC cost while unverified) · Get your Peanut card when eligible, otherwise Make your first payment. Hides itself when done. Interrupt cards (provider rejection, email block, and the region-restricted terminal explanation, which outranks the checklist) keep their dedicated rendering.
- Fully restricted residences no longer see the "Unlock payments" CTA; the card launch banner gates on residence too.

## Region-restricted rejection screens (from #2778, unchanged)
- The terminal "We can't verify IDs from your country" ending across all four surfaces, no fake retry, no support dead end. See #2778 for the full design discussion; its ActivationCTAs changes are semantically merged with the checklist (region-restricted outranks it, pinned by test).

## Localization
- All new copy in en, es-419, es-AR, pt-BR, with the duplicate-string drift guard satisfied (reused strings adopt each locale's existing translations, including fallback-chain resolution for es-AR).

## Verification
- 578 tests pass across Home, Kyc, Profile, hooks, and i18n suites post-merge, including 60+ new tests (residence step, unlock-payments derivation, checklist, residence-change modal, restriction sets, URL sync hook). Typecheck and lint clean for all touched files (pre-existing content-submodule/asset-declaration errors in this environment are unrelated).
- Deploy-order safe: without the BE PR, the residence anchor hides, restrictions fall back to the bundled mirror, and the region screens simply never render.


innolope-dev:
@coderabbitai review

PR #2835, opened 2026-08-26
hotfix(seo): crawl-allow /invite?invited_by= like /invite?code=

0xkkonrad:
## What

Adds `/invite?invited_by=` to `GOOGLE_DEINDEX_CRAWL_ALLOW_PATHS` next to `/invite?code=`, with tests.

## Why

peanut-ui emits `/invite?invited_by=<username>` once #2828 lands (`?code=` stays a permanent read alias). This policy is main-only, so #2828 can't touch it. Without it, prod robots.txt is `Allow: /invite?code=` / `Disallow: /invite`: the new URL shape is crawl-blocked, Google can't read its `X-Robots-Tag: noindex`, and shared links get indexed as URL-only entries — the problem this exception was added for.

`X-Robots-Tag` itself already covers both shapes (prefix-based; checked on prod).

## Tests

`seo-route-policy.test.ts` + `robots.test.ts` green locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated search engine handling for invite links using the `invited_by` query parameter.
  * Invite links with valid tracking parameters can now be crawled appropriately, while protected routes remain excluded from indexing.

### DIFF
--- a/src/constants/__tests__/seo-route-policy.test.ts
+++ b/src/constants/__tests__/seo-route-policy.test.ts
@@ -74,12 +74,15 @@ describe('SEO route policy', () => {
         expect(receivesNoindexHeader(route)).toBe(false)
     })
 
-    it.each(['/home', '/home/activity', '/invite?code=abc&lid=campaign', '/api/og?type=send&username=alice&amount=10'])(
-        'renders a noindex header for %s through Next route matching',
-        async (route) => {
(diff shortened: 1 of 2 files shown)


0xkkonrad:
Closing per TASK-22850: #2828 is closed and invite links keep `?code=`, so there is no `invited_by` shape to crawl-allow.

PR #2839, opened 2026-08-26
ci: stop re-judging authorship of commits already merged to main/dev

innolope-dev:
## The bug

`human-authors` judges **every commit the PR lists**. A back-merge lists every commit `main` has that `dev` lacks — so one bot-authored commit already on `main` fails the gate forever.

That is where [#2830](https://github.com/peanutprotocol/peanut-ui/pull/2830) is now:

```
aa6db97: authored by chip-peanut-bot[bot] — bot-authored commits are not allowed; commit as yourself
```

`aa6db97` (`chore: retrigger OTA build`, an empty commit) came in via #2823 and is an ancestor of `main`. Ruleset `6457257` applies `non_fast_forward` + `required_signatures` to `refs/heads/**` with **zero bypass actors**, so it can never be rewritten — and `ci-success` is a required check on `main`/`dev`, also with zero bypass. There is no branch shape that clears this and no one who can merge past it. Every future back-merge inherits the same wedge.

## The fix

A commit already reachable from `main` or `dev` is permanent, and the PR that first proposed it is where the gate belongs. Those are now reported with `core.warning` instead of failing the run.

Unchanged for new work: bot authors, unlinked emails and non-collaborators are still hard failures on the PR that introduces them.

Cost: the `compare` call runs **only** for a commit that already failed the author test, cached per SHA. A green PR makes zero extra API calls.

## Verified

Script extracted from the built YAML and exercised against a mocked Octokit:

| case | result |
|---|---|
| all-human PR | pass, 0 compare calls |
| bot commit already on `main` (the back-merge case) | pass + warning |
| bot commit not landed anywhere | **fail** |
| unlinked email already on `dev` | pass + warning |
| unlinked email, new | **fail** |
| non-collaborator, new | **fail** |

YAML parses; `prettier --check` clean.

## Follow-up

`main` has no `human-authors` job at all — that is how a bot-authored commit reached it through #2823 in the first place. Gating `main`-targeted PRs is a separate PR.

PR #2836, opened 2026-08-26
feat(press): inline asset previews + layout pass on /press

0xkkonrad:
## What

Improves `/press` (the page `peanut.me/presskit` redirects to) — inline previews of the actual assets, and a layout pass. Stacks on #2811 (`feat/press-brand-subpage`); base is that branch so this PR shows only its own delta. If #2811 merges first, GitHub retargets this to `dev`.

## Previews

- Every brand-asset group now has a preview well derived from the frontmatter hrefs by extension — no content changes. Logotype / Icon / Wordmark render their SVG (`next/image` `unoptimized`, no `dangerouslyAllowSVG`); Mascot shows all five `.webp` side by side in one well; the Font group renders a KNERD specimen via `Title` on pink (the three-layer offset the guidelines mandate); Brand Guidelines (PDF only) has no well and spans the row.
- Download pills are labelled by format (`SVG · EPS · AI`) when unambiguous within the group, keep filenames otherwise, and carry `download` on same-origin files (the GitHub link keeps `target=_blank`).
- Helpers live in `src/components/Marketing/pressAssets.ts` with a 6-case jest test.

## Layout

- Brand assets moved above Tagline & headlines; two jump pills (`#assets`, `#contact`) at the top, zero new strings.
- `items-start` on the card grids (kills the stretched-card whitespace), `text-h4` section headings on both press pages, `shadow-4` only on the asset cards + contact card.
- Tagline section: the flagship headline no longer appears twice; the one-liner sits inside the flagship card with a label.
- Team photos: `sizes` on the `fill` images (browser now fetches `w=384` instead of `w=3840`), 2 columns on mobile, note moved above the grid.
- Company + Media contact share one row; contact is a pink card.
- `MarketingHero`: `text-6xl sm:text-7xl md:text-9xl` — at 390px "PRESS"/"BRAND" were clipped by the hero's `overflow-hidden`. Only the base breakpoint changes.
- i18n: `Short / Medium / Press / Partner`, the photo alt text, the one-liner label and a one-line assets note are now keys in all four locales.

## Verified

- prettier, eslint, `tsc --noEmit` (0), jest 6/6.
- Local `next dev --webpack` DOM audit at 1000px and 390px on `/en/press`, `/pt-br/press`, `/en/press/brand`: no horizontal overflow, hero letterforms inside the viewport, all 16 previewed images load, 28 download links ≥ 44px tall, no English labels on pt-BR.
[...]

PR #2830, opened 2026-08-26
chore: backmerge main into dev (2026-08-26)

innolope-dev:
Backmerge of `main` into `dev`.

One conflict, in `src/config/peanut.config.tsx` — both sides rewrote the same provider `useEffect`. Resolved as the semantic union:
- kept **dev**'s structure: the dynamic `AppStateProviders` import, the marketing-route split, and the lazy country-locale load
- grafted in **main**'s passkey-shim install path from #2801 wholesale: async `installPasskeyShim` with the patched-flag verification throw, the bounded 1.5s retry, and `markPasskeyShimFailed()` on double failure
- the `scheduleDirectFetchCanary()` call and import from main's side were **not** carried over: dev retired the native transport-migration canary in #2737 and the `native-canary.ts` module no longer exists on this branch — this merge is where main's last reference to it dies

Typecheck clean; passkeyCeremony.utils, ClientProviders, user-query and usePWAStatus test suites all green (35 tests).

Brings in:
- PR #2808 — hotfix/reenable-manteca
- PR #2801 — fix/task-21782-login-ceremony-hang
- PR #2807 — hotfix/disable-all-manteca
- PR #2797 — hotfix/noindex-app-surfaces-v2
- PR #2765 — fix/eea-uplift-keep-sdk-open
- PR #2800 — fix/bridging-fee-note

PR #2840, opened 2026-08-26
ci: gate main-targeted PRs on human authorship too

innolope-dev:
## Why

`human-authors` was added on `dev` and has not reached `main` yet, so **every PR based on `main` runs without it**. That is not a theoretical gap — it is exactly how the current mess happened:

- #2823 (base `main`) carried `aa6db97` — `chore: retrigger OTA build`, an empty commit authored by `chip-peanut-bot[bot]`. Ungated, so it merged.
- `aa6db97` is now an ancestor of `main` and can never be rewritten (`non_fast_forward` + `required_signatures` on `refs/heads/**`, zero bypass actors).
- Every back-merge lists it, so [#2830](https://github.com/peanutprotocol/peanut-ui/pull/2830) is red and unmergeable. [#2839](https://github.com/peanutprotocol/peanut-ui/pull/2839) fixes that end (stop re-judging commits that already landed); **this PR closes the door it came through**.

## What

The same job, with the same script as `dev` has after #2839 — byte-identical text in the same position, so the next `dev` → `main` release merges it as the same change rather than a conflict. Added to `ci-success.needs` so it actually gates.

## Effect on open `main` PRs

I checked every one. Only [#2829](https://github.com/peanutprotocol/peanut-ui/pull/2829) fails: its single commit `4d393a7` is authored by `chip-peanut-bot[bot]`. That is the gate doing its job — the fix is to re-commit that one-line change from a human account. Everything else (#2833, #2835, #2836) is human-authored and unaffected.

The content pipeline is unaffected: `update-content.yml` commits as `Hugo Montenegro <h@hugomontenegro.com>`, a real account, so content-publish PRs pass and merge-on-green keeps working.

Depends on nothing — but merge #2839 first if you want the two files to converge cleanly.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -428,6 +428,110 @@ jobs:
                           });
                       }
 
+    # Human-only authorship gate. Every commit on the PR must be authored by a
+    # repo collaborator's own GitHub account — no `*[bot]` authors (claude[bot],
+    # github-actions[bot], …) and no emails that GitHub cannot map to an
+    # account. Policy set 2026-08-24: AI is a tool, not an owner; every commit
+    # has a human who is accountable for it. Cloud coding sessions that commit
+    # as claude[bot] are exactly what this blocks. Content-publish PRs pass:
+    # update-content.yml authors and signs its commit as a human account.
+    #
+    # No allowlist to maintain: "has write access to this repo" IS the team
+    # roster, and GitHub org/repo access management keeps it current. The
+    # per-user permission endpoint works with the default GITHUB_TOKEN.
+    human-authors:
+        name: human-authors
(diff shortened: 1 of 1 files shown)


innolope-dev:
Same Dependabot caveat as [peanut-api-ts#1436](https://github.com/peanutprotocol/peanut-api-ts/pull/1436): there is no `.github/dependabot.yml` here either, so GitHub's automatic security-update PRs target `main` and are authored by `dependabot[bot]`. None are open right now, but the next one will fail this gate. If the team wants those to keep flowing, the fix is one line — an `ALLOWED_BOTS` set checked before the `[bot]` rejection — and it should land in the same shape in both repos and on both `main` and `dev`.

PR #2843, opened 2026-08-27
fix(fx): stop the add-money and withdraw screens freezing on a rate fetch

### LINKED ISSUE #1848: fix: add timeout/error handling for currency rate and limits API calls

innolope-dev:
Fixes #1848. Replaces #2707 (see below for why).

## The freeze

`useCurrency` seeded `code` from the prop on the first render and then ignored the prop for the rest of its life — there was no sync effect. Both callers derive the currency from `useSearchParams()`:

- `withdraw/manteca/page.tsx` → `useCurrency(selectedCountry?.currency ?? null)`
- `MantecaAddMoney.tsx` → `useCurrency(selectedCountry?.currency ?? 'ARS')`

On the first render of a statically exported page the search params are empty, so the withdraw page passed `null`, the hook latched it, never fetched, and `price` stayed null forever. Its gate is `if (isCurrencyLoading || !currencyPrice || …) return <PeanutLoading />` — an unresolvable loader with no header and no way out. A deterministic hang, not a slow network, which fits "reproducible on G31 and G23" better than a timeout would.

The same latch also meant add-money would price a Brazil flow in ARS (its fallback) whenever the country arrived a render late.

Timeouts and retries were already handled (`fetchWithSentry` aborts at 20s; `useLimits` inherits `RETRY_STRATEGIES.FAST`), and the null-rate crash in `InputAmountStep` was fixed in #2325 — so the remaining freeze was not a missing timeout.

## Changes

**`useCurrency`**
- Sync `code` when the prop changes, tracking the previous prop so the unsupported-currency branch (which deliberately nulls `code`) isn't clobbered on the next render.
- Report the gap between the prop landing and the fetch starting as `isLoading`, so consumers can't read it as a settled "no rate" and flash an error.
- Add `refetch()`. Previously the effect only re-ran on a currency change, so a rate outage stranded the user until they left the screen entirely.
- Ignore a superseded in-flight response so a slow request can't overwrite a newer currency's rate.
- Clear `price`/`symbol` on a failed fetch, so an in-place currency change whose fetch fails can't leave the previous currency's number on a money-display path.

**Failure states** — new shared `RateUnavailable` (alert + retry), wired into all **four** `useCurrency` callers:
- `InputAmountStep`: retry button instead of a bare alert; header stays mounted while loading so back always works.
- `withdraw/manteca`: had no error branch at all — a failed fetch fell into the same infinite loader. Now an error screen with back and retry.
- `MantecaReviewStep`: rendered `1 USD = undefined ARS`; now shows the retry state.
- `add-money/[country]/bank`: showed a hardcoded, untranslated English string with no way to recover.

**i18n** — moved `rateUnavailable` from `addMoney.errors` to the shared `errors` namespace across all four catalogs rather than duplicating the string.

## Why this replaces #2707

#2707 conflicted with `dev` and could not be updated in place. Resolving the conflict requires a merge commit, and merging `dev` into that branch (forked 2026-08-17) drags in 7 pre-2026-08-18 unsigned `src/content` bumps that the signature ruleset rejects with `GH013`; `non_fast_forward` blocks the rebase escape hatch. This branch is cut fresh from `dev` and carries the identical resolved tree.

The conflict itself is worth noting: `dev` commit 23c47f79f independently fixed the bank flow's hardcoded string by pointing it at `addMoney.errors.rateUnavailable` — the key this PR deletes. That was the single conflicting line in the whole diff, and it is exactly the breakage @kushagrasarathe flagged in review. The resolution keeps this PR's `RateUnavailable` (shared `errors` namespace), which supersedes dev's fix and adds the retry.

## Review follow-ups from #2707

- **Blocking** — bank flow reading the moved key: fixed, and upgraded to the shared `RateUnavailable` with a retry.
- **Minor 1** — stale rate on failed refetch: fixed, with a regression test.
- **Minor 2** — `withdraw/manteca/page.tsx` churn hotspot: out of scope here, filed as TASK-21934. Verified 62 commits in 6 months, 10 of the last 20 are `fix` (needs `git log --follow`; the file was renamed).

## Verification

- `useCurrency` suite 7/7 (late-arriving currency, refetch recovery, out-of-order response, stale-rate clear) and `add-money-states` 59/59 — 66 passing across the two touched suites.
- `tsc` clean on the touched files, `eslint` clean (remaining warnings in `withdraw/manteca` are pre-existing), `prettier --check` passes.

## Not covered

No page-level test for `withdraw/manteca` — it has no existing suite and would need a large mock surface. The hook test covers the root cause; the page's new branch is not directly exercised. That gap is part of what TASK-21934 covers.

## Summary by CodeRabbit

* **New Features**
  * Added retry actions when exchange rates are unavailable during add-money and withdrawal flows.
  * Preserved back navigation while exchange rates are loading.
  * Improved handling when switching currencies, including clearer loading and error states.
  * Added localized exchange-rate error messaging across supported languages.

* **Bug Fixes**
  * Prevented outdated exchange-rate results from appearing after currency changes.
  * Cleared stale rates when a currency update fails.

### DIFF
--- a/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
+++ b/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
@@ -19,6 +19,7 @@ import { useCallback, useEffect, useMemo, useState } from 'react'
 import countryCurrencyMappings, { isNonEuroSepaCountry, isUKCountry } from '@/constants/countryCurrencyMapping'
 import { formatUnits } from 'viem'
 import PeanutLoading from '@/components/Global/PeanutLoading'
+import RateUnavailable from '@/components/Global/RateUnavailable'
 import EmptyState from '@/components/Global/EmptyStates/EmptyState'
 import AddMoneyBankDetails from '@/components/AddMoney/components/AddMoneyBankDetails'
 import { getCurrencyConfig, getCurrencySymbol, getMinimumAmount, railJurisdictionForBank } from '@/utils/bridge.utils'
@@ -192,7 +193,12 @@ function BridgeBankOnrampPage() {
     // deposit-side price (localCurrency per USD) for limits validation — deposits
     // execute at buy, so the USD equivalent must derive from buy, not the sell-side
     // display quote served by /fx/rate (useCurrency handles USD as 1:1)
-    const { price: localPrice, isLoading: isRateLoading, isError: isRateError } = useCurrency(localCurrency)
+    const {
+        price: localPrice,
+        isLoading: isRateLoading,
+        isError: isRateError,
+        refetch: refetchRate,
+    } = useCurrency(localCurrency)
 
     // convert input amount to USD for limits validation
     // bridge limits are always in USD, but user inputs in local currency
@@ -471,7 +477,7 @@ function BridgeBankOnrampPage() {
                     {error.showError && !!error.errorMessage && !limitsValidation.isBlocking && (
                         <ErrorAlert description={error.errorMessage} />
                     )}
-                    {localCurrency !== 'USD' && isRateError && <ErrorAlert description={t('errors.rateUnavailable')} />}
+                    {localCurrency !== 'USD' && isRateError && <RateUnavailable onRetry={refetchRate} />}
                 </div>
 
                 <OnrampConfirmationModal
--- a/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
+++ b/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
@@ -469,7 +469,7 @@ function BridgeBankOnrampPage() {
                             (localCurrency !== 'USD' && (isRateLoading || isRateError))
                         }
                         className="w-full"
-                        loading={isCreatingOnramp}
+                        loading={isCreatingOnramp || (localCurrency !== 'USD' && isRateLoading)}
                     >
                         {tCommon('continue')}
                     </Button>
--- a/src/app/(mobile-ui)/add-money/__tests__/add-money-states.test.tsx
+++ b/src/app/(mobile-ui)/add-money/__tests__/add-money-states.test.tsx
@@ -1953,6 +1953,30 @@ describe('GROUP 8: InputAmountStep Component', () => {
         expect(refetch).toHaveBeenCalledTimes(1)
     })
 
+    // The rate block disables Continue, so hiding its retry behind `error` or a
+    // blocking limits card leaves the user stuck with no way to clear it.
+    test('rate retry stays available even when another error is showing', () => {
+        const refetch = jest.fn()
+        renderWithProviders(
+            <InputAmountStep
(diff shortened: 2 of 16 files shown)


innolope-dev:
@coderabbitai review

innolope-dev:
Addressed all three CodeRabbit findings in 233fbd528 — each one re-opened the dead end this PR exists to close, so they were on-theme rather than out of scope.

**1. `withdraw/manteca` — back control lost during retry.** Confirmed. After `refetchCurrency()`, `isCurrencyLoading` flips true, the new error branch is skipped, and the old gate returned a bare `PeanutLoading` with no `NavHeader`. A retry that stalls put the user back in the exact frozen state of #1848. Both rate states now render the header-bearing layout; the bare loader is left only for the case where the country itself has not resolved.

**2. `InputAmountStep` — retry hidden behind other errors.** Confirmed, and the sharper point is that Continue is `disabled={… || rateUnavailable}`, so when `error` or a blocking limits card was also showing, the rate block held Continue down with its own retry hidden — no way out. The retry now renders whenever the rate failed. It can stack with the generic alert, which is correct: two distinct problems, one of which has an action.

Added a regression test (`rate retry stays available even when another error is showing`) and verified it actually fails against the old condition rather than passing either way.

**3. `add-money/[country]/bank` — no loading feedback.** Confirmed. Continue is disabled while `isRateLoading` but `loading` only reflected `isCreatingOnramp`, so it read as an unexplained dead button. Took the proposed one-liner.

Verification: full suite 293/293 suites, 3633 tests. `eslint` clean (the 5 remaining warnings in `withdraw/manteca` are pre-existing `exhaustive-deps`), `prettier --check` passes.

One note for transparency: an earlier full-suite run showed a single failure I could not reproduce across four subsequent clean runs and did not manage to capture by name — most likely a timing flake, but flagging it rather than pretending I saw only green.

PR #2844, opened 2026-08-27
ci: dispatch trusted PR reviews to Chip on dev

Hugo0:
## Summary

`pull_request_target` reads its workflow from the **base branch** of the PR. The copy going to `main` therefore only covers `main`-targeted PRs. Nearly all work targets `dev`, so `dev` needs the same bridge or it gets no AI review at all.

Identical to the canonical file in mono (`ops/schedulers/pr-review/chip-pr-review.yml`), verified by `PR_REVIEW_CHECK_DEPLOYED=1 node --test`.

## Why this is safe

- `permissions: {}` — the job holds no GitHub token.
- It never checks out PR code and never forwards PR title, body, or comment text. Only the repository name, PR number and event ids cross the SSH bridge.
- The dispatcher on chip re-verifies the author against an explicit allowlist **and** live repository permission before any review starts.
- Chip never approves. On `dev` it always comments, so it cannot block a green merge.

## Task

Follow-up to the Chip PR review rollout.

### DIFF
--- a/.github/workflows/chip-pr-review.yml
+++ b/.github/workflows/chip-pr-review.yml
@@ -1,4 +1,12 @@
 # Trusted event bridge. It never checks out PR code or forwards PR text.
+#
+# Only pull_request_target and issue_comment may live here. Both resolve their
+# workflow file from a branch a PR author cannot write: the PR base, and the
+# repository default branch. pull_request_review does NOT -- GitHub runs it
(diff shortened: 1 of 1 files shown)


Hugo0:
Closing: redundant. `pull_request_target` resolves its workflow from the repository default branch, so the copy on `main` (#2833) covers dev-targeted PRs too. Caught by Chip's own review on this PR.

PR #1443, opened 2025-11-10
Add missing beneficiary details

Zishan-7:

PR #2847, opened 2026-08-27
ci: keep a passer-by from dispatching a manual Chip review

Hugo0:
## Summary

Follow-up to #1443 / #2846.

Dropping the association gate on the manual job left `/chip review` open to anyone on the public repo: type it and an SSH session opens, three GitHub calls run, and the dispatcher then refuses. Cheap per attempt, free to repeat.

It now excludes `NONE` and `MANNEQUIN` — the two values that mean pure stranger.

It deliberately still admits `CONTRIBUTOR`, because that is what an org admin with **private** org membership reads as. That is the exact trap that silently skipped their reviews earlier today, so the guard must never be used to decide trust.

Live repository write permission still decides that, server side.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restricted manual pull request reviews to comments from approved repository contributors, improving review workflow security.

### DIFF
--- a/.github/workflows/chip-pr-review.yml
+++ b/.github/workflows/chip-pr-review.yml
@@ -67,7 +67,7 @@ jobs:
         if: >-
             github.event_name == 'issue_comment' &&
             github.event.issue.pull_request &&
-            !contains(fromJSON('["NONE","MANNEQUIN"]'), github.event.comment.author_association) &&
+            contains(fromJSON('["OWNER","MEMBER","COLLABORATOR","CONTRIBUTOR"]'), github.event.comment.author_association) &&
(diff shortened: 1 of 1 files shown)


Hugo0:
Closing: superseded, not abandoned. This change — allowlisting `OWNER`/`MEMBER`/`COLLABORATOR`/`CONTRIBUTOR` on the manual `/chip review` guard — is already on `main`. It shipped inside the later notify-prefs PR (peanut-ui#2853 / peanut-api-ts#1446), which was cut from the canonical file in mono after this fix had landed there.

Verified: `PR_REVIEW_CHECK_DEPLOYED=1 node --test ops/schedulers/pr-review/test.mjs` passes, so the deployed workflow on each default branch matches the canonical one byte-for-byte. This branch now only conflicts with what already merged.

PR #2849, opened 2026-08-27
fix(qr-scanner): keep the start attempt alive while camera permission is pending

innolope-dev:
Closes TASK-21927.

## The bug

The web/PWA camera path races `getUserMedia` against a hard 5-second deadline and, on expiry, destroys the scanner and shows permission-denied.

That deadline exists for a real reason: in an iOS PWA (WKWebView) `getUserMedia` can hang forever after a denial instead of rejecting, and without a deadline the user is stranded on a spinner. But a hang after denial is indistinguishable from a prompt the user simply has not answered yet — so the same timeout also tore down cameras that were one tap away.

- 30-day production class: **251 events across 178 users**.
- Monitor specimen (2026-08-26): one iOS PWA user hit `NotAllowedError: Camera start timed out` at 18:19:27Z, then scanned four QR codes from 18:22:10Z. `$0` moved, lost, held or stuck — it is a UX failure, not a money one.

## The fix

The deadline no longer ends the attempt.

- At 5s the permission modal still goes up, so nobody stares at a spinner and the **retry and paste fallbacks stay reachable** — unchanged from today.
- The pending `start()` keeps running underneath instead of being destroyed. When permission settles in the user's favour the camera comes up on its own.
- Recovery **restarts** rather than adopting the resolved stream: the modal unmounts the `<video>` the pending scanner is bound to, so the element the user would see is a fresh one that needs its own start.
- A late rejection now reports the real error (`NotFoundError` / `NotReadableError`) instead of a blanket "permission denied".
- A grant that lands while the app is backgrounded is left to the existing visibility listener — rebuilding on a hidden document resolves without a stream, which would have swapped the modal for a false "camera start failed".

## Native: nothing changes

The Capacitor branch never had a deadline. On native, `ensureNativeCameraPermission()` settles the OS dialog *before* `getUserMedia`, and `getUserMedia` then rejects cleanly on denial — the short timeout was removed from that path earlier precisely because it made the first-run prompt look like a failure. This PR does not touch it, and a test pins it: `native keeps waiting on the OS dialog with no deadline at all` fails if a deadline is ever introduced there.

## Tests

Six new cases in `useQRScanner.test.ts` under an iOS device profile, modelling the `<video>` unmount/remount the permission modal causes. Four of them fail on `dev` and pass here; the other two are the no-regression pins and pass both ways.

| case | on `dev` | here |
|---|---|---|
| the deadline shows the permission modal without ending the attempt | ✗ | ✓ |
| a grant that lands after the deadline brings the camera up on its own | ✗ | ✓ |
| a grant that lands while backgrounded is picked up on the way back | ✗ | ✓ |
| a hardware failure that lands after the deadline reports the hardware error | ✗ | ✓ |
| a denial that lands after the deadline stays on the permission modal | ✓ | ✓ |
| native keeps waiting on the OS dialog with no deadline at all | ✓ | ✓ |

The 5 existing TASK-21862 resume/backgrounding tests pass unchanged.

## Two follow-ups in the second commit (no behaviour change)

**Extracted the deadline race.** The complexity bot flagged `startCamera` at CC 23 / SLOC 99 once the fix grew it. `raceStartDeadline` now owns the race *and* the timeout handle, so the outer `startTimeoutId` and its `clearTimeout` in the catch are gone — the function is back under where it started.

**De-flaked the suite.** While verifying, I hit a ~4% failure in `does not re-enter the camera flow while the error view is up` (expects 2 starts, sees 3). It is **pre-existing** — reproducible on `dev` at 1/25, so it is not from this change, but it is in the file I am touching. Cause: every mount schedules a 100ms element-retry before the `<video>` exists, and the file's first test left that one on a real clock; it outlived its own test and fired `startCamera` into whichever test was running 100ms later, inflating the module-level `startCalls` there. Fake timers are now installed file-wide and that test drains its own retry, as the describes already did. 40 consecutive runs clean.

## Local verification

`src/components/Global` jest suite green (339 tests), prettier clean, eslint clean on the touched files, typecheck shows no new errors (only the pre-existing SVG `TS2307` noise).

CodeRabbit replied **"Review rate limited"** and reviewed nothing — not re-requested.

### DIFF
--- a/src/components/Global/QRScanner/__tests__/useQRScanner.test.ts
+++ b/src/components/Global/QRScanner/__tests__/useQRScanner.test.ts
@@ -10,6 +10,7 @@ import { useQRScanner } from '../useQRScanner'
 let onDecode: (result: { data: string }) => void
 let startBehaviour: () => Promise<void> = () => Promise.resolve()
 let startCalls = 0
+let destroyCalls = 0
 let attachStream = true
 let lastOptions: { preferredCamera?: unknown } = {}
 
@@ -44,7 +45,9 @@ jest.mock('qr-scanner', () => ({
             this.active = false
             this.video.srcObject = null
         })
-        destroy = jest.fn()
+        destroy = jest.fn(() => {
+            destroyCalls++
+        })
         setCamera = jest.fn()
         setInversionMode = jest.fn()
     },
@@ -53,11 +56,15 @@ jest.mock('@sentry/nextjs', () => ({ captureException: jest.fn() }))
 jest.mock('@/components/0_Bruddle/Toast', () => ({
     useToast: () => ({ error: jest.fn(), info: jest.fn() }),
 }))
+// jest hoists these factories, so the switches they read must be `mock`-prefixed
+let mockDeviceType = 'web'
+let mockCapacitor = false
 jest.mock('@/hooks/useGetDeviceType', () => ({
-    useDeviceType: () => ({ deviceType: 'web' }),
+    useDeviceType: () => ({ deviceType: mockDeviceType }),
     DeviceType: { IOS: 'ios' },
 }))
-jest.mock('@/utils/capacitor', () => ({ isCapacitor: () => false }))
+jest.mock('@/utils/capacitor', () => ({ isCapacitor: () => mockCapacitor }))
+jest.mock('@/utils/camera-permission', () => ({ ensureNativeCameraPermission: jest.fn(async () => true) }))
 jest.mock('next-intl', () => ({ useTranslations: () => (key: string) => key }))
 
 // jsdom implements neither, and the hook's cleanup calls both on unmount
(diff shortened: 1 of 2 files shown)


innolope-dev:
> **MAJOR** · `useQRScanner.ts:279` · Handle qr-scanner's actual rejection shape

Confirmed, and fixed in `fece1d9e5`. Verified against the pinned library rather than the types — `qr-scanner@1.4.2`'s `_getCameraStream`:

```js
_getCameraStream(){
  if(!navigator.mediaDevices) throw "Camera not found.";
  ...
  for(let d of [...c,...b]) try{ ...getUserMedia... return {stream,facingMode} } catch(f){}   // swallows every DOMException
  throw "Camera not found.";                                                                  // a string, not an Error
}
```

So `err instanceof Error` is always false, `errName` is always `''`, and the `NotFoundError` / `NotReadableError` branch never fired. The test passed only because it rejected the mock with a `DOMException` — a shape the pinned library never produces. Good catch.

I dropped the classification instead of trying to reconstruct the cause. By the time `start()` rejects, denial, missing hardware and a busy camera are one indistinguishable failure, and preserving the browser cause would mean either calling `getUserMedia` ourselves or probing `enumerateDevices()` on the error path — both a rewrite of the acquisition path, well outside a P3 UX fix. The permission modal is already up and is the one state that offers **both** retry and paste, so the user stays there. The test now rejects with the library's real value (`'Camera not found.'`) and pins that outcome.

Net effect on this PR: the "reports the real error" bullet is withdrawn (it never worked), and the outcome for a late failure is identical to today's — permission modal with retry + paste. The core of TASK-21927 is untouched.

**Wider, pre-existing, not fixed here:** the same shape mismatch makes the *main* catch's whole `NOT_ALLOWED` / `NOT_READABLE` / `NOT_FOUND` switch and its `NotReadableError` retry loop dead for library-originated failures, and two TASK-21862 tests on `dev` mock the same fictional `DOMException`. That predates this PR and deserves its own change.

innolope-dev:
Follow-up opened: [#2864](https://github.com/peanutprotocol/peanut-ui/pull/2864) — it fixes the wider version of the finding above.

The same shape mismatch makes this PR's *pre-existing* neighbours dead too: the main catch's whole `NOT_ALLOWED` / `NOT_READABLE` / `NOT_FOUND` switch, its camera-busy retry loop, and `setIsPermissionDenied(true)` firing even for a device with no camera. #2864 adds `classifyCameraFailure`, which recovers the reason from the browser (`enumerateDevices`, then a repeat of the ladder's last rung) instead of the library.

Two notes for whoever merges:
- Both PRs touch `startCamera`'s catch and `useQRScanner.test.ts`, so the second one in needs a small conflict fix.
- Once both land, `adoptPendingStart`'s catch here should use `classifyCameraFailure` as well. It currently declines to classify precisely because the reason was unrecoverable — which #2864 changes.

PR #2858, opened 2026-08-27
security: require a human code owner so a bot approval cannot count

Hugo0:
## Summary

`main` already sets `require_code_owner_review: true`, but neither repo has a CODEOWNERS file — so the rule has no owners to require and is vacuous. Today a bot approval would count toward the one required approval.

Chip has never approved a PR (checked across 50 recent PRs in both repos) and the publisher now refuses to submit anything but `COMMENT`/`REQUEST_CHANGES`. But Chip holds a token with `pull_requests: write`, and until now the only thing preventing an approval was an instruction in a skill — policy, not structure.

With this file, a human has to approve `main`, whatever else appears on the PR. That is enforced by GitHub, not by us.

Reconciliation also now dismisses any approval wearing Chip's name and alerts in Discord.

## Summary by CodeRabbit

* **Chores**
  * Added repository ownership rules requiring review from designated maintainers.
  * Clarified that automated accounts and apps are not eligible as code owners.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,10 @@
+# Chip must never be able to satisfy a review requirement.
+#
+# main requires code-owner review. Without this file that rule has no owners
+# to require, so any approval counted -- including one from a bot holding a
+# write token. Listing humans here means a human has to approve, whatever else
+# happens to appear on the pull request.
+#
+# Keep this list to people. Never add an app or a bot.
+
+* @Hugo0 @jjramirezn @kushagrasarathe @0xkkonrad @innolope-dev @abalinda


Hugo0:
Both findings acted on — this is Chip reviewing the PR that stops Chip approving, and it was right twice.

**MAJOR — invalid or non-writable owners are silently skipped.** Correct, and it would have failed silently with more confidence than before. Verified before merge: all six handles are `admin` on **both** repos, so every entry is honoured.

```
Hugo0 jjramirezn kushagrasarathe 0xkkonrad innolope-dev abalinda  → admin (peanut-ui, peanut-api-ts)
```

On the team suggestion: the only org team is `collaborators`, which currently contains one member, so a team would narrow the owner set rather than widen it. Staying with explicit handles, now watched (below).

**MINOR — the only thing stopping a bot becoming an owner is a comment.** Also correct, and the same policy-not-structure pattern this PR exists to replace. Implemented your suggested fix: reconciliation now compares CODEOWNERS on `main` against an expected set in config and posts an `@here` on any account added or removed, or the file going missing. A PR quietly making Chip a code owner is now as loud as a branch-protection change.

Merging.

PR #2859, opened 2026-08-27
fix(rewards-nav): safe-back everywhere, one Send header, honest sent-money badge

innolope-dev:
Fixes four Issue-Inbox cards in the rewards/send navigation + profile area (TASK-21673, TASK-21201, TASK-21929, TASK-21394-FE).

## TASK-21673 — Rewards navigation dumps users on Home (remainder)
The rewards page and InvitesGraph halves already landed on dev; this covers the send-flow half:
- `SendRouter.view.tsx` / `Contacts.view.tsx`: the base-view back handler used a hardcoded `router.push('/home')`. It now goes through `useSafeBack('/home')`, so Rewards → invitee → send → back lands on Rewards. The byte-identical duplicated handler in `ContactsView` is gone — the router passes its single handler down as a prop.
- `PublicProfile.tsx`: back relied on `document.referrer` (`checkIfInternalNavigation`), which is empty for SPA pushes and in the Capacitor export, so it reliably fell through to `/home`. Replaced with `useSafeBack('/home')`.
- `SendInputView.tsx`: safe-back fallback pointed at the marketing root `/`; now `/home`.

## TASK-21201 — flow flashed "Send" → "Pay" → "Send"
The direct-send input view was the only screen in the flow titled `headers.pay`. It now uses `headers.send` end to end (product call: sending to a contact reads Send, not Pay). `headers.pay` stays — qr-pay/contribute-pot/semantic-request use it legitimately. No new i18n keys.

## TASK-21929 — contact trust indicator inconsistent between Send and profile
The profile derived "you've sent money to them" from `totalUsdReceivedFromCurrentUser`, whose SQL misses send-link claims and requester-less direct sends — so the same contact showed the indicator in Send contacts but not on their profile. The profile now uses the interaction-status endpoint (`useUserInteractions` → `checkInteractions`), the most complete relationship source, staying neutral until the query resolves and never querying for guests.

## TASK-21394 (FE half) — points preview must never block a flow
- Removed the dead `INVITE_KYC_VERIFIED` member from `PointsAction` (zero call sites; the API's member is `KYC_VERIFIED`).
- `pointsApi.calculatePoints` now resolves `null` on any failure (logged once per session) instead of throwing. All consumers already optional-chain `estimatedPoints`, so a failed preview degrades to "no points card". The backend half of TASK-21394 (kind mapping) ships separately.

## Tests
- `send-states.test.tsx`: cold deep-link → `/home` fallback; in-app history → `router.back()`.
- `PublicProfile.test.tsx`: both back branches + indicator-from-interactions, neutral-while-loading, no-guest-query.
- `points-calculate.test.ts` (new): success, null-on-failure, logs-once.
- 37/37 pass via `--runTestsByPath`; `pnpm typecheck` clean.

Notion: TASK-21673 · TASK-21201 · TASK-21929 · TASK-21394

### DIFF
--- a/src/components/Profile/components/PublicProfile.tsx
+++ b/src/components/Profile/components/PublicProfile.tsx
@@ -12,16 +12,18 @@ import { useTranslations } from 'next-intl'
 import Image from 'next/image'
 import posthog from 'posthog-js'
 import ProfileHeader from './ProfileHeader'
-import { useState, useEffect, useMemo, useRef } from 'react'
+import { useState, useEffect, useRef } from 'react'
 import { invitesApi } from '@/services/invites'
 import { usersApi } from '@/services/users'
 import { useRouter } from 'next/navigation'
 import { isCapacitor } from '@/utils/capacitor'
 import { requestUrl } from '@/utils/native-routes'
 import Card from '@/components/Global/Card'
-import { checkIfInternalNavigation, saveToCookie, toInviteCode } from '@/utils/general.utils'
+import { saveToCookie, toInviteCode } from '@/utils/general.utils'
 import { useAuth } from '@/context/authContext'
 import { useGuestStoreHandoff } from '@/hooks/useGuestStoreHandoff'
+import { useSafeBack } from '@/hooks/useSafeBack'
+import { useUserInteractions } from '@/hooks/useUserInteractions'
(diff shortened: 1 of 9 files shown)


innolope-dev:
@coderabbitai review

PR #2864, opened 2026-08-27
fix(qr-scanner): recover the camera failure reason qr-scanner discards

innolope-dev:
Follow-up to [#2849](https://github.com/peanutprotocol/peanut-ui/pull/2849), where chip-review found this shape mismatch in one new branch. It is much wider than that branch.

## Every branch of the camera error handling was dead

`qr-scanner@1.4.2`'s `_getCameraStream` walks a ladder of `getUserMedia` constraint sets, swallows each DOMException on the way, then rejects with a **string**:

```js
_getCameraStream(){
  if(!navigator.mediaDevices) throw "Camera not found.";
  let b=[{width:{min:1024}},{width:{min:768}},{}],
      c=b.map(d=>Object.assign({},d,{[a]:{exact:this._preferredCamera}}));
  for(let d of [...c,...b]) try{ ...getUserMedia... return {stream,facingMode} } catch(f){}  // swallowed
  throw "Camera not found.";                                                                 // not an Error
}
```

So in `useQRScanner.ts`, `err instanceof Error ? err.name : ''` was **always** `''`, and therefore:

| consequence | effect on users |
|---|---|
| `getErrorMessage` always hit `default:` | every camera failure read "camera unavailable"; the NotAllowed / NotReadable / NotFound copy was unreachable |
| `shouldRetry` never true | the camera-busy retry loop (`MAX_CAMERA_RETRIES`) never ran once |
| `setIsPermissionDenied(true)` for every failure | a device with no camera got the "enable camera access" modal — help it cannot act on |

Two tests hid it by rejecting the scanner mock with a `DOMException` the pinned library cannot emit.

## The fix: ask the browser, not the library

`classifyCameraFailure` (new, in `src/utils/camera-failure.ts`) recovers the reason the library discarded:

1. An `err` that is already an `Error` came from our own code (chunk load, scanner constructor, `applyConstraints`) and keeps its name — no probe.
2. `enumerateDevices()` — no `videoinput` means genuinely absent hardware. Identified **without opening a camera**; entries are listed even without a grant, only labels are hidden.
3. Otherwise repeat the request that just failed. The library's ladder ends on a plain `{video:{}}` ≡ `{video:true}`, so this one call reproduces its last rung and surfaces the DOMException it threw away. Raced against a 2s deadline, because the probe inherits the iOS PWA hang it is diagnosing.
4. A probe that **succeeds** means the device is present and permitted and the failure was transient (contention, hardware settling) — `NotReadableError`, the one classification the caller retries on. The stream is stopped immediately.

Never throws; returns `''` when it cannot tell, which is the previous behaviour.

The `await` lands inside the catch, so `startCamera` re-checks `superseded()` / `isScanningRef` before committing state — a newer start can slip past an await that did not exist before.

### Alternatives weighed

- **Acquire the stream ourselves and hand it over.** Viable — `start()` skips `_getCameraStream` entirely when `$video.srcObject` is already set. Rejected: matching the library's constraint ladder (exact-facingMode → width floors → plain) is a compatibility risk on exactly the low-end devices this serves, and it splits mirror/`setCamera` handling between us and the library.
- **Patch `_getCameraStream` to rethrow.** Rejected: a patch against a minified bundle that no reviewer can read.

## Tests

`camera-failure.test.ts` (8 cases) covers the classifier directly: denial, busy, absent hardware, no `mediaDevices` at all, a probe that hangs, a probe that succeeds (and is released), a real Error passed straight through, and the never-throws guarantee.

Four new hook cases prove the previously-dead paths are reachable — **all four fail on `dev`** with exactly the symptoms above:

| case | on `dev` | here |
|---|---|---|
| a denial reaches the permission modal with its own copy | `cameraUnavailable` | `cameraPermissionDenied` |
| a missing camera is no longer reported as a denied permission | `isPermissionDenied: true` | `false` |
| a busy camera retries, and recovers when the device frees up | `isPermissionDenied: true`, no retry | retries, recovers |
| a camera that stays busy stops retrying and says so | `cameraUnavailable` | `cameraStillBusy` |

The three tests that mocked the fictional `DOMException` now reject with the string the library really throws, and the browser stub they run against models `enumerateDevices` / `getUserMedia` rather than a rejection shape the library cannot produce.

## Native

Untouched. `ensureNativeCameraPermission()` still gates the Capacitor path before `getUserMedia`; the classifier only runs after a start has already failed, and its probe works the same in a WebView.

## Local verification

`src/components/Global` + `src/utils` green (92 suites, 1681 tests), 12 consecutive runs of the touched suites clean, prettier and eslint clean on all four files, typecheck shows no new errors.

## Overlap with #2849

Both touch `startCamera`'s catch and the same test file, so whichever lands second needs a small conflict fix. #2849's `adoptPendingStart` deliberately does **not** classify its late rejection, citing this exact library limitation — once both land, that catch should adopt `classifyCameraFailure` too. Noted there.


innolope-dev:
@coderabbitai review

innolope-dev:
Rebuilt on `dev` now that [#2849](https://github.com/peanutprotocol/peanut-ui/pull/2849) has landed — `8aac7220d`. Conflict resolved, and the follow-up promised there is done in the same commit.

**The conflict.** The hook merged cleanly. The test file's conflict was just two independent `describe` blocks interleaved by git, so I took `dev`'s file and re-applied this branch's additions on top rather than untangling the hunks.

**`adoptPendingStart` classifies now.** #2849 deliberately did *not* classify its late rejection, on the grounds that the reason was unrecoverable — which is precisely what `classifyCameraFailure` changes. The 5s deadline still guesses "denied" (the common case, and it needs to put actionable UI up immediately), but a late failure now walks that guess back instead of offering permission help to a device with no camera. It asks for the copy that does not promise a retry, since a start that already outlived its deadline gets none.

**Tests.** The last three cases mocking the fictional `DOMException` are converted to the string qr-scanner really throws, and #2849's single pending-deadline failure case is split in two:

| case | without the `adoptPendingStart` wiring | here |
|---|---|---|
| a late denial keeps the permission modal the deadline put up | ✓ | ✓ |
| a late failure on a device with no camera walks the guess back | `isPermissionDenied: true` ✗ | `false` ✓ |

18 cases in the hook suite, all green. `src/components/Global` + `src/utils`: 92 suites / 1689 tests. 10 consecutive runs of the touched suites clean, prettier + eslint clean, typecheck no new errors.

Not re-triggering CodeRabbit — it replied "Review limit reached" on this PR.

innolope-dev:
Both findings confirmed and fixed in `69998e5e6`.

**MAJOR · `camera-failure.ts` · late probe stream** — right, and it is the worse of the two: when the 2s deadline won the race, `finally` ran with `stream` undefined, so a `getUserMedia` resolving afterwards handed back a MediaStream nobody held a reference to and nobody stopped. The camera and its indicator stay live behind an error screen. And that is exactly the delayed-grant case this path exists for, so it is the likely outcome, not the edge one.

Fixed by releasing from the acquisition promise instead of the race, so the stream is stopped whenever it arrives:

```ts
const acquisition = navigator.mediaDevices.getUserMedia({ video: true, audio: false })
void acquisition.then((stream) => stream.getTracks().forEach((t) => t.stop())).catch(() => {})
```

The race then only decides the classification. Test: `stops a probe stream that arrives after the deadline has passed` — resolves the grant *after* the timeout and asserts `track.stop()`.

**MINOR · `useQRScanner.ts` · stale busy retry** — also right, and a consequence of this PR making that loop reachable at all. The retry timer was not tied to the start generation, so a manual Retry that recovered inside the 1s delay got torn down by the stale callback. It now takes the same `superseded()` guard every other deferred start in the file already uses. Test: `a manual retry that recovers first is not torn down by the pending busy retry`.

Both tests fail without their fix — verified by reverting each in isolation (`Expected: 3, Received: 4` on the retry one).

92 suites / 1691 tests green, prettier + eslint clean, typecheck no new errors.

PR #2865, opened 2026-08-27
fix(kyc): stop showing "try again" to users who have nothing to try

innolope-dev:
Frontend half of the residence/KYC work. Backend: peanutprotocol/peanut-api-ts#1455 (and #1450 for `geoProhibited`).

## Three screens that told users something untrue

**1. The Verify button that did nothing (TASK-21882).** One payload-build failure marks all four Bridge rails FAILED in a single `updateMany`, and nothing in the product re-enables them. The backend answered with a bare approved + null-token body — byte-identical to "you're done" — so [useSumsubKycFlow.ts:329](src/hooks/useSumsubKycFlow.ts) fired `onKycSuccess`, opened no SDK, set no error, and closed. The user pressed Verify and *nothing happened*. Support then told them to press it again.

The backend now distinguishes the two cases with an explicit `rails-unavailable` actionType. This routes it exactly like the existing `unsupported-region` branch: honest terminal message, no success callback, `userInitiatedRef` cleared so a late websocket APPROVED cannot re-open "all set" on top of the error.

**2. "Let's try that again" for users who cannot.** `UnlockedRegions.view.tsx` hardcoded `rejectLabels` / `rejectType` / `failureCount` to null, with a comment explaining the view had no per-verification Sumsub history. It does — `identityVerification` carries exactly that, and `useIdentityVerification` already exposes it. Nulling them made `isTerminalRejection` degrade to retryable for **every** user, so a terminally rejected user got a retry button above a call that could not help them.

**3. A type that hid the bug.** `KycActionType` was missing two values the backend already emits — `bridge-uplift` (since the cross-region work) and now `rails-unavailable`. The union narrowed to `'manteca' | undefined` at the comparison site, which is how the drift stayed invisible.

## Also

`geoProhibited`'s docstring said "Rain's prohibited-issuance list"; it covers our own GB block too, and a UK resident is blocked just as permanently. The field had been typed here but never sent by the backend, so `cardState.utils.ts:139`'s `geo-blocked` branch was unreachable — api-ts#1450 emits it.

## Verification

- 2 new cases in `useSumsubKycFlow.test.ts`, including the inverse: a genuinely finished user (approved, no token, no actionType) must still be treated as success, so the fix cannot turn every approval into an error. 37/37 across the touched suites.
- i18n: `railsUnavailableError` added to `en`, `es-419` and `pt-BR`. `es-AR` is a delta-over-`es-419` locale and correctly needs no entry — 196/196 i18n tests pass, including locale parity.
- Typecheck back to the branch baseline.

## Deploy ordering

Merge **after** api-ts#1455, or alongside it. The `rails-unavailable` branch is inert until the backend sends it, and the reject-label change reads a field that already exists — so a FE-first deploy degrades to today's behaviour rather than breaking. The one hard dependency runs the other way: #1455 makes a LATAM initiate require a `targetCountry`, and this PR is where that error surfaces as a readable message rather than a raw code.

### DIFF
--- a/src/app/actions/__tests__/sumsub-refusals.test.ts
+++ b/src/app/actions/__tests__/sumsub-refusals.test.ts
@@ -1,0 +1,88 @@
+/**
+ * Wire-level tests for how a backend refusal becomes a client-side result.
+ *
+ * These assert against the RESPONSE BODY, not against a mocked return value of
+ * `initiateSumsubKyc`. That distinction is the point: the two failure modes
+ * below both live in the translation from body to result, so a test that mocks
+ * the action away cannot see either of them.
+ *
+ * The backend's `error` field is overloaded — a machine code on these routes,
+ * human prose on older ones — while `userMessage` is always prose. Getting
+ * that wrong either shows a raw code to the user or silently loses the
+ * terminal classification and restores a futile retry.
+ */
+
+import { initiateSumsubKyc, isTerminalActionCode } from '@/app/actions/sumsub'
+import { serverFetch } from '@/utils/api-fetch'
+
+jest.mock('@/utils/api-fetch', () => ({ serverFetch: jest.fn() }))
+
+const mockFetch = serverFetch as jest.MockedFunction<typeof serverFetch>
+
+const respondWith = (status: number, body: unknown) => {
+    mockFetch.mockResolvedValue({
+        ok: status >= 200 && status < 300,
+        status,
+        json: async () => body,
+    } as unknown as Response)
+}
+
+beforeEach(() => mockFetch.mockReset())
+
+describe('initiateSumsubKyc — backend refusals', () => {
+    it('classifies a permanent Manteca nationality refusal as terminal AND keeps its message', () => {
+        return (async () => {
+            respondWith(400, {
+                error: 'manteca_us_nationality_restricted',
+                userMessage: 'Payments from this country are not available for US citizens at this time.',
+            })
+
+            const result = await initiateSumsubKyc({ regionIntent: 'LATAM', crossRegion: true })
+
+            expect(isTerminalActionCode(result.code)).toBe(true)
+            expect(result.error).toMatch(/US citizens/i)
+            // the machine code must never be the thing the user reads
+            expect(result.error).not.toMatch(/manteca_us_nationality_restricted/)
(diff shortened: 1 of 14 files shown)


innolope-dev:
@coderabbitai review

innolope-dev:
### Not acting on the `'use server'` finding — the premise doesn't hold

Re: **BLOCKING · src/app/actions/sumsub.ts:41 · [moonshotai/kimi-k3] Sync helper exported from a 'use server' actions module breaks the Next.js build**

The rule it invokes is real, but it doesn't apply here: `src/app/actions/sumsub.ts` is not a `'use server'` module.

**1. The directive doesn't exist — not in this file, not anywhere.**

```
$ grep -rn "use server" src/app/actions/
(no matches)

$ grep -rln "'use server'" src/
(no matches)
```

Line 1 of `sumsub.ts` is `import { type InitiateSumsubKycResponse, ... }`. The finding infers the directive rather than observing it ("the import sits at line 9, leaving room for the directive above") — an understandable read of a file under `app/actions/`, but this repo doesn't use React server actions there.

**2. These aren't server actions.** `serverFetch` is `export const serverFetch = callApi` in `src/utils/api-fetch.ts` — a plain `fetch` wrapper around `fetchWithSentry`. The client hook calls these functions directly; nothing crosses a server boundary, so no directive is required and none is present.

**3. A real production build passed on the exact SHA.** `Deploy-Preview` runs `vercel build --target=preview` (see `.github/workflows/preview.yaml`), which is a full `next build`:

```
Preview deploy | cdd5f0bc | success
```

`cdd5f0bc` is the head that introduced `export const isTerminalActionCode` (line 39). If the constraint applied, that build would have failed with *"A 'use server' file can only export async functions"*.

The finding notes that typecheck + jest can't see this class of error, which is correct — but the preview build can, and did.

Happy to move `isTerminalActionCode` and `TERMINAL_ACTION_CODES` into a separate module anyway if we'd rather be defensive against someone adding the directive later; just say the word. As things stand it isn't a bug, so I've left it.

*(Flagging for context: chip marks `moonshotai/kimi-k3` findings as diff-only advice. Its other finding this round — `error` being read as both a machine code and the display message — was correct and is fixed in `cdd5f0bc`.)*

PR #2867, opened 2026-08-27
feat(telemetry): triage network-layer send failures — offline vs edge vs CORS-layer

innolope-dev:
Follow-up to #2860 / TASK-21956.

## Why the diagnosis needed evidence

#2860 concluded Hugo's device "lost connectivity". Hugo reports the opposite: his connection was good and the PWA worked at the same moment — only the native app failed. Both stories are consistent with the only signal we had, because `TypeError: Failed to fetch` carries no status and is produced identically by:

- a **WAF/bot challenge** on `api.peanut.me` (Cloudflare scores an Android WebView UA very differently from real Chrome — a mechanism where native fails while the PWA succeeds on the same network),
- a **CORS-less edge error page** (Cloudflare 502/522/524 and challenge pages carry no `Access-Control-Allow-Origin`, so a cross-origin fetch from the native `https://localhost` origin surfaces them as a bare TypeError),
- a **blocked preflight** (native requests always preflight — `Authorization` header),
- and a genuinely **dead radio**.

We have no Cloudflare access, so this PR makes every future occurrence label itself from the client side instead.

## How

**`src/utils/network-triage.ts`** — a `no-cors` fetch resolves with an opaque response whenever the network transaction completes: ANY status, no ACAO required, no preflight. So when a flow fails on a network-layer error, two bounded (2.5s) parallel probes run:

1. `HEAD {PEANUT_API_URL}/healthz` — unauthenticated, rate-limit-exempt, `no-store`. Resolving means the edge answered *something* — even a 403 challenge page counts, which is exactly right: the real request died anyway, so the failure was in the CORS/response-visibility layer, **not** connectivity.
2. `HEAD https://www.gstatic.com/generate_204` — Android's own captive-portal endpoint — separating "our edge is unreachable" from "device is offline".

Observation → `net_probe` tag: `api_reachable`, `edge_unreachable`, or `offline`, plus `net_error_class` (`fetch_rejection` / `timeout` / `service_unavailable`), `net_online`, `net_effective_type`, and `net_probe_ms` in extra/props. Deliberately framed as a post-failure *observation*, not a causal verdict: a transient radio drop can recover inside the ~2.5s probe window and still read `api_reachable` — per event it is a hint, and the causal signal is the distribution.

**Wiring** — `captureNetworkTriagedFailure()` replaces the bare `posthog.capture` + `captureException` pairs at both send-link sites (`create` and background `persist`) and in `useDirectSendFlow`, so the verdict rides on the **same** Sentry and PostHog events rather than needing a cross-referenced probe event. Triage triggers on the three engine-exact rejection messages **and** on `ServiceUnavailableError`/`ConnectionTimeoutError` — `fetchWithSentry` rethrows under those names, so flow-level catches only see the raw TypeError for calls that bypassed it.

**Shared predicate** — the engine-message list moved into `network-triage.ts` and the `connectionLost` classifier in `friendly-error.utils.tsx` now imports it, so copy and probes can't drift.

**Ambient signal** — `beforeSendHandler` stamps `net_online` on every kept browser event (guarded for server/edge runtimes). Free coarse evidence across all flows.

**CSP** — `connect-src` entry for `https://www.gstatic.com` (policy is report-only, but the probe shouldn't be the loudest violation).

## Tradeoffs

- The failure reports are now **fire-and-forget behind the probes** (≤2.5s). UI error state is set before the helper is invoked and `finally` cleanup is not delayed; the risk is a user killing the app inside that window loses the report — the same in-memory-queue loss that already existed, widened by up to 2.5s. Accepted: the verdict is the entire point.
- A challenge page resolving the `no-cors` probe is *why* `api_reachable` is trustworthy, but it means the verdict can't distinguish "challenged" from "CORS-less 5xx". Both are our side, not the user's — enough to act on.

## How this resolves the question

Once deployed, the existing Sentry issue for send failures gains `net_probe`. If Hugo's pattern is Cloudflare challenging WebViews, the population will show `net_probe:api_reachable` + `net_error_class:fetch_rejection` + `net_online:true` — and the fix moves to Cloudflare config (exempt `api.peanut.me` from challenge actions). If it's genuinely connectivity, `offline` verdicts confirm #2860's original story.

## Not done here

- `qr-pay` / `withdraw` still have the untagged-capture gap noted in #2860 — same follow-up as before; the helper is ready for them.
- No copy change: `connectionLost` still says "check your connection" even when the verdict says otherwise. Deliberate — copy should follow evidence, not precede it.

## Verification

- `pnpm typecheck` clean
- `network-triage.test.ts` — 12 new tests (verdict matrix, exact-message predicate incl. the wrapped-copy trap, no-cors probe options, tag merging on both capture paths)
- `friendly-error.utils.test.tsx` — 32 passed (unchanged, now exercising the shared predicate)
- `Initial.link.send.view.test.tsx` — 7 passed
- `prettier --check` clean

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -126,6 +126,10 @@ function contentSecurityPolicyReportOnly(includeReporting = true) {
             'https://*.analytics.google.com',
             'https://stats.g.doubleclick.net',
             'https://www.googletagmanager.com',
+            // Network-failure triage probe (network-triage.ts): Android's own
+            // captive-portal endpoint, used to tell "device offline" from "our
+            // edge unreachable" when a fetch dies without a status.
+            'https://www.gstatic.com',
             // Every Google country domain, because the remarketing beacon goes
             // to the user's own — see csp-google-domains.js for why this cannot
             // be a wildcard. Supersedes the former lone 'https://www.google.com'
--- a/sentry.utils.ts
+++ b/sentry.utils.ts
@@ -446,6 +446,11 @@ export function beforeSendHandler(event: ErrorEvent): ErrorEvent | null {
     }
     collapseNoisyFingerprint(event)
     cleanSensitiveHeaders(event)
+    // Whether the device believed it was online at capture time — the free
+    // half of the TASK-21956 network triage (navigator is absent server-side).
+    if (typeof navigator !== 'undefined' && typeof navigator.onLine === 'boolean') {
+        event.tags = { net_online: String(navigator.onLine), ...event.tags }
+    }
     return event
 }
 
--- a/src/components/Send/link/views/Initial.link.send.view.tsx
+++ b/src/components/Send/link/views/Initial.link.send.view.tsx
@@ -13,7 +13,7 @@ import { useWallet } from '@/hooks/wallet/useWallet'
 import { sendLinksApi } from '@/services/sendLinks'
 import { useFriendlyError } from '@/hooks/useFriendlyError'
 import { isAmountWithinBalance } from '@/utils/balance.utils'
-import { captureException } from '@sentry/nextjs'
+import { captureNetworkTriagedFailure } from '@/utils/network-triage'
 import { criticalFlowTags } from '@/utils/sentry-critical-flow'
 import { useQueryClient } from '@tanstack/react-query'
 import { useCallback, useContext, useEffect, useMemo } from 'react'
@@ -130,7 +130,7 @@ const LinkSendInitialView = () => {
                 } catch (error) {
                     // we want to capture any errors here because we are already in the background
                     console.error(error)
-                    captureException(error, {
+                    void captureNetworkTriagedFailure(error, {
                         tags: { ...criticalFlowTags('send-link'), send_link_step: 'persist' },
                     })
                 }
@@ -139,22 +139,24 @@ const LinkSendInitialView = () => {
             // handle errors
             const errorString = toFriendlyError(error)
             setErrorState({ showError: true, errorMessage: errorString })
-            // `error_message` is the LOCALIZED copy the user saw, so it can't be
-            // grouped on — carry the raw class/message alongside it, which is what
-            // actually identifies the failing substep (TASK-21956).
-            posthog.capture(ANALYTICS_EVENTS.SEND_LINK_FAILED, {
-                amount: tokenValue,
-                error_message: errorString,
-                error_name: error instanceof Error ? error.name : 'unknown',
-                error_raw: error instanceof Error ? error.message : String(error),
-            })
             // Tagged, or the noise filters drop it: a send link that dies on a
             // failed fetch matched `networkIssues` in sentry.utils.ts and was
-            // silently discarded, leaving the user on "contact support" and us
-            // with nothing to look at — the exact gap in TASK-21956.
-            captureException(error, {
+            // silently discarded (TASK-21956). Fire-and-forget: the triage
+            // probes may take up to 2.5s and must not hold up `finally`.
+            // `error_message` is the LOCALIZED copy the user saw, so it can't be
+            // grouped on — error_name/error_raw carry the raw class alongside.
+            void captureNetworkTriagedFailure(error, {
                 tags: { ...criticalFlowTags('send-link'), send_link_step: 'create' },
                 extra: { amount: tokenValue, hasAttachment: !!attachmentOptions?.rawFile },
+                analytics: {
+                    event: ANALYTICS_EVENTS.SEND_LINK_FAILED,
+                    props: {
+                        amount: tokenValue,
+                        error_message: errorString,
+                        error_name: error instanceof Error ? error.name : 'unknown',
+                        error_raw: error instanceof Error ? error.message : String(error),
+                    },
+                },
             })
         } finally {
             setLoadingState('Idle')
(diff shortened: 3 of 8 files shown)


innolope-dev:
@coderabbitai review

innolope-dev:
Chip's round-2 finding (**Native API probe is blocked by CORP**) is real and is fixed API-side in peanutprotocol/peanut-api-ts#1456: the global `same-site` default now yields to route-set CORP, and `/healthz` (GET+HEAD) serves `cross-origin` — it exposes no user data by design. No client change needed here.

Deploy-order note: until #1456 is live, native probes will read `net_probe:edge_unreachable` for a healthy edge (the PWA origin is same-site and unaffected). If this PR ships first, treat early native `edge_unreachable` data as unusable until the API deploy lands.

innolope-dev:
Round-3 finding addressed in `9c3f5e5fd` — and you were right that the previous answer was incomplete: parking the fix in api-ts#1456 left this PR's telemetry wrong until that deploy landed, which is not a property a diagnostic should have.

The probe now runs in **both modes** against `/healthz`:

- **cors** — CORP is enforced on `no-cors` requests only, so a cors-mode probe is untouched by the `same-site` header, and `https://localhost` is explicitly in the API's CORS allowlist (`src/app.ts`). This reports a healthy edge from native **today**, with or without #1456.
- **no-cors** — still there, because it resolves on responses the cors probe can't see: challenge pages and CORS-less error pages.

The pair is a better discriminator than either alone, so this added an observation rather than just unblocking one: **`api_cors_blocked`** = cors fails while no-cors succeeds = something answered but the browser wouldn't let the app read it. That is *direct* evidence of the WAF-challenge / CORS-less-error-page failure mode, where before I could only infer it from `api_reachable` + a healthy-looking client.

Two new tests cover exactly the response-policy behaviour you flagged: a CORP-blocked no-cors probe still yielding `api_reachable` via the cors probe, and the answered-but-unreadable case landing on `api_cors_blocked` instead of `edge_unreachable`. 17 passing in the suite.

api-ts#1456 stays worth merging — it makes the no-cors half work on native, which is what turns `api_cors_blocked` from detectable-in-principle into reliably observable — but this PR is no longer gated on it.

innolope-dev:
## Evidence update — the theory in this PR's description is contradicted

Ran the data analysis this PR was built to enable, using the retired native canary (`native-canary.ts` on `main`) as a substitute instrument. **The Cloudflare WAF / CORS theory does not survive it**, and neither does #2860's connectivity theory. Recording it here so nobody reviews this against a description I no longer believe.

### First: Sentry cannot see these failures at all

`title:"*Failed to fetch*"` and `"*Load failed*"` return **0 events in 30d**. `IGNORED_ERRORS.networkIssues` drops them by substring in `beforeSend`, client-side. The only survivor is Firefox's `NetworkError when attempting to fetch resource.` (224 events, 100% `environment:production`, 100% Firefox) — it escapes only because the drop list says `'Network Error'` *with a space*.

So the native population is deleted before it is sent, and **any reasoning from these events' presence or absence is invalid — including #2860's**. Also `has:critical_flow` = **0 events in 30d**, so #2860's escape hatch is not yet producing data.

### The canary: same device, same second, self-normalizing

30d Android, all to `api.peanut.me`:

| probe | transport | reached server | network-error | fail rate |
|---|---|---|---|---|
| `GET /healthz` | WebView | 623 | 692 | **52.4%** |
| `GET /users/me` | WebView | 729 | 686 | **48.2%** |
| `POST /healthz` | WebView | 1300 | 5 | **0.38%** |
| `GET /healthz` | OS HTTP client | 1289 | 20 | **1.5%** |

**iOS: 0% failure on all three.** `online: true` on **2136 of 2136** network-errors. A sampled failure carries `errorMessage: "Failed to fetch"`, `errorName: "TypeError"` — Hugo's exact error.

### Why both theories die

- **Connectivity**: at the same instant on the same device, POST reaches the server 99.6% of the time and the identical GET over the OS HTTP client 98.5%, while the WebView GET fails 52%. A dead radio cannot be method- and transport-selective.
- **WAF / CORS (this PR's premise)**: iOS WebView — same app, same edge, same WebView-class UA — is **0/324**. A UA or fingerprint rule would hit it too. And a challenge or edge error page is an HTTP *response*: the probe would record `http-403`/`http-503`, not `network-error`. Argentina, the best candidate for ASN reputation, fails **below** the Android average (41.5% vs 52.4%) — its 63% share of raw failures is pure volume.

**What it does support:** an Android-WebView / GET-path defect. Confined to Android, to the WebView-proxied transport, and to GET.

### What this means for this PR

The triage still earns its place — it is what makes any of this visible from inside a real send, and `api_cors_blocked` remains the only way to positively rule the edge in or out per-event. But the **expected** verdict is no longer `api_cors_blocked`; on the canary's evidence Android failures should land on `edge_unreachable` or `unknown`, and that would be consistent with the WebView-GET defect rather than the edge.

Two changes already made in response (`cd74467b3`): the canary shows `get-healthz-nocors` failing **108/108 on iOS** while every other iOS probe succeeds 100% — WKWebView serves no opaque responses — so both no-cors probes here always fail on iOS. The old fallthrough would have labelled every iOS failure `offline`, the exact mislabelling this module exists to remove. Nothing-answered now splits on `navigator.onLine` and otherwise reports a new `unknown`.

Filing the Android-GET defect and the drop-list erasure separately — the latter has by far the widest blast radius, since it is currently deleting the entire money-flow failure population.

PR #2869, opened 2026-08-27
feat(telemetry): triage qr-pay and withdraw failures too

innolope-dev:
Follow-up to #2867 (**stacked on it** — merge that first; this diff reads clean against it). Closes the `qr-pay` / `withdraw` gap both #2860 and #2867 flagged as out of scope.

## What I expected to find, and what was actually there

The expected gap was the send-link one: `captureException` untagged, so `IGNORED_ERRORS.networkIssues` silently drops any "Failed to fetch". That was there. Two worse things were not expected:

**1. Two sites stopped reporting entirely when #2860 landed.** `qr-pay` sign and `manteca` withdraw sign only call `captureException` inside the `classified.code === 'genericSupport'` branch:

```ts
} else if (classified.kind === 'code' && classified.code === 'genericSupport') {
    captureException(error)          // ← the only report in this catch
    setErrorMessage(t('errors.signFailed'))
} else {
    setErrorMessage(toFriendlyError(error), ...)   // ← no report
}
```

Before #2860 a `TypeError: Failed to fetch` classified as `genericSupport`, so it was captured (then dropped by the noise filter). After #2860 it classifies as `connectionLost` — which lands in the `else` — so **no event is created at all**. Same net observability (zero), but the fix has to add the report, not just tag it. Nothing regressed for users; it's the diagnosis path that got quieter.

**2. Two money-leg catches never reported to Sentry at all.** `manteca` submit and `crypto` execute emitted only a PostHog event whose `error_message` is the **localized** copy — ungroupable across locales, and nothing in Sentry. A withdrawal failing at the money leg left nothing queryable anywhere.

## Sites wired (7)

| flow | step | was |
|---|---|---|
| qr-pay | `initiate` | untagged capture |
| qr-pay | `sign` | **captured only for genericSupport** |
| qr-pay | `submit` | untagged capture |
| withdraw-manteca | `lock-rate` | untagged capture |
| withdraw-manteca | `sign` | **captured only for genericSupport** |
| withdraw-manteca | `submit` | **no Sentry capture at all** |
| withdraw-crypto | `execute` | **no Sentry capture at all** |

All now go through `captureNetworkTriagedFailure` with `criticalFlowTags(flow)` + a step tag, so the noise filters keep them and the `net_probe` / `net_error_class` observation rides along. The two PostHog-only sites keep their existing event and gain `error_name` / `error_raw` beside the localized string.

## New export: `isNetworkLayerFailure`

The two selective catches must not start reporting everything — their other branches are deliberate non-reports (backend wire codes, or the user declining a passkey). So the `else` branch reports **only** when the error is network-class:

```ts
if (isNetworkLayerFailure(error)) {
    void captureNetworkTriagedFailure(error, { tags: {...} })
}
```

## Deliberately not changed

- The `SessionKeyGrantRequiredError` and `'not allowed'` branches stay silent — a user declining biometric auth is not a defect.
- `PIX_MIN_AMOUNT` in qr-pay stays out of Sentry (existing comment says why: deterministic rail-minimum rejection).
- The `captureMessage` calls in the crypto withdraw's recordPayment paths are untouched — different concern (funds already moved), already reported.
- Camera failures in `QRScanner` / `QRScannerOverlay` are out of scope; #2864 classifies those.

## Verification

- `pnpm typecheck` clean
- `network-triage.test.ts` — 23 passed (6 new for `isNetworkLayerFailure`, incl. the `Failed to fetch charges: 500` false-positive trap and a non-`Error` throw)
- `withdraw-states.test.tsx` + `qr-pay-states.test.tsx` — 63 passed, 3 skipped (unchanged)
- `prettier --check` clean

### DIFF
--- a/src/app/(mobile-ui)/qr-pay/page.tsx
+++ b/src/app/(mobile-ui)/qr-pay/page.tsx
@@ -44,7 +44,8 @@ import { EHistoryUserRole } from '@/hooks/useTransactionHistory'
 import { loadingStateContext } from '@/context/loadingStates.context'
 import { getCurrencyPrice } from '@/app/actions/currency'
 import { PaymentInfoRow } from '@/components/Payment/PaymentInfoRow'
-import { captureException } from '@sentry/nextjs'
+import { captureNetworkTriagedFailure, isNetworkLayerFailure } from '@/utils/network-triage'
+import { criticalFlowTags } from '@/utils/sentry-critical-flow'
 import posthog from 'posthog-js'
 import { ANALYTICS_EVENTS, REFERRAL_SOURCES } from '@/constants/analytics.consts'
 import { isPaymentProcessorQR, EQrType, NAME_BY_QR_TYPE, type QrType } from '@/components/Global/DirectSendQR/utils'
@@ -671,7 +672,9 @@ export default function QRPayPage() {
                     // minimum) — actionable copy, not a Sentry-worthy surprise.
                     setErrorMessage(pixMinAmountErrorMessage)
                 } else {
-                    captureException(error)
+                    void captureNetworkTriagedFailure(error, {
+                        tags: { ...criticalFlowTags('qr-pay'), qr_pay_step: 'initiate' },
+                    })
                     setErrorMessage(t('errors.initiateUnexpected'))
                 }
                 setIsSuccess(false)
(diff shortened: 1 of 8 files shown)


innolope-dev:
@coderabbitai review

innolope-dev:
**CI note for reviewers — updated, superseded by the retarget below.**

~~Once #2867 merges, GitHub retargets this PR's base to `dev` and normal CI applies.~~ That was wrong: GitHub only auto-retargets a stacked PR when the base branch is **deleted**, and `feat/send-failure-network-triage` still exists. #2867 merged at 20:21 UTC and this PR stayed pointed at the stale feature branch, so `tests.yml` (`pull_request: branches: [main, dev, develop]`) and `code-analysis.yml` still never queued.

Fixed by retargeting this PR to `dev` directly. A base change fires `pull_request: edited`, which is not in the default trigger set, so a close/reopen was needed to actually queue the run — full `pull_request` CI and the Chip pass are now running against `dev`.

**Scope grew as a result, and that is deliberate.** This PR now also carries `cd74467b3`, which was pushed to #2867's branch *after* that PR had already merged and was therefore stranded — never reviewed, never CI'd, not on `dev`. It is not cosmetic: on `dev` today the `observed` union has no `'unknown'`, so on iOS — where both no-cors probes always fail because WKWebView serves no opaque responses — any unreachable-API failure is labelled `offline` even when the device is online. That is exactly the mislabelling this module exists to remove.

Commits over `dev`:
- `b5b5e537c` feat(telemetry): triage qr-pay and withdraw failures too
- `d2702f89f` test(withdraw): mock captureException in the crypto confirm suite
- `cd74467b3` fix(triage): never infer offline from a channel iOS does not provide
- `841bee969` merge of the above

Note on the CORP finding from the earlier Chip pass: it was accurate — `https://api.peanut.me/healthz` still returns `cross-origin-resource-policy: same-site` in production. peanutprotocol/peanut-api-ts#1456 fixes that header and is merged but **not yet deployed**. The client does not depend on that deploy; the `cors` probe is unaffected by CORP.

PR #2871, opened 2026-08-27
feat(native): restore the transport canary, failure-only, iOS included

innolope-dev:
## Why

The retired v3 canary (`c45f7655`) answered the one question no per-failure error report can: on Android a WebView **GET** fails while a **POST** and an **OS-client GET** to the same host in the same second succeed. The comparison *is* the evidence — it is what ruled out both the connectivity theory and the Cloudflare/WAF theory.

It is gone from `dev`, so the next release ends the data. That matters more than it looked, because the failure is **release-dependent and the newest reporting bundle is the worst**:

| build | date | ok | failed | fail rate | users |
|---|---|---|---|---|---|
| `d0a9834` (1.0.50) | Aug 8 | 0 | 25 | 100% | 16 |
| `9ceeb94` (1.0.53) | Aug 17 | 14 | 0 | 0% | — |
| `8016c68` | Aug 21 | 201 | 6 | **2.9%** | 4 |
| `d4bd3ab` (ota-1.0.56) | Aug 26 | 68 | 18 | **20.9%** | 10 |

Android only, so this is not the platform confound. The newest bundle fails ~7× more than the Aug 21 one, across 10 distinct users. This is live and possibly regressing, not a historical artifact — and without the canary we cannot tell regression from noise.

Caveat worth stating: releases are not randomized cohorts, and `d4bd3ab`'s n (86) is smaller than `8016c68`'s (207), so some of that gap may be population rather than build. Confirming that is part of what this restores.

## What changed vs the retired version

It was retired for real cost — five `captureMessage` calls per launch on every platform (~550 events/day, all `level:info`) plus five extra cold-start round trips. Those reasons were legitimate, so this is not a straight revert:

- **One event per launch, and only when a probe actually failed.** A clean launch reports nothing, so steady-state cost is zero. Denominator comes from PostHog app opens rather than success events.
- **Three probes, not five.** `get-users-me` added an auth variable to a reachability question; `get-healthz-nocors` is unusable (below).
- **`level:warning`**, so it can never rejoin the `level:info` flood that made the original a deletion target.

Estimated ~20–40 events/day against the old ~550.

## Why there is no no-cors probe

This runs on **iOS as well as Android**, which is exactly why that probe is gone. WKWebView serves no opaque responses at all — the retired canary measured it failing **108/108 on iOS** while every other probe in the same batch succeeded 100%. Including it would report a failure that did not happen on every iOS launch.

It is separately blocked by `Cross-Origin-Resource-Policy: same-site` on `/healthz` from the native `https://localhost` origin. Verified against production while writing this:

```
$ curl -sSI https://api.peanut.me/healthz
HTTP/2 200
cross-origin-resource-policy: same-site
```

peanutprotocol/peanut-api-ts#1456 fixes that header and is **merged but not yet deployed**. Nothing here depends on that deploy — all three probes are ordinary CORS requests and mean the same thing on both platforms.

## Follow-up, not included here

PostHog's registered device context (`src/i18n/app/locale-store.ts`) carries `platform` but **not** the app version, so the per-build split in the table above is not reproducible from PostHog as it stands. Adding `app_version` to that `posthog.register` call would close it. Left out deliberately to keep this PR to one concern.

## Test plan

- `src/utils/__tests__/native-canary.test.ts` — 6 tests: silence on all-success, silence on non-2xx (a completed request is not a transport failure), the Android GET-fails/POST-ok/native-ok shape including the `net::` code capture, abort→timeout classification, total outage, and an assertion that exactly three probes fire and **none** is no-cors.
- `pnpm typecheck` clean.

### DIFF
--- a/src/config/peanut.config.tsx
+++ b/src/config/peanut.config.tsx
@@ -10,6 +10,7 @@ import 'react-tooltip/dist/react-tooltip.css'
 import { isCapacitor, getNativeRpId } from '@/utils/capacitor'
 import { authReady } from '@/utils/auth-token'
 import { installNativeAuthCapture } from '@/utils/native-auth-capture'
+import { scheduleTransportCanary } from '@/utils/native-canary'
 import { installCeremonyTelemetry } from '@/utils/webauthn-ceremony-telemetry'
 // Note: Sentry configs are auto-loaded by @sentry/nextjs via next.config.js
 // DO NOT import them here - it bundles server/edge configs into client code
@@ -35,6 +36,7 @@ export function PeanutProvider({ children }: { children: React.ReactNode }) {
         if (isCapacitor()) {
             void authReady() // start Preferences hydration before any API call needs it
             installNativeAuthCapture()
+            scheduleTransportCanary()
             import('@capgo/capacitor-passkey').then(({ CapacitorPasskey }) => {
                 const nativeRpId = getNativeRpId()
 
--- a/src/utils/__tests__/native-canary.test.ts
+++ b/src/utils/__tests__/native-canary.test.ts
@@ -1,8 +1,8 @@
 import * as Sentry from '@sentry/nextjs'
-import { runCanary } from '../native-canary'
+import { runCanary, scheduleTransportCanary } from '../native-canary'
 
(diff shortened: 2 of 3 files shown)


innolope-dev:
@coderabbitai review

PR #2887, opened 2026-08-31
feat(ota): five-tap beta switch that puts a tester on the staging channel

innolope-dev:
## Why

Every merge to `dev` publishes a bundle to the Capgo `staging` channel and **no device ever reads it**: `production` is the app's default download channel, the app never calls `setChannel()`, and `staging` has self-assignment turned off. So an OTA today is all-or-nothing — internal testing on a real device has no path at all.

## What

**Profile → About → five taps on the version line** reveals a Beta-updates switch:

- **on** → `setChannel('staging')` + an immediate update check, so the next launch runs the beta bundle;
- **off** → `unsetChannel()` **and** `reset()` back to the store bundle. Staging versions (commit-count band) outrank every production one, so no production OTA could ever replace a beta bundle — dropping to builtin is the only exit, and leaving it out would strand testers on beta forever;
- the row also prints the running bundle and the Capgo device ID (tap to copy), which is what someone needs to force a device onto a channel from the dashboard.

Hidden behind the tap gesture and behind `isNativeBridge()`, so it never renders on web and can't be reached by accident.

A refused channel is reported as its own case (`OtaChannelClosedError`) with a message naming the fix, because until an admin flips the toggle below, **that is what every tester will hit**.

## Dashboard prerequisite (not in this PR)

`staging` → Information → **"Allow devices to self dissociate/associate"** must be enabled in Capgo; it is off today, and changing it needs a Super Admin. Until then the switch shows "ask an admin to allow self-assignment".

Unrelated but adjacent, also for a Super Admin: the app's **default upload channel is `ios-mobile-release`**, a dead lane from the retired `feat/mobile-release` branch. Nothing in CI uploads without `--channel`, so it is inert, but it should point at `staging` before that channel and `android-mobile-release` are deleted. Default *download* channel is correctly `production`.

## Tests

- `capgo-updater`: join stages a bundle, a rejected/`error`-carrying `setChannel` surfaces as a closed channel, leave unsets **and** resets.
- `About.view`: four taps reveal nothing, the fifth does, and a lapsed window forgets the streak.
- `BetaUpdatesCard`: hidden off native, closed-channel and restart-needed toasts.

`docs/NATIVE-RELEASE.md` §9 gains the tester path and its two prerequisites.

### DIFF
--- a/docs/NATIVE-RELEASE.md
+++ b/docs/NATIVE-RELEASE.md
@@ -331,12 +331,18 @@ from the dashboard instead.
 
 Three prerequisites, all one-time:
 
-- The account must be in the **`beta-ota-channel`** PostHog cohort. The tap gesture only
-  hides the switch; this is what decides who may use it, because `staging` runs unreleased
-  code against production money and self-assignment is open to any device once enabled.
-  Non-prod builds bypass the flag so QA doesn't need a cohort edit.
+- The account must be in the **`beta-ota-channel`** PostHog cohort, which is what keeps
+  the switch off customer devices. Read what it is, though: a client-side flag that
+  decides whether a React component renders. `setChannel` talks to Capgo directly with
+  the app key shipped in every binary, and non-prod builds skip the flag entirely, so
+  anyone running a modified client can self-assign regardless. It stops people who did
+  not mean to be on beta, not people who do. Move the join behind the backend (verify
+  the account server-side, assign the device through Capgo's API) if that ever needs to
+  be a real boundary.
 - `staging` must have **"Allow devices to self dissociate/associate"** enabled, or
-  `setChannel()` is refused and the app says so. Changing it needs a Super Admin.
+  `setChannel()` is refused and the app says so. Changing it needs a Super Admin — and
+  it is the decision that actually opens the channel: from that point any install that
+  can call the plugin can join, whatever the cohort says.
 - The tester's binary must not outrank the bundle: `disable_auto_update_under_native`
   makes a device on versionName 1.1.x refuse anything sorting below it. Staging's
   commit-count band sits far above production's, so this only bites right after a native
--- a/docs/NATIVE-RELEASE.md
+++ b/docs/NATIVE-RELEASE.md
@@ -352,6 +352,13 @@ Leaving the channel unsets it **and** calls `reset()` back to the store bundle: 
 versions outrank every production one, so no production OTA could ever replace a beta
 bundle on its own.
 
+One asymmetry to know before forcing anyone from the dashboard: `unsetChannel()` is
+local-only in the plugin (it drops a stored preference and returns ok), so a **dashboard
+device override survives it** and Capgo keeps serving beta. The switch detects that —
+it re-reads the effective channel and says an admin has to remove the override rather
+than resetting into an exit that undoes itself on the next launch — but the removal is
+dashboard-side. Self-assignment is the cleaner enrolment path for that reason.
+
 ---
 
 ## 10. Pre-submission verification
--- a/docs/NATIVE-RELEASE.md
+++ b/docs/NATIVE-RELEASE.md
@@ -329,8 +329,12 @@ Beta-updates switch that calls `setChannel('staging')` — no dashboard work per
 the row also prints the device ID for the times someone has to be forced onto a channel
 from the dashboard instead.
 
-Two prerequisites, both in Capgo, both one-time:
+Three prerequisites, all one-time:
 
+- The account must be in the **`beta-ota-channel`** PostHog cohort. The tap gesture only
+  hides the switch; this is what decides who may use it, because `staging` runs unreleased
+  code against production money and self-assignment is open to any device once enabled.
+  Non-prod builds bypass the flag so QA doesn't need a cohort edit.
 - `staging` must have **"Allow devices to self dissociate/associate"** enabled, or
   `setChannel()` is refused and the app says so. Changing it needs a Super Admin.
 - The tester's binary must not outrank the bundle: `disable_auto_update_under_native`
--- a/docs/NATIVE-RELEASE.md
+++ b/docs/NATIVE-RELEASE.md
@@ -357,7 +357,9 @@ local-only in the plugin (it drops a stored preference and returns ok), so a **d
 device override survives it** and Capgo keeps serving beta. The switch detects that —
 it re-reads the effective channel and says an admin has to remove the override rather
 than resetting into an exit that undoes itself on the next launch — but the removal is
-dashboard-side. Self-assignment is the cleaner enrolment path for that reason.
+dashboard-side. Self-assignment is the cleaner enrolment path for that reason. When Capgo
+cannot be reached to answer that question, the switch refuses to reset at all and asks the
+tester to retry online: only a confirmed answer licenses the reset.
 
 ---
 
--- a/docs/NATIVE-RELEASE.md
+++ b/docs/NATIVE-RELEASE.md
@@ -321,6 +321,27 @@ own `out/` under the binary's versionName, then assert the channel serves it.
 - **Boundary:** OTA ships web assets only. New plugins, Gradle, permissions, versionCode
   → Play release.
 
+### Internal testing (the `staging` channel on a real device)
+
+Every merge to `dev` already publishes to `staging`; the missing half was a way for a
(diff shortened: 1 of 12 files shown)


innolope-dev:
Addressed all three Chip findings in 97a0f8a.

**MAJOR — five taps exposes staging to every customer.** Correct, and the tap gesture was never an eligibility check. The card is now gated on the **`beta-ota-channel`** PostHog cohort (`nonProdBypass`, so previews and local builds still show it without a cohort edit). The gesture hides it; the flag decides who may use it. Device-ID forcing from the dashboard stays available for anyone outside the cohort — the row prints the ID for exactly that.

**MAJOR — failed beta downloads reported success.** Fixed. `checkAndStageUpdate()` now returns `staged | up-to-date | failed`, `joinBetaOtaChannel()` passes it through, and the switch has a distinct message per outcome: restart-now only after `next()`, "nothing newer right now" for an up-to-date join, and a warning when the channel was joined but no bundle could be fetched — which is where the `disable_auto_update_under_native` case lands. Tests cover the version-gate and download-failure paths.

**MINOR — every failure labeled as a closed channel.** Fixed. Only the private-channel codes (`channel_private`, `disabled_by_config`, `channel_not_found`, `cannot_set_channel`) map to `closed`, read from both the `CapacitorException` `data.error` code and the message; everything else falls through to `failed`. Added a network-error test and one for the `data.error` shape.

`docs/NATIVE-RELEASE.md` now lists the cohort as the first of three prerequisites.

innolope-dev:
Second review pass addressed in 538da04.

**MAJOR — cohort removal hides the only beta exit.** Right, and it was the worst version of the bug: a device already on `staging` cannot be recovered by any production OTA, so hiding the card on a cohort miss stranded the tester permanently. The card now renders whenever `isBeta` is true regardless of the flag; only *joining* needs the cohort (the toggle is disabled for a non-cohort account that is already off). Test covers the offboarded-tester case.

**MAJOR — docs presented a client-side flag as access control.** Fair, and the doc now says the opposite of what it said: `setChannel` talks to Capgo directly with the app key shipped in every binary, non-prod builds skip the flag entirely, so **enabling self-assignment is the decision that opens the channel** and the cohort only keeps unintended devices off it. Routing the join through the backend is named as the option if that ever needs to be a real boundary.

**MINOR — Android private-channel codes fall through.** Confirmed against the plugin source (`CapacitorUpdaterPlugin.java:1227`, `CapacitorUpdaterPlugin.swift:930`): added `cannot_update_via_private_channel` and `channel_self_set_not_allowed`, dropped `cannot_set_channel` which does not exist. Parameterized test over all three real codes.

**MINOR — half-failed leave reads as a clean exit.** `reset()` is now retried once, and a second failure throws `OtaResetFailedError` → `left-still-beta`, which tells the tester the app is still on the beta build and has to be reinstalled. That state is unrecoverable by OTA, so it had to stop reading as success.

**MINOR — unhandled rejection copying the device ID.** Now async/try-catch with a failure toast.

innolope-dev:
**MINOR — join fires an unguarded concurrent update check.** Real, fixed in 869bb11.

The window is wider than the 5s delay suggests: the launch check can still be *downloading* when the tester flips the switch, and both paths called `checkAndStageUpdate()`, so both could reach `next()` — the bundle that boots is whichever write lands last, possibly the one resolved against the channel the device just left.

Checks now queue through a module-level `pendingCheck`. Deliberately a queue rather than a shared in-flight promise: the join needs a check made *after* its `setChannel`, so handing it the launch check's verdict on the old channel would have been the wrong fix — it would report `up-to-date` for production and never look at staging at all.

Test drives the real ordering: hold the launch check unresolved, join, assert no second `getLatest` starts, release it, then assert the beta bundle is the one staged.

The trade-off, stated plainly: a join during a slow bundle download now waits for it, showing as the switch's busy state, instead of racing it.

innolope-dev:
Both new findings are right — fixed in c739c0f.

**MAJOR — leave races an in-flight check.** Same hazard as the join, worse outcome: a check that already chose its bundle against `staging` can call `next()` after the reset cleared it, so the device boots beta code with the channel unset — exactly the state `reset()` exists to prevent. `queueUpdateCheck` is now a general `queueOtaWork`, and `leaveBetaOtaChannel` runs through it, so unset + reset can never interleave with a check. Test holds a launch check unresolved, calls leave, asserts `unsetChannel` has not fired, then releases it.

**MINOR — a leave with no reload was silent and snapped the toggle back on.** Also right, and the path is easy to hit: join while up-to-date (nothing downloaded), toggle straight off, `reset()` resolves without reloading because the device is already on the builtin bundle. `setBeta` now refreshes the status after a leave, and `left` gets its own confirmation toast — unobservable in the usual reload case, correct in this one.

For the record on the first comment in this thread (the pre-`538da04` "five taps exposes staging to every customer"): the cohort gate landed in 97a0f8a and the docs correction — that the cohort is not a security boundary and enabling self-assignment is what opens the channel — in 538da04.

innolope-dev:
Correct, and worse than described — fixed in 1f50c4f.

I checked the plugin source rather than assuming: in 8.45.9 `unsetChannel` never contacts the backend at all. iOS (`CapgoUpdater.swift:1632`) and Android (`CapgoUpdater.java:1324`) both just drop the persisted `defaultChannel` key and return `ok`. `setChannel`, by contrast, POSTs to the channel endpoint. So *any* server-side assignment outlives the unset — not only a dashboard override — and the app has no way to clear one.

The leave now re-reads the effective channel with `getChannel()` (a PUT to the same endpoint, so it reports what Capgo will actually serve). If it still says `staging`, it raises `OtaChannelOverrideError` → `left-override`, and the toast says an admin has to remove the device's channel override.

Deliberately **no reset in that branch**: resetting would clear the bundle, reload the app — taking the explanation with it — and the next launch would stage beta again. A tester who cannot leave should be told why, not shown an exit that silently undoes itself. When `getChannel()` fails (offline), the leave proceeds to reset as before rather than blocking on a network round-trip; test covers that path too.

`docs/NATIVE-RELEASE.md` now flags the asymmetry: self-assignment is the cleaner enrolment path precisely because dashboard-forced devices can only be released from the dashboard.

innolope-dev:
Re-raise acknowledged and fixed in e38ff13 — the previous fix left exactly the hole you name.

`getChannel().catch(() => null)` fell through to `reset()`, so a dashboard-forced tester who was offline got reloaded onto the store bundle, the reload ate the explanation, and the surviving override routed them back to beta on the next check. That is the failure the check was added to prevent, reintroduced through its own error path.

Now only a confirmed answer licenses the reset. An unreachable or error-bearing response raises `OtaChannelUnknownError` → `left-unconfirmed`: no reset, the switch stays on, and the toast asks the tester to retry online.

Replaying one adjacent variant found a second instance of the same shape: the plugin returns `{ error: 'rate_limit_exceeded' }` **with no channel field**, which my `effective?.channel === BETA` test read as "clear" and would also have reset. The gate is now `!effective || effective.error` → indeterminate; an empty `channel` on a clean response still resets, which is the normal self-assigned exit. Both cases are covered by a parameterized test asserting `reset` is not called, plus one asserting it still is once Capgo confirms no assignment.

innolope-dev:
Right, and it is the residual I had accepted rather than fixed. Fixed in 406bafe.

The half-finished exit is the worst state on this surface: channel cleared, beta bundle still running, and — because `isBeta` was read off the channel alone — the card gone for anyone outside the cohort, on a device no production OTA can reach by definition. Silent, permanent, and invisible to the one control that could undo it.

- The attempt is recorded in local storage **before** `unsetChannel`, since everything after that point can fail and nothing else would remember an exit is owed.
- `isBeta` is now `channel === staging || pendingExit`, so the card and its retry survive a restart, a cohort removal, and a cleared channel.
- The marker clears on the first status read where the running bundle is `builtin` **and** the channel is no longer beta — the channel alone cannot prove the exit finished, which is what `onBuiltinBundle` is for.
- It also survives a failed `reset()`, which previously reported `left-still-beta` and then left nothing behind to retry from.

Four tests on the marker (recorded before the clear, kept on an unreadable channel, kept on a failed reset, cleared on a reset that resolves) plus a new `useOtaChannel` suite for the three hook states.

One deliberate leftover, stated rather than narrowed: nothing retries the exit automatically on launch. It waits for the tester to tap the switch again, which the card now guarantees is visible. An automatic retry would reload the app during startup, and I would rather not add that to the boot path in this PR.

PR #2914, opened 2026-09-01
fix: bounded read in ds-shots publisher — CodeQL race (TASK-22044)

kushagrasarathe:
Mirror of the CodeQL fix pushed to the main hotfix PR #2912 (`js/file-system-race` on the report size gate): a single `readSync` capped at MAX_BYTES+1 replaces the stat-then-read pair — no TOCTOU, and a hostile oversized file is never fully read into memory. Plus one jest case for the oversized-file path (24 total).

Files are byte-identical to the #2912 branch, keeping dev and main aligned so the next dev→main release merges clean.

**Task:** TASK-22044

Screenshots: N/A (CI tooling only).

### DIFF
--- a/.github/workflows/ds-shots-comment.yml
+++ b/.github/workflows/ds-shots-comment.yml
@@ -28,20 +28,22 @@
 #   any operational API failure               → red job, comment untouched
 #
 # Supersession is enforced by head-SHA binding, not by cancellation. The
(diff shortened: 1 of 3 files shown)


kushagrasarathe:
Superseded by the merged main hotfix #2912 — these exact bytes now reach dev via the back-merge PR instead of this mirror. Closing.

PR #2915, opened 2026-09-01
fix(profile): stop claiming a Brazilian bank-transfer rail in Unlock payments

innolope-dev:
Brazil has no bank-transfer rail — the active catalog seeds `MANTECA × PIX_BR` only (`peanut-api-ts/scripts/seed-rails.ts`; `BANK_TRANSFER_BR` was never seeded). Chip flagged this claim class on #2903, where the signup residence copy was fixed; this PR fixes the same stale claims on the profile Unlock payments screen.

Three row labels change in `profile.unlockPayments.rows` (all four catalogs, `es-AR` overrides included):

| key | before | after |
|---|---|---|
| `saBank` | PIX, QR & bank transfers (Brazil & Argentina) | PIX (Brazil), QR & bank transfers (Argentina) |
| `pixBank` | PIX & bank transfers (Brazil) | PIX payments & transfers (Brazil) |
| `brBank` | Bank transfers (Brazil) | PIX transfers (Brazil) |

`brBank` wasn't in the original finding but is the same falsehood in the same block: it's the Manteca bank-side row shown to Bridge-verified users who already hold BR QR, and what that unlock actually delivers for Brazil is PIX. Argentina's rows (`arBank`, `arQrBank`, `arQr`) are untouched — AR genuinely has bank transfers and MercadoPago QR.

Where the labels render: `unlock-payments.utils.ts` builds the rows, `UnlockPayments.view.tsx` renders `rows.{labelKey}` and feeds the same label into the Unlock modal title. Nothing keys on the wording; no keys added or removed.

Verified: full i18n messages suite green (key-set parity across catalogs + the duplicate-value drift test — the new English values are unique, so no cross-key translation coupling), prettier run, and no code or test references the old literals.

### DIFF
--- a/src/i18n/app/messages/en.json
+++ b/src/i18n/app/messages/en.json
@@ -359,10 +359,10 @@
             "rows": {
                 "p2p": "Peanut-to-Peanut payments",
                 "card": "Peanut card",
-                "saBank": "PIX, QR & bank transfers (Brazil & Argentina)",
-                "pixBank": "PIX & bank transfers (Brazil)",
+                "saBank": "PIX (Brazil), QR & bank transfers (Argentina)",
+                "pixBank": "PIX payments & transfers (Brazil)",
                 "pixQr": "PIX QR payments (Brazil)",
-                "brBank": "Bank transfers (Brazil)",
+                "brBank": "PIX transfers (Brazil)",
                 "arQrBank": "QR payments & transfers (Argentina)",
                 "arQr": "QR payments (Argentina)",
                 "arBank": "Bank transfers (Argentina)",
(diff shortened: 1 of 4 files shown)


innolope-dev:
Re the review's MINOR (signup residence comparison still says "PIX & bank transfers"): acknowledged, and deliberately not fixed on this branch. #2903 owns the residence step and already corrects it — commit 5f48301f5 changes `compare.items.pix` to "PIX" in all four catalogs with the matching unit-test update. Duplicating the one-line change here would guarantee a merge conflict on Residence.test.tsx for whichever PR lands second (and repo rules make branch-side conflict resolution the only way out). Whichever order these merge, dev ends up with the corrected string.

PR #2925, opened 2026-09-02
fix(app): DS review critical findings — rails summary, re-verify, country input, locale, forced logout

innolope-dev:
## What

Critical/high items from the staging DS review — TASK-22114 ([Notion](https://app.notion.com/p/3cf8381175798170b443e3331ad27884)). The design-system items ship separately in `fix/task-22114-ds-review-design`; the backend half (activation rule + `POST /users/identity/restart`) is peanut-api-ts `feat/task-22114-activation-restart-identity`.

| Finding | Fix |
|---|---|
| Residence summary omitted SEPA / USD / GBP for Portugal and Brazil | `residenceAvailability` is now additive over the same Bridge/Manteca knowledge Unlock payments renders: Bridge-served residences list EUR SEPA, GBP Faster Payments and USD ACH/Wire; BR adds PIX, AR its QR rail, MX SPEI. A cross-check test asserts every listed rail is a live row in `buildUnlockGroups` for that residence, so the two screens cannot disagree again. |
| "Save & verify" opened a "Not available yet / Not Found" popup | The button is now **Submit documents** and a failed re-verification reads as retriable ("Verification couldn't start" + Try again → `handleRestartIdentity`). The 404 itself is the missing backend route, added in the api-ts PR. |
| Scrolling the country list pulled to refresh; dropdown instead of input | New `CountryCombobox` (typeahead over the residence option list, inline listbox, keyboard nav, diacritic-insensitive filter) used in onboarding and the change-residence modal. The list, every `BaseSelect`, and the setup layout now use the same pull-to-refresh guards as the app layout. No fetch happens on selection. |
| "Make your first payment" opened the QR scanner and promised "any @username" | Routes to `/send`; copy says what the app accepts (Peanut user, ENS name, wallet address). Completion is the backend `isActivated` flag, widened in the api-ts PR. |
| Policy titles on About were English in every locale | Titles live in `profile.about.policies.*` (4 locales) via a shared `LEGAL_POLICIES` constant; the re-consent modal reads the same keys. |
| Help Center reset the PWA to English | `DocsLink` navigates same-tab in a standalone PWA (no `_blank` round-trip through `start_url`), and the startup-resolved locale is persisted so a full document load re-derives correctly. |
| Unexpected logout at 23:15 Lisbon | Root cause from Sentry: `kernelClient.context.tsx` called `logoutUser()` after 3 failed RPC client builds (9 hits in 30 days, 8 production users). Now only a stale-key error logs out; transient failures keep the session, report to Sentry, and rebuild on `online` / foreground / a 60 s fallback (`reconnect.utils.ts`). |

## Tests

New/updated: `residence-availability`, `CountryCombobox`, `ResidenceChangeModal`, `Residence`, `UnlockPayments` (retriable re-verify title), `GettingStartedChecklist`, `About.view`, `ReConsentModal`, `DocsLink.standalone`, `locale-store`, `reconnect.utils`. Typecheck clean, ds-lint ratchet unchanged, prettier clean.

## Not verified in a browser

The change-residence modal (needs a live API) and the combobox clear-button placement after its last fix are unit-tested only. Please eyeball the residence step and the modal on staging.

## Merge note

Touches `UnlockPayments.view.tsx`, `Residence.tsx`, `About.view.tsx` and the message catalogs, as does the design PR — whichever merges second needs a rebase.

## Added after review

- **Unlock payments rebuilt on the design system** (option D of the Unlock Payments Rebuild study): PageStack shell, residence anchor as a `ListItem` + `StatusBadge`, outage / in-review notices as `Notification`s, each region a `Section` over a `ListGroup` of `ListItem`s (IconBubble leading, StatusBadge trailing, disabled rows for unavailable rails), and a per-section limits `Card` (P2P no-limit line, per-transfer cap and monthly allowance with locale grouping — no more `$100000`). The "Payment limits" link is gone; limits live next to the methods they govern.
- **Home checklist is the empty state**: hidden once money has moved on the account (product decision). Its "Make your first payment" row completes from the new `firstPaymentAt` on `/users/me` (any outgoing peer payment), while activation itself stays on the canonical card/QR spend rule.
- Review follow-ups: kernel-client failure branch has its own test; the residence re-verification carries the new country's intent (validated client-side, wire-tested) into `POST /users/identity/restart`; Colombia maps to ROW.


innolope-dev:
Second round addressed on `fix/task-22114-ds-review-critical`:

- **Colombia** no longer maps to LATAM (`regionIntentForResidence('CO') === 'ROW'`, tested) until its Manteca rail is back.
- **Server action validates the intent**: `restartIdentityVerification` only forwards `LATAM | ROW | EU | NA`; anything else is dropped before the request. A wire-level test in `sumsub-refusals.test.ts` pins the JSON body for the intent, the no-intent and the bogus-intent cases.
- **Backend half**: peanut-api-ts #1492 validates and honours `regionIntent` for all four values on `POST /users/identity/restart` (TypeBox schema, 400 on anything else) and mints the level from it. The two land together.
- **Funded activation steps**: keeping `GettingStartedChecklist` hidden once money has moved is a product decision from Slava for this task — the checklist is the Home empty state and gives way to the activity list. The "first payment" fact itself moved to a dedicated `firstPaymentAt` on `/users/me` (any outgoing peer payment, payer-keyed), so activation and Rewards stay on the canonical spend rule.
- **product/countries.md drift**: agreed the doc is the stale side (REGION_RAIL_MAP enrols the four Bridge rails for both EU and NA intents); that file lives in the mono repo, so it is a follow-up there rather than in this PR.

Also in this push: the Unlock payments screen is rebuilt on the design system (option D of the Unlock Payments Rebuild study — PageStack, Section, ListGroup/ListItem, StatusBadge, per-section limits Card, Notifications for the outage and in-review notices), which retires the hand-rolled group boxes and pills the earlier "Payment limits" comment touched.

innolope-dev:
Third round addressed on `fix/task-22114-ds-review-critical`:

- **UK residence → ROW**: `regionIntentForResidence` consults `RESTRICTED_RESIDENCE_ISO2` before Bridge's document map, so GB (and the rest of the block) never targets a Bridge level; tested.
- **`firstPaymentAt` producer**: it is emitted by peanut-api-ts #1492 (`GET /users/me` → `firstPaymentAt`, payer-keyed off the ledger). The pair lands together; the FE falls back to `activationMilestone === 'activated'` when the field is absent.
- **Funded steps / checklist**: unchanged on purpose — the checklist is Home's empty state by product decision (TASK-22114); the remaining first-payment / card guidance lives in the carousel and the Card entry points. Noted in a code comment so it does not read as an oversight.
- **Signup rail promise**: the summary follows peanut-api-ts `REGION_RAIL_MAP`, which enrols SEPA/ACH/FPS/SPEI for both EU and NA intents; the product/countries.md matrix is the stale side (as this review's previous round concluded). UK residents still read banking-unavailable.

Also in this push: the Home **Get started** checklist is rebuilt on the DS (`Section` › `ListGroup` › `ListItem`, `StatusPill status="completed"` as the done marker) with the same items, copy, completion logic, taps and analytics.

innolope-dev:
Fourth round addressed in `0bad49e6`:

- **Funded accounts keep an activation CTA** (`ActivationCTAs.tsx`). The checklist stays the empty state, but `outbound` and `card` now fall through to the single remaining step card instead of returning `null`, so a funded-but-unactivated account still has a route to the spend that activates it — and to Rewards. Its copy also had to move: the CTA routes to `/send` now, so `home.activation.steps.outbound.description` names a Peanut user / ENS name / wallet address in all four locales instead of "Pix and MercadoPago QR codes". Tests pin both step cards and the `/send` push for a user without card access.

- **Signup rails are derived from the residence's verification intent** (`residence-availability.ts`). `bankRailsFor` no longer asks "is this country in Bridge's document map" — it reads `INTENT_RAILS[regionIntentForResidence(iso2)]`, an FE mirror of `REGION_RAIL_MAP` in peanut-api-ts `src/kyc/rails.consts.ts`, with SPEI and the LATAM QR rails narrowed to their own residence because the provider issues those accounts to one country only. The promise is now residence-scoped by construction: GB, sanctioned and Bridge-unserved residences map to `ROW` and get `['bank']`, so `LOCAL_RAIL.GB = 'gbpFps'` — the contradicted promise flagged earlier — is gone rather than merely unreachable. For an EU/NA residence the set is unchanged, and that is the enrolment contract, not an inference: `REGION_RAIL_MAP.EU` and `.NA` both enrol `ACH_US, SEPA_EU, SPEI_MX, FASTER_PAYMENTS_GB`, and `isUkResident` is the only residence gate on Bridge rails (`src/kyc/capabilities/resolver.ts`). `product/countries.md` is the side that is stale here; that doc fix is filed separately.

- **The cross-check test says what it proves.** It is a drift guard between signup and Unlock payments, not evidence that a rail serves a residence — that now comes from the intent assertions above. It also sweeps every country `buildResidenceCountryOptions` offers rather than eleven hand-picked ones.

- **Same class as the GB block, pre-empted**: a Bridge *banking* exclusion (JP, DZ, BI, TN) also routes to `ROW`. A Bridge or Manteca level for those residences could only end on a terminal rejection.

Also folded in the setup DS nits from `fix/ds-ui-nits-22121` that only existed on that branch (thanks @kushagrasarathe): `FieldError` for the notify-email validation, 44px hit areas + focus rings on the underlined links, centered "Choose a different country" with a `gap-4` CTA stack, the chrome's heading/body spelling, `space-y-2` on the compare-card list, and background/foreground tokens in `CountryCombobox` — applied to `BaseSelect` too, so the two listboxes stay identical.

Typecheck, prettier and ds-lint clean; ds-lint counts are at or below baseline on every metric.

innolope-dev:
Fifth round addressed in `dca3f5cf`.

**Route the funded CTA to an activation spend** (`ActivationCTAs.tsx`) — fixed, you were right and the previous round made it worse.

`GET /users/me` marks an account activated on exactly two intents: `CARD_SPEND_AUTH`, or `QR_PAY` with `provider: MANTECA` (`src/routes/user/get-user.ts`, the `firstSpendIntent` query). A `/send` is volume. So the fourth round handed a funded user without card access a card whose CTA could never clear it, and which would come back after every payment they made.

- Without card access the CTA now opens the QR scanner — the one activating spend that cohort has — and the copy goes back to naming Pix and MercadoPago QR codes in all four locales. With card access the card/QR chooser is unchanged.
- The rail is checked, not assumed: `operationStatus(rail.id, 'pay') === 'enabled'` on a `qr-only` rail, so a Manteca pool-tier rail that reads `enabled` overall while `pay` still needs an upgrade does not keep the card alive.
- A funded user with **neither** card access nor a QR rail has no activating spend at all. That step now renders nothing and the activity list keeps the slot. It still satisfies what the fourth round was for — a funded account keeps a route *to the spend* — while dropping the card in the one case where no such route exists. Six tests pin it: QR scanner not `/send`, chooser for card users, nothing for bank-only, nothing for a QR rail whose `pay` op is blocked.

`product/activation-funnel.md` in mono describes the fall-through; the narrowing needs a matching edit there, filed separately.

**Stale-key logout leaves the rejected passkey in user preferences** (`kernelClient.context.tsx`) — fixed, correct as written. All three forced-logout paths (address mismatch in `initializeClients`, `assertClientOwnedByUser`, and both `isStaleKeyError` branches) now go through one `purgeStoredCredential()` before `logoutUser()`. `logoutUser` already hard-navigates to `/setup`, which is the login/recovery choice, so the remaining half was the credential. Test asserts `updateUserPreferences('u1', { webAuthnKey: undefined })` on the AA24 path.

**Scope advertised rails to the residence** (`residence-availability.ts`, `Residence.tsx`) — taken as the presentation fix, not the model narrowing. The enrolment claim holds: a Bridge customer is created with all four endorsements and Peanut only marks Bridge KYC approved once all four are `APPROVED`, so `REGION_RAIL_MAP.EU` really does open SEPA/FPS/ACH from one verification. What was missing is your second point — those rails only pay out into an account on their own network. A card whose set spans more than one currency now carries that as a condition: *"Transfers in another currency need a bank account in that currency."* Deliberately currency-level rather than per-residence home rails: naming a home currency per country would need a eurozone taxonomy the app has no other use for, and Portugal-vs-Poland is exactly where such a map drifts.

**Ship the restart route before calling it** (`sumsub.ts`) — merge-order item, no code change. The `/users/identity/restart` call is already on `peanut-ui` `dev` (`src/app/actions/sumsub.ts:140`); this head only adds an optional `regionIntent` body, which a backend that predates it ignores, and `RestartIdentityResponse.regionIntent` is optional for the same reason. So this is not a new dependency — but the route genuinely does not exist on api-ts `dev` yet, and peanut-api-ts #1492 is the fix for both halves. It merges and deploys before this one. A 404 today surfaces as the `errorRestartFailed` message rather than a crash.

**Expose firstPaymentAt before consuming it** (`GettingStartedChecklist.tsx`) — no code change; the fallback is already there and the row is unreachable in the state described. `done: milestone === 'activated' || hasSentPayment`, so an absent `firstPaymentAt` falls back to the milestone. And the checklist is only rendered by `ActivationCTAs` when the user is *not* funded (`isFundedNotActivated` returns the step card instead), so a user who can complete a `/send` is never looking at this row. #1492 still ships first.

Typecheck 0, prettier clean, ds-lint ratchet green on every metric (`offScaleSpacing 186/204`, `iconOffScale 74/81`, rest unchanged). 323 tests pass across Home, Setup/Views, context, utils and i18n.

innolope-dev:
Sixth round addressed in `5e8358c5`. One new finding, fixed; three re-raises, answered below with the same evidence as before.

**`hasQrSpendRail` excludes Pix, the rail it names** (`ActivationCTAs.tsx`) — fixed, and you're right about the mechanism as well as the outcome.

`METHOD_CHANNELS` (peanut-api-ts `src/kyc/capabilities/channel.ts`) puts `PIX_BR` in `bank` — its own comment says "Bank-shaped methods (deposit + withdraw, **and for PIX also pay**)" — and `MERCADOPAGO_QR_AR` is the only `qr-only` entry in the map. `channelOf` on this side is just `rail.channel`, so the gate was reading a dimension that never contains Pix.

The failure mode is real and not hypothetical: `enableQrPoolRails` (`src/kyc/rails.ts`) skips any rail already in a non-PENDING state (`if (existing && existing.status !== 'PENDING') continue`) and enables the rest individually, so `pix_br` ENABLED alongside a rejected `mercadopago_qr_ar` is a state the backend produces. That user had `hasCardAccess === false` and `hasQrSpendRail === false`, so the step returned `null` — no CTA and no checklist — while a Pix QR pay is precisely what `firstSpendIntent` counts as activation.

The gate now asks the provider and the operation, not the channel:

```ts
const hasQrSpendRail = canDo('pay', { provider: 'manteca' })
```

Two notes on the shape, since you offered `canDo('pay')` as an alternative. I kept the provider scope: `canDo` falls back to `rail.status` when a rail carries no per-op map, so an unscoped `canDo('pay')` returns true for an enabled `bridge.sepa_eu` — a rail that cannot activate anyone. Scoped to `manteca` it matches `firstSpendIntent` exactly. And `canDo` reads `operations.pay ?? status`, which is the same per-op refinement the previous round added, so a pool-tier rail reading `enabled` overall while `pay` still needs an upgrade continues to render nothing.

Fixed at the call site rather than in `channelOf`: the channel classifier is correct and two other consumers depend on it as-is (the `bank`/`qr-only` `rejectableRails` filter in this same file, and `WelcomeUnlockModal`'s `qrRails`). The defect was this predicate picking the wrong dimension — `WelcomeUnlockModal:33` already asks the channel-blind question via `canDo('pay')`.

You were also right that the old test could not catch it. `enabledQrRail` was `manteca.qr_br` on `qr-only`, a combination the resolver never emits. Fixtures are now the two real shapes — `manteca.pix_br` on `bank` with `operations: { pay: 'enabled', deposit: 'requires-info' }`, and `manteca.mercadopago_qr_ar` on `qr-only` — plus a case pinning the exact scenario you described (paying Pix, rejected MercadoPago) and one for the AR rail. Reverting the one-line fix fails 3 of the 22 cases; the mock also gained `canDo` so it stays faithful to the hook. `dev/journey/journeyData.ts` carried the same wrong assumption in its condition text and is corrected.

---

**Scope advertised rails to the selected residence** (`residence-availability.ts`) — re-raise, and the answer is unchanged: this is a presentation question, already answered in the presentation layer.

The enrolment claim is not derived from "the broader onboarding intent" as a convenience — it is what one verification actually opens. `REGION_RAIL_MAP` (peanut-api-ts `src/kyc/rails.consts.ts`) enrols `['ACH_US','SEPA_EU','SPEI_MX','FASTER_PAYMENTS_GB']` for both EU and NA; a Bridge customer is created with all four endorsements and Peanut only marks Bridge KYC approved once all four come back `APPROVED`; the single residence gate on Bridge rails is `isUkResident` (`src/kyc/capabilities/resolver.ts`). Production holdings bear it out as one cohort rather than four national ones (SEPA 1,714 / ACH 1,698 / SPEI 1,684 / GBP 1,470, `product/providers/fiat/bridge-exit-cost.md`). Narrowing the FE mirror to one rail per country would make the signup card under-promise against what the user's own verification grants.

Your second point is the true one and it is already implemented: those rails only pay out into an account on their own network. A set spanning more than one currency carries that as an explicit condition — *"Transfers in another currency need a bank account in that currency."* — instead of listing the extras as flat benefits of living somewhere. That is the agreed presentation-level fix from the previous round.

On the specific hazards you name: no UK overpromise is reachable, since GB is in `RESTRICTED_RESIDENCE_ISO2` and `residenceAvailability` short-circuits on `restrictions.banking` before any rail is listed. `product/countries.md` is the side that is out of date here; that edit is filed separately rather than made in this PR.

**Ship the restart route before calling it** (`sumsub.ts`) — re-raise; merge-order item, no code change, unchanged from the last round.

`POST /users/identity/restart` is already called on peanut-ui `dev` (`src/app/actions/sumsub.ts`). This head adds an optional `regionIntent` body field, which a backend predating it ignores, and `RestartIdentityResponse.regionIntent` is optional for the same reason — so this head introduces no new dependency. The route genuinely is absent on api-ts `dev`; peanut-api-ts #1492 adds it and merges and deploys before this PR. A 404 in the interim surfaces as `errorRestartFailed`, not a crash.

**Expose firstPaymentAt before consuming it** (`GettingStartedChecklist.tsx`) — re-raise; no code change, unchanged from the last round.

The fallback is already in place: `done: milestone === 'activated' || hasSentPayment`, so an absent `firstPaymentAt` degrades to the milestone rather than pinning the row open. And the row is unreachable in the state described — `ActivationCTAs` only renders the checklist when the user is **not** funded (`isFundedNotActivated` returns the step card instead), so a user in a position to complete a `/send` is not looking at this checklist. #1492 ships the producer first regardless.

---

Verification on `5e8358c5`: `tsc --noEmit` exit 0, prettier clean, eslint clean on the touched files, ds-lint ratchet green with no metric increased (`offScaleSpacing 186/204`, `iconOffScale 74/81`, `rawDuration 5/9`, rest unchanged), 105 tests pass across Home and `useCapabilities` (22 in `ActivationCTAs`).

innolope-dev:
Seventh round addressed in `0b18b774`. One new finding — a real hole in my own previous fix — plus one doc item filed as a follow-up. Three re-raises answered unchanged.

**Do not treat a bank-only Manteca rail as QR pay** (`ActivationCTAs.tsx`) — fixed, and you caught a regression the last round introduced.

`canDo` resolves an operation as `operations[op] ?? status`. `MANTECA_METHOD_OPERATIONS` (peanut-api-ts `src/kyc/capabilities/manteca-operations.ts`) is:

```
PIX_BR:            ['pay', 'deposit', 'withdraw']
BANK_TRANSFER_AR:  ['deposit', 'withdraw']   // bank rail, no merchant-QR
MERCADOPAGO_QR_AR: ['pay']
```

so an enabled `manteca.bank_transfer_ar` emits `{ deposit, withdraw }` with no `pay` key, the `?? status` fallback read `'enabled'`, and an Argentine holding only that rail got "Make your first payment" and a QR scanner with nothing that can pay a merchant QR. Worse than the `null` the round before it — a dead end instead of a missing card.

The backend guards this exact case and says so: the US-nationality branch derives per-op status from the method's real op list "so a bank rail (no `pay` op) can't get a phantom pay capability" (`resolver.ts`). The status fallback on this side defeated that guard.

The gate now requires `pay` to be present **and** enabled:

```ts
rails.some((rail) => {
    if (rail.provider !== 'manteca') return false
    if (rail.operations) return rail.operations.pay === 'enabled'
    return channelOf(rail) === 'qr-only' && rail.status === 'enabled'
})
```

Deliberately not `canDo`/`operationStatus` — the fallback is the defect, so this reads `operations` directly. `mantecaEnabledOperations` lists every op a method supports, so on a rail carrying the map an absent `pay` means "no merchant QR"; the map is only absent for an unknown method or a response predating it, where the `qr-only` channel is pay by construction, and that stays the fallback. `operations` is only attached to rails the resolver already resolved to `enabled`, so no rejected rail can slip through.

I did not push the fix down into `canDo`: `operations[op] ?? status` is correct for Bridge and Rain, which have no per-op split and leave `operations` undefined by design, so changing the primitive would flip every other consumer's semantics. Absent-op-means-unsupported is only sound on a rail that carries a map, so it belongs at this call site. If it turns out to be wanted generally, the honest version is a separate `supportsOperation` helper on the hook rather than a change to `canDo` — filed as a follow-up, not smuggled into this PR.

Coverage: Pix (bank-channel, pay enabled) keeps the CTA, MercadoPago (qr-only) keeps it, pool-tier Pix with `pay: 'requires-info'` still renders nothing, the new `manteca.bank_transfer_ar` shape renders nothing, and that rail alongside a paying Pix rail still keeps the step. Reverting the predicate to the `?? status` form fails the bank-only case; reverting to the `qr-only` channel form fails 3 of the Pix cases.

**activation-funnel.md no longer matches the funnel shape** (MINOR) — agreed on both counts, and agreed the code is the right half. That file lives in mono, not in this repo, so it is not a change I can make in this PR; filed as a follow-up to add the two decision-log entries (first-spend push suppressed when no activating spend is open; the home checklist's third row tracks first peer payment, not first spend) so the `journeyData.ts` citation resolves. Worth noting the shape moved again this round: the suppressed cohort is now "no Manteca rail with `pay` present and enabled", which is what the doc entry should say.

---

The three re-raises are unchanged from the last round, with the same evidence.

**Scope advertised rails to the selected residence** — `REGION_RAIL_MAP` enrols `['ACH_US','SEPA_EU','SPEI_MX','FASTER_PAYMENTS_GB']` for both EU and NA; a Bridge customer is created with all four endorsements and Bridge KYC is only marked approved once all four return `APPROVED`; the sole residence gate on Bridge rails is `isUkResident`. Narrowing the FE mirror per country would under-promise against what the user's own verification actually grants. The true half of the finding — those rails only pay out into an account on their own network — is already implemented as an explicit condition on any multi-currency set ("Transfers in another currency need a bank account in that currency"). No UK overpromise is reachable: GB is in `RESTRICTED_RESIDENCE_ISO2` and `residenceAvailability` short-circuits on `restrictions.banking` before listing any rail. `product/countries.md` is the stale side; that edit is filed separately.

**Ship the restart route before calling it** — merge-order item, no code change. `POST /users/identity/restart` is already called on peanut-ui `dev`; this head only adds an optional `regionIntent` body field that an older backend ignores, and `RestartIdentityResponse.regionIntent` is optional for the same reason, so this head adds no new dependency. peanut-api-ts #1492 adds the route and merges and deploys before this PR. A 404 in the interim surfaces as `errorRestartFailed`, not a crash.

**Expose firstPaymentAt before consuming it** — no code change. `done: milestone === 'activated' || hasSentPayment`, so an absent `firstPaymentAt` degrades to the milestone rather than pinning the row open, and `ActivationCTAs` only renders the checklist for a user who is not funded — a user in a position to complete a `/send` is not looking at this row. #1492 ships the producer first regardless.

---

Verification on `0b18b774`: `tsc --noEmit` exit 0, prettier clean, eslint clean on the touched files, ds-lint ratchet green with no metric increased, 107 tests pass across Home and `useCapabilities` (24 in `ActivationCTAs`).

innolope-dev:
Eighth round on `0b18b774` — no new code findings. Recording the state so the thread is unambiguous.

Chip confirms the activation gate is closed: *"P24/P28 fixed: the QR-spend gate includes Pix via operations.pay and excludes bank-only Manteca rails without pay."* That was the one live code defect across the last two rounds, and it is settled — Pix (bank-channel, `pay` enabled) and MercadoPago (`qr-only`) both keep the CTA; a bank-only `manteca.bank_transfer_ar` does not.

The four remaining findings are the standing set, each consolidated by Chip itself into a single item this round (P25, P26, P27, P29), and each already answered with evidence above:

- **Residence rails / ACH scope** (P25, previously P9/P11/P15/P19/P21) — declined; the enrolment claim matches `REGION_RAIL_MAP` and Bridge's four-endorsement approval gate, and the genuine half (a foreign-currency rail needs an account on that network) already ships as an explicit condition on multi-currency sets. `product/countries.md` is the stale side; that edit is filed separately.
- **Restart route** (P26, previously P7/P17/P22) — declined; merge order, not a code change. The call already exists on peanut-ui `dev`; this head only adds an optional body field. peanut-api-ts #1492 merges and deploys first.
- **`firstPaymentAt`** (P27, previously P12/P18/P23) — declined; the milestone fallback is already in place and the row is unreachable for a user who could complete a `/send`, since the checklist only renders pre-funding. #1492 ships the producer first.
- **`product/activation-funnel.md`** (P29) — agreed, and agreed the code is the right half. The file lives in mono, so it is not changeable from this PR; filed as a follow-up covering both entries (first-spend push suppressed when no activating spend is open — now stated as "no Manteca rail with `pay` present and enabled" — and the checklist's third row tracking first peer payment rather than first spend).

Two follow-ups recorded from this review that are deliberately not in this PR: the `product/countries.md` and `product/activation-funnel.md` decision-log edits in mono, and a `supportsOperation` helper on `useCapabilities` if absent-op-means-unsupported turns out to be wanted beyond this call site (changing `canDo` itself would flip Bridge/Rain semantics, which have no per-op split by design).

CI green at this head: unit, typecheck, eslint, format, ds-lint, ds-shots, analyze, preview, human-authors, bot-approval.

innolope-dev:
## The four open threads all resolve outside this PR — here is the map

Every finding still marked unresolved on `0b18b774` is gated on one of two PRs that are pushed and green but not yet merged, so a reviewer reading this PR against `main` will keep re-deriving them. Consolidating rather than answering each again in-thread.

### Gated on peanut-api-ts #1492 (merge order, not a code change)

**"Ship the restart endpoint before calling it"** and **"Expose `firstPaymentAt` before consuming it"** are both correct about `main` and both already built. #1492 adds `POST /users/identity/restart` and the `firstPaymentAt` projection on `/users/me`; it is at `dbb32e3f` with unit, all three integration shards, typecheck, eslint, format, onchain and audit green.

Neither degrades badly if the order slips: the restart call already exists on `dev` (this head only adds an optional `regionIntent` body, and the response field is optional), so a 404 surfaces the `errorRestartFailed` message rather than crashing; and the checklist row falls back to `milestone === 'activated'` when `firstPaymentAt` is absent, which is the pre-PR behaviour. **But the order does matter for the feature to work at all — #1492 should merge and deploy first.**

### Gated on mono #125 (the doc is the stale side, and it is being fixed)

**"Keep ACH scoped to US residence verification."** The country tables are the side that is out of date, and #125 corrects them. Bridge entitlement is per-**endorsement**, not per-country, and the chain is unambiguous in this codebase:

- Every Bridge customer is created with `endorsements: ALL_BRIDGE_ENDORSEMENTS` — base + sepa + spei + faster_payments — with no residence branch (`peanut-api-ts src/kyc/provider-submission/bridge-adapter.ts`).
- Peanut marks the Bridge KYC `approved` **only when all four are APPROVED**; one rejection makes the whole KYC `rejected` (`src/bridge/status-processors.ts`). So "Bridge-approved" and "all four rails endorsed" are the same state by construction.
- Enrolment matches: `REGION_RAIL_MAP` gives both EU and NA `[ACH_US, SEPA_EU, SPEI_MX, FASTER_PAYMENTS_GB]`.
- The only residence gate on any Bridge rail is `isUkResident`.
- Production agrees: the four rails hold 1,714 / 1,698 / 1,684 / 1,470 enabled users — one ~1,700-user cohort, not four residence-scoped ones (`product/providers/fiat/bridge-exit-cost.md` §1, evidence 2026-08-24). If SPEI were Mexico-only it would be a fraction of SEPA, not 98% of it.

So an EU-verified resident really can receive USD. #125 redefines `deposit_rails` as the country's **local** rails and states the cohort rule once, with the honest caveat that three of the four are dead by volume (27 lifetime on-rampers on MXN+GBP against 3,154 enablements) so marketing should still lead with the local rail.

**"Update the canonical activation funnel."** Agreed, and done in the same PR — `product/activation-funnel.md` now carries a dated decision entry covering exactly the three things named here: the checklist as the pre-funding empty state, its payment row completing from a peer send while activation stays a card/Manteca-QR spend, and the conditional suppression of the funded step. It also records the rule that forced the entry — no surface may leave a funded-but-unactivated account without an activation CTA, **unless no activating spend is available to that account at all**, which is the carve-out this PR implements.

### Summary

| Thread | Resolves in | State |
|---|---|---|
| Restart endpoint | api-ts#1492 | pushed, green, needs merge first |
| `firstPaymentAt` | api-ts#1492 | pushed, green, needs merge first |
| ACH scoping | mono#125 | pushed, doc corrected with evidence |
| Activation funnel doc | mono#125 | pushed, decision entry added |

Nothing here is unaddressed in code on this branch. Suggested merge order: **api-ts#1492 → this PR**, with mono#125 landing whenever convenient since it is documentation only.

PR #2926, opened 2026-09-02
fix(ds): DS review design findings — modal tones, notifications, emoji sweep, nav, OTA row

innolope-dev:
## What

Design-system items from the staging DS review — TASK-22114 ([Notion](https://app.notion.com/p/3cf8381175798170b443e3331ad27884)). Companion to #2925 (critical/high items) and peanut-api-ts `feat/task-22114-activation-restart-identity`.

### Icon colour rule
`ActionModal` gains `tone: 'error' | 'warning' | 'success' | 'info'` (red / yellow / green / blue bubble + default glyph, mirroring `Notification`). Every yellow icon bubble in the app was reclassified against the rule **yellow = warnings only, red = errors, green = success, blue = information**: the Unlock-payments error modals are red, the blocked/rejected KYC states are red, the "send with a link" / request / deposit-method / saved-account icons are blue, the cancel-card and lock-card confirmations stay yellow. The rule is documented on the icon-bubble and modal DS pages.

### Components
- Signup "We're having trouble setting up your account" → `Notification(error)` above the buttons.
- White bordered explainer boxes (residence "Which country goes first?", account-ready notes, passkey help, KYC prep checklist) → `Notification(info)` without icon.
- QR result modals (payment confirmation + 5 siblings) rebuilt on `ActionModal` with the DS `Checkbox`; the ℹ️ title emoji is gone. `ActionModal`'s raw checkbox input is the DS `Checkbox` now.
- Emoji removed from `home.welcomeUnlock.title`, `badges.toastMultiple`, mascot alts, the recover-wallet title (now `common.fundsOnTheWay`, hence the regenerated `*.marketing.json`), and the fix-card-signature status glyphs (→ icon bubbles).
- Generated avatars: the frown mouth is replaced by a wide grin (array length unchanged, so other users' faces don't change).
- Profile: Bio / phone / website hidden behind `SHOW_COMING_SOON_FIELDS`, the self-profile share pill hidden (`showShareButton={false}`), "Your card" rows aligned to the profile menu sizing.
- QR scanner: "Click to paste" sits directly under the scan square instead of pinned above the drawer.
- Home ↔ Card: horizontal slide keyed on the tab order (`TabSlide` inside `#scrollable-content`, respects reduced motion).
- Bottom nav: hard offset shadow `3px 3px 0` in the border ink on the bar and the QR button, `1px 1px 0` at 55% on the active pill (per the Bottom Nav Contrast Studies "hard" variant).
- Setup: no "Log in" link on the PWA-install steps.
- pt-BR: "Envie dinheiro com link".

### "Update available" row (native)
`OtaUpdateContext` replaces the discarded `useOtaUpdates()` call: it seeds from `CapacitorUpdater.getNextBundle()` (survives reloads), learns about newly staged bundles from the updater, and flags `disable_auto_update_to_major/minor` as "store update required". The profile row under **About Peanut** opens a modal whose CTA applies the bundle in place: marker → `set()` (immediate reload); if `set()` rejects, re-stage + `reload()`; if the page is still alive 3 s later, `App.exitApp()` on Android or "close the app from the app switcher" copy on iOS. On the next launch the marker is compared with the running bundle and a failed apply is reported at error level. The store branch opens `STORE_URL` — **the iOS entry is still the placeholder URL** and needs the real App Store id before that branch is useful.

### About page 5-tap
The five taps only ever revealed the beta toggle, which is hidden by the `beta-ota-channel` flag for most accounts, so nothing visible happened. The 5th tap now always toasts: "revealed" (+ scroll into view), "not enabled for this account", or "only in the app" on web. Joining still requires the toggle.

## Tests

17 suites, 186 tests: `ActionModal` (tone mapping), `OtaUpdateContext` (seed, store flag, apply, reject fallback, iOS manual restart, next-launch report), `tab-order`, `DotFaceAvatar` (no downward mouth), `QRScanner` placement, `QRScannerOverlay`, `About.view`, `ClientProviders`, marketing-messages parity. Typecheck clean; ds-lint ratchet **down** on six metrics; prettier clean.

## Needs visual verification on staging

Per Slava's instruction the browser pass was skipped to land the code first: DS pages `/dev/ds/patterns/modal` and `/dev/ds/primitives/icon-bubble`, setup residence and account-ready screens, `/card` vs `/profile` row heights, QR paste position on short viewports (< 730 px), Home ↔ Card slide, bottom nav shadow, fix-card-signature icons, the OTA row/modal on a device (Slava verifies OTA himself), About 5-tap toasts.

## Merge note

Overlaps #2925 on `UnlockPayments.view.tsx`, `Residence.tsx`, `About.view.tsx` and the catalogs — whichever merges second needs a rebase.


kushagrasarathe:
Heads-up from the TASK-22121 DS-nits branch (`fix/ds-ui-nits-22121`). We fixed the same setup-flow ground in parallel, compared both sides, and reverted ALL our setup changes so your #2925/#2926 own that surface — your CountryCombobox and the Notification info boxes are the right component choices. A few fixes existed only on our side; please fold them into your PRs so they don't get lost, plus some merge heads-ups.

## Fixes to add (only we had these — now reverted)

**`Setup/Views/SignTestTransaction.tsx`**
1. The sign-step column `flex h-full flex-col justify-end gap-2 text-center` — drop `text-center`. The h1 above is `text-left`; setup copy is left-aligned (Kush QA 17.1). Your tree still centers it.
2. Same file: container `gap-10` → `gap-6` (XL/24 section gap; 40px is off-anatomy here).
3. The footer "learn more" underlined button needs a 44px hit area + focus ring: `relative after:absolute after:inset-x-0 after:-inset-y-3.5 focus-visible:outline-[3px] focus-visible:outline-action-focus` (LinkButton pattern; design.md touch law).

**`Setup/Setup.consts.tsx`**
4. The `sign-test-transaction` step has `contentClassName: 'flex flex-col items-center justify-center gap-6'` — siblings (signup/residence/passkey-permission) all use `items-end`. Align it.

**`Setup/Views/Residence.tsx`**
5. Notify-email validation is a raw `<p className="text-body-s text-error">` — `text-error` is the legacy palette token and field validation is `<FieldError>` per design.md (see JoinWaitlist.tsx:98 / Signup.tsx:150 for the pattern).
6. The four underlined links (3× "Choose a different country", the multi-doc disclosure) need the same 44px `after:` hit area + focus ring as (3).
7. Kush's ruling (QA item 23): "Choose a different country" is **centered** (`self-center text-center`) on the heads-up/congrats views, and the CTA stack gap is `gap-4` (16px), not `gap-2`.
8. In-view headings miss `w-full text-left leading-tight` (SetupWrapper.tsx:363 spelling); in-view descriptions are `text-body-s` where the chrome renders every other step at `text-body-m`; sub-view containers `gap-4` → `gap-6`; compare-card list `space-y-0.5` → `space-y-2` (S/8 list gap).

**`Setup/components/PasskeyInfoModal.tsx`**
9. `text-black` ×2 → `text-foreground-primary` (same raw color also sits in `SetupWrapper.tsx:372`).

**`Common/CountryCombobox.tsx` (#2925)**
10. Minor DS nits: `bg-gray-200`, `text-white`, `shadow-lg` are raw utilities — swap for tokens. And consider genericizing it later (a `SearchableSelect` without the `ResidenceCountryOption` coupling) so `DynamicBankAccountForm` can reuse it.

## Merge heads-ups (our branch vs yours — whoever merges second hits these)

- **`Global/DotFaceAvatar.tsx` is DELETED on our branch** (team decision 2026-09-02: DS `AvatarWithBadge` only), and `common.userAvatarAlt` is removed from all 4 catalogs. #2926's frown→grin change and its tests land on a deleted file — drop that hunk.
- **`AddMoney/views/AddMoneyMethodSelection.view.tsx` is DELETED on our branch** (the old "How would you like to add money?" screen is gone; bare `/add-money` renders the country list). #2926's bubble-color change there is moot.
- **`Global/QRScanner/index.tsx`**: both sides moved the paste/icons block under the scan square. Ours carries Kush's ruling: payment icons in a `grid grid-cols-2 gap-2`, block at `mt-6` below the square — keep the grid when resolving.
- **`Global/BottomNav/index.tsx`**: our branch sets `PILL_SPRING` bounce `0.3` (Kush wants the bounce) and `shadow-4` on bar + QR circle (his explicit ruling from the screenshot). #2926's `tab-order.ts` extraction re-hardcodes the old damping and uses a different shadow token — needs one decision, but bounce 0.3 must survive.
- **`Global/ReConsentModal/index.tsx`**: we retokenized the `DocsLink` line (`text-foreground-primary underline`) and made "Not now" a stroke button; #2925 rewrites the same DocsLink line for i18n — hand-merge, both are wanted.
- **i18n catalogs ×4**: our branch removed `qrScanner.myQrCollapsedTitle`/`myQrExpandedTitle` (→ single `myQrTitle`), `addMoney.howWouldYouLikeToAdd`, `common.userAvatarAlt`; #2926 edits the same `qrScanner` block.
- **`Setup/Views/PasskeySetupHelpModal.tsx`**: TASK-21904 wants this converted to a scrollable drawer (content too long for a modal). We built that conversion but dropped it to avoid colliding with your Notification change — consider folding the drawer conversion into #2926, otherwise it returns as a TASK-21904 follow-up.

innolope-dev:
Third round addressed on `fix/task-22114-ds-review-design`:

- **OTA watchdog vs recovery**: the 3 s watchdog is cancelled the moment `set()` rejects and re-armed only after the re-download + `reload()` sequence has run; a delayed-download test pins that no `exitApp` fires mid-recovery.
- **QR paste stack on short viewports**: below 730 px the paste actions render in the strip pinned above the My QR drawer (`QR_DRAWER_PEEK_PX + QR_DRAWER_PASTE_GAP_PX`), otherwise under the scan square — one render, driven by `matchMedia`, with a viewport test.
- **iOS store prompt**: the App Store listing for app 6786373552 is not published yet, so the store branch of "Update available" is hidden on iOS behind `IOS_APP_STORE_LISTING_LIVE = false`; Android keeps its Play link.

innolope-dev:
Fourth round addressed in `84329da5`:

- **OTA readiness is above every lazily-loaded provider** (`ClientProviders.tsx`). Correct catch: on app routes `PeanutProvider` renders its children only through the dynamically-imported `AppStateProviders` chunk, so a slow or failed chunk load took `notifyAppReady()` past the 15 s app-ready timeout with it. `OtaUpdateProvider` is now the outermost provider, above `NuqsAdapter` and `PeanutProvider` — it reads no context of its own, so nothing else had to move. The provider-order test pins it at index 0 on `/home`, `/setup` and `/`, and pins that the `(dynamic)` app providers sit below it.

- **A failed apply no longer asks for a restart** (`capgo-updater.ts`, `OtaUpdateContext.tsx`). `applyStagedBundle` returns `'reloading' | 'failed'` instead of `void`. Only a re-stage that actually produced a bundle earns a `reload()`: an offline / up-to-date / store-update-required check, a check that throws, and a rejected `reload()` all abandon the apply. The provider re-arms the watchdog **only** after the page was really handed to the plugin, so no Android `exitApp()` and no iOS "close the app" instruction for an update that was never staged; the modal drops out of `applying` into an error-tone state with a Try again.

  Two follow-ons found while fixing it: the abandon path also clears the `capgoPendingApply` marker, or the next launch reports a phantom failed apply at error level; and a successful re-stage mints a *new* bundle id, so `onRestaged` now feeds it back into `pendingBundle` — otherwise a retry would hand `set()` exactly the dead id it just rejected. The apply also latches on a ref, since two taps in the same tick both read `applyState === 'idle'` from their render's closure.

  Six new cases: abandon on an offline re-stage, abandon on an up-to-date re-stage, abandon on a rejected `reload()`, adoption of the re-staged id, single-apply on a double tap, and retriability after a failure.

@kushagrasarathe — folded in from `fix/ds-ui-nits-22121`, thank you for the list:
- `SignTestTransaction`: dropped `text-center`, `gap-10` → `gap-6`, 44px hit area + focus ring on the "learn more" link.
- `Setup.consts`: `sign-test-transaction` is `items-end pt-8` like its siblings.
- `PasskeyInfoModal` ×2 and `SetupWrapper.tsx:372`: `text-black` → `text-foreground-primary`.
- `QRScanner`: payment badges are a `grid grid-cols-2 gap-2` at `mt-6`.
- `BottomNav`: the pill has its own spring at bounce 0.3 — `TAB_SPRING` stays near-critically damped because it also drives the full-page `TabSlide`, where overshoot reads as a glitch. Bar and QR circle now carry the DS `shadow-4` token per your ruling, replacing the arbitrary 3px shadow.
- Residence.tsx items (FieldError, hit areas, centered link, headings, compare-card gap) and the `CountryCombobox` tokens went to #2925, which owns that file's current shape — `BaseSelect` got the same token swap so the two listboxes stay identical.
- Left as-is deliberately: the `DotFaceAvatar` grin and the `AddMoneyMethodSelection` bubble colour. Both files are live on `dev` today; if your branch lands first the hunks resolve by deletion, and if it lands second the fixes were correct in the meantime. The `PasskeySetupHelpModal` → drawer conversion stays as the TASK-21904 follow-up.

Typecheck, prettier clean; ds-lint counts at or below baseline on every metric.

innolope-dev:
Fifth round addressed in `2742be7c`.

**MAJOR — retarget the apply marker after re-staging** (`capgo-updater.ts:201`). Confirmed and fixed. `applyStagedBundle('b-2')` writes `capgoPendingApply = b-2` before `set()`. When `set()` rejects, the recovery re-stage downloads a *new* bundle and `next()`s it, so the `reload()` that follows applies `b-3` — but the marker still named `b-2`. On the next launch `reportPendingApply` compared the running `b-3` against the expected `b-2`, took the mismatch branch and logged `[capgo-apply] restart did not apply bundle b-2; running b-3` at **error** level. Every successful recovery reported itself to Sentry as a failed apply, which is exactly the signal that path exists to raise.

The re-stage's `onUpdateAvailable` now captures the bundle alongside forwarding it to `onRestaged`, and the marker is rewritten to that id immediately before `reload()` — after the `outcome !== 'staged'` gate, so an abandoned apply still clears the marker rather than retargeting it. The rejected-`reload()` path is unchanged: nothing was handed to the plugin, so `abandon()` removes the marker outright.

New case: *retargets the marker to the re-staged bundle, so the recovered launch reads as a success* — asserts the marker is `b-3` after the recovery reload, then mounts a fresh provider on `b-3` and pins that it logs `restart applied` at warn level with `console.error` never called. Verified non-vacuous: with the one-line retarget reverted it fails on `Expected "b-3", received "b-2"`.

`typecheck` 0, `OtaUpdateContext` + `capgo-updater` suites 50/50 green, `prettier --check .` clean, ds-lint ratchet ok on all 16 metrics (nothing raised; `rawDuration` 9→5, `offScaleSpacing` 204→186).

innolope-dev:
Sixth round addressed in `4298d28f`.

**MINOR — hide the close button while OTA apply is locked** (`OtaUpdateModal.tsx:38`). Confirmed and fixed. `preventClose` is not the whole dismissal surface: in `Global/Modal/index.tsx` it gates Escape (`:41`), the Dialog's own `onClose` (`:52`) and the backdrop (`:75`), but the close button's `onClick={onClose}` (`:127`) is ungated. So while `applyState === 'applying'` both CTAs were disabled and the incidental dismissals were blocked, yet the X still closed the modal. During a slow recovery re-download that is a real window: if the apply then resolved `failed`, the error tone and **Try again** rendered into a modal nobody could see, and the silent dismissal read as success — the user waits for a restart that is never coming.

`hideModalCloseButton` (already on `ActionModal:54`, maps to `classButtonClose: '!hidden'`) now tracks the same `applying` flag, which is the `preventClose` + `hideModalCloseButton` pairing the other fully-locked modals already use (`EnableAutoBalanceBanner`, `DeleteAccountButton`). `idle`, `failed` and `manual-restart` stay dismissable, so the lock covers only the in-flight operation.

**Why the call site and not the primitive.** Gating `Modal:127` on `preventClose` looks like the general fix, but it would be a regression: `preventClose` carries two different intents across its ~10 consumers. `Global/PostSignupActionManager/index.tsx:66` passes it with the comment *"Prevent closing the modal by clicking outside"* — the X is that modal's only dismissal affordance besides the CTA, so gating the primitive would trap users in it. The real fix is splitting the prop into `preventClose` (incidental) vs `preventDismiss` (all paths) and auditing all ten surfaces for which they meant; that is a DS change well outside this PR, so it is **noted as a follow-up** rather than done here — flagging it so a later reviewer doesn't re-raise this as a missed generalisation.

New suite `Profile/components/__tests__/OtaUpdateModal.test.tsx` (5 cases): stubs `ActionModal` to surface the lock props as data-attributes, asserts both are set while `applying`, asserts both are clear for `idle` / `failed` / `manual-restart`, and pins that `failed` still offers an enabled **Try again** so the lock can never strand the user. Verified non-vacuous: with the one-line `hideModalCloseButton` reverted the applying case fails on `Expected "true", received "false"`.

`typecheck` 0, the three OTA suites (`OtaUpdateModal` + `OtaUpdateContext` + `ClientProviders`) 26/26 green, `eslint` clean on both touched files, `prettier --check .` clean, ds-lint ratchet ok on all 16 metrics (nothing raised). `ds-shots` and `chip-review` were both green on the previous head `2742be7c`.

innolope-dev:
**MAJOR: External-URL QR flow passes javascript:/data:/file: through to openExternalUrl** — fixed in `df0b785ff`.

Confirmed reachable, and worse than the report on one point. `recognizeQr`'s URL pattern is

```
/^(https?:\/\/)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\+.~#?&//=]*)$/
```

The leading character class contains `:` and the trailing one contains `(`, `)`, `?`, `.` and `/`, so a schemed payload matches as long as it carries a dot plus a short TLD-like token. `javascript:self.co?0:eval(atob(location.hash.slice(1)))` matches — verified against the literal regex — and is a working payload, so `recognizeQr` returns `EQrType.URL`, `processQRCode` calls `setRedirectTo(data)` with the raw scan (index.tsx:355), and `openExternal` forwarded it because it only asked "does this have a scheme".

The web fallback is the worse leg, not the `window.open` one. `openExternalUrl` (src/utils/capacitor.ts:232-242) is:

```ts
} else if (!window.open(url, '_blank')) {
    window.location.assign(url)
}
```

A blocked popup — the common case — falls through to `location.assign`, which executes a `javascript:` URL in the **current** document, not a new `about:blank`. That is the app's own origin directly.

Fixed at the call site with an allowlist rather than a denylist, so no scheme has to be enumerated:

```ts
const candidate = /^[a-z][a-z0-9+.-]*:/i.test(redirectTo) ? redirectTo : `https://${redirectTo}`
return /^https?:\/\//i.test(candidate) ? candidate : undefined
```

The scheme-less upgrade is kept — that branch exists because `Browser.open` throws on `example.com/x` and the tap died silently. A payload that cannot be resolved to http(s) is no longer offered as a link at all: `EModalType.EXTERNAL_URL` renders the unrecognised-code content instead, so there is no "Open link" button to press. Also took the second half of the suggestion — the modal now renders the resolved destination between the intro and the trust line, so the warning is actionable.

New pin: `src/components/Global/QRScannerOverlay/__tests__/external-url-scheme.test.tsx` — https opens with the destination shown, a scheme-less payload still upgrades to https, and the `javascript:` payload reports as unrecognised with `openExternalUrl` never called. Verified non-vacuous: reverting the allowlist to `return candidate` fails the third case.

innolope-dev:
**MAJOR: Unlock the modal when Android exit fails** — fixed in `df0b785ff`.

Correct, and the stuck state was total. The watchdog's Android leg was:

```ts
import('@capacitor/app')
    .then(({ App }) => App.exitApp())
    .catch((err) => console.warn('[capgo] exitApp failed:', err))
```

With `applyState` still `'applying'`, `OtaUpdateModal` sets `preventClose={applying}` and `hideModalCloseButton={applying}` and disables both the primary and the "Not now" CTA — so on an `exitApp` rejection or a failed chunk load there was no Escape, no backdrop, no X, and no enabled button. Every exit was closed at once.

The catch now sets `manual-restart`, which is the state iOS already lands in for exactly this situation (the page outlived a reload that was genuinely issued), so the user gets the app-switcher instruction and a single dismissing "Got it". `applyingRef` is deliberately left latched and `applyNow` keeps its `applyState === 'manual-restart'` guard: the bundle *was* handed to the plugin here, so a retry is wrong — this is not the `'failed'` path.

New case in `src/context/__tests__/OtaUpdateContext.test.tsx`: "falls back to the manual-restart instruction when the Android exit fails". Verified non-vacuous — restoring the log-only catch fails that case and only that case.

innolope-dev:
**MINOR: Clear the stale apply marker during re-stage** — fixed in `df0b785ff`, exactly as suggested.

`removeStoredValue(PENDING_APPLY_KEY)` now runs the moment `set()` rejects, before `queueUpdateCheck()`, and the re-staged id is written immediately before `reload()` as it already was. A kill during the recovery download therefore leaves no marker, so the next launch says nothing rather than reporting an error-level failed apply for a bundle that never reached the plugin.

Checked the gap this opens in the other direction: clearing early could in principle lose a *real* failure signal if the code reached `reload()` with no marker written. It cannot. `checkAndStageUpdate` only returns `'staged'` on the path that has just called `callbacks.onUpdateAvailable?.(bundle)` synchronously (capgo-updater.ts:113-116), and the caller guards with `if (outcome !== 'staged') return abandon(...)`, so `restaged` is always set by the time `writeStoredValue(PENDING_APPLY_KEY, restaged.id)` runs. Every path that issues a reload still carries a marker.

`abandon()` keeps its own `removeStoredValue` — it is idempotent for the two re-stage failures and still load-bearing for the `reload()`-rejected case, which runs after the marker has been rewritten.

New case: "drops the marker while the recovery re-stage is in flight, so a kill is not reported as a failed apply" — it samples the marker from inside the `getLatest` mock, so the assertion happens mid-re-stage without leaving a promise pending. (First draft held the download open with an unresolved promise; when the assertion failed it aborted before resolving it, wedging the module-level `pendingOtaWork` queue and cascading into 7 unrelated failures. Rewritten so a failure stays local.) Verified non-vacuous: dropping the new `removeStoredValue` fails that case alone.

innolope-dev:
**MINOR: Add Argentine copy for OTA apply failures** — fixed in `df0b785ff`.

Confirmed the delta was exactly as described: es-AR's `profile.update` block overrode `title`, `description`, `restartNow`, `notNow` and `manualRestart` but stopped short of `applyFailed`, so the failure state fell through to the es-419 tuteo string ("Revisa tu conexión e inténtalo de nuevo") in a block whose neighbours are already voseo ("Cerrá", "volvé", "Reiniciá").

Added the sibling:

> No se pudo aplicar la actualización. Revisá tu conexión y probá de nuevo: Peanut sigue con la versión con la que iniciaste.

`Revisa → Revisá` and `inténtalo de nuevo → probá de nuevo` are the two verbs that carried tuteo; "iniciaste" is already shared between both forms. The full i18n suite passes (`src/i18n` — 14 suites), including the locale-coverage checks.

innolope-dev:
**MAJOR: Restore a cancel path for direct-send confirmation** — fixed in `a5f1a7d84`, with one correction to the framing.

The trap is real and the fix is right, but it is not a regression from this PR. On `dev` the same modal was already inescapable — `git show origin/dev:src/components/Global/QRScannerOverlay/index.tsx` renders it with `preventClose={modalContent !== EModalType.QR_NOT_SUPPORTED}` and `classButtonClose="hidden"` (lines 446, 449), and `DirectSendContent` (lines 85-117) contains exactly one `<Button>`, the disabled-until-acknowledged Continue. So "unlike the previous modal, the CTA list also has no Close or Cancel action" is not accurate — the previous modal had none either. This PR inherited the trap rather than introducing it.

Fixing it anyway: the PR rewrote this modal, the defect is squarely in the rewritten surface, and a user who scans someone's EVM address by mistake should not have to accept a cross-chain warning and be pushed into the send flow to get out. Added the same secondary the external-link modal already carries:

```ts
{ text: tCommon('close'), variant: 'stroke', onClick: onClose },
```

`onClose` is the overlay's `setIsModalOpen(false)`, so it closes without navigating, and the acknowledgement is reset for the next scan by the existing effect (pinned by the "starts a new scan unticked" case in `base58-case.test.tsx`).

innolope-dev:
**MAJOR: TabSlide force-remounts the whole page tree on every navigation** — fixed in `a5f1a7d84`. Diagnosis accepted; the suggested key is not, because I measured it reintroducing a remount.

The diagnosis is right and this one *is* ours — `TabSlide.tsx` is new in this PR. `key={pathname}` remounted the entire route subtree on every navigation, including the same-component ones the app router preserves state across.

The suggested `key={direction ? pathname : 'tab-content'}` does not hold, because `direction` is not stable for the lifetime of a route. `previous.current` is updated in a `useEffect`, so one commit after the slide lands `tabSlideDirection('/card', '/card')` returns `null` and the key flips from `/card` back to `'tab-content'` — remounting the page it just finished sliding in. Any re-render on the destination route triggers it.

Measured, not argued. I built the suggestion verbatim on the original component shape and ran it against the new pins:

```
✓ keeps the subtree mounted across a same-component navigation
✓ keeps the subtree mounted when navigating to a non-tab route
✕ remounts once on a home -> card tab slide, and not again while it stays there
    Expected: 2
    Received: 3
```

It fixes the collateral remounts and adds a third mount on the tab slide itself.

What landed instead: the key is a counter that only advances when a slide actually plays, with the direction stored next to the path it was computed for.

```ts
const slide = useRef<{ key: number; direction: TabSlideDirection | null; path: string | null }>(...)
if (slide.current.path !== pathname) {
    const direction = reduceMotion ? null : tabSlideDirection(slide.current.path, pathname)
    slide.current = { key: direction ? slide.current.key + 1 : slide.current.key, direction, path: pathname }
}
```

Derived during render, not in an effect, because framer-motion reads `initial` at mount — a direction computed after paint arrives too late, which is why the original needed the pathname key at all. Storing `path` alongside makes both values stable across a double render, so StrictMode's second pass reuses the stored direction instead of recomputing it to `null` and dropping the animation. The `useEffect` that maintained `previous` is gone.

New pin: `src/components/Global/AppShell/__tests__/TabSlide.test.tsx` counts child mounts across three navigations. Non-vacuous — restoring `key={pathname}` fails the first two cases, and the suggested key fails the third (above).

innolope-dev:
**Re: `OtaUpdateContext.tsx:102` — applyNow's reload fallback can deadlock the apply lock**

Declining: the premise ("the 3s exit timer is only started in the catch branch") does not match the file.

`armWatchdog()` is called **unconditionally at `OtaUpdateContext.tsx:116`, before the `try`** — the comment on 111-115 says why, in exactly these words: "a set() that neither reloads nor rejects would otherwise leave the promise pending and the fallback never scheduled". It is armed a second time at :137 once the outcome is `'reloading'`. So a `set()` that **resolves without reloading** returns `'reloading'` (capgo-updater.ts:186-187), re-arms, and `RELOAD_GRACE_MS` later leaves `'applying'`: `manual-restart` off Android (:96-98), `exitApp()` on Android with a `manual-restart` fallback if that rejects (:103-108). There is no path where the modal stays locked in `'applying'`.

That case was untested, so it is now pinned rather than argued — `src/context/__tests__/OtaUpdateContext.test.tsx`, `describe('a set() that resolves instead of reloading')`: Android closes the app after the grace period, off-Android lands on `manual-restart`. Non-vacuity check: restructuring the provider to arm the watchdog only inside `catch` — the shape this comment describes — fails 7 tests in that suite, including both new ones.

On reachability, both halves check out:
- **Web / preview builds**: `applyNow` returns at :90 unless `pendingBundle` is set, and the only writers are inside the `if (!isCapacitor()) return` effect (:53, :61, :68). New test: *"has nothing to apply off native, so no restart can be offered there"* — with `isCapacitor()` false the plugin queue is ignored, `applyNow` never reaches `set()`, and `applyState` stays `idle`. Deleting the `isCapacitor()` gate fails it.
- **Unsupported WebView**: that branch (`ClientProviders.tsx:73-82`) renders `OtaUpdateProvider > AppIntlProvider > UnsupportedWebViewScreen` and nothing else. The provider is there only so `notifyAppReady` runs; `applyNow`'s only consumer is `OtaUpdateModal`, rendered from `Profile/index.tsx:186`, which that branch never mounts.

innolope-dev:
**Re: `OtaUpdateContext.tsx:137` — 3-second fallback exits the app even when the bundle was never applied**

Declining: the guard being asked for is already there, on both legs.

`applyStagedBundle` returns `'failed'` when the re-stage does not come back `staged` (`capgo-updater.ts:216`) and when `reload()` itself rejects (:221-225). The provider then clears the timer, drops the apply lock and sets `'failed'` without arming anything (`OtaUpdateContext.tsx:131-136`) — so `exitApp()` cannot run for an apply that never reached the plugin. `armWatchdog()` at :137 is reachable only after an outcome of `'reloading'`.

Existing coverage in `src/context/__tests__/OtaUpdateContext.test.tsx`: *"abandons the apply when the re-stage comes back offline / up to date"* and *"abandons the apply when reload() itself rejects"* both assert `applyState === 'failed'`, no `reload()`, and no `exitApp()` after 3s.

The marker sub-point does not hold either, and for a reason worth stating: the marker is **removed before** the re-stage (`capgo-updater.ts:198`), so a retarget write that silently fails leaves **no** marker rather than the dead id — `writeStoredValue` swallows the failure (`safe-storage.ts:36-44`) and `reportPendingApply` returns immediately on a null marker (`capgo-updater.ts:234`). Nothing is logged at error level, so no phantom Sentry event. Now pinned: *"reports no failed apply when the marker retarget cannot be written"* makes storage throw on the `b-3` write specifically, and asserts the recovered launch is silent. Removing the pre-re-stage `removeStoredValue` fails it (the marker stays `b-2`, and the launch running `b-3` logs the phantom error).

No code change here.

innolope-dev:
**Re: `QRScannerOverlay/index.tsx:93` — external-URL modal forwards schemed payloads when the scheme check is bypassed**

Accepted and fixed in e727059b8. A regex over the raw scan is the wrong instrument on this boundary, so the verdict now comes from the parser (`QRScannerOverlay/index.tsx:80-91`):

```ts
const candidate = /^[a-z][a-z0-9+.-]*:/i.test(redirectTo) ? redirectTo : `https://${redirectTo}`
try {
    const parsed = new URL(candidate)
    return parsed.protocol === 'https:' || parsed.protocol === 'http:' ? parsed.href : undefined
} catch {
    return undefined
}
```

The scheme-less `https://` prepend stays, and what is displayed and handed to `openExternalUrl` is now the **parsed href**, so the address the user approves in the modal is the one that gets opened.

Two corrections to the report, from reading the code rather than the diff:

1. Whitespace and control characters cannot reach this modal in the first place. `EXTERNAL_URL` is only shown for `EQrType.URL`, whose pattern (`DirectSendQR/utils.ts:85-86`) is anchored `^...$` and admits no whitespace in any of its classes — `  javascript:…`, `java\nscript:…`, `jav\tascript:…` and `data:text/html;…` are all classified as unrecognised one layer earlier. They are in the test file now as guarantees, but they pass on either implementation; so does `JaVaScRiPt:` (the old `/^https?:\/\//i` already rejected it).
2. What the old check genuinely waved through was a payload that *starts* with `https://` but parses to nothing openable — `https://#x.co` (no host) and `%.co` (invalid host, reached via the scheme-less prepend). Both were offered with a live "Open link" button that could only fail inside `openExternalUrl`.

`src/components/Global/QRScannerOverlay/__tests__/external-url-scheme.test.tsx` covers all seven refusals plus the two openable cases. Non-vacuity: reverting to the regex fails exactly the two cases in (2) — so those are the behaviour change, and the rest document the layer above.

innolope-dev:
**Re: `AppShell/TabSlide.tsx:34` — TabSlide still remounts the route subtree for every home↔card navigation**

Declining, on a measurement rather than an argument.

The key increments **only** on home↔card (`TabSlide.tsx:41-48` via `tabSlideDirection`) — and those are precisely the navigations where the app router has already swapped the subtree itself. In `next@16.2.3`, `layout-router` renders each segment's tree under its own router cache key (`TemplateContext.Provider` carries `stateKey` as its React key), and the bfcache keeps only the active entry unless `cacheComponents` is on (`bfcache-state-manager.js`: `MAX_BF_CACHE_ENTRIES = 1`), which it is not here. `/home` and `/card` are different segments, so they unmount each other with or without this key.

`src/components/Global/AppShell/__tests__/TabSlide.test.tsx` now counts mounts across `/home → /card → /home` through TabSlide and through a keyless animate-only wrapper, with a router-shaped child (keyed per segment):

| wrapper | mounts across the trip |
| --- | --- |
| `TabSlide` (key present) | 3 |
| keyless wrapper | 3 |

Identical — the key adds no remount on the only navigations it fires for, so no mount effect re-runs and no page state is discarded that the router was not discarding anyway. To show the harness can see a difference at all, the same measurement with a child the router would have preserved gives 3 vs 1; that divergence is also the evidence that the motion node itself remounts on a tab change, which is the only thing that replays `initial`.

The existing tests still pin the inert cases (same-component navigation, non-tab route, re-render at an unchanged route), and a new one pins the animation contract directly: `initial` is `{x: 40}` on home→card, `{x: -40}` on card→home, and `false` on non-tab navigations and under `prefers-reduced-motion`.

Swapping to `useAnimationControls` would trade a key that measurably costs nothing for an imperative effect that has to fire after the route commit — later than `initial`, which is read at mount. Keeping the key.

innolope-dev:
## Re-checked all five threads on `e727059b8` — no code change needed on any

Three are re-raises whose new specifics are refuted by construction, and the two OTA ones were already answered with tests. Evidence per thread.

### TabSlide — the `/card → /card/pin → /card` scenario cannot happen

The new claim is that this sequence "increments the key twice in quick succession and remounts the card page mid-slide". It increments it **zero** times.

`tabSlideDirection` returns `null` unless `pageTab(from)` and `pageTab(to)` are both non-null **and different** (`tab-order.ts:31-34`). `pageTab` resolves through `isSameRoute`, which is exact match after a trailing-slash strip (`constants/routes.ts:236-240`) — so `pageTab('/card/pin')` is `null`. Both legs of that sequence therefore return `null`, and `TabSlide.tsx:44` only bumps the key `if (direction)`. The key moves solely on a genuine home↔card crossing, which is also the only navigation where the router swaps the subtree anyway — the 3-vs-3 measurement already posted.

### External-URL QR — `router.push(redirectTo!)` can never receive a scanned string

The remaining concern is that "the raw `redirectTo` string remains the value used elsewhere". `redirectTo` has exactly two writers, and they are disjoint from the two readers:

- `setRedirectTo(data)` — the raw scan — is only ever paired with `showModal(EModalType.EXTERNAL_URL)` (`index.tsx:383-386`). That is the branch now gated on `new URL()` + a `parsed.protocol` allowlist; a refused payload renders the unrecognised modal, whose only CTA is `okay → onClose`.
- `setRedirectTo(toConfirmUrl)` — an app-built internal path — is only ever paired with `EModalType.DIRECT_SEND` (`index.tsx:411-414`), which is the sole branch containing `router.push(redirectTo!)` (`index.tsx:148`).

So the raw scan reaches only the guarded external modal, and `router.push` receives only a URL this app constructed. There is no path from a scanned payload to `router.push`.

On the suggested tightening — "refuse anything whose raw scan contains a scheme at all" — that would refuse every legitimate `https://example.com` scan, which is the primary case this modal exists for. The parsed-protocol allowlist is the correct shape: it accepts exactly `http:` and `https:` as the URL parser resolves them, and forwards `parsed.href`, so the address shown is the address opened.

### The two OTA threads

Both were answered on the previous head with tests rather than prose, and nothing in the restated text engages with that evidence. Briefly: `armWatchdog()` is called unconditionally **before** the `try` (`OtaUpdateContext.tsx:116`, with the comment naming this exact case) and again on a `'reloading'` outcome (:137) — so a `set()` that resolves without reloading is covered, and a test now pins that (restructuring to arm only inside `catch`, the shape the finding describes, fails 7 tests). The exit guard likewise exists on both legs (`capgo-updater.ts:216` and :221-225 return `'failed'`; the provider then clears the timer and never arms).

### The new MINOR — retry after a failed apply

Verified; it is safe, and the truncated diff hid the answer.

- The lock is released before `'failed'` is set: `applyingRef.current = false` (`OtaUpdateContext.tsx:132-134`), so the retry passes the guard at :90.
- The marker is **cleared on every failure path** — `abandon()` calls `removeStoredValue(PENDING_APPLY_KEY)` (`capgo-updater.ts:182`, :198) — and rewritten fresh on the retry (:176). There is no stale marker to carry forward.
- Double-staging is prevented structurally: every re-stage goes through `queueUpdateCheck` → `queueOtaWork`, serialized on the module-level `pendingOtaWork` promise. And `'failed'` is only set after the previous `applyStagedBundle` has settled, so two applies cannot overlap in the first place.

---

All checks are green on this head. Nothing outstanding in code.

PR #2928, opened 2026-09-02
fix(sentry): one issue for fetch timeouts, and a shorter QR scan budget

innolope-dev:
## Why

Two Sentry issues prompted this — PEANUT-UI-SZQ (`/manteca/qr-payment/init` timed out after 20000ms) and PEANUT-UI-SZT (`/fx/card-markup` timed out after 10000ms). Neither is a server fault:

| endpoint | reqs (7d) | avg | p95 | **max** | client budget |
|---|---|---|---|---|---|
| `GET /fx/card-markup` | 12,001 | 87ms | 662ms | **3.0s** | 10s |
| `POST /manteca/qr-payment/init` | 13,572 | 397ms | 1.96s | **12.9s** | 20s |

The slowest *completed* server request in a week sat well under both budgets, so these are mobile clients giving up on a request in transit — the same shape as the `/rain/cards` finding.

Looking at the wider population turned up the real problem: **30 separate timeout issues in 7 days for one phenomenon**, each too small to alert on and collectively invisible. Same endpoint, multiple issues (`/rain/cards` = QD5 + SYD, `/users/me` = QHG + SZ2), plus a single-event issue for every dynamic path (`/rhino/status/0xfe73…`, `/users/username/desravinekel`).

## What changed

**Grouping — one issue instead of thirty**
- Timeouts fingerprint on `['timeout']` alone with a stable `Request timed out` title. The endpoint moves to a `route` tag that Sentry can break down and alert on *within* the issue; the raw url stays in `extra`. Ranked endpoints in one tag breakdown beats 30 issue rows.
- `route` + `http.method` tags added to the non-2xx and network-error captures too.
- `sanitizeUrl` now collapses hex path segments (addresses, tx hashes, link ids) and profile usernames, which were opening one issue per address/username.
- ⚠️ `isFetchSiteMutationFailure` in `beforeSend` read the method from `fingerprint[2]`. Dropping it from the fingerprint would have made that money-flow rescue **silently inert for exactly the events it exists to keep** — a POST that dies on the network. It now reads the `http.method` tag, with the positional slot as fallback for the captures that still fingerprint that way.

**Reporting — stop paging on designed fallbacks**
- New `silentTimeout` opt-out, used by `fetchCardMarkup`. `useCardMarkupRate` already catches this, falls back to the static markup table and renders the row anyway ("Never block a payment surface on a comparison number") — so the error-level issue was reporting a path behaving exactly as specified. The throw is unchanged so the fallback still runs, a breadcrumb keeps it on the trail, and a non-2xx / bad JSON / broken contract on that endpoint still reports.
- `FEATURE_TAGS` gains `qr-pay`, `manteca` and `fx`.

**QR scan — the one with real user impact**
- Scan-time init gets a 10s budget instead of the 20s default. The server has never needed more, so the extra 10s only decided how long someone stands at a till: ~89s worst case before, ~49s now.
- Terminal "provider is down" copy now waits for the retries to actually be exhausted. It was rendering off `failureReason` — i.e. the **first** failure — so a scan that recovered on attempt 2 had already told the user the rail was down. A new `Still fetching details` state covers the retry window (en / es-419 / pt-BR; es-AR inherits from es-419).

## Worth a second opinion

The retry policy's comment claimed *"Max 3 attempts: immediate, +1s delay, +2s delay"*. It is actually **4 attempts with a flat 3s delay** — `failureCount` is 0-based where react-query consults it, so `failureCount < 3` permits three retries. I corrected the comment and left the behaviour alone, but if 3 was the intent, that's a one-character fix (`< 2`) and takes the worst case to ~39s. Deliberately not doing it here — it's a payment retry policy, not a comment.

## Testing

- New `sentry-timeout-reporting.test.ts` — one fingerprint across endpoints, url out of the title, route/method/feature tags, `silentTimeout` suppresses the capture but still throws and still reports non-2xx.
- `sentry-fingerprint-url.test.ts` — hex and username collapsing, short hex-looking slugs left intact.
- `sentry.utils.test.ts` — the mutation rescue off the `http.method` tag, both directions.
- `qr-pay-states.test.tsx` — the retry window now holds the loading state and only shows terminal copy once the attempts are spent.
- `pnpm typecheck` clean; 88 util/hook suites (1735 tests) and 23 i18n suites (281 tests) green.


innolope-dev:
All three findings addressed in `9f3b758` + `982ff1f`.

**MAJOR — end-to-end budget.** Confirmed and fixed at the transport layer. `fetchWithSentry` now sizes **one budget pool per call** (`timeoutMs * maxAttempts`) instead of handing `timeoutMs` fresh to every leg:

- The WebView retry keeps its own per-attempt budget, so the silent GET retry (PEANUT-UI-R44) still gets two full transport attempts.
- The OS-client fallback draws on the remainder. In the case it exists for (PEANUT-UI-R5F, edge rejecting the WebView at the TLS layer) the WebView fails fast, so the fallback is still handed nearly the whole pool — asserted at >4s of a 5s budget.
- When the WebView leg genuinely times out, the pool is spent and the fallback is skipped rather than starting a second full-length leg.

The ~49s scan bound now holds on Capacitor too, so the constant's claim is true on both platforms. Three tests in `sentry-prefer-native.test.ts` pin all three behaviours.

**MINOR — dead caption.** Correct, and it defeated the point of the state. `Still fetching details` was set on the context but the page renders a bare `<Loading variant="mascot" />` while the lock is absent, so the user saw the same unlabeled spinner. It now renders through `QrPayPageLoading`, the captioned loader this page already uses for the waiting-for-merchant state, and the test asserts the **visible caption** instead of the context value.

**MINOR — untested recover path.** Correct, including the reasoning about the `failureReason` invariant. Added the sibling test as specified: `mockRejectedValueOnce` then resolve, asserting two calls, the merchant name on screen, no spinner and no error copy.

One correction to the review's framing, for the record: this PR did not introduce the native `2x` — native was already ~89s before it. The change improved web and left native as it was. It's genuinely fixed now on both.

innolope-dev:
Addressed in `18984ac`. Three fixed, one declined with data.

**MAJOR — OS-client-first POST loses its WebView fallback.** Correct, and mine. The budget pool *gated* the two rescue legs but only *floored* the WebView leg, so on the `preferNativeTransport` ordering a stalled OS client spent the pool and the fallback was armed with 250ms it could never complete in. Now gated like the others — skip the leg and throw the timeout — which is what `MIN_TRANSPORT_LEG_MS`'s own rationale asks for. A *fast* OS-client failure still leaves the pool whole and still gets its full second chance; both orderings now have tests.

Two further defects surfaced while fixing it, both mine:
- The floor was absolute, so it refused **every** leg of any call whose total budget was under 250ms. It is now `min(MIN_TRANSPORT_LEG_MS, timeoutMs)`.
- The pool didn't cover the 300ms backoff *between* attempts, so the silent GET retry was being squeezed out by its own delay. The pool is now legs + backoff.

**MAJOR — deterministic init rejections behind the retry caption.** Correct, and a regression I introduced. Verified all five against the sibling route: `MANTECA_SOURCE_OVER_MONTHLY_CAP`, `MANTECA_MERCHANT_VOLUME_NEAR_CAP`, `MANTECA_MERCHANT_RECENT_REFUND` (all `cap-check.ts`, 422 via `error: <code>`), `MANTECA_USER_NOT_PROVISIONED` (422), and the KYC 400 whose `error` is the sentence `User KYC not approved`. All added, with a case per code asserting one call and no caption.

Also took the earlier `isPaused` finding: a device dropping offline mid-retry parks the query, so no AbortController fires and the terminal error never arrives — that would have pinned the scan on the caption indefinitely. `retriesPending` now excludes `fetchStatus === 'paused'`.

**MINOR — assert the timeout reaches serverFetch.** Added `manteca-qr-init-transport.test.ts` covering both hops, plus that omitting the option sends no `timeoutMs` key at all.

**MINOR — 10s cutoff below the observed 12.9s max. Declining, with the measurement.** I pulled the distribution rather than argue it: over **13,708 requests in 7d, p99 is 6.59s, and exactly 10 — 0.073% — crossed 10s**. Those are retried, not failed. Raising the budget to clear the 12.94s outlier would make *every* genuinely stalled scan wait 50% longer to rescue 0.073%. That trade is the point of the constant, so I've put the numbers in its comment instead of changing it — happy to revisit if you read the tail differently.

innolope-dev:
Addressed in `7fd55d1`. Three fixed, the 10s cutoff still declined.

**MAJOR — native leg not bounded by the deadline.** Correct, and it made my own claim false. Handing the remaining budget to `nativeHttpRequest` bounded nothing: CapacitorHttp splits it into `connectTimeout` and `readTimeout`, two separate phase limits, and on Android the read limit resets as data arrives — so one leg could take 2x, and a slow drip was unbounded. The await is now raced against the wall clock and rejects with an `AbortError` the caller already classifies as a timeout. The underlying request is left to finish and discarded — CapacitorHttp exposes no abort, and an aborted `fetch` leaks the server's work the same way, so this adds no new class of problem on a non-idempotent init.

**MAJOR — paused retries recovering behind a latched error.** Correct, and a path I created. `errorInitiatingPayment` gates the whole render and nothing cleared it, so a scan that reconnected and succeeded sat behind a permanent outage message holding a valid lock. A lock in hand now clears any earlier init error. Covered by driving react-query's own `onlineManager` through offline → pause → reconnect → success.

**MINOR — cap/KYC shown as an outage.** Taking this one, and thank you for the honest framing — you're right that the copy predates the PR, and equally right that this is the diff that named the codes and pinned the wrong string in a test. Cap, merchant-limit and KYC now have their own branches and copy in en / es-419 / pt-BR. Also took the wire-code point: `initiateQrPayment` now throws `ApiError`, so the KYC rejection matches on `MANTECA_KYC_REQUIRED` rather than the English sentence — with a test asserting a *reworded* prose message still routes correctly. Message order is unchanged, so every existing matcher still sees what it did.

**MINOR — 10s cutoff. Still declining, same measurement.** Over 13,708 requests in 7d: p99 6.59s, and exactly 10 — 0.073% — crossed 10s, and those are retried rather than failed. On the new argument (sustained upstream slowness failing all four attempts): if the rail is sustainably answering slower than 10s, four attempts at 20s means ~89s at a till before the same failure, which I don't think is the better outcome for that user. The numbers are in the constant's comment so the trade is visible rather than assumed. Genuinely happy to be overruled by the product owner here — it's a judgment call about which tail matters more, not a correctness question.

innolope-dev:
All four addressed in `00eb644`.

**MAJOR — wire-code KYC still retried.** Correct, and mine. The result effect read `wireErrorCode` but the retry predicate matched only `error.message`, so a 400 carrying `MANTECA_KYC_REQUIRED` with reworded prose showed the right copy on attempt one and then scheduled three more POSTs. Predicate now checks the code too, via a `NON_RETRYABLE_QR_PAY_CODES` list beside the prose one.

Your point about the test is the sharper half: it asserted one call *before* the 3s backoff could fire, so it passed against the bug. That's the third "assertion that cannot fail" in this PR after the `http.method` tag and the timeout forwarding. It now waits past the backoff, and I verified it fails without the predicate fix.

**MINOR — cap copy.** Verified both claims and both were wrong in my copy:
- `mono product/kyc.md:52` — `scope: "Deposits, withdrawals, and QR payments combined"`. Calling it a "monthly QR payment limit" implies deposits and withdrawals still work.
- `peanut-api-ts src/manteca/cap-check.ts:123` — `if (attempted <= available) continue`, so the 422 fires when *this payment* exceeds remaining headroom, not when headroom is zero. My copy dropped the backend's own "try a smaller amount" and replaced it with a hard month-end reset — sending away a user who could still pay less.

Now: *"This payment is over your remaining monthly limit, which is shared across deposits, withdrawals and QR payments. Try a smaller amount, or raise your limit in the app."* — en / es-419 / pt-BR, with a test pinning both the affordance and the shared scope.

**MINOR — KYC ETA promise.** Correct, and I invented that reassurance. `mono product/lessons-from-corrections.md:109` records "verification takes 2 minutes" as a correction, and `MANTECA_USER_NOT_PROVISIONED` reaches this copy for partial-onboarding users who may already be in manual review. Sentence dropped, with a test asserting no completion time is offered. Taking the scope call your way — it's the only wait-time guidance product/ holds and the copy contradicted it directly.

**MINOR — native race untested.** Correct, and the reasoning is exactly right: `sentry-prefer-native.test.ts` mocks `@/utils/native-http`, so the race could have been deleted with every test in the repo still green. Added to `native-http.test.ts`: a never-settling plugin promise rejects at the budget with `name: 'AbortError'` (which `fetchWithSentry` classifies on), plus the inverse that an in-budget response still resolves.

innolope-dev:
Taking the one MINOR, fixed in `8218e741`.

Verified the claim in the sibling repo and it's right — the copy stated something false. `checkMerchantBeneficiaryHistory` is reached only when `isQr3ArgPoolPayment` (`peanut-api-ts src/routes/manteca/qr-payment.ts:374-376`), i.e. `mantecaUserId === MANTECA_USER_ID_ARG`, the shared AR corporate pool. So `MANTECA_MERCHANT_VOLUME_NEAR_CAP` fires when *every Peanut pool user combined* has pushed 2,000,000 ARS through that merchant in 30 days, and `MANTECA_MERCHANT_RECENT_REFUND` can fire on a refund belonging to another user. The merchant accepts QR fine — our pooled source account is what can't pay it. The backend scopes its own message accordingly (`cap-check.ts:292`, "for our payments") and I'd dropped that scoping.

Now: *"We can't complete QR payments to this merchant right now. Try another payment method, or a different merchant."* — en / es-419 / pt-BR, keeping the actionable second sentence, with a test asserting the copy doesn't claim the merchant is the one refusing.

That's the last outstanding finding. Recap of where the review landed across five rounds:

| Area | Outcome |
|---|---|
| Sentry timeout grouping, route/method tags, mutation rescue, sanitization | fixed, no findings since |
| Shared native/WebView timeout budget | fixed across 4 rounds — the model was wrong in several distinct ways before it was right |
| QR retry state (caption, recover, paused/offline, deterministic fail-fast) | fixed |
| Error copy vs product truth (cap scope, cap semantics, KYC ETA, merchant blame) | fixed, all four verified against `mono/product/` and the sibling repo |
| 10s scan cutoff | refuted with measurement — accepted as an explicit latency policy |

Three of the defects found along the way were assertions of mine that could not fail (the `http.method` tag, the timeout forwarding, the wire-code retry). Worth noting for anyone reading this thread later: that pattern, not the transport logic, was the most productive thing this review caught.

innolope-dev:
Both actionable items fixed in `9c1df73`.

**MINOR — cap copy promised an in-app limit raise.** Verified and correct. `isBrUserEligibleForLimitIncrease` (`src/features/limits/utils.ts:264-272`) requires a BRL/BRA limit at or under 1000 USDT; every Argentina QR3/MercadoPago user — this branch's entire population — falls through to `openSupportWithMessage`. `mono product/kyc.md:209` documents the path as "Contact support or submit additional documentation". And as you note, the QR error screen renders a bare string with no link, so "in the app" pointed nowhere from a till.

Now *"Try a smaller amount, or contact support to raise your limit"* in en / es-419 / pt-BR, with the existing cap-copy case extended to assert no self-service promise. Agreed this is distinct from the earlier cap finding — that one was the QR-only naming and the month-over framing, both fixed in the prior revision.

**P7 — pin the 10s scan budget.** Taking this inline rather than filing a task; it's one assertion. You're right that the forwarding tests pass just as well against the 20s default, so nothing pinned the number the constant argues for from measured latency. `expect(MANTECA_QR_INIT_SCAN_TIMEOUT_MS).toBe(10_000)` now makes raising it a deliberate edit.

**Not fixed here — cap rejections on the amount-entry re-init path.** `handleMantecaPayment` still routes those to `errors.initiateUnexpected`, which is the same harm class the scan-time path just stopped doing. A human closed that thread and it has a Notion task, so it stays out of this PR by that decision rather than by oversight. Flagging it plainly because the task title understates it: a capped AR user who types an amount still gets "unexpected error", which is the exact framing `product/providers/fiat/README.md` records as having blocked someone for three days.

innolope-dev:
Both fixed in `6632821`.

**MAJOR — cap rejections after amount entry.** Correct. The mapping only ran in the scan-time effect, and an open-amount QR gets a lock with an *empty code* from the scan — so its cap verdict only exists once the amount is re-submitted from the Pay button, and that catch handled `PIX_MIN_AMOUNT` alone. The actionable guidance this PR added was unreachable on the one path where the amount is the thing being rejected.

Both call sites now share `deterministicInitErrorMessage`. It returns null for everything else, so each path keeps what genuinely differs — the scan path owns decode/expiry/recurring and the retry state machine, the re-init path owns its network-triage capture.

**MINOR (kimi) — offline scan blamed the rail.** Good catch, and it was mine. A paused query is a *positive* signal that the device lost connectivity — nothing else parks it — and falling through to `providerIssues` told a user with no signal that MercadoPago was down. That's the same misdiagnosis the rest of this screen's copy exists to remove. Paused now gets its own connection copy in all three locales; `providerIssues` stays for genuinely exhausted retries.

Also fixed while there: the result effect was missing `paymentLockFetchStatus` from its dependency array, so it only re-ran on pause by accident of `isLoadingPaymentLock` changing alongside it.

**On coverage, being precise about what I did and didn't do.** You asked to cover the open-amount cap rejection end to end. I tried and backed out: the mocked `AmountInput` leaves the Pay button disabled, so driving that path needs scaffolding this suite doesn't have. Instead the classifier is covered directly — 12 cases over every code, the wire-code path, and the null cases that keep each caller's own handling intact. That proves the mapping and, since one function now serves both call sites, the shared behaviour; it does not prove the re-init wiring end to end. If you'd rather have the full flow test, that's a reasonable ask and I'd add the harness scaffolding for it — just say so.

innolope-dev:
Closed in `cb0a021` — this was the gap I flagged last round rather than fixed, and you were right to push on it.

The blocker I hit was self-inflicted, not the harness: the mocked `AmountInput` drives the USD leg with the same string, so typing `5000` asked for $5,000 against the $100 test balance and the Pay button stayed disabled. With a sane amount the flow runs fine, and no scaffolding was needed.

The test does what you specified: renders an open-amount QR (scan resolves a lock with an empty `code`), enters an amount, rejects the second `initiateQrPayment` with `MANTECA_SOURCE_OVER_MONTHLY_CAP`, and asserts `/remaining monthly limit/` is shown, the generic copy is not, and `captureNetworkTriagedFailure` did not fire — a deterministic rejection is not a transport surprise.

**I verified it actually catches the regression** rather than trusting the green tick: reverted line 865 to the old `PIX_MIN_AMOUNT`-only check, confirmed the test fails, restored the fix, confirmed it passes. Given this review found four of my assertions that couldn't fail (`http.method`, timeout forwarding, wire-code retry, and the budget's literal value), a new test claiming to close that exact class of gap shouldn't be taken on trust either.

That's every finding across eight rounds either fixed or refuted with evidence. Thanks — the review materially improved this, and not mainly on the transport logic: the most valuable thing it caught was a repeated habit of writing assertions that pass whether or not the code works.

innolope-dev:
Both addressed in `53a1c90` — one fixed, one done for a different reason than stated.

**MAJOR — capped open-amount payments couldn't retry.** Correct, and it's the sharpest finding of the whole review: I wrote copy telling the user to try a smaller amount and then built a screen that refuses to accept one. The cap rejection set `errorMessage` with no code, `isBlockingError` stayed true, and Pay was disabled for the rest of the scan — the only way out was abandoning the QR at the till.

The classifier now reports whether a rejection is *amount-shaped*. A cap (payment over the remaining headroom) and a Pix minimum (payment under the floor) both invite another try; a merchant limit or unfinished KYC doesn't care what the user types. Amount-retryable ones carry an `amountRetryable` code that `isBlockingError` treats the way it already treats `balanceSettling`, and editing the amount clears the last init error so a stale message never sits beside a new number.

The call-site test now follows the advice end to end: cap rejection → enter a smaller amount → assert the message clears and Pay re-enables → submit again. Verified it fails with the recovery half reverted.

**MAJOR (kimi) — page export breaks `next build`. Doing the move, but the stated failure mode does not reproduce.** `Deploy-Preview` ran a full **4m37s `next build` on `cb0a021` and passed**, and `next.config.js` sets no `ignoreBuildErrors` — so the export was not failing app-router page validation. I'd rather say that than accept a build-breakage claim the evidence contradicts.

The move is still right for the *other* reason given, which I think is the real one: a page should export only what Next.js allows, and a test importing a pure function shouldn't drag the page's whole graph (wallet hooks, zerodev SDK, posthog) in behind it. So the classifier now lives in `qr-pay/init-error-classifier.ts`, imported by both.

innolope-dev:
Fixed in `faf6c32`. Verified before acting: I dumped every string in that `qrPay.errors` block and checked the register. Every register-marked neighbour is tuteo — `Pídele`, `escanea`, `Inténtalo`, `Contacta`, `Pregúntale` — and my four were the only voseo in the file. The finding is right on both counts: region-marked for Mexico/Colombia, and inconsistent with the copy rendered beside it.

Fixed the way the repo already models this rather than by flattening it. `es-AR` exists as a delta over `es-419` and already overrides exactly this class of string (`pixDecode`, `sessionExpired`, `completeFailed`, `initiateUnexpected`) with voseo. So:

- **es-419** takes the neutral tuteo forms — "Revisa tu conexión y escanea de nuevo", "Prueba con un monto menor o escríbenos", "Termina de verificar tu identidad".
- **es-AR** takes the voseo variants as a delta.

Argentina — the corridor actually hitting these QR errors — still reads natural Rioplatense, and nobody else gets grammar aimed at Buenos Aires. Confirmed after the change that es-419 has zero voseo keys and es-AR's overrides sit alongside the existing ones.

Not a regression I'd have caught myself: I wrote the Spanish for the users I had in my head (AR QR payers) and dropped it in the file that serves the whole continent.

innolope-dev:
Fixed in `d947c8a`. Both halves verified against the sibling repo first.

**Terminal errors were being cleared.** Correct, and this is a defect my *previous* round's fix created — the amount-edit clear I added to make cap rejections recoverable was unconditional, so an unfinished KYC or a merchant refund block re-enabled Pay on a single keystroke. The user could then re-POST a rejection no amount can change, and on the refund path that mints another Manteca price lock before the backend applies the block. The clear is now gated on the error carrying `amountRetryable`.

**The merchant split is right too, and I had it backwards.** Checked `cap-check.ts`:
- `MANTECA_MERCHANT_VOLUME_NEAR_CAP` (line 288) — `rolling30dTotal + attempted >= LIMIT`. Amount-dependent; a smaller amount can fit under the merchant's remaining headroom. I had classified it terminal.
- `MANTECA_MERCHANT_RECENT_REFUND` (line 259) — keys off `lastRefundAt` age and `windowRefundCount`, never reads the amount. Genuinely terminal.

Both directions covered: amount-retryable clears and re-enables Pay; terminal stays latched with the button disabled and no further POST possible. Verified the latch test fails against the unconditional clear.

**One thing I left alone deliberately:** the volume cap is now amount-retryable but still shows `merchantNotAvailable` ("try another payment method, or a different merchant"), which doesn't mention a smaller amount. Not false — those alternatives work — but it under-sells the option the backend actually leaves open. I didn't split the copy because that's a fourth locale pass on a merchant that is by definition nearly maxed for the month, so "try a smaller amount" is marginal advice. Flagging it rather than silently deciding; happy to split it if you'd rather.

Worth noting the pattern: rounds 7, 9 and 10 each found a defect introduced by the previous round's fix. The review is catching them, but that's three consecutive fixes with unexamined blast radius on my side.

PR #2930, opened 2026-09-02
fix(about): append the CI build number to the release version

innolope-dev:
## What

The About screen showed `Version 1.1.1 (34534)`. It now shows `Version 1.1.34534`.

The patch digit in that string came from `MARKETING_VERSION`, which is typed by hand into `ios-release.yml` / `android-release.yml` at release time and can sit unchanged for months. The build number — the run number of the release workflow, and the only value that identifies a specific build — sat in parentheses next to it. So the part that read as "the version" was the part that had not moved, and the part that had moved read as an aside.

`formatBinaryVersion` now composes `<major>.<minor>.<build>`, so the digit a user reads out to support is the one that changes every release.

## Edge cases

- A binary reporting fewer than two version segments keeps the old `<version> (<build>)` shape, rather than rendering `undefined` into the one number a support conversation depends on.
- Web is untouched: `getBinaryInfo()` returns null with no binary to ask, so the bundled `package.json` fallback still renders as-is.

## Tests

New `src/utils/__tests__/app-version.test.ts` covers the patch-position swap, a marketing patch digit being ignored (`1.1.0` and `1.1.99` both → `1.1.<build>`), a multi-digit minor, and the two malformed-version fallbacks. `useAppVersion.test.tsx` now stubs only the bridge call and runs the real formatter, so the string About renders is exercised end to end.

Verified non-vacuous: corrupting the formatter fails 4 of the 7 cases.

Typecheck, prettier and the ds-lint ratchet are clean.

## Note

Split out of #2925 deliberately — that PR is unrelated, several review rounds deep, and gated on peanut-api-ts#1492 merging first. No reason to hold a one-line display fix behind it.

### DIFF
--- a/src/hooks/__tests__/useAppVersion.test.tsx
+++ b/src/hooks/__tests__/useAppVersion.test.tsx
@@ -19,8 +19,8 @@ describe('useAppVersion', () => {
 
         const { result } = renderHook(() => useAppVersion('1.0.53'))
 
-        // the build number takes the patch position — it is the digit that moves
-        await waitFor(() => expect(result.current).toBe('1.1.412'))
+        // the release version is kept whole; the CI build is a fourth segment
+        await waitFor(() => expect(result.current).toBe('1.1.0.412'))
     })
 
     it('keeps the bundled version on web, where there is no binary to ask', async () => {
--- a/src/hooks/__tests__/useAppVersion.test.tsx
+++ b/src/hooks/__tests__/useAppVersion.test.tsx
@@ -2,7 +2,12 @@ import { renderHook, waitFor } from '@testing-library/react'
 import { useAppVersion } from '../useAppVersion'
 
 const mockGetBinaryInfo = jest.fn()
-jest.mock('@/utils/app-version', () => ({ getBinaryInfo: () => mockGetBinaryInfo() }))
+// Only the bridge call is stubbed — the formatter runs for real, so the shape
+// the About screen renders is exercised end to end.
+jest.mock('@/utils/app-version', () => ({
+    ...jest.requireActual('@/utils/app-version'),
+    getBinaryInfo: () => mockGetBinaryInfo(),
+}))
 
 describe('useAppVersion', () => {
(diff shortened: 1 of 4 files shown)


innolope-dev:
Fixed in `e10675e9` — the finding was right, and the original change was wrong in exactly the way described.

`scripts/release-version.mjs` documents the scheme verbatim:

```
Scheme: <major>.<build>.<ota>
  build   native build counter. Every store build bumps it and resets ota to 0.
  ota     OTA counter within one native build. 0 is the JS baked into the binary.
```

So the segment I substituted was the OTA counter, and `1.1.0` + build `34534` rendered as `1.1.34534` — a string naming OTA revision 34534, which has never existed. Worse on a support call than the format it replaced, since it looks like a legitimate version.

The version line is now `<major>.<build>.<ota>.<ci-build>` — `1.1.0.34534`. `appVersion` is passed through whole, and the CI build is appended as a fourth segment rather than substituted for any of the three. That keeps the canonical identifier intact while still surfacing the build number as part of one string a user can dictate, which was the point of the change.

On the cross-platform gap: also correct, and I have left it visible rather than papered over. `android-release.yml` sets `ANDROID_VERSION_CODE=$((10000 + GITHUB_RUN_NUMBER))` while `ios-release.yml` passes `CURRENT_PROJECT_VERSION="$IOS_BUILD_NUMBER"` (`github.run_number`) as-is, so the same release genuinely shows a ~10000 gap between an Android and an iOS device.

I did not normalise it. Subtracting 10000 on Android would be guesswork about which builds predate the offset — the `+10000` exists to clear legacy version codes — and it would make the displayed number stop matching what Play Console and TestFlight show, which is what someone reading it off a device needs it to match. The number identifies a build; it does not order them across platforms. That is now stated in the code comment so the next reader does not try to "fix" it.

Tests cover the OTA counter surviving (`1.1.4` → `1.1.4.34534`, not `1.1.34534`), both platforms' build numbers rendering verbatim, and the two missing-component cases degrading without a leading dot or an `undefined`.

PR #2934, opened 2026-09-03
fix(kyc): localize the tax_id_unresolved rail reason

innolope-dev:
## What

peanut-api-ts#1492 adds a rail reason code the app has never seen:

```
{ code: 'tax_id_unresolved', userMessage: 'Local deposits and withdrawals need a local tax ID…' }
```

Rail reasons resolve through `REASON_CODE_KEYS` (`src/constants/capability-reason-labels.consts.ts`), and an unknown code falls back to the backend's English `userMessage`. That fallback is deliberate, but here it lands on precisely the population that can reach the branch: the resolver only emits this reason for a **first-party** Manteca rail, and Brazil's only geo rail is the QR-pool `PIX_BR` (promoted back to ENABLED, never a block) — so today it is Argentina only. An es-419 audience reads the feature's headline message in English.

Render sites affected: `src/components/Kyc/InitiateKycModal.tsx:88` and `src/components/Profile/views/UnlockPayments.view.tsx:259`.

## Changes

- `tax_id_unresolved: 'reasons.tax_id_unresolved'` in `REASON_CODE_KEYS`.
- Catalogue entries in `en`, `es-419`, `es-AR` (voseo) and `pt-BR`, matching how the sibling Manteca codes (`manteca_us_nationality`, `country_not_supported`) are carried.

`identity` is not in `marketing-namespaces.json`, so no `generate:marketing-messages` run is needed.

## Testing

`pnpm tsc --noEmit`; the 17 i18n suites (222 tests) and the `constants` / `Kyc` / `Profile` suites (539 tests) all pass.

## Merge order

Independent of api-ts#1492 in both directions — the backend copy is already English-safe via the fallback, and this key is inert until the backend ships the code. Either order is fine.

### DIFF
--- a/src/constants/capability-reason-labels.consts.ts
+++ b/src/constants/capability-reason-labels.consts.ts
@@ -52,6 +52,10 @@ export const REASON_CODE_KEYS = {
     // restrictions
     manteca_us_nationality: 'reasons.manteca_us_nationality',
     country_not_supported: 'reasons.country_not_supported',
+    // A Manteca first-party rail whose submission found no CUIT/CPF. Reachable
+    // today only in Argentina, i.e. an es-419 audience — without this entry the
+    // one population that can see it gets the backend's English userMessage.
+    tax_id_unresolved: 'reasons.tax_id_unresolved',
     uk_resident_blocked: 'reasons.uk_resident_blocked',
 } as const
 
--- a/src/i18n/app/messages/en.json
+++ b/src/i18n/app/messages/en.json
@@ -2988,6 +2988,7 @@
             "card_rejected": "Your card application couldn't be approved. Message support if you think this is a mistake.",
             "manteca_us_nationality": "Payments from this country are not available for US citizens at this time.",
             "country_not_supported": "This rail needs a document from a supported country. You can verify with a different ID.",
+            "tax_id_unresolved": "Local deposits and withdrawals need a local tax ID, and we couldn't read one from your documents. You can verify again with an ID that carries one. QR payments aren't affected.",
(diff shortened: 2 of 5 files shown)


innolope-dev:
Agreed on the verdict — the code is the correct side and the docs are wrong — and the copy in this PR is unchanged as you asked.

The doc correction was already underway in **mono#125** (`70bd511c`), covering `product/kyc.md` (the `regional_id_policies` global policy, the Argentina and Brazil per-country notes, the eligibility line at :171, and the tourist note at :240) plus `product/quick-ref.md:82`.

But your citation of **`countries.md:25` and `:36` caught a real miss** — that file was not in that commit, so it was still carrying the same promise in two more places. Fixed in `cedd12b6` on the same branch:

- `local_id_required: false` now says "sufficient **to verify**", with a comment naming the AR/BR local fiat rails as the one exception and pointing at `providers/provider-requirements.md`.
- Both `foreigner_access` notes now say the local rails are a per-user Manteca account requiring residence **and** a document carrying a CUIT/CUIL/DNI or CPF, while keeping the passport promise exactly where it is true — verification and QR, which settle through the pooled account.

That is the same shape as the `kyc.md` correction, so the four files now agree with each other and with `provider-requirements.md:90`.

One note on your reachability analysis, which matches what I found in the api-ts PR: this reason is Argentina-only today, and not just because of the tax-ID rule. Brazil's only geo rail is `PIX_BR`, which is itself a QR-pool method that `enableQrPoolRails` promotes back to ENABLED, so a BR first-party failure never surfaces as a block at all. That gap is filed separately as TASK-22158 — which is also why the `es-419` catalogue entry here is the one that matters.

PR #2955, opened 2026-09-03
fix(native): OTA beta switch off PostHog, Peanut-only receive, claim settles on CLAIMED

innolope-dev:
## Why

Three fixes found while investigating native "trouble reaching Peanut" reports.

1. **OTA beta switch never appeared.** The five-tap beta switch was gated on a `beta-ota-channel` PostHog flag that was never created, so `isFeatureEnabled` returned false on every prod device — the switch stayed hidden and no device could self-assign to the Capgo `staging` channel.
2. **Receive screen offered non-Peanut rails.** The claim/receive screen showed bank / mercadopago / pix / exchange-wallet options beside the Peanut button.
3. **Claim screen stuck on "Processing".** A user got the "claimed" push notification while the screen kept spinning: the backend marks the SendLink `CLAIMED` and sends the notification before the on-chain `claim.txHash` projects, but the UI only settled once the poll observed that hash.

## What

**1. OTA beta switch without the cohort** (`875d231b`)
- Drop the `beta-ota-channel` flag gate. Five taps now reveals the switch on any native build.
- The off switch stays reachable, so a device on staging can always return to the store bundle.
- Remove the now-dead `notEnabled` toast and its locale strings.

**2. Peanut-only receive screen** (`ae8b6838`)
- Hide the alternate claim rails so the only receive option is Peanut, for logged-in and guest users.
- Rail code is kept intact behind a `SHOW_ALT_RAILS` flag (mirrors the file's `SHOW_INVITE_MODAL_FOR_DEVCONNECT` pattern) — one-line re-enable. Rail-specific tests are `describe.skip`'d, not deleted.
- Guest "Continue with Peanut" button and the devconnect event flow are untouched.

**3. Settle the claim screen on CLAIMED** (`de1aa6f0`)
- Treat `status === CLAIMED` as terminal success — the same point the backend marks the claim done and notifies — with or without the `txHash`.
- The poll reports the hash when present, `null` when it has yet to project; the success view tracks a `claimConfirmed` flag instead of gating on the hash. FAILED/CANCELLED and the give-up fallback are unchanged.

## Tests

- `useClaimSuccessPolling`: updated the CLAIMED-without-hash case to settle instead of poll on; other cadence/failure/ceiling tests unchanged.
- `SuccessClaimLinkView`: added a case for CLAIMED-without-hash rendering the success card.
- `BetaUpdatesCard` / `About.view`: dropped the cohort cases.
- Local gate: prettier clean, changed files typecheck-clean, affected suites pass.

## Notes / risks

- **Money surface.** #2 removes bank/mercadopago/exchange as claim destinations from this screen (kept reversible behind the flag). #3 changes when a claim reads as "success" — verified against `peanut-api-ts`: `SendLink.status` is set to `CLAIMED` and `processPostClaim` (which sends the push) run together, after the claim tx is handled; the failure path writes `FAILED` via `rollbackClaimOnError`.
- The `claimed` push is sent to the **sender**; the claimer's screen now settles on the same `CLAIMED` moment, so both track one signal.
- Follow-up for backend (not in this PR): `processPostClaim` awaits the receipt but does not flip `FAILED` on a revert, so a revert-after-CLAIMED would show success in both the notification and now the UI.

https://claude.ai/code/session_01Fe2ritzXbCgcxnWLLdJyfg

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fe2ritzXbCgcxnWLLdJyfg)_

### DIFF
--- a/src/components/Claim/Link/Onchain/Success.view.tsx
+++ b/src/components/Claim/Link/Onchain/Success.view.tsx
@@ -43,6 +43,9 @@ export const SuccessClaimLinkView = ({
     // The optimistic claim path lands here before the broadcast is known to
     // have succeeded, so a failure can only arrive through the poll below.
     const [claimFailure, setClaimFailure] = useState<{ code: string | null } | null>(null)
+    // CLAIMED can settle before the on-chain txHash has projected, so success is
+    // its own flag rather than "we have a hash".
+    const [claimConfirmed, setClaimConfirmed] = useState(false)
     const { user: authUser } = useUserStore()
     const { fetchUser } = useAuth()
     const router = useRouter()
@@ -64,8 +67,11 @@ export const SuccessClaimLinkView = ({
     }, [queryClient])
 
     const handleClaimConfirmed = useCallback(
-        (txHash: string) => {
-            setTransactionHash(txHash)
+        (txHash: string | null) => {
+            setClaimConfirmed(true)
+            // The hash may still be projecting when CLAIMED settles; set it once
+            // it is there so any hash-dependent path downstream still gets it.
+            if (txHash) setTransactionHash(txHash)
 
             // Force immediate refetch of balance and transactions,
             // bypassing staleTime; only currently mounted queries.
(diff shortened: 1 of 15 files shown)


innolope-dev:
Superseded by #2956 — same three commits, re-authored locally under my git identity with signed commits (cloud sessions hardcode the author to Claude, which this repo's CLAUDE.md forbids). Tree is byte-identical.

PR #2957, opened 2026-09-03
fix(claim): block sub-minimum cross-chain claims/transfers to 0x (Rhino)

innolope-dev:
## Problem

When claiming a send-link to an external \`0x\` wallet cross-chain (Rhino SDA), a claim **below the route minimum was silently lost**. Rhino accepts the on-chain SDA deposit but **never bridges a sub-minimum deposit** — the funds strand at the deposit address with **no auto-refund** (manual Rhino support recovery; see the 2026-07-15 incident referenced in \`cross-chain-fee.utils.ts\`).

Withdrawals already guard this (static per-chain floor + dynamic Rhino min). The **claim-to-external-wallet** and **pay-request cross-chain** paths had **no minimum check** at all.

## Fix (frontend)

- **Claim flow** — `Initial.view.tsx` and `Onchain/Confirm.view.tsx`: pre-flight guard blocks a sub-minimum cross-chain claim **before any SDA is provisioned**, using the same per-chain floor as withdraw (`getMinWithdrawUsdForChain` → \$0.50 default / \$5 Ethereum / \$10 Tron), with a new localized message `claim.errors.belowNetworkMinimum` (en / es-419 / es-AR / pt-BR).
- **`executeClaimXChain`**: threads `amountUsd` through and adds a dynamic backstop against Rhino's live `minDepositLimitUsd`, thrown **before signing** (no funds have moved at SDA-provision time).
- **`useCrossChainTransfer`**: sends `payAmountUsd` so the backend guard also covers **pay-request** cross-chain, not just claim.
- **`rhino-sda`**: surfaces the backend's specific error message + wire `code` instead of a `<label>: 400 {json}` blob; `friendlyError` maps `BELOW_MIN_BRIDGE_AMOUNT` so the confirm view shows the real reason verbatim.

## Pairs with

Backend PR (peanut-api-ts) that rejects sub-minimum SDA provisioning in `POST /rhino/sda-transfer` as the authoritative backstop.

## Testing notes

- Backend typechecks clean; prettier clean on all touched files.
- ⚠️ This local `peanut-ui` worktree's `node_modules` was broken (symlinked to a deleted sibling), so FE `tsc`/`jest` could not be run here. Verified by hand against real types and the existing `messages.test.ts` / `claim-campaign-isolation.test.ts` invariants — please let CI run `typecheck` + `test`.


innolope-dev:
Thanks for the review — addressed in the latest commit.

**MAJOR — guard blocks claiming to the Peanut balance, with dead-end copy.** Fixed: the pre-flight guard is now scoped to **external-wallet claims only** (`claimToExternalWallet`). A claim into the user's own Peanut balance is forced down the same `isXChain` branch but settles to Arbitrum USDC with no product minimum, so it's no longer blocked — and the "claim to your Peanut balance" advice is now accurate for the external-wallet cases that do hit the guard. Also stopped sending a client-reported amount: claim-xchain now forwards the **deposit identity** (chainId/depositIdx/version) so the backend reads the amount from chain (authoritative).

**MAJOR — no money-path regression test.** Added `useClaimLink.minGuard.test.ts`: sub-minimum blocks before `signWithdrawalMessage` and the `/claim` POST, boundary (equal) passes, priceless / no-route-minimum cases don't block, a backend refusal propagates before signing, and the deposit identity is forwarded to `provisionSdaTransfer`. 11 claim tests pass locally.

Pairs with peanutprotocol/peanut-api-ts#1511, which makes the backend guard server-authoritative and returns the specific message + `BELOW_MIN_BRIDGE_AMOUNT` code that this PR surfaces via `postRhino` + `friendlyError`.

Note: I could not run the full FE suite locally (this worktree's node_modules was broken) — CI will run typecheck + tests.

PR #2963, opened 2026-09-03
fix(ota): stop losing bundles to a background apply the OS freezes

innolope-dev:
Stacked on #2960 (it owns `canRestartInPlace`). Base it on `dev` once that lands.

## The failure

Capgo consumes a `next()`-staged bundle only from `installNext()`, which runs from `appMovedToBackground()`. The reload therefore lands in a process the OS is about to freeze. Freezing stops threads but not the clock, so on resume every overdue `setTimeout` fires at once — webpack's 120 s `chunkLoadTimeout` included — and every in-flight `import()` rejects with a `ChunkLoadError` for a local file that was never missing. `notifyAppReady` sat behind two of those imports plus React hydration, so it was never called and the plugin rolled the bundle back.

This is **not** the plugin-thread deadlock. `installNext()` has always wrapped `set()`/`_reload()` in `startNewThread`, in 8.45.9 and 8.51.14 alike — I diffed both.

`PEANUT-UI-SVT` is the whole chain in one trace (Honor PLG110, Android 16):

```
13:46:04.029  navigation — new bundle's document begins loading
13:46:04.571  fetch x4 in flight, boot proceeding normally
              ... 23m 46s, zero breadcrumbs ...
14:09:50.784  failed to init app listeners: ChunkLoadError (timeout: ...)
14:09:50.805  [capgo] ota init failed: ChunkLoadError: Loading chunk 8789 failed
14:09:50.823  [CapgoUpdater] notifyAppReady was not called, roll back
```

Chunk 8789 is the `@/utils/capgo-updater` chunk. `PEANUT-UI-SVE` (Samsung SM-A505G, Android 11) is the milder shape: `notifyAppReady` **was** called, 6 ms after the deadline.

## What this changes

- **`notifyAppReady` moves to a raw inline script** in the root layout, calling the bridge stub Capacitor injects at document start (`JSExport.java` / `JSExport.swift`). No chunk, no hydration, no React. `initCapgoUpdater` still calls it too — it is idempotent, and it covers any binary whose bridge does not expose the stub.
- **A boot-incomplete counter replaces Capgo's rollback net**, which the early call gives up. It is the better net: Capgo could not tell a bundle whose JS is broken from one the OS froze and rolled back on both. A frozen boot resumes and clears the counter; a broken one never does. Three consecutive dead launches fall back to builtin.
- **A bundle staged by an earlier launch is applied in the foreground**, while the splash still covers the reload, instead of being left to the background apply. Once per bundle, and only where `canRestartInPlace()`. Outside that window nothing changes — the background apply is now survivable.
- **`import()` retries once on a chunk-load failure.** The second attempt runs with the process scheduled again.

## Why the counts looked like n=1

Capgo's logs reach Sentry only through `bridge.eval("console.error(...)")` into the WebView (`Logger.java:252`), and a background apply destroys the page that would receive them — the rollback line is emitted ~20 ms before `performReset` tears the next one down. Whatever survived, the `[capgo]` / `[CapgoUpdater]` prefix filter dropped. Over 90 days the entire population reported as **one event, one user**; `ChunkLoadError` and `[capgo] ota init failed` reported **zero**.

So: the two rollback strings are now in `CAPGO_ACTIONABLE`, and `getFailedUpdate()` is reported at launch under the `[capgo-apply]` prefix the filter does not match. That record lives in the plugin's SharedPreferences and outlives the reload, the rollback and a process kill — it is the only way to get a real fleet-wide rate.

## Related

`appReadyTimeout` 15 s -> 120 s lives in #2968 (which replaces #2961), since it is binary-only and that PR already touches `capacitor.config.ts`. Android silently floors it at 30 s for a pending bundle, so 15 s was inert there and binding on iOS.

## Testing

- `pnpm tsc --noEmit` clean.
- Full jest suite: 439 suites / 5631 passing.
- New coverage: the inline script executed as it ships (bridge stub, no-op off native, the counter and its builtin fallback, storage-less WebView), the launch apply (applies behind the splash, once per bundle, stands down on deadlocking binaries and when the staged bundle is already running, still surfaces a rejected apply for a manual restart), `importWithChunkRetry`, the `getFailedUpdate` report, and the two new Sentry carve-outs.

Not verified on-device — the inline-script path needs a native build to confirm the bridge stub is reachable that early on both platforms. It fails safe: if the stub is missing the script returns and the existing React path calls `notifyAppReady` exactly as today.

### DIFF
--- a/sentry.utils.test.ts
+++ b/sentry.utils.test.ts
@@ -101,6 +101,23 @@ describe('shouldIgnoreError — Capgo updater noise', () => {
         expect(shouldIgnoreError(eventWith({ message }))).toBe(false)
     })
 
+    /*
+     * An update that silently un-happens leaves no other trace: Capgo's own logs
+     * only reach Sentry through an eval into a WebView the rollback is about to
+     * tear down, so suppressing these made the whole population read as one
+     * event in 90 days (PEANUT-UI-SVT).
+     */
+    it('keeps a bundle the plugin rolled back for want of notifyAppReady', () => {
+        expect(
+            shouldIgnoreError(
+                eventWith({ message: '[CapgoUpdater] 🔴 notifyAppReady was not called, roll back current bundle: E3J' })
+            )
+        ).toBe(false)
+        expect(shouldIgnoreError(eventWith({ message: '[CapgoUpdater] 🔴 Update to bundle: 1.0.56 Failed!' }))).toBe(
+            false
+        )
+    })
+
     it('keeps a checksum mismatch — the bundle arrived corrupt, not merely late', () => {
         expect(shouldIgnoreError(eventWith({ message: '[CapgoUpdater] 🔴 Checksum mismatch' }))).toBe(false)
     })
--- a/sentry.utils.ts
+++ b/sentry.utils.ts
@@ -104,10 +104,19 @@ const IGNORED_ERRORS = {
  * (~95/day on native). The user never sees them: the updater just retries on
  * the next launch. Suppress those, but keep the failures that mean OTA is
  * genuinely broken rather than merely flaky — a bundle that semver-sorts below
- * the installed binary, or one that arrived corrupt.
+ * the installed binary, one that arrived corrupt, or one the plugin rolled back
+ * because notifyAppReady never landed. That last class is the reason this list
+ * is not just the two it started with: an update that silently un-happens
+ * leaves no other trace, and suppressing it made the whole population read as
+ * one event in 90 days (PEANUT-UI-SVT).
  */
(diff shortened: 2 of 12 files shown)


innolope-dev:
CI does not auto-run on a stacked base, so Tests was dispatched manually against this branch: [run 33796398832](https://github.com/peanutprotocol/peanut-ui/actions/runs/33796398832) — eslint, ds-lint, typecheck, format, native-export, unit and ci-success all green. `human-authors` is skipped on a `workflow_dispatch` event; the single commit is authored by innolove-dev. Chip has no dispatch trigger, so this PR gets no Chip pass until it is rebased onto `dev`.

PR #2969, opened 2026-09-03
feat(spend): one-tap mixed spend for the sign-only engine, dark behind its own flag

innolope-dev:
> **Stacked on #2959** (`feat/session-key-spend-build-gate`) — it needs the PostHog-flag gate from there. Retarget to `dev` once #2959 merges. Stacked PRs get no automatic tests/format/chip run; `tests.yml` is triggered by hand on this branch.

## Why

QR pay, Manteca withdraw and the card lock/cancel modals sign a mixed spend through `useSignSpendBundle` and hand the UserOp to the backend to broadcast. That engine still costs two passkey sheets (Rain admin EIP-712, then the UserOp). #2959 covers only the broadcasting engine (`useSpendBundle`). Item 3 of `mono/engineering/passkey-prompt-pairs-2026-09-03.md`.

## What

- `signMixedEphemeralSpend` (`src/hooks/wallet/mixedEphemeralSign.ts`): sign-only twin of `tryMixedEphemeralSpend`. Same single enable-signature tap inside `createEphemeralSpendSession`; the ephemeral key then signs the Rain admin EIP-712 and the UserOp silently, self-uninstall as the last call, and returns the unbroadcast op in the existing `SignedUserOpData` shape. `signUserOperation` from the ZeroDev SDK, exactly what `useSignUserOp` uses.
- `useSignSpendBundle` mixed path tries it first when `sessionKeySignEnabled()`; on any failure it falls through to the two-tap path with the **same** prep (nothing was broadcast, nothing at stake) and reports `session_key_spend_fallback` with `flow: 'sign-only'`.
- **Own flag `session_key_spend_sign`** (build gate shared with #2959). Why not reuse `session_key_spend`: here a permission that does not validate on-chain surfaces later, as the backend's broadcast reverting, with no client-side retry. So this flag stays off until #2959 has proven the ERC-1271 ordering on production contracts, then it is one click.
- `createEphemeralSpendSession` takes an optional `ttlSeconds`; this path uses 10 min (default stays 180 s). The backend broadcasts after Manteca settles inside the same request, with a receipt re-poll on timeout, so 3 min was tight. The op stays single-use through its nonce and the batch still ends in the self-uninstall.
- `/dev/session-key-spend` shows the second flag.
- `buildWithdrawCall` exported from `mixedEphemeralSpend.ts` (shared with the twin).

## Rollout

1. #2959 merges, flag `session_key_spend` proves one real mixed send on prod (one sheet, no fallback).
2. Then create `session_key_spend_sign`, internal cohort; one real QR pay on a mixed balance; expect one sheet and `session_key_spend_attempted{flow:'sign-only'}` with no fallback and a successful backend broadcast.
3. Widen.

## Verification

- New tests: `mixedEphemeralSign` (artifact shape, TTL, silent admin sig, uninstall last, signed through the session client and never sent, dispose on every exit; preflight and signing failures as `ok:false`), `sessionKeySignEnabled` gate matrix, and three `useSignSpendBundle` cases (flag off unchanged, flag on returns the ephemeral artifact with no passkey admin signature, fallback reuses the same prep and reports the reason).
- `pnpm typecheck` clean; ephemeralSpendKey, useGrantSessionKey, session-key-spend.consts, useSignSpendBundle suites green; prettier clean.

### DIFF
--- a/src/app/(mobile-ui)/dev/session-key-spend/page.tsx
+++ b/src/app/(mobile-ui)/dev/session-key-spend/page.tsx
@@ -16,6 +16,7 @@ import { useState } from 'react'
 import { Button } from '@/components/0_Bruddle/Button'
 import {
     SESSION_KEY_SPEND_BUILD_ENABLED,
+    SESSION_KEY_SIGN_FLAG,
     SESSION_KEY_SPEND_FLAG,
     sessionKeySpendDeviceOptIn,
     setSessionKeySpendDeviceOptIn,
@@ -28,6 +29,7 @@ export default function SessionKeySpendPage() {
     const isFlagEnabled = useFeatureFlags()
     const deviceOn = sessionKeySpendDeviceOptIn()
     const flagOn = isFlagEnabled(SESSION_KEY_SPEND_FLAG)
+    const signFlagOn = isFlagEnabled(SESSION_KEY_SIGN_FLAG)
     const effective = SESSION_KEY_SPEND_BUILD_ENABLED && (deviceOn || flagOn)
 
     const toggle = () => {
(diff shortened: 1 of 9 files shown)


innolope-dev:
Folded into #2959: with no rollout gate the sign-only engine port no longer needs anything from a flag branch, so both engines ship in one PR. Commits 0bc310f08 and e0d0ca73a were cherry-picked there and the flag stripped.

PR #2973, opened 2026-09-04
fix(card,kyc): card surfaces where the card is unavailable + device-feedback copy round

innolope-dev:
Device-test feedback round from 2026-09-04, three commits.

## 1. Card surfaces shown to users the card is not available to

`hasCardAccess` is the waitlist's **inner gate** — `cardAccessGrantedAt` or a `skip:card-queue` badge — not "holds a card". `releaseUsersFromWaitlist` stamps that grant with **no geo or eligibility check**, so reading it as a holder test offered a Peanut card to every released user resident in a Rain-prohibited country (UA, IN, TR, VE, VN, IL, IQ, NP, NI): the profile row showed, and the bottom nav's card tab led to /card's `geo-blocked` dead end — while Unlock payments correctly greyed the same card out on the same account.

- **`useCardSurfaceAccess`** — one predicate for every card surface. The holder test is now an issued, non-canceled card (`findActiveCard`, the same predicate `/card`'s own gate uses); everyone else needs an unrestricted residence and a server eligibility that is not `false`.
- The profile menu row moves onto it (its inline guard was the bug).
- The nav's **middle slot falls back to the exchange-rates page** when the card is unavailable. `TabId 'card'` becomes `'middle'`: one pill box, two destinations, identical geometry.
- The card variant of that tab now routes `hasCardAccess ? /card : /shhhhh`, matching the profile row. It used to link `/card` unconditionally, which `notFound()`s anyone without a `flowEarlyAccess` stamp.

The hook is mounted app-wide by BottomNav, which is the pattern `utils/support-cache.ts` warns off. Both of its reasons are addressed in a comment on the hook: BottomNav renders only for logged-in users, neither query polls for the users that rule protects, and Home mounts both already so react-query dedupes.

## 2. Copy and affordances

**Unlock drawer** — the intro no longer states a duration ("— about a minute", "Takes about a minute"): the prep checklist gives the real answer under HOW LONG and the two disagreed. "just" dropped with it. The "one more document may be asked for" note moves **above** HOW LONG, against the requirements it qualifies. The privacy footnote's ⓘ is now a link to `/help/verification`, whose "your data and privacy" section is that claim in full — an icon that looked tappable and did nothing was the report.

**Exchange rate & fees** — "Try it!" → "Send now", "You Send" → "You send", "Recipient Gets" → "Recipient gets". The arrival line loses its ⓘ (nothing behind it) and reads "Usually arrives in minutes/hours".

**Withdraw method picker** (where that CTA lands) — "Choose withdrawing method" → "Choose sending method", the crypto row's "Withdraw to a wallet address" → "Send to a wallet address", and the confirm CTA says "Send" on every entry rather than only from the send flow. Page titles keep their entry point's framing.

All four catalogs updated; es-AR keeps its voseo deltas.

## 3. Profile menu shell

The nav bar and QR button floated over the profile menu — a full-screen list of destinations, so the bar duplicated navigation on the one screen that is navigation. The existing hide is a drawer-scoped hold (`hideBottomNav`, built for the home Add/Send sheet) and does not apply to a route, so the shell's `nav` slot is gated on the route instead (exact match — `/profile/*` sub-pages keep the bar). With no bar to clear, the 6rem bottom reservation was dead space under Log out: 1rem for that route, and the button's own `pb-10` → `pb-2`.

## Tests

New: `useCardSurfaceAccess` (7 cases, incl. the waitlist-released-in-a-restricted-country regression), BottomNav middle slot (5), KycPrepChecklist ordering (2). Full suite green locally — 450 suites, 5714 tests.

## Not done, needs a ruling

The Notification component's vertical padding and gaps were asked about. They are **DS-defined and were ruled a day before this branch** — `mono/design/design.md` line 472, "compact inline (ruled 2026-09-03, kush): Body/S 14 text, title included, 16px icon, tighter padding". Loosening them reverses that ruling, so it needs kush rather than a PR.


innolope-dev:
## Round 2 response

Pushed as `980024b86`. Three of the findings contradicted decisions already made on this branch, so they went back to Slava before anything was touched — his rulings are recorded below.

### Fixed

**BLOCKING · first-save `showFullName`** — reverted entirely. Ruled: *default private on first save*. `ProfileEdit.view.tsx` and `ShowNameToggle.tsx` are back at their `dev` state and `shouldActivateShowFullName` is gone with its test. The finding was correct: the visibility toggle only renders once a saved `fullName` exists, so writing `showFullName=true` on the save that creates the name published it on public profiles and counterparty transaction surfaces with no visible opt-out.

**MAJOR · existing applications lose the card slot** — fixed. `useCardSurfaceAccess` now treats an in-flight application as a card relationship (`overview.status.hasApplication`), because `computeCardState` renders `rejected` / `requires-support` / `requires-info` / `pending` / `manual-review` **above** its geo and eligibility gates. The destination moved into the hook as `cardHref` so the profile row and the nav tab cannot drift apart, and an applicant or holder now routes to `/card` rather than the `/shhhhh` door. Two new cases cover it.

**MAJOR · CTA names the opposite flow** — fixed. The page derives the label from the computed destination instead of hard-coding it. `getExchangeRateWidgetRedirectRoute` sends a zero balance, and any pair that is not USD → local, to `/add-money`, so the label is now `exchangeRate.addMoneyCta` on that branch and `tryIt` on the withdraw branch. The pair comes from the same nuqs query state the widget writes, so no new prop was needed. Two new tests pin both branches.

**MAJOR · "held in USD" vs the Lexicon's USDC** — the denomination is not mine to pick, so this went to Slava. Ruled: *neutral wording, no denomination*. The note now names no asset at all. The same rewrite also narrows the withdrawal clause, which addresses the `claude-opus` finding — it is conditional on a local rail existing rather than promising local-currency withdrawal to the 14 widget currencies that have no fiat rail.

**MINOR · es-AR voseo** — fixed. Added an `exchangeRate.widget.recipientGets` override (`"Recibís"`); inheriting es-419's `"Recibes"` mixed tú with the `"Vos enviás"` override in the same widget.

### Accepted, not fixed

**MINOR · Notification spacing reversal.** This is a deliberate product decision, not drift. Ruled by Slava after device review: *keep the wider spacing*. Variant A changed four things in one commit — no border, one type step down, a 16px icon, and halved padding and gaps — and on device the first three read well while the fourth did not; with the border gone, 8px padding left the tint block against its own text. Only the spacing half is reversed; border, type step and icon size stand. The `indent` is recomputed from the 16px icon (`pl-6`) rather than restored to the 20px-icon `pl-7`.

On "land the approved design-system update in the same change": the DS record lives in a different repo, so it is peanutprotocol/mono#135 rather than a file in this PR. Tracked for kush's sign-off as TASK-22269 — if he wants the compact tokens back, that reverts both together.

### Not fixed, not blocking

`ds-shots` fails in Next 16.3.1's webpack (`WasmHash._updateWithBuffer`) after Playwright `install-deps` fails three times — infra, no source anchor, and it reproduces on earlier heads. It is not in `ci-success`'s `needs:` list, so it does not gate merge.

innolope-dev:
## Round 3 response — both remaining findings accepted, no code change

Chip reports no blocking findings at `980024b86` and closes the eight prior ones. The two MINORs that remain are both deliberate decisions, tracked rather than reworked.

**MINOR · Notification spacing** — unchanged, ruled by Slava after device review. Reasoning and the exact token table are in the component's block comment and in TASK-22269. Chip asks to "land the design-system ruling together with this change"; the DS record lives in a different repo, so that is peanutprotocol/mono#135, awaiting kush. If he wants the compact tokens back, that reverts both together.

**MINOR · "Supported currencies" header** (`claude-opus`) — accepted as-is; filed as TASK-22270. The finding is correct and well-evidenced: that section lists 13 currencies Peanut does not settle in (ALL, CZK, DKK, HUF, ISK, MDL, MKD, NOK, PLN, RON, RSD, SEK, CHF), Peanut pays out only in USD, EUR, GBP, MXN, ARS and BRL, and the header renders on the public landing page as well as `/profile/exchange-rate`. The rename was a product request to stop the two sections reading as overlapping sets; the wording call stays with Slava, and the task records the gap with the `product/countries.md` citations so it is revisited deliberately.

**`ds-shots`** remains red for the same reason as every earlier head: a `WasmHash._updateWithBuffer` crash inside Next 16.3.1's webpack, before capture, with no source anchor. It is not in `ci-success`'s `needs:` list, so it does not gate merge. Every functional check on this head is green.

PR #2974, opened 2026-09-04
fix(status): show red when the page cannot confirm we are up

innolope-dev:
## Why

https://peanut.me/en/status read **all systems operational** through yesterday's production outage. The frontend half of that.

The page renders on Vercel and its feed comes from Render, so the fetch it makes is the one measurement it can still take when the backend is the thing that is broken. Three ways that measurement was being thrown away:

**A failed fetch rendered as amber.** "Live status could not be loaded. Please try again shortly." — a hiccup notice, on the page whose entire job is to report that the backend is unreachable.

**Next's Data Cache served a green board over a dead origin.** `fetch(..., { next: { revalidate } })` stores the response, and when a revalidating fetch *fails* Next keeps serving the last good body rather than surfacing the error. Verified locally against a feed refusing every connection: the dev server logged `TypeError: fetch failed / ECONNREFUSED` on request after request while the page went on rendering the last green board, and the `catch` never ran — because `fetch` never rejected. Any amount of red further down the stack is defeated by that alone.

**Nothing checked how old the answer was.** A CDN serving a cached body over a dead origin looks identical to a healthy one.

## What changes

- **Red when the feed is unreachable.** The fallback now renders the same banner, in the same red, as an outage the feed reported itself: "Status unavailable — We can't reach Peanut's systems right now… This usually means an outage is in progress."
- **A summary older than 10 minutes is refused** and shows that same red. This is what catches the Data Cache: a body Next keeps re-serving ages, and the page stops believing it. The budget is derived, not picked — a legitimately healthy body can be ~7 minutes old (`max-age=60, stale-while-revalidate=300` at the CDN, plus up to 60s in Next's cache), so ten leaves headroom against a false outage. `parseStatusSummary` now requires `generatedAt`, since the page makes a decision with it.
- **An outage gets a headline.** Previously the only sign of one was the colour of a 2px dot some rows down. The banner appears **only when something is wrong**, so a healthy page still opens straight at App & Account as designed. `unknown` is styled as an outage rather than a neutral third thing: not knowing whether Peanut is up is a bad state to be in, and softening it into grey is how this page read green on 2026-09-03.
- **A bar with no checks says so** — "No data — monitoring was unavailable" rather than "0/0" — which is what the companion API PR starts emitting for hours the collector could not write.

## What I tried first, and why it is not here

`cache: 'no-store'` fixes the Data Cache problem directly. It also turns `/[locale]/status` from `●` SSG into `ƒ`, alone among the marketing pages — confirmed in the build output. Every visitor during an incident would then reach `/status/summary` and its three queries directly, on the API least able to take it. That trades a cosmetic failure for a load amplifier, so the cache stays and the freshness gate does the work.

The cost is stated plainly: an unreachable backend reads green for up to 10 minutes rather than instantly. Happy to flip it if you'd rather have instant red and take the load — it is one line.

## Verification

- Three states driven in a browser against a local feed: **healthy** (unchanged, opens at App & Account, no banner), **gap** (red bars mid-window, "No data — monitoring was unavailable" on hover, "Service outage" headline), **unreachable** (red "Status unavailable").
- 33 tests in `status/`, 8 of them new: the banner appears once anything is not operational, `unknown` renders as an outage, a `checks: 0` bar reports unmonitored rather than 0/0, and the freshness gate accepts/refuses/handles an unparseable timestamp.
- Full jest suite: **5,706 passed**, 448 suites. `next build --webpack` clean, route confirmed back to `●` for all four locales. `tsc` and `prettier --check` clean.
- Copy added to all four locales; `generate:marketing-messages` produces no diff.

## Merge order — this one goes first

peanutprotocol/peanut-api-ts#1521 starts emitting `checks: 0` for hours nothing was recorded in. Without the `bucketDetail` change here those bars render as red with a **"0/0"** tooltip, so this PR should merge and deploy ahead of it. Nothing here depends on that one — the red-when-unreachable and freshness work stands on its own against today's feed.

## Companion

peanutprotocol/peanut-api-ts#1521 — the collector half: `backfillWindow` deleted, a missing hour read as `down`, 30-day bucket retention, and `api.peanut.me/healthz` folded into the `app` probe.


innolope-dev:
Fixed in 934867f28 — you were right, and it's the same bug one layer up from where I was looking.

Under `revalidate = 60` the page sat in the Full Route Cache, which answers the first request after expiry from cached HTML and regenerates behind it. A locale that last rendered green, then went an hour without traffic, hands that hour-old green to the next visitor while `isFresh` runs *after* they've read it — and if they're the only visitor, that's all they ever see. A data-age gate is worth nothing when nothing evaluates it before the response.

Took your first suggestion: route cache out via `force-dynamic`, feed Data Cache retained. That's the split I should have had from the start — the verdict recomputed per request, the origin still shielded.

I'd previously rejected opting out of the route cache on load grounds, and I'd asserted that in a code comment without measuring it. So this time I measured, against `next start` and a counting origin:

- **25 consecutive page loads → 1 request to the feed.** `force-dynamic` does not disable the fetch cache, so the once-a-minute-per-region shield holds. My earlier load objection was simply wrong.
- **A feed returning 200 with an all-green body stamped 30 minutes ago → red banner on the first request.** The age gate runs on the response path.

Residual staleness is now bounded by the Data Cache's 60s rather than by route-cache age, and by the 10-minute budget once the origin stops answering entirely. Happy to tighten that budget if you think 10 minutes is still long — the constraint is the feed's own `stale-while-revalidate=300`, which puts a legitimately healthy body at up to ~7 minutes old, so there isn't much room below 10 without risking a false outage. Lowering the API's `swr` would buy some, at the cost of more origin load during exactly the incident we're shielding against, which seemed like the wrong trade.

innolope-dev:
Switched to `revalidate = 0` in f664fea1b — but the mechanism in the finding isn't right, and it's worth recording why, because the conclusion still holds for a different reason.

`force-dynamic` is not unconditionally `fetchCache = 'force-no-store'` in Next 16. From `next/dist/server/lib/patch-fetch.js`:

```js
// If no explicit fetch cache mode is set, but dynamic = `force-dynamic` is set,
// we shouldn't consider caching the fetch. […]
const noFetchConfigAndForceDynamic =
    !pageFetchCacheMode && !currentFetchCacheConfig
    && !currentFetchRevalidate && workStore.forceDynamic
```

The opt-out is gated on `!currentFetchRevalidate`, and this fetch passes `next: { revalidate: 60 }`. Measured both settings against `next start` and an origin that counts requests: **25 consecutive page loads produce 1 request to the feed under `force-dynamic`, and 1 under `revalidate = 0`.** So the load amplifier described here wasn't present.

Taking the suggestion anyway, because the finding identifies a real fragility even though the stated cause is wrong: under `force-dynamic` the shield survives *only* as a consequence of that explicit `revalidate` on the fetch. Anyone who later removes it — reasonably, thinking the route is dynamic so the fetch config is redundant — silently turns every visitor into an API request, during exactly the incident this page exists for. `revalidate = 0` renders per request without making the cache contingent on a line elsewhere in the file, so it's the better thing to say regardless. Noted in a comment so the next reader doesn't 'simplify' it back.

Re-verified on the new setting: 25 loads → 1 origin request, and a feed answering 200 with an all-green body stamped 30 minutes ago still renders the red banner on the first request.

Flagging the measurement explicitly since I'd rather be corrected than have a wrong number stand: fixture origin on localhost, `next start` against the webpack production build, counter read via a side channel on the same server. If that setup misses something about how this behaves on Vercel's runtime, I'd want to know.

innolope-dev:
Fixed in 04db40a35. You're right and it's a hole in the gate itself, which makes it worse than a normal edge case: age was the *only* thing keeping a cached green board from outliving the system it describes, and I capped it in one direction only. A feed host an hour ahead that then goes unreachable buys a dead backend seventy minutes of green — the exact failure this PR exists to close, reachable through the mechanism I added to close it.

Bounded both ways now, with a minute of tolerance for a host whose clock runs slightly ahead of ours. Regression cases added next to the stale and unparseable ones: an hour ahead is refused, just-inside-skew is accepted.

Also added a merge-order note to both descriptions — this PR should land before peanutprotocol/peanut-api-ts#1521, since that one starts emitting `checks: 0` buckets that today's UI would render with a "0/0" tooltip. Nothing here depends on that PR, so the order is free to take.

PR #2999, opened 2026-09-05
fix(qr-scanner): stop teaching native users to change a browser permission (design-system)

innolope-dev:
Companion to #2997 (based on `dev`) — same change, applied to `feat/design-system`.

## What

The QR scanner's camera-permission modal shows per-browser screenshots — *tap the lock icon in the address bar*, *open Website Settings*, *set Camera to "Allow"*. In the Capacitor app there is no address bar and no site permission: the camera grant is an OS permission attached to the app, so every step points at a surface the user cannot reach. Native gets text steps naming the OS path plus a button that opens the app's own settings page; web and PWA are untouched.

Full rationale, the OTA-vs-binary split, the dependency review and the verification log are in **#2997** — please read that one first. This description covers only what differs here.

## What differs on this branch

**The paste CTA is kept.** `dev` deleted it in `60c56ef11` (2026-09-03) as a second secondary that broke the modal recipe; this branch still has it, along with the `LinkButton` footer. The cherry-pick conflicted exactly there, and I resolved it in this branch's favour rather than dragging `dev`'s ruling backwards: `onPaste` stays in the props and still renders its CTA. So on native here the modal reads Open Settings → Click to paste → Dismiss, whereas on `dev` it reads Open Settings → Dismiss.

That is the only intentional divergence between the two PRs. Everything else — the util, the copy, the four catalogs, both test files, and the `cap update` output — is identical.

## Ordering

These two are independent and can land in either order. Whichever merges second will conflict on the same CTA array when `dev` and `feat/design-system` next reconcile; the resolution is whatever the paste ruling is by then, not anything about this change.

## Verification (re-run on this base)

`tsc --noEmit` clean, `eslint` clean on all four changed source files, `prettier --check` clean, 250 tests green across `i18n` + `QRScanner` + `native-settings` (including the 6 new modal tests and 9 for the util). `npx cap update` on this branch produced byte-identical native registration to the cherry-picked hunks.

Not verified on a device — same caveat as #2997: the `app-settings:` fallback is read off Capacitor's source, and this machine has no Xcode.

## One thing #2997 needs and this does not

#2997 carries a second commit registering `capacitor-native-settings` in `NATIVE_DEPENDENCIES` (`scripts/native-fingerprint.mjs`) — the guard that stops a bundle calling the plugin being OTA'd onto a binary built before it. `feat/design-system` predates that script entirely, so there is nothing to register here.

## Screenshots (head `6f3fee3`, 375×667)

Captured from this branch running locally; native variants forced by stubbing the Capacitor bridge (`isNativePlatform` / `getPlatform` / `isPluginAvailable`), web variants by UA. Images live on the [`shots/pr-2999`](https://github.com/peanutprotocol/peanut-ui/tree/shots/pr-2999) branch.

### Native (new)

| iOS (plugin present) | Android (plugin present) | Old binary, no plugin (fallback) |
| --- | --- | --- |
| iOS native: text steps + Open Settings | Android native: 3 text steps + Open Settings | Android native without NativeSettings plugin: same steps, Try again CTA |
| 2 steps, `Open Settings` deep-links via plugin (or `app-settings:` fallback) | 3 steps (Permissions level exists on Android) | Same text steps; CTA falls back to `Try again` because no deep link is possible |

This branch keeps the `Click to paste` secondary CTA (see "What differs on this branch" above), so native reads Open Settings → Click to paste → Dismiss.

### Web / PWA (unchanged)

| iOS Safari | Android Chrome |
| --- | --- |
| Web iOS Safari: browser screenshot carousel, unchanged | Web Android Chrome: browser screenshot carousel, unchanged |

Browser-instruction carousel still renders exactly as before on web.

### DIFF
--- a/android/app/capacitor.build.gradle
+++ b/android/app/capacitor.build.gradle
@@ -24,6 +24,7 @@ dependencies {
     implementation project(':capgo-capacitor-passkey')
     implementation project(':capgo-capacitor-updater')
     implementation project(':onesignal-capacitor-plugin')
+    implementation project(':capacitor-native-settings')
 
 }
 apply from: "../../node_modules/.pnpm/@sumsub+cordova-idensic-mobile-sdk-plugin@1.42.0/node_modules/@sumsub/cordova-idensic-mobile-sdk-plugin/src/android/build-extras.gradle"
--- a/android/capacitor.settings.gradle
+++ b/android/capacitor.settings.gradle
@@ -46,3 +46,6 @@ project(':capgo-capacitor-updater').projectDir = new File('../node_modules/.pnpm
 
 include ':onesignal-capacitor-plugin'
 project(':onesignal-capacitor-plugin').projectDir = new File('../node_modules/@onesignal/capacitor-plugin/android')
+
+include ':capacitor-native-settings'
+project(':capacitor-native-settings').projectDir = new File('../node_modules/.pnpm/capacitor-native-settings@8.2.0_@capacitor+core@8.2.0/node_modules/capacitor-native-settings/android')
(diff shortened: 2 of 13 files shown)


kushagrasarathe:
Closing as superseded: the twin #2997 merged to `dev` on Sep 5 (plus the OTA fingerprint registration in 0e63d158b), and #3064 converted the modal to `CameraPermissionDrawer`, which carries all the native behavior with identical copy in all 4 locales (`native-settings.ts` is byte-identical). The one delta here — keeping the paste CTA — was deliberately removed on `dev` (two secondaries broke the drawer recipe), so dev's ruling stands. Screenshots of every modal variant are in the PR body above. TASK-21699 is cancelled with the same note.

PR #3001, opened 2026-09-05
design(surfaces): backup, identity, card and add-money polish

innolope-dev:
Twelve surface fixes from a design pass over the app's modals, drawers and notification-bearing pages.

## What changed

**Backup** — platform steps and the change-phone sequence use `NumberedList` (pink `action-primary` markers). The cloud icon becomes a wallet on a brand-pink bubble. The three FAQ sheets take uppercase mini-headers; change-phone's three stacked `Notification`s become prose sections; export-keys' bullets are pink via `marker:text-action-primary`.

**Identity → additional** and **Language** centre their content through `PageStack.Center`, the former inside a `Card` to match QR Pay.

**Add money** carries a green plus mark on all five of its surfaces; **Card limit** and **Card PIN** a pink card mark.

**Reject labels** render on the `error` tone rather than `info` — a rejection is a failure, not information.

**Residence** names the Peanut card on the congrats screen. See the policy note below.

## Bug fixed along the way

`/card-recovery` was throwing a client-side exception — a blank white screen on a page whose entire audience is people trying to recover funds.

`amountCents` and `dustWei` go straight into `BigInt()` during render, which throws on a missing or non-integer value. The fetch path was guarded; the parse path was not, so a partial `200` took the whole route down instead of landing in the error banner the page already had. Now validated into that banner.

`demo-api` also had no handler for `/rain/cards/recover-funds/preview`, so the route fell through to the `{}` default — which is how the harness surfaced this.

## Policy note — please read

The congrats screen deliberately did **not** name the card: its closed beta had to stay unnamed in onboarding (product direction), and any card mention had to state every access gate (compliance). No sentence satisfies both.

Slava's call on 2026-09-05 was to name it with the gates stated. The clause carries both remaining gates — the ID check and the waitlist — so the compliance half holds. But "waitlist" names the closed beta by implication, which is what the other half was protecting. Whoever owns that policy should see this.

It reads off `residenceAvailability()` rather than assuming that reaching the screen implies an unrestricted card, so a routing change upstream cannot turn it into a false claim. The date and the decision are in the code comment so this doesn't get reverted as drift.

## Refactors, and why

The `ds-lint` `classNameSitesInPages` ratchet caught utility strings landing in page files. Rather than rebaseline, pages now compose views: the FAQ sheets became `BackupFaqModals`, the screen marks a `ScreenMark` recipe. `MiniHeader` replaces two duplicate local `MINI_HEADER` constants.

Side benefit: those FAQ sheets are now a named component the shot harness can find — they were invisible to it before, being anonymous `ActionModal`s inside a page.

`IconBubble` gains a `brand` colour mapped to the existing `--color-background-brand` token. No new figma variable was invented — the semantic set claims 1:1 parity with figma and that stays true.

## Verification

Built and shot at 375×900 under fixture mode: Backup and all three sheets, card recovery, Language, Identity/additional, and both add-money surfaces.

Two items have no reachable fixture, so they are covered by test instead:

- **Reject labels** — new `RejectLabelsList.test.tsx` asserts reason rows and the no-labels fallback both carry `bg-background-badge-error`.
- **Congrats copy** — two existing tests asserted the card was *absent*. Rewritten to require the card **and both gates**, so silently dropping "waitlist" fails CI.

**Not visually verified:** `/card/limit` and `/card/pin` sit behind `ActiveCardGate` and no fixture has an active card, so the harness renders "No active card". The change is a one-line `ScreenMark` in a known flex-column parent, but I have not seen it render.

Typecheck clean, prettier clean, `ds-lint` ratchets all pass, every suite touching a changed file green.

`e2e/shots/pages.spec.ts` is new: it captures real routes and the overlays that have no `open`/`visible` prop, recording the URL it actually landed on so a redirect shows up as a note rather than a mislabelled screenshot.

### DIFF
--- a/e2e/shots/pages.spec.ts
+++ b/e2e/shots/pages.spec.ts
@@ -120,6 +120,8 @@ const CAPTURES: Capture[] = [
     },
     { id: 'p39-language', name: 'Language', route: '/settings/language', fixture: 'settings-language' },
     { id: 'p40-withdraw', name: 'Withdraw — saved accounts', route: '/withdraw', fixture: 'withdraw' },
+    { id: 'p41-home-add-drawer', name: 'Home — Add money drawer', route: '/home?drawer=add' },
+    { id: 'p42-home-send-drawer', name: 'Home — Send drawer (no mark)', route: '/home?drawer=send' },
 ]
 
 function seenOnceModals(): void {
--- a/e2e/shots/pages.spec.ts
+++ b/e2e/shots/pages.spec.ts
@@ -1,0 +1,226 @@
+/**
+ * One PNG per app ROUTE, and per modal that only exists inside one.
+ *
+ * src/dev/surfaces/registry.tsx can only mount a surface that takes an
+ * open/visible prop. The overlays declared inline in a page — the three Backup
+ * FAQ sheets, the qr-pay KYC branches, the delete-account sequence — have no
+ * such prop, and neither do the `Notification` blocks that sit in a screen
+ * rather than in an overlay. Those are reachable only by loading the real route
+ * and, where needed, clicking the thing that opens them.
+ *
+ * Writes `<PAGES_OUT>/<id>.png` plus `<id>.json` recording the URL that was
+ * actually shot and whether every click landed, so a page that redirected or a
+ * row that moved shows up as a note instead of a mislabelled picture.
+ *
+ *   pnpm exec playwright test --config=playwright.shots.config.ts \
+ *     --project=375 e2e/shots/pages.spec.ts
+ */
+
+import { expect, test } from '@playwright/test'
+import { mkdir, writeFile } from 'node:fs/promises'
+import { join } from 'node:path'
+import { FIXTURE_STORAGE_KEY } from '../../src/dev/fixtures/active'
+
+const OUT_DIR = process.env.PAGES_OUT ?? 'e2e/__shots__/pages'
+const FIXTURE = 'profile-edit'
+const FROZEN_NOW = new Date('2026-08-15T12:00:00.000Z')
+
+type Capture = {
+    id: string
+    name: string
+    route: string
+    /** Fixture to serve this capture. Defaults to FIXTURE (the demo baseline). */
+    fixture?: string
+    /** Accessible names clicked in order before the shot. */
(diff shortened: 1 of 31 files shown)


innolope-dev:
Before/after screenshots for all twelve items: https://claude.ai/code/artifact/65d087e2-fc2d-4393-a432-30d595f00ddd

Eight have a real pair. The other four have no screen the harness can reach and are labelled as such — item 7 (`/card/limit`, `/card/pin`) is the one genuinely unverified by eye, since both sit behind `ActiveCardGate` and no fixture has an active card.

innolope-dev:
Re: **MAJOR: Remove the card country-availability claim** — I traced the policy to `mono/content/_system/guidelines/partners/rain/marketing-compliance.md` §7, and the internal restatements in `product/card.md:57` and `skills/update-content/references/guardrails.md:28`. The rule and the acceptance-vs-issuance split are as you describe.

The scope is where this lands differently. §1/§3 bound that guide to content that is **public, persistent, and discoverable**. This string is none of those: `residence` is setup step 4, *after* `signup` (`Setup.consts`), so it renders only for a user who already has an account — not on the website, not indexable, not reachable anonymously. It also reports back the residence the user just typed rather than anything Peanut disclosed.

Product owner ruled on 2026-09-05 to keep the phrasing on that basis. The rationale is recorded in the `Residence.tsx` comment, along with the two conditions that would flip it: putting this screen into app-store or marketing screenshots (§3 counts UI visuals), or moving the step ahead of signup. The comment also notes that the claim tracks the *declared* residence while the card gate reads the KYC-time geo, so a declaration that fails verification can still be refused later.

Not marking this resolved unilaterally — flagging the scope reading so you or a human can push back if you read §3 differently.

innolope-dev:
Took your suggested phrasing — the clause now reads:

> The Peanut card is also available — it needs an ID check, and you'll join a waitlist.

No country framing, so §7 is satisfied on its own terms and the screenshot exposure you flagged is gone rather than argued around. Being self-contained, one string now serves both the rail and no-rail branches, so `cardClauseNoRail` is retired across all four locales.

Both MAJOR findings should be resolved by this. Thanks for the push on the second round — the screenshot angle was the right objection.

PR #3002, opened 2026-09-05
fix(passkey): one signup tap, one ceremony — and keep the Android error name

innolope-dev:
Investigating [PEANUT-UI-T09](https://peanut-c34d84c05.sentry.io/issues/7709110627/) — a signup that could not be completed on Android native (binary 1.3.0, OTA `d5f8d49`, Motorola g86, Android 16).

## What the session actually looked like

From the breadcrumbs, on `/setup/?screen=passkey-permission`:

| time (UTC) | what happened |
|---|---|
| 16:21:57, 16:21:59 | two taps on **Set it up**, 2.4s apart |
| 16:22:01.194 | tap A reaches `handleRegister` → resets ZeroDev state, clears the passkey cookie |
| 16:22:02.534 | tap B reaches `handleRegister` → resets state **again**, then `CeremonyConflictError` |
| 16:22:03–16:22:12 | four more taps, six more resets, six more `CeremonyConflictError` |
| 16:22:23.448 | tap A's ceremony finally POSTs `/passkeys/register/options` |
| 16:22:32.868 | ceremony #3 ends: `Error: User cancelled the selector` after 9.4s |
| 16:23:14.883 | retry → ceremony #4 ends the same way after 2.0s |

## Two defects, both fixed here

**1. The button is live during the pre-ceremony awaits.** `isRegistering` only flips once `handleRegister` runs, and `handlePasskeySetup` awaits twice before that — the live `checkPasskeySupport()` re-check and the username-availability request. On a phone that is a second or more of a fully enabled button.

That alone would be cosmetic, except each losing tap did real damage:

- it ran `resetZeroDevState()` + `removeFromCookie(WEB_AUTHN_COOKIE_KEY)` **before** the conflict guard, wiping the address and cookie the in-flight registration was about to fill;
- its catch dispatched `setIsRegistering(false)` — releasing the *winner's* flag and re-enabling the button mid-ceremony, which is what turned one double-tap into a self-sustaining run of conflicts.

Fixed by latching the whole handler in `SetupPasskey` (and surfacing the busy state for the awaits), bailing out of `handleRegister` before the destructive reset when a ceremony already owns the window, and never clearing a flag this call did not set — in login as well as registration.

**2. Android passkey errors arrive nameless.** `@capgo/capacitor-passkey` loses the DOM error name twice over:

- `CapacitorPasskeyPlugin.rejectCreateException` maps every non-`CreatePublicKeyCredentialDomException` to `"UnknownError"` — which is exactly the bucket `CreateCredentialCancellationException` lands in;
- the JS shim's `normalizePluginError` short-circuits on `error instanceof Error` and returns Capacitor's rejection unchanged, and `CapacitorException` never sets `name`.

So dismissing the Credential Manager sheet reaches us as `Error("User cancelled the selector")`, `err.name === 'Error'`, and every `switch (err.name)` falls through to its unknown-failure branch: the generic troubleshooting modal, plus a `passkey_setup_error` Sentry event for something the user did deliberately. That is also why the ceremony telemetry on this issue reads `errorName: "Error", errorCode: "LOGIN_ERROR"`.

`normalizeNativePasskeyError` recovers the name from the plugin's `code`/`data.name` and maps androidx's cancellation and no-create-option messages to `NotAllowedError`. Verified still broken on the latest published plugin (8.5.1), so this stays a client-side normalization — no plugin bump, no binary.

## What this does *not* explain

The terminal failure on both attempts was a genuine `CreateCredentialCancellationException` from Android. After this change it is reported and worded correctly ("cancelled") instead of as an unknown crash, but if the sheet was never actually dismissed by hand, there is a third bug underneath that needs a device to pin down. The `assetlinks.json` side checks out: `me.peanut.wallet` is listed with `delegate_permission/common.get_login_creds` and three fingerprints, and `capacitor-passkey.xml` points at the right statement list.

## Scope

OTA-shippable — JS only, no native dependency change.

Tests: `SetupPasskey.test.tsx` (new) locks down one-tap-one-ceremony and the busy state; both cases fail against the pre-fix component. `webauthn.utils.test.ts` covers the four normalization outcomes.

### DIFF
--- a/src/components/Setup/Views/SetupPasskey.tsx
+++ b/src/components/Setup/Views/SetupPasskey.tsx
@@ -53,6 +53,21 @@ const SetupPasskey = () => {
     // hydrates asynchronously from a stale web-authn-key cookie (native cookie
     // jar) can never be mistaken for a fresh registration and skip the step.
     const registrationInitiatedRef = useRef(false)
+    /*
+     * Synchronous latch around the WHOLE handler. `isRegistering` only flips
+     * once handleRegister runs, and two awaits precede it — the live support
+     * re-check and the username-availability request — so on a phone the button
+     * stayed enabled for a second or more after the first tap. Every extra tap
+     * in that window started its own registration, and the losers surfaced
+     * CeremonyConflictError ("Something interrupted the passkey prompt") while
+     * the real ceremony was still coming up (PEANUT-UI-T09). A ref, not state:
+     * the latch also has to hold for callers that aren't the button — the help
+     * modal's retry action re-enters the handler directly.
+     */
+    const setupInFlightRef = useRef(false)
+    // Covers those same two awaits in the UI, so the button reads as busy
+    // instead of dead while the username check is in flight.
+    const [isPreparing, setIsPreparing] = useState(false)
 
     // preflight check for common passkey issues
     useEffect(() => {
@@ -68,6 +83,18 @@ const SetupPasskey = () => {
 
     // handle passkey registration with retry logic
     const handlePasskeySetup = async () => {
+        if (setupInFlightRef.current) return
+        setupInFlightRef.current = true
+        setIsPreparing(true)
+        try {
+            await runPasskeySetup()
+        } finally {
+            setupInFlightRef.current = false
+            setIsPreparing(false)
+        }
+    }
+
+    const runPasskeySetup = async () => {
         // clear any previous inline errors
         setInlineError(null)
         setErrorName(null)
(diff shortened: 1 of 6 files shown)


innolope-dev:
Fixed the MAJOR finding in 0703587 — `src/hooks/__tests__/useZeroDev-ceremony-conflict.test.tsx`, built on the existing `useZeroDev-login-failure` harness with only `currentCeremonyId()` stubbed (the guard's own window stays real, so a test that doesn't claim a ceremony still runs one).

Four cases, covering both rules you named:

- **ordering** — `handleRegister` with a ceremony active rejects with `CeremonyConflictError` and `removeFromCookie`, `zerodev/reset` and `toWebAuthnKey` are all untouched, so a refactor that moves the bail below the prologue reintroduces the auth-state wipe with a red suite.
- **flag ownership, registration** — a conflict does not dispatch `zerodev/registering: false`.
- **flag ownership, login** — a conflict does not dispatch `zerodev/logging-in: false`.
- **two controls** — an ordinary `NotAllowedError` on each path still clears its flag.

Verified against the pre-fix hook: the two guard cases fail, the two controls pass either way, so an inverted condition can't hide behind them.

PR #1538, opened 2025-12-09
fix: take account owner name and use it to create external account on bridge

kushagrasarathe:
- contributes to TASK-17164 : take account owner name from user in ui always when creating an external account

PR #3003, opened 2026-09-06
fix(limits): render the Manteca cap-nudge, which never reached a screen

innolope-dev:
## What

The Manteca cap-nudge has never appeared on a screen. The backend has been emitting `mantecaCapNudge` on capped rails for a while and `railVerdict` correctly attaches it to the enabled rail's `verdict.nextAction` — but the only downstream consumer of a hint, `firstAdvisory`, skips any action without an `effectiveDate`, and that field is only ever set on Bridge future-dated requirements. The nudge was dropped there, silently.

Consequence: a capped user had no in-app route to raising their limit. The one limit-raising CTA we ship (`IncreaseLimitsButton`) drives a different endpoint (`POST /users/increase-limits`) and is scoped to BR legacy users, so an AR user who hit the cap saw nothing at all.

## Surface: the limits page

Of the two candidates, this one. The limits page already owns limit raising and sits next to the existing button; the payment-time route would interrupt a transfer to advertise a limit raise.

That choice also decided the plumbing. `firstAdvisory` keeps its `effectiveDate` requirement — loosening it is precisely what would have pushed the nudge into the add-money/withdraw pre-empt modal — so the cap-nudge gets its own selector, `selectMantecaCapNudge`, and Bridge advisory behaviour is byte-identical (the pre-existing test pinning "a hint WITHOUT effectiveDate is not an advisory pre-empt" still passes untouched).

## Both states render

| marker | hint | card |
| --- | --- | --- |
| fresh cap block | `sumsub:source_of_funds`, `raise-manteca-limit` | description + CTA → source-of-funds Sumsub flow |
| RFI completed | `manteca:limit-review`, kind `wait` | review copy, **no control on it** |

The `wait` state matters: Sumsub accepting the document does not mean Manteca raised the cap (still a manual support step), so it must neither re-ask for a document already submitted nor claim a raise that hasn't happened. A genuinely new cap block re-seeds the marker and the actionable CTA comes back on its own — and a fresh block outranks an in-review sibling, so the CTA wins.

Both branches are pinned to the action `kind` the FE can actually honour, so a future backend that re-purposes either key cannot make the review state tappable.

Between the SDK submit and the backend stamping `submittedAt` there is a webhook round-trip, so the card flips to the review copy optimistically and drops the flag once the backend's own state takes over — otherwise the CTA lingers and invites a second upload of the same document.

The component is deliberately tiny, for the reason the card's PoA self-heal gives: the multi-phase KYC machinery is bank-onboarding shaped and completes on rail semantics this lifecycle never reaches.

## Verification

- `capability-gate.test.ts` — 8 new selector cases (hint with no `effectiveDate` surviving; review state; re-block precedence; verdict-vs-`hintActions` sourcing; Bridge advisory not matching; miskinded action not tappable). All 53 in the file pass.
- `CapNudgeCard.test.tsx` — 8 cases, one per state plus the start/failure/submit transitions.
- `tsc --noEmit` clean; prettier clean; ds-lint ratchet green (no metric increased).

## Not in this PR

The open question on peanut-api-ts#1423 — whether Manteca documents `POST /v2/documentations/uploadUrl` with `docType: FUNDS` and an async `OPERATION_LIMIT_UPDATED` confirmation — is unresolved and needs someone with Manteca dashboard access. Worth noting the asserted path *moved between review rounds* (`/v2/…` in the first two, `/crypto/v2/…` in the third); a documented vendor endpoint doesn't move, so the exact path needs confirming rather than copying. If it is real, the raise could be automated end-to-end and the `wait` state would clear on a *confirmed* raise rather than on Sumsub GREEN. Nothing here blocks that: it would change what clears the review state, not the states themselves.

Copy (`limits.capNudge`, en / es-419 / pt-BR) landed separately in #2852 and is already on `dev`.

### DIFF
--- a/src/features/limits/components/CapNudgeCard.tsx
+++ b/src/features/limits/components/CapNudgeCard.tsx
@@ -9,7 +9,20 @@ import { startKycAction } from '@/app/actions/sumsub'
 import { useAuth } from '@/context/authContext'
 import { useCapabilities } from '@/hooks/useCapabilities'
 import { useLimits } from '@/hooks/useLimits'
+import { markSubmitted } from '@/hooks/useSubmissionWindow'
+import { getUserPreferences, updateUserPreferences } from '@/utils/general.utils'
 import { selectMantecaCapNudge } from '@/utils/capability-gate'
+
+/**
+ * How long a local submission suppresses the CTA on its own.
+ *
+ * The backend only flips the hint from `raise` to `wait` when the RFI webhook
+ * lands, which is asynchronous and can be lost. Long enough to cover the
+ * ordinary webhook, short enough that a lost one re-offers the upload instead
+ * of hiding it forever — and short enough that it can never be the reason a
+ * genuinely NEW cap block goes unanswered.
+ */
+const CAP_NUDGE_SUBMITTED_TTL_MS = 10 * 60 * 1000
 
 /**
  * The Manteca cap-nudge — the surface for "you hit your monthly cap, verify your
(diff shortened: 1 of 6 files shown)


innolope-dev:
### On the 10-minute submission TTL — fixed in the backend, not here

The advisory on `CapNudgeCard.tsx` (10-minute TTL vs the documented 1–3 business day review turnaround) was real, and the fix belongs on the API side. It is open as peanutprotocol/peanut-api-ts#1538.

**Why the frontend couldn't fix it.** The TTL was never racing the Manteca limit review — it was racing the *Sumsub document review*. `markCapNudgeSubmitted` is called from `rfi-level-reviewed.ts`, whose guard reads "Sumsub verdict — only GREEN counts as completion", so the hint only flipped `raise → wait` once a reviewer had approved. Our own copy sets the expectation for that step: *"a few minutes, or 1–3 business days if a reviewer takes a thorough look."* So anyone whose document went to a human had the CTA come back after 10 minutes and was re-prompted to upload a document they had already sent — minting a second Sumsub action.

Extending the TTL doesn't fix it either: a **RED** verdict also leaves the hint at `raise`, so a multi-day TTL would park a *rejected* user on "under review" for days with no way back to the upload. The client can't tell "still pending" from "rejected", and that's the actual problem.

**What #1538 changes.** `applicantActionPending` — document submitted, no verdict yet — now retires the nudge, so the `wait` state begins at submission. A rejection clears `submittedAt` and returns the CTA; the support alert moves onto the marker's own `alertedAt` so it stays exactly-once now that retiring and alerting are driven by different events.

**No change needed in this PR.** With #1538 in, the 10-minute marker here covers only the seconds between the SDK reporting a submission and the pending webhook landing — which is what a lost-webhook backstop should be, rather than a race against a human reviewer. The value stays as-is.

**Merge ordering:** #1538 should land before or with this. Until it does, this PR ships the behaviour the advisory describes.

PR #3009, opened 2026-09-07
feat(migration): QR payload hand-off through /app, store pair fixes, app-link coverage (TASK-21788)

0xkkonrad:
## Summary

PR 1 of the peanut.me landing → app-store migration (TASK-21788): the *mechanism* only, no page layout. It makes the download QR carry deferred context, teaches `/app` to hand that context to the right store on the right platform, fixes the store-button pair so it lays out correctly on desktop and can't overflow a phone, and claims `/app` for App Links on both platforms so an already-installed user who scans a QR lands in the app instead of the store page.

Everything here is inert until the `pwa-sunset` PostHog flag is on, except the App Links files (which are additive) and the `StoreBadges` → `StorePair` rename.

## What changed

- **`DownloadQR`** takes `size?: 160 | 192 | 224` and **one** context input: either `payload` (a querystring the caller already holds) or `handoff` (`{dest, invite}`), mutually exclusive in the props type. Given `handoff` it derives the payload itself after mount and passes the same `handoff` to the `StorePair` beneath it, so the code and the buttons can never carry different context. The encoded URL is `${origin}/app?${payload}&s=<surface>` (or `?s=<surface>` alone), so every scan is attributable to the QR that produced it and carries locale / invite / badge campaigns / destination through the install.
- **`size` is the QR frame width; the module area is 36px narrower** (`p-4` + `border-2`): 160 → 124px modules, 192 → 156px, 224 → 188px. Documented on the constant. **PR 2 should pass `size={192}` for the hero and `size={224}` for the app fold and the footer** to hit the brief's stated module areas.
- **`MIGRATION_QR_SHOWN`** moved from mount to a real impression: an `IntersectionObserver` at 50% visibility, once per mount, with `{ surface, hasContext }`. The app fold and the footer render their QR far below the fold, so counting mounts would have made the scan rate look broken. Falls back to counting the mount where `IntersectionObserver` is missing (jsdom, old WebViews); in a modal the QR is already in view on open, so that is the open event. The capture waits for the derived payload, so a contextful QR never reports `hasContext:false`.
- **`/app` (`src/app/app/page.tsx`)** reads `window.location.search` after mount and validates it with `parseDeferredPayload` (marker check):
  - *Android + payload* → auto-redirects to `playStoreUrlWithReferrer(payload)` and fires `trackDeferredHandoffCreated('android')`.
  - *iOS + payload* → **no** auto-redirect. The clipboard hand-off only succeeds inside a user gesture, so the visitor gets the pink App Store button + stroke Google Play button, and the App Store tap calls `copyIOSHandoff(payload)` inside the click handler before the anchor navigates.
  - *No payload* → today's behaviour, unchanged.
  - *Inside the native app* → `applyDeferredPayload(parsed)` then `router.replace(dest ?? '/home')`.
  - **`DEFERRED_LINK_HANDOFF_CREATED` is counted once per visit** (ref guard shared by the auto-redirect and the tap). Without it, a bounce that never took over — blocked, offline, Play missing — re-enabled the buttons after 4s and a tap counted a second hand-off, deflating the restore match rate on exactly the flaky devices.
  - Store taps report `trackStoreClick(store, 'smart_link', hasPayload, qrSurface)`. The `s=` tag is **read** here (validated against `MIGRATION_SURFACES`) and reported as `qr_surface` on the click and hand-off events, which is what joins a smart-link click back to the hero / app fold / footer / rates QR that produced the scan. It is still stripped before the payload is re-emitted, so it never rides into the Play install referrer.
- **`StoreBadges` → `StorePair`** (rename, imports updated). Anchors use `storeAnchorHref(store, handoff)` + `onStoreAnchorClick(store, surface, handoff)`. Container widths are per-appearance: `hero` is `max-w-[27.5rem]` (440px) so two `sm:w-52` anchors (208px each) plus `gap-3` (12px) = 428px sit on **one line** at every width ≥ `sm`, including 1366 and 1440; the compact row keeps `max-w-[26rem]`. Below `sm` the hero anchors are `w-full` and stack by design. The two hrefs are memoized per handoff identity — the android one runs `buildDeferredPayload` (cookie + badge-campaign + `window.location` reads) and the hero pair lives inside `LandingPageClient`, which re-renders every scroll-driven animation frame until PR 3 lands.
- **`storeAnchorHref` / `onStoreAnchorClick`** take an optional `StoreHandoff`, matching what `openStore` already accepted.
- **`StickyMobileCTA`**: `deviceType === WEB` (desktop-mode phone, or a UA we don't recognise) now renders the compact store pair instead of guessing iOS. Everything else is unchanged.
- **App links**: `/app` + `/app/*` added to all three appIDs in `public/.well-known/apple-app-site-association`, and `android:path="/app"` + `android:pathPrefix="/app/"` to the verified intent filter. `app` added to `NATIVE_EXPORT_ROOTS`, and `mapDeepLinkPath` collapses the whole `/app/*` claim onto `/app` — the wildcard has no route behind it on either the web app or the static export, so a passthrough would open the app onto a missing route (see Divergences).
- **`MIGRATION_SURFACES`** gains `SMART_LINK`, `LANDING_APP_FOLD`, `LANDING_FOOTER`, `LANDING_RATES`, `LANDING_COUNTRIES`, `LANDING_DOOR`; new `MIGRATION_SURFACE_PARAM = 's'` and `isMigrationSurface()` are shared by the QR builder and `/app`.
- **New tests**: `DownloadQR` URL building, handoff-derived payload reaching both channels, sizes, impression threshold and `hasContext`; `StorePair` hrefs, handoff, tap tracking and the hero row's one-line arithmetic; `/app` payload branching (android bounce, iOS no-bounce + clipboard-on-tap, bare visit, native apply-and-route, flag-off 404), `qr_surface` attribution, an unknown surface tag, and the once-per-visit hand-off count; `StickyMobileCTA` WEB case; `/app/*` collapsing in the deep-link mapper; an App Links parity guard (iOS AASA ↔ Android intent filter).

No new user-facing copy — no i18n keys added.

## How verified

- `pnpm typecheck` → clean on every touched file. 6 errors remain, all `TS2307` in files this PR does not touch (`src/utils/web-vitals-shim.ts`, `src/utils/app-review.ts`, `src/utils/native-settings.ts` and the web-vitals test): `web-vitals`, `@capacitor/app-launcher`, `@capgo/capacitor-in-app-review` and `capacitor-native-settings` are absent from this sandbox's shared `node_modules`. Same 6 before and after the change.
- `pnpm test -- src/components/Migration/__tests__/DownloadQR.test.tsx src/components/Migration/__tests__/StorePair.test.tsx src/app/app/__tests__/smart-store-link.test.tsx src/utils/__tests__/app-links.test.ts src/utils/__tests__/native-routes.test.ts src/components/LandingPage/__tests__/StickyMobileCTA.test.tsx src/components/Migration/__tests__/MigrationDownloadModal.test.tsx` → **7 suites, 246 tests, green**. `pnpm test -- src/utils/__tests__/deferred-link.test.ts src/utils/__tests__/migration.utils.test.ts` → **2 suites, 63 tests, green**.
- `pnpm lint --quiet <touched files>` → clean. `prettier --check <touched files>` → clean.
- Hero row geometry: asserted in `StorePair.test.tsx` against the resolved values rather than the class string (jsdom has no layout). Tailwind v4 with no `--spacing` or `--breakpoint-sm` override in `src/styles/globals.css`, so `w-52` = 13rem = 208px, `gap-3` = 0.75rem = 12px, `sm` = 640px, and the `max-w-[27.5rem]` = 440px cap clears the 428px the row needs. The previous `max-w-[26rem]` (416px) wrapped the pair at every width ≥ `sm`.
- No dev-server/browser check in this PR by design — the runtime pass (screenshots, QR decode, flag-off pixel diff) runs on the page PR, which is where the hero lockup actually renders.

## Divergences

1. **`/app` also had to become a mappable native route.** The brief asked only for the AASA + manifest entries. Adding them immediately failed the existing AASA drift guard, which asserts every claimed root resolves through `deepLinkToNativePath`. That is not a test technicality: an App Link the mapper drops cold-boots the app to `/home`, so the brief's own native branch (`applyDeferredPayload` → `router.replace(dest)`) could never run. `/app` is already built by the native static export, so nothing new ships — only the mapper's verdict changed. `app` was therefore added to `NATIVE_EXPORT_ROOTS` and removed from `WEB_ONLY_EXPORTED` in the export-drift test, with both comments updated. Side effect, accepted: an in-app anchor to `/app` would now push in-app rather than open the browser; there are none in the codebase.
2. **`handoff` threaded through the store helpers, and it is now the *only* context channel on `DownloadQR`** (a caller passes `payload` **or** `handoff`, never both; `handoff` derives the payload internally). PR 2 opens the hoisted modal with `dest=/send` on fold 4 and destination context on fold 5, and that modal's content is `DownloadQR` → `StorePair`. As first written the two props were independent: `handoff` alone left the QR context-free (and reported `hasContext:false`) while the buttons beneath it carried the destination, and `payload` alone did the reverse, with nothing catching it. One input, one derivation. The brief's reason for the caller building the payload — it reads `window` — is met by deriving it in a mount effect.
3. **`size` is the QR frame width, not the module area.** 160 is exactly the frame width `QRCodeWrapper` ships today (`max-w-[160px]`); the modules inside it are 124px. Reading "160 default" as the module area would silently resize every existing QR on merge, so the reading that keeps the stated default a no-op is the frame width. The frame→module mapping is now documented on the constant and a `224` option added, so PR 2 can hit the brief's module areas: `192` for the hero (156px modules), `224` for the app fold and footer (188px).
4. **The `/app/*` App Links wildcard is collapsed in the deep-link mapper.** The claim itself is exactly what the brief asked for on both platforms, but `/app/<anything>` is not a route on the web app or the static export. `mapDeepLinkPath` now returns `/app` for any `/app/*` deep link, so an installed user opening `peanut.me/app/anything` lands on the real page instead of the SPA's missing-route → home bounce. Neither the AASA drift guard nor the app-links parity test can see this class of bug — both only compare the two hand-maintained lists to each other.
5. **`/app` reads the `s=` surface tag instead of only stripping it.** The brief had the QR write the tag and `/app` strip it before re-emitting the payload. It still strips it (the tag must never ride into the Play install referrer), but it is now also validated against `MIGRATION_SURFACES` and reported as `qr_surface` on this page's click and hand-off events. Without that read the tag was costing QR modules for nothing and the landing surfaces were indistinguishable in the smart-link funnel — the one attribution question the param exists to answer.

## Flag-off impact

None on the page: every behaviour change is inside a `pwa-sunset` branch (`/app` still 404s with the flag off, `StickyMobileCTA`'s flag-off arm is untouched, `StorePair`/`DownloadQR` only render on migration surfaces). Three changes are live regardless and are additive by design: the `/app` App Links claim on both platforms, `app` joining `NATIVE_EXPORT_ROOTS`, and the mapper's `/app/*` collapse.

**Release gate — the two halves of the `/app` claim ship on different vehicles.** The AASA is a static file served by the web deploy, so iOS picks up the `/app` claim for already-installed apps within days of merge. The Android intent-filter lives in the APK and only takes effect in the next Play release. `app-links.test.ts` asserts the two *files* agree and stays green through the entire window in which the two *platforms* do not. So: **do not flip `pwa-sunset` on for Android traffic until an Android build containing this manifest is live in Play**, and run the runtime checklist's "Android UA" item against that binary, not against the web deploy. Until then an installed Android user who scans a download QR still lands on web `/app` (404 with the flag off; store buttons for an app they already have once it is on).

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -37,6 +37,8 @@
 
                 <data android:scheme="https" android:host="peanut.me" />
 
+                <data android:path="/app" />
+                <data android:pathPrefix="/app/" />
                 <data android:path="/home" />
                 <data android:pathPrefix="/home/" />
                 <data android:path="/card" />
--- a/public/.well-known/apple-app-site-association
+++ b/public/.well-known/apple-app-site-association
@@ -5,6 +5,7 @@
             {
                 "appID": "4K73GMPZP8.com.squirrellabs.peanut-app",
                 "paths": [
+                    "/app", "/app/*",
                     "/home", "/home/*",
                     "/card", "/card/*",
                     "/claim", "/claim/*",
@@ -26,6 +27,7 @@
             {
                 "appID": "web.app.peanut",
                 "paths": [
+                    "/app", "/app/*",
                     "/home", "/home/*",
                     "/card", "/card/*",
                     "/claim", "/claim/*",
@@ -47,6 +49,7 @@
             {
                 "appID": "PW388G893L.me.peanut.wallet",
                 "paths": [
+                    "/app", "/app/*",
                     "/home", "/home/*",
                     "/card", "/card/*",
                     "/claim", "/claim/*",
--- a/src/app/app/__tests__/smart-store-link.test.tsx
+++ b/src/app/app/__tests__/smart-store-link.test.tsx
@@ -1,0 +1,169 @@
+/** @jest-environment jsdom */
+/**
+ * /app — the smart store link a download QR encodes.
+ *
+ * The branching here is the whole deferred-deep-link hand-off: android may
+ * bounce automatically because the payload rides the Play install referrer,
+ * iOS may NOT, because the clipboard write only succeeds inside the tap. A
+ * regression that auto-redirects iOS loses the context for every scan without
+ * anything visibly breaking.
+ */
+import React from 'react'
+import { fireEvent, render, screen } from '@testing-library/react'
+import { IntlWrapper } from '@/test-utils/intl'
+import { STORE_URL } from '@/constants/migration.consts'
+
+let mockDeviceType = 'web'
+jest.mock('@/hooks/useGetDeviceType', () => ({
+    ...jest.requireActual('@/hooks/useGetDeviceType'),
+    useDeviceType: () => ({ deviceType: mockDeviceType }),
+}))
+
+const mockReplace = jest.fn()
+const mockNotFound = jest.fn()
+jest.mock('next/navigation', () => ({
+    useRouter: () => ({ replace: mockReplace }),
+    notFound: () => mockNotFound(),
+}))
+
+// flags land immediately: the page waits for this callback (or a 4s timeout)
+// before it is willing to 404
+jest.mock('posthog-js', () => ({
+    capture: jest.fn(),
+    onFeatureFlags: jest.fn((cb: () => void) => {
+        cb()
+        return jest.fn()
+    }),
+}))
+
+let mockNative = false
+jest.mock('@/utils/capacitor', () => ({
+    isNativeBridge: () => mockNative,
+    isCapacitor: () => mockNative,
+    isAndroidNative: () => false,
+    isIOSNative: () => false,
+    openExternalUrl: jest.fn(),
+}))
+
+let mockFlagOn = true
+jest.mock('@/utils/migration.utils', () => ({
+    isPwaSunsetOn: () => mockFlagOn,
+    trackStoreClick: jest.fn(),
+}))
+
+// the marker check is the real one (pinned in deferred-link.test.ts); only the
+// side-effecting ends are stubbed so the test can see what changed hands
+const mockCopyIOSHandoff = jest.fn().mockResolvedValue(undefined)
+const mockTrackHandoffCreated = jest.fn()
+const mockApplyDeferredPayload = jest.fn<{ dest: string | null; locale: null }, [unknown]>(() => ({
+    dest: null,
+    locale: null,
+}))
+jest.mock('@/utils/deferred-link', () => ({
+    ...jest.requireActual('@/utils/deferred-link'),
+    copyIOSHandoff: (...args: unknown[]) => mockCopyIOSHandoff(...args),
+    trackDeferredHandoffCreated: (...args: unknown[]) => mockTrackHandoffCreated(...args),
+    applyDeferredPayload: (payload: unknown) => mockApplyDeferredPayload(payload),
+}))
+
+jest.mock('@/components/Migration/MigrationHero', () => ({
+    __esModule: true,
+    default: () => <div data-testid="hero" />,
+}))
+
+import { trackStoreClick } from '@/utils/migration.utils'
+import SmartStoreRedirect from '../page'
+
+const PAYLOAD = 'pnutdl=1&lang=pt-br&invite=ABC123&dest=%2Fsend'
+const locationReplace = jest.fn()
+
+function visit(search: string) {
+    Object.defineProperty(window, 'location', {
+        configurable: true,
+        value: {
+            origin: 'https://peanut.me',
+            pathname: '/app',
+            search,
+            href: `https://peanut.me/app${search}`,
+            replace: locationReplace,
+        },
+    })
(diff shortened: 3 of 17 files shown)


0xkkonrad:

PR #3010, opened 2026-09-07
feat(landing): app-store funnel — S-full hero, hoisted QR modal, Get-the-app fold, footer block (TASK-21788)

0xkkonrad:
## Summary

Turns the landing page into the app-store funnel for the `pwa-sunset` window (TASK-21788, PR 2 of 4 — the page). Every download CTA on the page now ends at a store or at a QR that a phone can scan; nothing routes to `/setup` or `/send` while the flag is on. With the flag off the page is unchanged: the flag is false until mount, so SSR and the first client render are today's page byte for byte.

Depends on PR 1 (mechanism) for `/app`'s payload branching and the app-links files. The two were built in parallel from the same brief; where PR 1's component API does not exist on `dev` yet, this PR added its own equivalents rather than editing PR 1's files — see Divergences.

## What changed

- **Hero (fold 1).** The CTA slot becomes a device-specific lockup. Desktop: the S-full lockup — a 160px-module QR encoding `/app` with the deferred payload and `s=landing_hero`, the scan hint, the white App Store / Google Play pair, and the content-system subtext; horizontal from `lg` up, stacked below. Phone: one full-width white "Download now" with the detected store's mark, an href that starts at `/app` and upgrades to the store deep link on mount, an "Other store" text link to `/app`, and the same subtext. No two-button row exists on a phone, so nothing can overflow at 320px.
- **Hero geometry.** From `md` up and **flag-on only** (the `compactArtwork` prop), the artwork clamp moves from `calc(100svh-23rem)` to `calc(100svh-29.5rem)` and the mascot's from `calc(100svh-28rem)` to `calc(100svh-34.5rem)`. The desktop lockup is ~164px where the single button was ~76px, and that 88px is what the clamp buys back so the marquee stays in the fold at 1440x900 and 1366x768. Below `md` both keep today's values: the phone CTA barely changes height, so the laptop clamp there only cut hero artwork (up to ~100px on a 320x568 phone) for nothing.
- **Mascot.** A `ResizeObserver` on `#hero` now re-runs `position()`. `resize` alone missed every height change that leaves the window alone — the CTA swapping in on mount, a late webfont, an image decoding — each of which left the mascot floating off the headline.
- **One hoisted QR modal.** `AppModalProvider` wraps the whole page inside `LandingPageClient`, above the server-rendered slots, and owns a single `ScanToDownloadModal`. Its `interceptAppCta(surface, handoff)` is the contract every re-pointed CTA calls: desktop opens the modal tagged with the caller's surface **and carrying the hand-off**, phones bounce to their store with the same hand-off, flag-off returns false and the CTA keeps its old behaviour.
- **Fold 4, zero fees.** "Send Money" calls the intercept with `landing_rates` and `dest=/send` before the signed-in branch.
- **Fold 5, global cash.** The SIGN UP inside the countries illustration is now a small `'use client'` child; with the flag on its href is `/app` and the click opens the modal (`landing_countries`).
- **Fold 10, send in seconds → Get the app.** A client child swaps the whole body on the flag: "GET THE APP." over the existing tagline, then the scanner capture in a 2px/28px-radius outline with a 6px hard shadow, a 192px QR with the hint, and the store pair stacked with the subtext. Phone gets the title, the outline and one store button. The capture is `store-listing-v2/capture/shots-out/en/01-qr-scan.png` downscaled to 360px wide and palette-optimised — 12.7 KB. The 3-up is `md:gap-8 lg:gap-14` so the row fits the 736px content box at exactly 768px instead of silently shrinking the QR below its specified 192px.
- **Fold 12, footer.** A new `FooterGetTheApp` block on top of the footer chrome, **opt-in via `showGetTheApp` and passed only by the landing page**: heading, hint, the store pair in white-on-black with the border kept and the shadow dropped (the LocaleSwitcher's treatment), and a 192px QR on desktop; one download button on a phone. It renders `null` with the flag off. The SEO footer's Learn More column gains two crawlable store-listing links **from `MIGRATION_CUTOVER_DATE` on** — every other download CTA is a script-driven bounce a crawler cannot follow, and this footer is a server component on every marketing page, so a date is the only gate it can read.
- **Code splitting.** `HeroAppLockup`, `PhoneAppCta` and `GetTheAppFold` load through `next/dynamic` with `ssr: false`. None can render before mount, so nothing is lost, and `AppQrCode → QRCodeWrapper → react-qr-code` plus the scanner capture leave the landing page's main chunk for every visitor, flag off included.
- **Analytics.** `MIGRATION_SURFACES` gains the six keys from the brief. `MIGRATION_QR_SHOWN` fires once per QR, at 50% visibility, with `{surface, hasContext}` — so the three QRs on the page are counted where they are seen, not where they mount.
- **Dormant sub-flag.** `LANDING_HERO_QR_FLAG = 'landing-hero-qr'` is exported and documented; nothing reads it.
- **Copy.** All **copy pass pending**: `migration.otherStore` and `migration.getTheApp` in the app catalog (`en` / `es-419` / `pt-BR`, marketing subsets regenerated), plus five URL-locale keys in the marketing catalog — `migrationGetTheApp`, `migrationQrTitle`, `migrationScanHint`, `migrationDownloadNow`, `migrationOtherStore` in `en` / `es-419` / `es-ar` / `pt-br`. The landing folds read the URL-locale ones so /pt-br stays Portuguese on a Spanish phone.

## Review round 1 — what changed

Fix-then-ship. Every blocker and major addressed, plus all six minors.

- **blocker, e2e flag.** `isPwaSunsetOn` honoured its localStorage override only under `IS_DEV`, so against `next start` (which is what the spec and `playwright.regression.config.ts` run) every "flag on" case rendered the flag-OFF page and passed vacuously. The override is now live on every non-production domain (`IS_DEV || BASE_URL !== 'https://peanut.me'`) — the boundary `isFeatureFlagEnabled`'s `nonProdBypass` already draws — and each flag-on case asserts its lockup is visible, so a silently-inert override fails loudly. Both directions pinned in `migration.utils.test.ts`.
- **blocker, desktop never exercised.** `useDeviceType` branches on the user agent, not the viewport, and the shared project is a Pixel 7 — so the 1440/1366 cases rendered `PhoneAppCta`, never the QR lockup the re-tuned constants exist for. The laptop and 768 cases now run under a desktop UA via per-describe `test.use`, and assert the QR lockup before measuring the marquee.
- **major, dropped hand-off.** `AppModalProvider` kept only the surface on desktop, so `dest=/send` was phone-only and the modal had no way to accept one. It now keeps `{surface, handoff}` and threads it into `ScanToDownloadModal`, which carries it to both the QR payload and the store anchors.
- **major, scroll lock.** Removing `#sticky-button-target` on the flag-on branch could strand `document.body.style.overflow = 'hidden'` with no event able to clear it (a visitor already at fold 10 when the flag resolves — scroll restoration on reload or back-navigation). The scroll effect now bails and cleans up when the flag is on.
- **major, phone artwork.** The `compactArtwork` clamp was applied at every width but tuned for the desktop lockup. It is `md:`-scoped now, matching its mascot twin.
- **major, flag-off store links.** The SEO footer's two store links shipped on every marketing page with the flag off. Gated on the cutover date; the flag-off SSR HTML carries zero store hrefs again.
- **minors.** URL-locale copy for the fold and footer headings; `md:gap-8 lg:gap-14` on the 3-up plus a 768x1024 gate case; `next/dynamic` for the three lockups; `showGetTheApp` scoping the footer block to the landing page; the hero Log In link kept and re-pointed at `/app` instead of removed; `AppStorePair` back to the locked `max-w-[26rem]`.

## How verified

- `pnpm typecheck` → 6 errors, all pre-existing `TS2307` module-resolution failures in four files this branch never touches (`src/utils/web-vitals-shim.ts`, its test, `src/utils/app-review.ts`, `src/utils/native-settings.ts`), caused by the worktree's symlinked `node_modules`. Nothing from the diff appears in the output.
- `pnpm test -- src/components/LandingPage src/components/Migration src/i18n src/utils/__tests__/migration.utils.test.ts` → **31 suites, 321 tests, all passing.** New this round: the two `isPwaSunsetOn` production-build cases, the modal hand-off assertion in `AppModalProvider.test.tsx`, and the store-listing pair in `Footer.test.tsx` re-cast as before/after the cutover date.
- `pnpm lint --quiet <touched files>` → clean. `prettier --check` over the same set → clean.
- SSR check against `next dev`: `curl` of `/` and `/pt-br` with the flag off contains **zero** `apps.apple.com` / `play.google.com` hrefs — the flag-off page no longer advertises the listings.
- Layout gate: `e2e/flows/landing-migration.spec.ts` — 4 locales x {1440x900, 1366x768, 768x1024, 390x844} x {flag off, flag on}, desktop UA on the first three, asserting the right lockup rendered (and that neither exists with the flag off), no horizontal overflow anywhere, and the marquee above the fold at both laptop sizes; plus one check that no `/setup` or `/send` href survives with the flag on. Screenshots go to `LANDING_SHOTS_OUT`. **Not executed in this session** — see divergence 17: this box could not keep `next dev` alive with a browser attached (V8 OOM mid-stream, pages never hydrated), so no flag-on capture from here would be honest evidence. CI runs it against a production build, where the override now works.

## Divergences

1. **PR 1's API is not on `dev`, so this PR added its own components and touched none of PR 1's files.** `AppQrCode` (bare framed QR, 160/192px, payload + surface in the url, visibility-gated impression) and `AppStorePair` (row/column/footer) do what the brief specifies for `DownloadQR`'s new props and `StorePair`. Reconciling the two branches means collapsing them.
2. **`MIGRATION_SURFACES` gains all six keys verbatim**, including the two only PR 1 and PR 4 read, so both branches produce an identical hunk instead of two partial lists that conflict.
3. **The re-tuned hero constants are gated on the flag** (and now on `md`). Applying them unconditionally changes the flag-off hero wherever the clamp binds (1366x768 does), which breaks "renders identically to today when the flag is off". Rule (a) wins.
4. **The hero keeps its content-system headline.** The artboards print "TAP. SCAN. ANYWHERE."; the h2 is `strings.heroTapScan` and the locked table changes only the CTA slot, so it is untouched.
5. **The two footer store links are untranslated English constants**, in a column that already renders authored English across every locale by design (`RESOURCE_NAME_OVERRIDES` documents the rule in place), and the store names are untranslated constants already.
6. **`StickyMobileCTA` is left to PR 1**, which lists the WEB → both-stores case in its own mechanism section.
7. **No unit test for the rates-widget re-point.** `NoFees` renders `ExchangeRateWidget` (router, `useSearchParams`, a rate fetch); the one changed line is covered by the Playwright no-`/send` assertion, and the intercept it calls is unit-tested directly.
8. **The screenshot gate takes its output path from `LANDING_SHOTS_OUT`** rather than hard-coding a scratchpad path into a repo file.
9. **The `pwa-sunset` localStorage override now covers every non-production domain**, not just dev — the only way to turn the flag on in a build with no PostHog key, which is what the layout gate runs against. peanut.me is unchanged.
10. **The SEO footer's store links ride the cutover date, not the flag**, because a client-only flag in a server component either loses the crawlable HTML or keeps the flag-off exposure. Pre-cutover crawlability is the trade.
11. **`ScanToDownloadModal` grew a `handoff` prop, and `storeAnchorHref` / `onStoreAnchorClick` gained an optional `handoff` argument.** With a hand-off the modal renders PR 2's hand-off-aware QR and pair; without one it renders `DownloadQR` verbatim, so the home banner, guest flow and dev registry are untouched. Both helper changes are additive and neither PR 1 nor PR 4 rewrites those functions.
12. **The migration download copy is duplicated into the marketing (URL-locale) catalog.** next-intl resolves the DEVICE locale, which put a Spanish H1 on a Portuguese fold. Teaching the provider the route locale is an `IntlCore` change far outside this PR. es-AR is voseo, consistent with its siblings.
13. **The footer download block is opt-in.** `FooterChrome` is shared by /careers, /m/[slug], /shhhhh and the marketing layout; the locked structure scopes the block to fold 12, so `showGetTheApp` is passed only by `LandingPageContent`.
14. **The hero Log In link is kept and re-pointed at `/app`** rather than removed — it is the only re-entry path a returning logged-out visitor has, and removing it also changed the CTA column's height on mount.
15. **Runtime re-verification was not possible in this session.** `next dev` plus a browser is OOM-killed on this box (a second session holds another dev server); pages never hydrated, so every flag-on capture would have been a disguised flag-off page. Screenshots taken in that state were deleted rather than filed. Static gates and the SSR check above are what this round is verified by.

Full text with the reasoning for each: `run/DIVERGENCES-page.md` (round 2 entries are numbered 11–17 there).

## Flag-off impact

**None.** Every change is inside the `pwa-sunset` branch, and the one exception from round 1 — the SEO footer's store-listing links, which rendered on every marketing page in every locale regardless of the flag — is now gated on `MIGRATION_CUTOVER_DATE`, verified by curling the SSR HTML for `/` and `/pt-br`. `AppModalProvider` wraps the tree and returns `false` from its intercept while the flag is off; `FooterGetTheApp` returns `null` and is only mounted on the landing page at all; the hero geometry, the fold-10 body and the countries CTA keep their exact current markup in the flag-off path, which `SendInSecondsBody.test.tsx`, `CountriesSignUpCta.test.tsx`, `FooterGetTheApp.test.tsx` and `Footer.test.tsx` each assert.


0xkkonrad:

PR #3011, opened 2026-09-07
chore(landing): delete the Send-in-seconds scroll-jack (TASK-21788)

0xkkonrad:
## Summary

The "Send in seconds" fold froze the page. On the way down, once the CTA reached the sticky-bar band, `LandingPageClient` set `document.body.style.overflow = 'hidden'`, pinned `window.scrollY`, and swallowed `wheel` and `touchmove` with `preventDefault` until the reader had pushed 500px of virtual delta into growing the button to 1.5x. This deletes that machinery. Scrolling is now ordinary scrolling everywhere on the landing page.

Part of the PWA-sunset landing work (TASK-21788), split out as its own PR because it is a straight deletion and touches no migration surface.

## What changed

- `LandingPageClient.tsx`: removed `isScrollFrozen`, `buttonScale`, `animationComplete`, `shrinkingPhase`, `hasGrown`, their mirror refs, `frozenScrollY` / `virtualScrollY` / `touchStartY`, `handleScrollDelta`, and the `scroll` / `wheel` / `touchstart` / `touchmove` effect with its `document.body.style.overflow` writes. `buttonVisible` and its footer-visibility effect stay — that is the separate fade-out on the footer.
- `hero.tsx`: dropped the `buttonScale` prop, its default, and `--cta-scale` from `getCtaStyle`. The `--cta-x` / `--cta-y` / `--cta-r` entrance and the hover transform are untouched.
- `sendInSeconds.tsx`: removed the `<div id="sticky-button-target">` wrapper — the only thing that read it was the deleted handler. `id="send-in-seconds"` on the section stays, and fold 10 keeps its `.cta-motion .cta-enter` entrance.
- `globals.css`: dropped the now-unsettable `scale(var(--cta-scale, 1))` from `.cta-motion` and the `cta-enter` keyframe (no-op; see Divergences).
- New tests: `__tests__/LandingPageClient.scrollJack.test.tsx` and `__tests__/ctaScrollJackRemnants.test.tsx`.

## How verified

- `pnpm test -- src/components/LandingPage` → 7 suites, 27 tests, all green (includes the two new suites and the pre-existing Footer / SEOFooter / CurrencySelect ones).
- The new suite is a real guard, not a vacuous one. jsdom gives every element a zero rect, so the old freeze window could never open by accident; the test re-inserts a `#sticky-button-target` with `getBoundingClientRect` stubbed to `{top: 600, bottom: 660}`, which is inside the band the old handler froze on at jsdom's 768px viewport. Run against the pre-deletion component (copied out of `HEAD` into a scratch file), its tests fail: body overflow becomes `hidden`, `wheel` and `touchmove` come back `defaultPrevented`, and the hero receives a `buttonScale`. Against this branch all four pass.
- Round-1 review fix: the fourth test used to assert `#sticky-button-target` was absent from a slot that never contained it, so it could not fail either way. It now asserts the thing this PR actually removed — the anonymous `<div ref={sendInSecondsRef}>` around `sendInSecondsSlot` — by checking the slot's `parentElement` is the render container (`LandingPageClient` returns a fragment). Re-adding `<div>{sendInSecondsSlot}</div>` to the component makes exactly that test fail, and reverting makes it pass again.
- Round-1 review fix: `ctaScrollJackRemnants.test.tsx` split the fold-10 block in two. One `it` holds only the scroll-jack regression assertions (`#send-in-seconds` present, `#sticky-button-target` null, `.cta-motion` present); a second, separately named `it` holds the `/send` href and the `.cta-enter` entrance, so PR 2's "GET THE APP" rewrite of fold 10 can retarget or delete that one without touching the regression guard or reading as a scroll-jack regression.
- `pnpm exec eslint --quiet` on the six touched files → clean, exit 0.
- `pnpm exec tsc --noEmit` → 6 errors, all pre-existing `TS2307` in untouched files, none in this diff (see Divergences #3).
- `pnpm exec prettier --write` on the touched files, then re-run clean.
- **Browser pass, flag off** (round-1 review fix — this is the one change in the set that ships to 100% of landing traffic, so it is verified in a real engine, not only jsdom). `next dev --webpack -p 3053` in the worktree + Playwright 1.58.2 chromium, on `/` and `/pt-br` at 1440x900 and 390x844 — four combinations, all clean:
  - `getComputedStyle(document.body).overflow` is `visible` and `document.body.style.overflow` is empty both while fold 10 is centred in the viewport and after scrolling past it; same for `<html>`.
  - a dispatched cancelable `wheel` (deltaY 400) and `touchmove` at fold 10 both come back `defaultPrevented === false`.
  - a real `mouse.wheel` sequence at fold 10 moves the page 1472 / 1860 / 1499 / 2021 px on the four combinations — the page never pins.
  - `#sticky-button-target` is absent from the live DOM; `scrollWidth - clientWidth` is 0 (no horizontal overflow).
  - the `.cta-enter` entrance still fires after the keyframe edit: replaying the class gives a running `CSSAnimation` named `cta-enter`, duration 450ms, computed transform `matrix(0.999914, 0.0130896, -0.0130896, 0.999914, 0, 4)` — the 0.75deg rotation and 4px lift of the `from` frame, resolving correctly now that `scale(var(--cta-scale, 1))` is gone.
  - fold-10 centring survived the wrapper removal: the `.cta-motion` element's centre X equals `#send-in-seconds`'s centre X to **0.00px** on all four combinations (720.00 vs 720.00 at 1440, 195.00 vs 195.00 at 390).
  - console: no React `#418` / `#423`, no `pageerror` from application code. The only console errors are dev-server infrastructure (`webpack-hmr` websocket handshake failures, and `ERR_CONNECTION_REFUSED` on the run where earlyoom killed the server mid-pass — that combination was re-run clean afterwards).
  - screenshots: `run/shots-scrolljack/fold10-{en,pt-br}-{1440x900,390x844}.png`, raw measurements in `run/shots-scrolljack/report.json` and `report-ptbr-390.json`.

## Divergences

1. **Also removed the dead `--cta-scale` from `globals.css`.** The brief scoped the deletion to the components. Once `hero.tsx` stops writing the property, nothing in `src`, `e2e`, `public`, `scripts` or `docs` sets it, so `.cta-motion`'s `scale(var(--cta-scale, 1))` and the identical call in the `cta-enter` keyframe were permanently `scale(1)`. Both removed — a no-op at runtime that stops the CSS from carrying a variable no component can set.
2. **Removed the `#sticky-button-target` wrapper and the ref wrapper around the send-in-seconds slot.** The brief made the first conditional on a clean grep; the grep is clean, and the `<div ref={sendInSecondsRef}>` in `LandingPageClient` existed only to feed the freeze handler. Both were plain class-less block elements around block / inline-block children, so the layout is unchanged.
3. **`pnpm typecheck` is not clean on this worktree, for reasons this PR cannot fix.** It reports exactly 6 `TS2307` "Cannot find module" errors and nothing else: `web-vitals` (`src/utils/web-vitals-shim.ts` x2, its test), `@capacitor/app-launcher` and `@capgo/capacitor-in-app-review` (`src/utils/app-review.ts`), `capacitor-native-settings` (`src/utils/native-settings.ts`). None are in a file this PR touches, and none are of any other error class. This is the documented symlinked-`node_modules` worktree artifact — the native-only packages are absent from the shared store the worktree links to. CI installs properly and should be clean.

4. **Browser verification proved fold-10 centring geometrically instead of by screenshot diff against `origin/dev`.** The review asked for a pixel diff of the send-in-seconds fold against `origin/dev`; everything else on that list was run as asked. The diff was replaced by a direct measurement in the same run — section centre X vs `.cta-motion` centre X, equal to 0.00px on all four route x width combinations — which answers the question exactly rather than by proxy and does not need a second `next dev` on `origin/dev`. This box cannot afford one: three sibling TASK-21788 sessions compile in parallel and earlyoom SIGTERM'd the single dev server four times during the pass (`next-server` badness ~1100 at VmRSS 2.1-2.5 GiB). Two host-level tweaks, neither a repo change, were needed to keep one server alive: `fs.inotify.max_user_watches` 92678 -> 524288 (the ENOSPC watcher errors were putting `next dev` into an endless config-changed restart loop) and a temporary 3 GB swapfile at `/swap-verify.img`, which should be removed once the parallel runs finish.

## Flag-off impact

The scroll-jack ran in both flag states, so this is the one landing change in the set that is visible with `pwa-sunset` off: the page no longer freezes at fold 10 and the CTA no longer grows. That is the intent of the ticket. Nothing else about the flag-off page changes — same DOM, same copy, same entrance animations, same sticky mobile bar.

### DIFF
--- a/src/components/LandingPage/LandingPageClient.tsx
+++ b/src/components/LandingPage/LandingPageClient.tsx
@@ -1,7 +1,7 @@
 'use client'
 
 import { useFooterVisibility } from '@/context/footerVisibility'
-import { Suspense, useEffect, useMemo, useState, useRef, useCallback, type ReactNode } from 'react'
+import { Suspense, useEffect, useMemo, useState, type ReactNode } from 'react'
 // Imported directly, not through the barrel: `export *` pulls every sibling
 // into this chunk, including dropLink's nine repeat: Infinity animations,
 // which the landing page never renders.
@@ -102,25 +102,6 @@ export function LandingPageClient({
     }, [migrationOn, deviceType, isDesktop, heroConfig.primaryCta, tMigration])
 
     const [buttonVisible, setButtonVisible] = useState(true)
-    const [isScrollFrozen, setIsScrollFrozen] = useState(false)
-    const [buttonScale, setButtonScale] = useState(1)
-    const [animationComplete, setAnimationComplete] = useState(false)
-    const [shrinkingPhase, setShrinkingPhase] = useState(false)
-    const [hasGrown, setHasGrown] = useState(false)
-    const sendInSecondsRef = useRef<HTMLDivElement>(null)
-    const frozenScrollY = useRef(0)
-    const virtualScrollY = useRef(0)
-    const touchStartY = useRef(0)
-
-    // Use refs to avoid re-attaching listeners on every state change
-    const isScrollFrozenRef = useRef(isScrollFrozen)
-    const animationCompleteRef = useRef(animationComplete)
-    const shrinkingPhaseRef = useRef(shrinkingPhase)
-    const hasGrownRef = useRef(hasGrown)
-    isScrollFrozenRef.current = isScrollFrozen
-    animationCompleteRef.current = animationComplete
-    shrinkingPhaseRef.current = shrinkingPhase
-    hasGrownRef.current = hasGrown
 
     useEffect(() => {
         if (isFooterVisible) {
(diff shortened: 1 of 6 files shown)


0xkkonrad:

PR #3012, opened 2026-09-07
feat(shhhhh): store hand-off instead of web signup when pwa-sunset is on (TASK-21788)

0xkkonrad:
## Summary

The `/shhhhh` door sends a signed-out visitor into `/setup` to create an account on the web. Once the `pwa-sunset` flag is on, that signup is the thing we are closing, so the door now hands the visitor to the app store instead: desktop opens the scan-to-download QR modal, phones deep-link their store, both carrying `/card` as the post-install destination so the app lands them on the door's real target.

Flag off, `/shhhhh` behaves exactly as it does on `dev`.

Part of the landing funnel work for TASK-21788 (PR 4 of 4).

## What changed

- `src/app/shhhhh/ShhhhhLandingPage.tsx`: `handleCTA` returns early while `isFetchingUser`, then captures `DOOR_TRY`, then queues any campaign tag, then `if (!user && interceptGuestCta({ dest: '/card' })) return`. Every door affordance (hero button, "or join the waitlist" link, §7 button, sticky mobile bar) calls `handleCTA`, so one branch covers all four, and both `/setup` routes below it (`/setup?redirect_uri=/card` and the campaign `/setup?step=signup`). The hook is asked for the guest impression with `trackImpressionWhenGuest: !isFetchingUser && !user && !isJoined`. The sticky mobile bar unmounts while `storeHandoffModal` is open, so it cannot paint over the QR sheet's own close / store buttons on a narrow desktop-mode phone.
- `src/hooks/useGuestStoreHandoff.tsx`: optional `surface` argument, default `MIGRATION_SURFACES.GUEST_FLOW`, labelling `openStore`, `ScanToDownloadModal` and the guest-impression event. The desktop branch now also builds the deferred payload at click time (`buildDeferredPayload(handoff.dest, handoff.invite)`) and hands it to the modal, so the phone that scans lands on the caller's `dest`.
- `src/components/Migration/ScanToDownloadModal.tsx` / `DownloadQR.tsx`: optional `payload` prop threaded to the QR URL (`${origin}/app?${payload}`; bare `${origin}/app` when absent). See divergence 5 — PR 1 owns this prop and supersedes it on merge.
- `src/constants/migration.consts.ts`: `LANDING_DOOR: 'landing_door'`.
- `src/app/shhhhh/ShhhhhLandingPage.test.tsx` (10 tests): surface + settled impression flag, no impression and no action during the pre-auth flash, campaign tag queued *before* the intercept, sticky bar hidden while the modal is open, hand-off from the door and from the waitlist link, campaign-tagged link, flag-off keeps `/setup?redirect_uri=%2Fcard`, signed-in visitor still joins the waitlist and is never intercepted.
- `src/hooks/__tests__/useGuestStoreHandoff.test.tsx` (9 tests): phone bounce carries the caller surface, default surface, desktop modal, handoff reaches the desktop QR payload, bare QR without a handoff, no interception inside Capacitor, no impression inside Capacitor, no interception with the flag off, impression event surface.

No new i18n strings, no new copy.

## How verified

- `pnpm test -- src/app/shhhhh/ShhhhhLandingPage.test.tsx src/hooks/__tests__/useGuestStoreHandoff.test.tsx` → 2 suites, 19 tests, all pass.
- `pnpm test -- src/components/Migration/__tests__/MigrationDownloadModal.test.tsx src/components/Invites/InvitesPage.test.tsx src/components/Profile/components/__tests__/PublicProfile.test.tsx` (the other `useGuestStoreHandoff` / migration consumers) → 3 suites, 53 tests, all pass.
- `pnpm lint --quiet` on the touched files → clean. `prettier --check` → clean.
- `pnpm typecheck` → 6 pre-existing `TS2307 Cannot find module` errors, none in a file this PR touches: `src/utils/web-vitals-shim.ts` (x3), `src/utils/__tests__/web-vitals-shim.test.ts`, `src/utils/app-review.ts` (x2), `src/utils/native-settings.ts`. The worktree's `node_modules` is a symlink to the shared checkout and `web-vitals`, `@capacitor/app-launcher`, `@capgo/capacitor-in-app-review` and `capacitor-native-settings` are physically absent from it (verified by `ls`). Environment gap, not a type error in source; the count is identical before and after this PR. CI runs a real install and should be clean.
- No dev server in this stage; the runtime check runs in the later stage of the workflow.

## Divergences

1. **The door does not call `openStore` itself.** The brief split the two branches between `useGuestStoreHandoff` (desktop) and a direct `openStore(store, LANDING_DOOR, {dest:'/card'})` (phone). The hook already owns both branches and hardcoded `GUEST_FLOW` in three places, so a direct call would duplicate the device split and leave the desktop modal reporting `guest_flow` while the phone reported `landing_door`. Instead the hook takes an optional `surface`, defaulting to the old value, and the door passes `LANDING_DOOR` once.
2. **`LANDING_DOOR` is added here as well as in PR 1.** PR 4 cannot compile without it and both branch off the same `origin/dev`. Same key, same value as the brief specifies. **For the merger: if PR 1 lands first, drop this one line on rebase — the key and the value are identical, there is nothing to reconcile.**
3. **The campaign-tagged door path is intercepted too, and the tag now rides the hand-off.** Verification checklist item 5 wants no `/setup` left on the door when the flag is on, and `/setup?step=signup` is the same closed web signup. `queuePendingBadgeCampaigns` is hoisted above the intercept (behaviour-preserving — the campaign branch already queued before its own `!user` check), so the tag is in the cookie queue that `buildDeferredPayload` reads and it rides the Android install referrer, the iOS clipboard hand-off and the desktop QR payload; `applyDeferredPayload` re-queues it inside the app. *Correction to the first round of this PR, which claimed the tag could not survive — it can, via the deferred-link mechanism built for exactly this.* It is still lost for a desktop visitor who ignores the QR and installs by hand from a store listing. Flag off, the campaign path is untouched.
4. **`pnpm typecheck` is not clean** — 6 pre-existing errors from the incomplete shared `node_modules`, itemized under "How verified".
5. **`payload` on `DownloadQR` / `ScanToDownloadModal` lands here, ahead of PR 1.** PR 1 owns this prop, but without it PR 4's desktop branch dropped `{dest:'/card'}` on the floor and the QR encoded a bare `/app`, so a desktop visitor who scanned landed on `/home` instead of the door's destination — the stated purpose of the PR. Same prop name and semantics as PR 1's spec. **PR-1 reconciliation: keep PR 1's version of `DownloadQR` (it also appends `&s=<surface>` and adds `size`) and drop this one on rebase.** The two other handoff-passing callers (`SendLinkActionList`, `SendWithPeanutCta`, both `{invite}`) get the same upgrade — their desktop QR now carries the invite code and the current page as `dest`, matching what `openStore` already does for them on phones. `InvitesPage` and `PublicProfile` pass no handoff and stay byte-identical.
6. **`MIGRATION_GUEST_CTA_SHOWN` for the door is gated on settled auth.** The brief says nothing about the impression event; every other migrated guest surface supplies `trackImpressionWhenGuest`, and without it `landing_door` would have hand-off clicks and no denominator. Gated with `isFetchingUser` (as in `SendWithPeanutCta` and `PublicProfile`) so the pre-auth flash never counts — and `handleCTA` returns early in that window, so a returning signed-in user who taps before auth resolves is not bounced to the store by a full-page navigation with no way back in-tab.

Full text: `run/DIVERGENCES-shhhhh.md`.

## Flag-off impact

None on `/shhhhh`: the only new behaviour is behind `interceptGuestCta`, which returns `false` whenever `useMigrationFlag()` is false or the app is running under Capacitor, and the routes below it are unchanged. The `isFetchingUser` early return applies in both flag states and is a strict improvement (a tap during the auth round-trip previously misrouted a signed-in user through `/setup`). `LANDING_DOOR` is a new constant that nothing reads while the flag is off.

Outside `/shhhhh`, also flag-gated: the `payload` prop is only ever non-empty on a path that `interceptGuestCta` already gated on the flag, so `SendLinkActionList` and `SendWithPeanutCta` see the richer QR only with the flag on.

### DIFF
--- a/src/app/shhhhh/ShhhhhLandingPage.test.tsx
+++ b/src/app/shhhhh/ShhhhhLandingPage.test.tsx
@@ -1,4 +1,4 @@
-import { fireEvent, screen, waitFor } from '@testing-library/react'
+import { cleanup, fireEvent, screen, waitFor } from '@testing-library/react'
 import type { ComponentProps, ReactNode } from 'react'
 import { renderWithIntl } from '@/test-utils/intl'
 import en from '@/i18n/app/messages/en.json'
@@ -10,8 +10,11 @@ const mockUseGuestStoreHandoff = jest.fn()
 const mockGetInfo = jest.fn()
 const mockJoinWaitlist = jest.fn()
 
-let mockAuth: { user: null | { user: { userId: string } }; fetchUser: jest.Mock }
+let mockAuth: { user: null | { user: { userId: string } }; isFetchingUser?: boolean; fetchUser: jest.Mock }
 let mockSearch = ''
+let mockStoreHandoffModal: ReactNode = <div data-testid="store-handoff" />
+const mockQueuePendingBadgeCampaigns = jest.fn((campaigns: readonly string[], _expiryDays?: number) => [...campaigns])
+const callOrder: string[] = []
 
 jest.mock('next/navigation', () => ({ useRouter: () => ({ push: mockPush }) }))
 jest.mock('@/context/authContext', () => ({ useAuth: () => mockAuth }))
@@ -22,11 +25,24 @@ jest.mock('@/hooks/useGuestStoreHandoff', () => ({
     useGuestStoreHandoff: (opts?: { surface?: string }) => {
         mockUseGuestStoreHandoff(opts)
         return {
-            interceptGuestCta: (handoff?: { dest?: string }) => mockInterceptGuestCta(handoff),
-            storeHandoffModal: <div data-testid="store-handoff" />,
+            interceptGuestCta: (handoff?: { dest?: string }) => {
+                callOrder.push('intercept')
+                return mockInterceptGuestCta(handoff)
+            },
+            storeHandoffModal: mockStoreHandoffModal,
         }
     },
 }))
+jest.mock('@/components/Invites/badge-campaign-context', () => {
+    const actual = jest.requireActual('@/components/Invites/badge-campaign-context')
+    return {
+        ...actual,
+        queuePendingBadgeCampaigns: (...args: [readonly string[], number?]) => {
+            callOrder.push('queue')
+            return mockQueuePendingBadgeCampaigns(...args)
+        },
+    }
+})
 jest.mock('@/services/badge-campaigns', () => ({
     claimAndSettlePendingBadgeCampaigns: jest.fn(async (campaigns: readonly string[]) => ({
         claims: campaigns.map((badgeCampaign) => ({ badgeCampaign, badgeCode: 'OTHER', outcome: 'awarded' })),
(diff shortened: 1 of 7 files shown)


0xkkonrad:

PR #3013, opened 2026-09-07
Split badge/avatar earn toast into two sequential toasts

innolope-dev:
## Summary
- Split the badge-unlock and avatar-unlock announcements into two separate toasts, the avatar one firing 500ms after the badge one, instead of stacking two links inside a single toast.
- Each toast's link now always sits on its own line (line break before it).
- Dropped the leading dash from "tap to view" copy and capitalized it ("Tap to view") across en / es-419 / es-AR / pt-BR.

## Test plan
- [x] `BadgeEarnToast.test.tsx` updated and passing (9/9), including new coverage for the 500ms delay between toasts
- [x] `tsc --noEmit` clean on touched files
- [x] `prettier --check` clean on touched files
- [ ] Manual check of the toast in a live session with a fresh badge that ships avatars (not verified live — needs a real earned-badge state)

### DIFF
--- a/src/components/Badges/BadgeEarnToast.tsx
+++ b/src/components/Badges/BadgeEarnToast.tsx
@@ -6,7 +6,9 @@
  * Globally mounted (ClientProviders), self-contained. When the signed-in user
  * lands on /home with freshly-earned badges they haven't seen, it fires ONE
  * coalesced toast ("Badge unlocked: X" / "You unlocked N badges") that taps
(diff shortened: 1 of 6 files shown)


innolope-dev:
Fixed in e9432442e — `avatarTimeoutsRef` is now a `Set` so each batch's delayed avatar toast is tracked and cancelable independently, instead of overwriting a single ref.

PR #3049, opened 2026-09-08
fix(android): stop support and QR camera preflight crashes

innolope-dev:
## Summary

- remove the unrelated camera-permission preflight from native Support so chat can always open
- bypass the crashing Capacitor Camera permission API on Android and let the WebView's real `getUserMedia` request own the QR-scanner prompt
- keep the existing iOS permission preflight and SupportDrawer cancellation/snapshot semantics

## Root cause

[Sentry PEANUT-UI-T30](https://peanut-c34d84c05.sentry.io/issues/7719621109/?project=4505827431415808) shows a fatal Android `RuntimeException` / `NullPointerException` in `CameraPlugin.getPermissionStates` on release `1.5.0`. Tapping Support called `Camera.checkPermissions()` before opening Crisp even though support does not need the camera. The same shared preflight was also used by the QR scanner.

The crash became visible in the optimized 1.5.0 Android shell after R8/minification was enabled. Because it happens in native code before the promise can reject to JavaScript, the existing catch cannot recover it.

## Validation

- `pnpm exec next typegen`
- `pnpm typecheck`
- 53 focused SupportDrawer, camera-permission, and QR-scanner tests
- targeted ESLint and Prettier checks
- production native static export
- `node scripts/native-fingerprint.mjs --diff v1.5.0` — native surface unchanged; OTA-safe

### DIFF
--- a/src/components/Global/SupportDrawer/__tests__/SupportDrawer.test.tsx
+++ b/src/components/Global/SupportDrawer/__tests__/SupportDrawer.test.tsx
@@ -20,6 +20,7 @@ import SupportDrawer from '../index'
 import { isCapacitor } from '@/utils/capacitor'
 import { SUPPORT_EMAIL } from '@/constants/crisp'
 import { dispatchBackPress, resetBackHandlersForTests } from '@/utils/back-handler'
+import { ensureNativeCameraPermission } from '@/utils/camera-permission'
 
 const render = (ui: Parameters<typeof rtlRender>[0]) => rtlRender(ui, { wrapper: IntlWrapper })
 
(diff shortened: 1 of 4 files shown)


innolope-dev:
Follow-up on the Chip finding: no additional Support change is needed.

We inspected the exact Android artifact shipped by this app (`@capgo/capacitor-crisp 8.0.27` → `im.crisp:crisp-sdk:2.0.19`), rather than relying on the historical source comment. Crisp's attachment handler:

- checks whether `android.permission.CAMERA` is declared;
- checks the current grant and launches `ActivityResultContracts.RequestMultiplePermissions` for CAMERA when it is not granted;
- starts the camera only from the successful grant callback.

That flow runs when the user chooses **Take a photo**. It is independent of the Capacitor `CameraPlugin.getPermissionStates` R8 failure that caused Support itself to close. Restoring the app-level `Camera.checkPermissions()` preflight would reintroduce the original crash in the 1.5.0 native shell and would prompt for an optional capability before chat opens.

We also manually validated the OTA on-device: Support opens and taking a photo from Crisp succeeds. The planned native R8 follow-up will preserve Capacitor permission annotations for QR/plugin calls; it does not need to change Crisp or put a camera preflight back into Support.

If this ever regresses independently, the correct reproduction is revoke CAMERA → open Support → choose Take a photo, then inspect the Crisp-side stack. For this PR, the finding is not applicable.

PR #3057, opened 2026-09-08
fix: Back navigation loop in Send to friend wallet flow (TASK-22424)

abalinda:
## Summary

Back in Send → Send to friend → Exchange or wallet (or bank) trapped users in a loop. The send-origin back handlers in `withdraw/page.tsx` pushed `/send` on top of the retained `/withdraw?method=…` history entry. Send's safe-back then popped straight back into the withdraw step, and `?method=crypto` re-selected crypto — Back never reached Home. One production reporter with a deterministic video; the monitor sweep confirmed the trace.

Fix: use the shared `useSafeBack('/send', { replace: true })` in both send-origin back handlers (crypto amount step, bank method-selection step). With in-app history it pops the withdraw entry off the stack; on a cold deep link it replaces to `/send` instead of minting an entry that Send's own safe-back would walk back into. Same pattern SendRouter already uses for its link/contacts subviews. `?method=…` stays on the URL — `useSendFlowOrigin` needs the marker to survive every hop.

## Task

[TASK-22424 — Fix Back navigation loop in Send to friend wallet flow](https://app.notion.com/p/3d583811757981999640edfb2abd3518) (P2, PRODUCTION). [TASK-22420](https://app.notion.com/p/Stop-Back-navigation-from-looping-between-Send-and-crypto-withdrawal-amount-3d583811757981fa86e8c7e326fbf71e) is the monitor's card for the same defect — close as duplicate when this merges.

## Risks / breaking changes

- FE only, no cross-repo impact. Blast radius is the two send-origin back handlers; non-send withdraw back paths are untouched.
- Base is `main` (production bug) — creates main→dev back-merge debt.

## QA

- 5 regression tests fail on the old code and pass on the fix (in-app pop + cold-deep-link replace, for both crypto and bank paths).
- Full local gate green: prettier, typecheck, all 347 jest suites (4270 tests), production build.
- Manual path to verify: Home → Send → Send to friend → Exchange or wallet → Amount → Back → Send → Back → Home. Also cold-load `/withdraw?method=crypto` → Back → Send → Back → Home.

Screenshots: N/A (no visible change — screens render the same; only where Back lands changes).

- Send and withdraw step state stays out of the URL and off nuqs (both files parse query params by hand, against the repo's URL-as-state rule). Pre-existing, out of this PR's blast radius — noted as a follow-up refactor opportunity, not done here.

### DIFF
--- a/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
+++ b/src/app/(mobile-ui)/add-money/[country]/bank/page.tsx
@@ -46,7 +46,7 @@ import { ANALYTICS_EVENTS } from '@/constants/analytics.consts'
 import { addMoneyCountryUrl, rewriteMethodPath } from '@/utils/native-routes'
 import { isMantecaSupportedCountryCode } from '@/constants/manteca.consts'
 import { useSafeBack } from '@/hooks/useSafeBack'
-import { getRegionIntent } from '@/utils/regions.utils'
+import { getBankRegionIntent } from '@/utils/regions.utils'
 import { useLocale, useTranslations } from 'next-intl'
 import { localizedCountryTitle } from '@/utils/country-name.utils'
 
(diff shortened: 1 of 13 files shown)


abalinda:
Superseded by #3059 — same fix rebuilt on `dev`. This branch was cut from `main`, so after repointing the base to `dev` the history carried 13 main-only commits and conflicted.

PR #3077, opened 2026-09-09
fix(kyc): TASK-22471 WebSDK fallback when the native Sumsub SDK cannot start

kushagrasarathe:
## TASK-22471 · P1 · Android native KYC broken on Play binary 1.5.0+21653381

**What**: when the native Sumsub SDK cannot start (plugin missing from the binary, init throw, launch rejection), render the WebSDK in the WebView with the same session token instead of a dead-end error modal.

**Why now**: the current Play binary shipped without the Cordova plugin's native class ("Class not found" at `SumsubNativeSdk.tsx` launch). Every Android user on it — 5+ real users and climbing since the 100% pwa-sunset flip — cannot KYC at all (Sentry [PEANUT-UI-T33](https://peanut-c34d84c05.sentry.io/issues/PEANUT-UI-T33)). This change is OTA-able: merge → production OTA dispatch → installed users heal on next launch, no Play review.

**Why the fallback is safe**:
- The wrapper's own design note records that "the WebSDK does run inside the Capacitor WebView" — native was preferred only because a Sumsub-side iframe init failure is silent. As a fallback that runs *only after native failed*, the trade-off inverts: silent-failure risk applies only to users who otherwise could not verify at all, and the web driver has its own load-error UI.
- Camera: `AndroidManifest.xml` declares `CAMERA`; there is no custom `WebChromeClient`/`onPermissionRequest` override, so Capacitor's default bridge grants WebView `getUserMedia` when the app holds the permission.
- Native stays primary: iOS and a future fixed binary never see the fallback (`does not render the fallback while the native SDK is up` test).
- `onError` is not fired on fallback — recovery isn't a terminal error; the web driver still calls it for its own failures. No current caller passes `onError`.

**Telemetry**: every native failure still captures to Sentry + PostHog (`kyc_sdk_init_failed`), plus new `kyc_web_fallback_used` so fallback usage is measurable (and alertable when the fixed binary should have made it zero).

**Verification**: Kyc suites 26/26 green; full jest: only 2 pre-existing timezone-flaky suites fail locally (IST machine; date-string assertions — CI/UTC green on base). Typecheck via CI (fresh-worktree svg-types env issue locally, known phantom class).

**Follow-ups (in TASK-22471, not this PR)**: rebuild the Android binary with the plugin synced (`release-native.yml` run 34240637046 shows no cordova/cap-sync lines — the packaging root cause), and add a build assertion that the APK contains the Sumsub native class.

### DIFF
--- a/src/components/Kyc/SumsubNativeSdk.tsx
+++ b/src/components/Kyc/SumsubNativeSdk.tsx
@@ -1,13 +1,12 @@
 'use client'
 
 import { useEffect, useRef, useState } from 'react'
-import { useLocale, useTranslations } from 'next-intl'
+import { useLocale } from 'next-intl'
 import * as Sentry from '@sentry/nextjs'
 import posthog from 'posthog-js'
-import Modal from '@/components/Global/Modal'
 import { ANALYTICS_EVENTS } from '@/constants/analytics.consts'
 import { toSumsubLocale } from '@/i18n/app/sumsub-locale'
-import { SumsubSdkErrorView } from './SumsubSdkErrorView'
+import { SumsubWebSdkModal } from './SumsubWebSdkModal'
 import type { SumsubSdkProps } from './sumsubSdk.types'
 
 /**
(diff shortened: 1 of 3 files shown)


kushagrasarathe:
Closing per Kushagra's decision: no web fallback — the fix ships as a corrected Android binary through Play instead (packaging fix + APK assertion, tracked on TASK-22471). Keeping the branch; reopen if the binary path stalls and a stopgap is needed.

PR #3098, opened 2026-09-10
fix(support): refresh unread badge on native push

abalinda:
## Summary

- refresh the existing support-unread badge when a push arrives while the native app is already foregrounded
- refresh notification consumers when a native app resumes, without relying on WebView `visibilitychange`
- keep the backend unread count as the source of truth; this adds event triggers, not client-side guessing or polling
- cover both native lifecycle and foreground-push paths with regression tests

## Regression postmortem

**Impact:** A support reply could produce the expected OS push notification while the pink unread indicator remained absent in the running app. Fully reopening the app caused the indicator to appear because the badge hook re-fetched on mount.

**What regressed:** The original badge implementation listened to browser `visibilitychange`. That worked in the PWA/browser environment, but Android WebViews do not reliably emit that event when the native app resumes. Separately, our OneSignal adapter exposed notification clicks but not foreground delivery, so a push received while the app stayed open produced no refresh signal at all.

**Why the navigation redesign was not the cause:** The unread hook, server-backed `support` count, pink `IndicatorDot`, and Support-tab rendering all survived the BottomNav migration. The missing link was between native notification/lifecycle events and the hook's existing `notifications:updated` refresh event.

**Why tests missed it:** Coverage proved that `useSupportUnread` responds to `notifications:updated`, and that OneSignal click events route correctly, but no test connected native resume or foreground push delivery to that event.

**Fix and prevention:** Both OneSignal adapters now expose foreground delivery. Native app wiring emits `notifications:updated` on foreground delivery and on active resume. Regression tests pin both paths, while the existing unread-count request still decides whether the support dot is actually shown.

## Verification

- 37 focused Jest tests passed across support unread, native app links, OneSignal adapter, and notification hooks
- changed-file ESLint passed
- Prettier and `git diff --check` passed
- full local typecheck reached pre-existing missing static-asset module errors in this worktree; no changed-file type error remained

Requested by @Hugo0 — https://discord.com/channels/972435984954302464/1547180308707676241/1547180650149183519

### DIFF
--- a/src/components/Global/SupportDrawer/index.tsx
+++ b/src/components/Global/SupportDrawer/index.tsx
@@ -18,6 +18,7 @@ import {
 } from '@/constants/crisp'
 import type { AppLocale } from '@/i18n/app/config'
 import { notificationsApi } from '@/services/notifications'
+import { notifyNotificationsUpdated } from '@/utils/notifications-events'
 import { isCapacitor } from '@/utils/capacitor'
 import { ensureNativeCrispConfigured, nativeCrispFields } from '@/utils/crisp'
 
@@ -137,7 +138,7 @@ const SupportDrawer = () => {
         if (!isLoggedIn) return
         notificationsApi
             .markAllRead('support')
-            .then(() => window.dispatchEvent(new CustomEvent('notifications:updated')))
+            .then(notifyNotificationsUpdated)
             // A failed mark-read only means the badge stays on a bit longer.
             .catch(() => {})
     }, [isLoggedIn])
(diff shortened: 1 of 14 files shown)


abalinda:
Byte-identical reopen of #3068 (same head `85cf13f44`, same base `dev`). Prior review context lives there: Chip's MAJOR (foreground refresh racing the unread-row commit) was fixed in `85cf13f44` via a bounded 2.5s recheck, and the review-fix summary is at https://github.com/peanutprotocol/peanut-ui/pull/3068#issuecomment-5620838514. One manual QA ask before merge: on-device check that foreground push banners still display (first `foregroundWillDisplay` listener changes the plugin's display path).

abalinda:
Closed in favor of #3099 — same change (byte-identical patch) rebuilt as fresh human-authored history on `fix/support-unread-native-refresh`.

PR #3102, opened 2026-09-10
PWA sunset sweep: native apps replace the PWA (ui-pwa)

0xkkonrad:
Peanut is on the App Store and Google Play now, and the web app is being retired, so the copy and metadata that still sold Peanut as a PWA are wrong. This fixes the sentences that make a distribution claim: the site-wide JSON-LD (WebApplication → MobileApplication with both store URLs), llms.txt/llms-full.txt, the README badges, the migration download prompt in all 5 locale files, and the merchant landing CTA ("INSTALL PEANUT" → "GET PEANUT").

Also collapses the two competing Play Store constants into `STORE_URL.android` so marketing surfaces and the deferred-deeplink referrer cannot drift, points the iOS `related_applications` entry at the real listing instead of peanut.me, drops the `us` storefront segment from the App Store URLs, and deletes the leftover `app.acme.com` template rewrite in vercel.json that was silently aimed at the real `/app` route.

Rulings applied:
- Store URLs: https://apps.apple.com/app/id6786373552 and https://play.google.com/store/apps/details?id=me.peanut.wallet, both verified live. No storefront segment, so Apple resolves to the visitor's own country.
- Markets: "in select markets" / "where available". No country lists, no global-availability claim.
- App name: "Peanut" in prose. Min-OS versions stay out of marketing copy.
- Install code untouched: InstallPWA setup view, IosPwaInstallModal, useBravePWAInstallState, usePWAStatus, the webmanifest, the Serwist service worker, the pwa-sunset flag, MIGRATION_CUTOVER_DATE and /app are all owned by the migration mechanism and are shown only to users still on the web app. Only the placeholder store URL constants moved. `IOS_APP_STORE_LISTING_LIVE` stays `false`, byte-identical to dev — no behaviour change to the store-update prompt.
- No cutover date anywhere. Peanut Split stays a PWA and is not touched.
- Card held: nothing card-related is in this branch. Card copy is prepared on the separate card branch as a held draft, because the waitlist is still wired in production.

## Verification

Two independent verifiers re-checked the fix commits; both pass, all four earlier findings resolved.

Gate re-check:

> `git diff 20fd5190c..HEAD -- src/constants/migration.consts.ts` shows `export const IOS_APP_STORE_LISTING_LIVE = false` — identical to base/origin/dev, so `storeUpdateOffered` at src/components/Profile/index.tsx:47 behaves exactly as on dev. Profile/index.tsx is untouched on this branch. […] `git diff 20fd5190c..HEAD -- public/.well-known/apple-app-site-association` produces NO output (exit 0): byte-identical to base. Both legacy appID blocks and web.app.peanut in webcredentials/activitycontinuation are restored.

Facts re-check:

> Line 57 now `export const IOS_APP_STORE_LISTING_LIVE = false`, identical to origin/dev […] The replacement comment is factually accurate (listing is live; the stale "store URL 404s" claim is gone) and does not re-assert the flip. […] Finding 3 (layout.tsx JSON-LD) RESOLVED. Line 75 is `name: 'Peanut'`. No card wording on a site-wide surface. The rest of the node (`'@type': 'MobileApplication'`, `operatingSystem: 'iOS, Android'`, `downloadUrl: [STORE_URL.ios, STORE_URL.android]`) is unchanged and correct.

Two nits left open, neither blocking:
- The non-English locales render "in select markets" as "en algunos países" / "em alguns países". No country is named so the MARKETS ruling holds, but the register drifts toward country-by-country. "en algunos mercados" / "em alguns mercados" would match English better — cosmetic.
- `src/app/(mobile-ui)/dev/home-ctas/page.tsx:161` still hardcodes the old "add to home screen / no download needed" carousel copy. Skipped deliberately: it mirrors `en.json:233-234`, which the INSTALL-CODE ruling preserves, and the route is dev-only. Delete both together when the migration owner retires the install paths.

One correction to an earlier note: dev has already added the `/app` universal-link paths to `apple-app-site-association`. This branch does not touch that file, so dev's version survives; only the team-id/legacy-appID pruning is still outstanding, and that belongs to the native-release chain.


0xkkonrad:
Closing: Kushagra is covering the same copy/constants change in his native-app work. The verified diff stays on branch sunset/pwa-native if any of it is useful.

PR #3105, opened 2026-09-10
fix(native): drop the iOS-only cross-chain withdraw gate (TASK-22250)

abalinda:
## Summary

The iOS app forced the withdraw token selector to USDC on Arbitrum, so iOS users could not pick any of the cross-chain destinations that web and Android have offered since the Rhino cross-chain withdrawal flow shipped. `disableXchainWithdraw` ORed the cross-platform ops kill-switch with `isIOSNative()` — a gate added for one App Store review in August (4564b401) that outlived the review.

The switch is now one plain value that every platform reads. Flipping it to `true` is still the whole emergency kill-switch.

## Task

[TASK-22250 — Enable cross-chain withdrawals in the native app](https://app.notion.com/p/3d0838117579819ea950d71c9d1ef493)

## What is NOT changed

- **The ops kill-switch** — same property, same consumer (`TokenSelector`), same effect: every platform locks to USDC on Arbitrum while it is on.
- **The Rhino destination allowlist** — `RHINO_WITHDRAW_SUPPORTED_TOKENS_BY_CHAIN` still decides which chains the withdraw selector offers.
- **The per-chain PostHog rollout flags** — `useChainRollout` still gates each destination; they were never part of this gate.
- **Web and Android** — they never read the iOS branch, so their behaviour is byte-identical.

## Risk

Low, and confined to the iOS app. iOS users can now reach the same cross-chain withdrawal paths Android and web already use — the same code, the same Rhino allowlist, the same rollout flags, the same per-chain minimums and fee checks at review time. Nothing else reads the removed condition (`disableXchainWithdraw` has one consumer). No backend change, so no deploy ordering.

The App Store consideration that motivated the original gate was review-time only; it does not apply to the shipped build.

## QA

1. In the iOS app, Withdraw → Exchange or Wallet → amount → the destination selector lists every rolled-out Rhino destination, not just USDC on Arbitrum.
2. Pick a non-Arbitrum destination, enter an address, Review — the route preview, fee and confirmation behave as they do on Android.
3. Android and web: unchanged.
4. Kill-switch: set `disableXchainWithdraw: true` in `underMaintenance.config.ts` — all three platforms fall back to USDC on Arbitrum with the info banner.

## Tests

`src/config/__tests__/ios-store-gating.test.ts` is replaced by `xchain-withdraw-gate.test.ts`, which reads the flag through the **real** platform detection on web, iOS native and Android native and requires the same answer — so a new platform condition fails a test instead of shipping in a store build.

Route preview → confirmation already has platform-agnostic coverage (`withdraw/crypto/__tests__/crypto-withdraw-confirm.test.tsx`, `features/withdraw/views/__tests__/ConfirmWithdrawView.test.tsx`); with the gate gone, that coverage now describes iOS too.

## Screenshots

The withdraw destination selector, captured at 375×667 with the Capacitor bridge reporting **iOS** — the only platform this PR changes.

| Before — iOS locked to one destination | After — the same list web and Android already get |
|---|---|
|  |  |
| "Cross-chain transactions are temporarily unavailable. You can use USDC on Arbitrum." · one available token | Network row (ARB / ETH / OP / BASE / more), token search, the full popular-token list |

Captured against the `withdraw-amount` fixture, so no backend was involved. The assets live on the `pr-assets-3105` branch and it is deleted after merge.

## Follow-up this creates (separate, not in this PR)

Chip's third reviewer found it and it is right: mono's App Store stance pack still names this switch as the iOS mitigation.

- `ops/legal-apple-stance/README.md` §5 recommends *"flip the swap kill-switch for iOS"* — after this PR that lever no longer exists per-platform; the switch is all-or-nothing.
- `ops/legal-apple-stance/02-exchange-surface-inventory.md` row 2 cites the same switch when inventorying what Apple could call an exchange service.

Neither is a code change and neither belongs in this PR. Dilemma 5 is still awaiting Konrad's pick, and whether the mitigation is being withdrawn is his call, not a doc edit I can make — but the pack must be updated before the next iOS submission so the Resolution Center reply is not drafted against a mitigation the build no longer has.

## Design notes

`disableXchainWithdraw` was a getter because `isIOSNative()` can only be read at render time. With the platform condition gone there is nothing to defer, so it collapses to a plain literal like every sibling flag in the file — one less thing to explain, and the ops switch is one line to flip.


abalinda:
Advisory MINOR from `claude-opus` — acknowledged, and it is right. `ops/legal-apple-stance/README.md` §5 recommends flipping this switch for iOS, and `02-exchange-surface-inventory.md` row 2 cites it; after this PR the lever is no longer per-platform.

Not fixed here on purpose: it is a mono docs change, and content and code never travel in one PR. It is also not a doc edit I can make on my own — dilemma 5 is still awaiting Konrad's pick, and whether that mitigation is being withdrawn is his ruling, not mine. Recorded in the PR body under "Follow-up this creates" so it is updated before the next iOS submission.

PR #3146, opened 2026-09-12
fix(claim): show the loading message for the whole loading state

### LINKED ISSUE #2759: Show the initial loading message while Claim fetches a send link
## Summary
The Claim loading state does not show the initial loading message during the first send-link request.

## Required change
Update `src/components/Claim/Claim.tsx` so the Claim component renders `loading.loadingYourLink` whenever `isSendLinkLoading` is `true`, including when `failureCount` is `0`.

Keep the existing failure-count translation behavior:
- Use `loading.loadingYourLink` for the initial request and early retries.
- Use `loading.takingLonger` after repeated failures.
- Keep the existing `loading.stillTrying` behavior for later retries.

## Rationale
The current `failureCount > 0` condition hides the primary loading message for the initial request. Users only see the loading mascot during this state.

## Affected area
- `src/components/Claim/Claim.tsx`
- Claim send-link loading state

## Acceptance criteria
- When `isSendLinkLoading === true` and `failureCount === 0`, the UI shows `loading.loadingYourLink`.
- When `isSendLinkLoading === true` and `failureCount` is greater than `0` but less than `3`, the UI shows `loading.loadingYourLink`.
- When `isSendLinkLoading === true` and `failureCount >= 3`, the UI shows `loading.takingLonger` and retains the existing `loading.stillTrying` message.
- The change does not modify the restyle-only scope of PR #2758.

## Backlinks
- PR: https://github.com/peanutprotocol/peanut-ui/pull/2758
- Review comment: https://github.com/peanutprotocol/peanut-ui/pull/2758#discussion_r3819683846
- Requested by: @kushagrasarathe

mariano-aguero:
## Summary

- the claim loading copy now covers the whole `LOADING` state, not just the window where the send-link query reports `isLoading`
- `ClaimLoadingView` drops its `isSendLinkLoading` prop. It only ever renders under `linkState === LOADING`, so the flag was a second and shorter gate on top of that one
- `failureCount` still picks the copy: `loading.loadingYourLink` up to the third failure, `loading.takingLonger` from there on, with `loading.stillTrying` next to it
- adds the first test for this view, one case per acceptance criterion on the issue

## Why

Closes #2759.

The `failureCount > 0` guard hid the primary message during the initial request, so a recipient opening a link saw the mascot with nothing under it.

Dropping that guard on its own would have left a second gap. `isSendLinkLoading` is React Query's `isLoading` and goes false as soon as `sendLinksApi.get` resolves, but `processLink` still awaits the user fetch, `fetchTokenDetails` and, for anything that is not a stablecoin, `fetchTokenPrice` before it sets `linkState` to `CLAIM`. The copy would have appeared, vanished for the length of those calls, and only then given way to the claim view. Gating on the view being mounted closes the original bug and that flash together.

`useClaimFlow` no longer returns `isSendLinkLoading`. Nothing else read it.

## Verification

- `pnpm exec jest src/components/Claim`: 94 tests in 20 suites passed
- the first-request case fails against the old condition, so the suite pins the behavior instead of restating it
- `pnpm exec eslint src/components/Claim`: 0 errors, and the same 9 pre-existing `exhaustive-deps` warnings as before the change
- `pnpm exec prettier --check` on the four changed files: clean
- `pnpm typecheck`: no new errors beyond the 5 that already fail on `dev` from the content submodule
- no manual QA on a deployed environment

The issue is assigned to @kushagrasarathe. Happy to close this if it is already in progress on your side.


mariano-aguero:
Closing this. A pull request from a fork cannot get a green CI run in this repo, so the branch has no way to reach review.
Happy to reopen it if fork PRs become viable here.

PR #3197, opened 2026-09-15
refactor: TASK-22665 CurrencySelect legacy palette -> DS tokens

kushagrasarathe:
## Summary

Token migration of `LandingPage/CurrencySelect` (rendered inside the product app via `Global/ExchangeRateWidget` → profile/exchange-rate) — part of the landing/marketing DS migration. Panel border/bg, coming-soon and selected rows, and text move to semantic tokens at the same rendered hex; the currency code de-stacks `text-base font-bold` → `text-body-m-semibold`. Behavior untouched, zero prop/consumer changes.

## Task

TASK-22665 — [Landing/marketing DS migration](https://app.notion.com/p/3dc8381175798124a987cf7756de3a79)

## Flagged: the HeadlessUI → radix swap is deliberately NOT here (needs a decision)

The migration checklist orders this component off HeadlessUI onto radix. Evidence gathered against it, verbatim from the attempt:

- `BaseSelect` cannot reproduce the UI: consumer-supplied trigger node, flag `<Image>` per option, coming-soon rows with `StatusBadge` — BaseSelect renders `{label, value}` strings only
- raw radix Select force-mounts `react-remove-scroll` (modal scroll-lock, no opt-out): the pinned pull-to-refresh test crashes in its touch handler, and in real browsers body scroll locks while open — today's HeadlessUI Popover is non-modal by design
- the faithful non-modal equivalent (`@radix-ui/react-popover`) is not a dependency — adding one is a decision, not a migration
- the /dev/ds audit already plans to fold CurrencySelect into a future `Global/Combobox` — that rebuild is the right vehicle for the headless swap

Also noted: `border-foreground-secondary` on the selected row — no `border-*` token carries #5f646d; the color token resolves and keeps the rendered color exact (inline comment explains).

## Risks / breaking changes

Renders inside the product app — but the diff is 6 class swaps at exact/nudge hexes, all verified against `@theme`. The 16px code line-height 24→20 shrinks the option row ~4px (token value).

## QA

typecheck, prettier green; 4 jest suites / 27 tests green (CurrencySelect pull-to-refresh test byte-identical, ExchangeRateWidget suites); zero legacy classes remain in the file.


kushagrasarathe:
Note for review: the red `publish / publish` check is the Screen-library screenshot-publish step failing with an R2 signature error (SignatureDoesNotMatch against the screenshots-library bucket). It is repo infra, fails identically on retry, references entries from unrelated PRs, and is advisory — it gates nothing in this PR. Tracked separately.

PR #3259, opened 2026-09-18
fix(deposit-accounts): settled-screen back controls exit the flow (VA QA)

Hugo0:
## Summary

Two navigation bugs Kushagra found while QA-ing the deposit-accounts (virtual accounts) feature. Both let the user step the wrong way out of a settled screen. Nav-only — no money or state logic changes.

**Bug 1 — post-deposit top-up screens back into the amount field.** After a bank/QR top-up is created, the details/QR screen's back/close returned to `?step=inputAmount`. The user landed on the amount field again and could start a second deposit. Three call sites, one uniform fix: each settled screen's back now uses the flow's outermost `onBack` (the same `useSafeBack` the amount step uses), so it leaves the flow to where the user came from.
- `MantecaAddMoney` PIX QR screen → back uses the flow `onBack`. Done is unchanged (`router.replace('/home')` — the terminal exit that stops device/browser back from popping into the finished deposit whose `step=showQR` would mint a new one).
- `MantecaAddMoney` ARS/CVU details screen → back uses the flow `onBack`.
- `BridgeBankOnrampView` bank-details screen → back uses the flow `onBack`. Removed the now-unused `setUrlState` from the destructure.

Both flows keep `history: 'replace'` for step params, so `useSafeBack` → `router.back()` leaves the whole flow rather than stepping to `inputAmount`. Using `useSafeBack` (not a hardcoded route) also satisfies the repo's back-control lint rule (no bare `router.push/replace` as `onBack`, ui#1997).

**Bug 2 — post-claim success screen back goes forward.** `AccountOpenedScreen`'s back chevron was wired to `onContinue`, the same handler as the Continue CTA (`setParams({ step: 'details' })`), so the back control moved the user forward into the details screen. It is a terminal success screen, so the back control is now hidden (`hideBackBtn`). Continue is the only forward move.

## Task

VA QA follow-up — no dedicated task yet. Relates to the deposit-accounts feature (branch `deposit-accounts-22589`, ui#3134). Assignee/task link is Hugo's call.

## Risks / breaking changes

- Bug 1 changes navigation in the money **top-up** flow (Manteca AR/BR, Bridge SEPA/GBP). **Nav only — no money logic, no amount/deposit-creation changes.** The deposit is already created when these screens render; only the exit destination changes.
- No cross-repo impact. No API/contract changes.

## QA

- **Bug 2 (success screen):** open a deposit account (EUR), reach the "account opened" success screen, confirm there is no back chevron; Continue still opens the account details.
- **Bug 1 (Manteca PIX/BRL):** Brazil top-up → reach the PIX QR screen → press back → leaves the flow, not the amount field. Done still goes home.
- **Bug 1 (Manteca CVU/ARS):** Argentina top-up → reach the deposit-details screen → press back → leaves the flow, not to the amount field.
- **Bug 1 (Bridge bank):** a SEPA/GBP country top-up → reach the bank-details screen → press back → leaves the flow, not to the amount field.

Automated: `AccountOpenedScreen.test.tsx` asserts the back control is absent; `MantecaAddMoney.exits.test.tsx` asserts both PIX and CVU back leave the flow (not `inputAmount`); `BridgeBankOnrampView.exit.test.tsx` asserts the bank-details back leaves the flow.

## Screenshots

N/A — nav-only. No rendered-output change except the removed back chevron on the terminal success screen (Bug 2), described above.


Hugo0:

PR #3265, opened 2026-09-18
feat(withdraw): currency-first cash-out selector + trimmed destination form

Hugo0:
## Summary

Rework the withdraw / cash-out entry to lead with **currency**, not country, and cut the destination form to essentials — the Aleks/Slava feedback that modern neobank users think in currency, not their bank's country of incorporation.

Four changes, each independently reversible:

1. **Currency-first selector.** The withdraw flow led with a full country list (`CountryList`). It now leads with the currencies you can cash out in — **EUR, GBP, USD, MXN, and local ARS / BRL / COP** — mirroring the add-money hub (`DepositAccountsListScreen`). Country is demoted to a secondary step: an **"Other countries"** fallback that reveals the full country list, plus an inline country list for a shared currency (EUR across SEPA) where disambiguation is actually needed. Single-country currencies route straight through.
2. **Trimmed the SEPA form.** BIC is derivable from the IBAN, so it is now fetched in the background and **hidden once resolved** — the common SEPA path asks only for **name + IBAN** (+ the Bridge-required beneficiary address). The field reappears, optional, only when derivation fails. Everything a rail actually needs to route money stays (routing number, sort code, CLABE, Colombia doc fields, address).
3. **Tap-to-paste.** Each bank field gets a paste-from-clipboard icon, reusing the existing `readClipboard` + `extractPaymentValue` smart-paste path (IBAN/routing/sort-code aware).
4. **Terminology.** The two lead headings move from "withdraw" to **"cash out"** (English only).

## Task

VA withdraw rework (Aleks/Slava feedback). 

- **Shared selector not extracted (deferred).** The add-money hub is coupled to VA deposit-accounts state (corridors / gates / claimable) that withdraw has no equivalent for, so a single shared component is too big to land safely. Withdraw now matches the add-money *pattern* and **reuses** the shared primitives (`CountryList`, `countryCurrencyMapping`, DS `ListItem`/`Section`, `liveRailsForCountry`). `CountryList` gained one additive optional prop, `currencyFilter`. Full extraction is a follow-up.
- **Routing kept in one place.** All withdraw routing stays in `WithdrawMethodView.handleCountrySelected`, shared by the currency list and its country fallback, so the two can never disagree (no second source of truth).
- **Only BIC was trimmed.** The API (`POST /users/accounts`) only strictly requires `accountType/accountNumber/countryCode/countryName/accountOwnerType/accountOwnerName`; BIC/routing/sortCode/address are API-optional. But `bank-corridors.ts` marks SEPA/US/GB/MX `needsAddress: true` (Bridge requires the beneficiary address) and routing/sortCode/CLABE are the rail's minimal identifier — removing them would break payouts. BIC is the one field that is both API-optional and auto-derivable, so it is the only safe trim.

## Risks / breaking changes

- **Withdraw entry UI changes** (money-out flow). No API/contract change, no new data sent to any third party (BIC still sent when present). No cross-repo change.
- **EUR now routes via a country the user picks in the disambiguation list** (unchanged behaviour — SEPA already routes by IBAN, the selected country is cosmetic per `DynamicBankAccountForm`).
- **Terminology change is English-only**; pt-BR / es-* still read "withdraw/retirar". Translating "cash out" is a content follow-up.

## Decisions for Hugo (picked the simplest coherent option)

1. **Chosen term: "Cash out"** (already used in the product, e.g. the withdraw layout description and `cashoutLimit`). Applied only to the two lead flow headings to stay minimal/reversible — nav label, `bankForm` heading, and localized strings left as-is. Say the word if you want the full sweep.
2. **BIC made optional + hidden-when-derived** rather than deleted outright, so a rare bank needing a manual BIC is still reachable. If you'd rather drop it entirely, it's a one-line change.
3. **Paste icon on every non-char-count field** (skips the address lines, whose trailing slot shows the character counter).
4. **Shared-selector extraction deferred** (see design notes) — flag if you want it in this PR instead.

## QA

- `npm test` — 39 affected suites / 394 tests green, incl. new `WithdrawCurrencyList.test.tsx` (currency-first + country fallback + EUR disambiguation) and `withdraw-currencies.test.ts`.
- Manual: `/withdraw` → currency list leads; tap **EUR** → SEPA countries expand; tap **GBP/USD/MXN** → straight to the form; **Other countries** reveals the full list; SEPA form shows name + IBAN only after a valid IBAN; paste icon fills each field.

## Screenshots

⚠️ Not captured in this pass (the withdraw flow needs an authed seeded user + the sandbox stack, contended on this box). Before/after is described above; Hugo QAs live. Happy to add a captured set if wanted.


Hugo0:

PR #3266, opened 2026-09-18
fix(receipt): resolve P2P receipt PDF by intent uuid, not tx hash

Hugo0:
## Summary
Sharing/downloading the receipt PDF from the P2P (DIRECT_TRANSFER) payment **success** screen failed with a 404 ("receipt PDF unavailable").

Root cause: `PaymentSuccessView` built the success drawer's synthetic transaction with
`id: paymentDetails?.payerTransactionHash ?? chargeDetails.uuid` — i.e. the **on-chain tx hash**, not the history-intent uuid. The receipt page and its PDF twin feed `transaction.id` straight into `GET /receipt/[entryId]/pdf`, which resolves the entry via the backend `GET /history/:id`. For a `DIRECT_TRANSFER` the backend accepts **only** `transaction_intents.id` (the charge uuid) as the lookup key — a tx hash fails the uuid regex and 404s. (QR pay was unaffected: `useQrReceipt` already uses a resolvable id.)

The resolvable id is available at success-render with no race: the intent row is created synchronously at charge-create (before the on-chain send), and `chargeDetails.uuid === transaction_intents.id`. So this is the "id is genuinely available" case — fix option (a): feed the uuid.

## Fix
`PaymentSuccessView.tsx`: `id: chargeDetails.uuid` (drop the tx-hash fallback). The on-chain hash still travels in the separate `txHash` field, so the Tx-ID receipt row is unchanged. One line + comment.

## Layer choice
Frontend-only. The backend already resolves `DIRECT_TRANSFER` correctly by intent uuid, and the authenticated-owner PDF path already exists on `dev` (`/receipt/[entryId]/pdf` forwards the session and renders private kinds for the owner). The only defect was the FE handing it an unresolvable id. No matched `peanut-api-ts` PR needed.

## Scope guards
- QR receipt path untouched (`useQrReceipt`).
- P2P-vs-QR privacy model unchanged (DIRECT_TRANSFER stays a private, owner-authenticated receipt).

## Tests
`src/features/payments/shared/components/__tests__/PaymentSuccessView.receipt-id.test.tsx` — renders the success view for a P2P send and asserts the receipt transaction carries the intent uuid as `id` (so `/receipt/[id]/pdf` resolves), the tx hash in `txHash`, and kind `DIRECT_TRANSFER`. Verified red before the fix (received the tx hash), green after.

## QA
Sandbox: complete a P2P send → success screen → "See receipt" → "Share receipt" (private-kind PDF share). Before: 404 / "receipt PDF unavailable". After: the receipt PDF resolves and shares.

## Docs / Legal
None. Pure id-correctness fix — no new data leaves our systems, no new processor/category/retention change.

## Related task
Follow-on defect of "Generate receipts pdfs for all transactions" (API #1599 + UI #3159 private-receipt PDF work). Formal TASK link left for Hugo to confirm — see Decisions in the handoff.


Hugo0:

PR #3263, opened 2026-09-18
fix(deposit-accounts): VA QA wave 2 — rate-pair bug, phantom account cap, hub order, receipt footer

Hugo0:
## Summary

VA QA follow-up **wave 2** — deposit-accounts + add-money polish and two real bugs. Several items in the original brief were already shipped on `dev`; this PR carries the ones that were still open, plus the receipt items.

**Task:** VA QA follow-up wave 2 _(no Notion `TASK-XXXXX` supplied by the requester — please attach one; draft PRs are not auto-reviewed)._

## What changed (by brief item)

- **#4 BUG — EUR "see exchange rates" opened the reversed pair.** `DepositFeeLine` built `?from=USD&to=<currency>` (USD→EUR); on a foreign-currency account the rate the user wants is `<account currency>→USD`. Fixed to `?from=<currency>&to=USD`. Test assertions updated (`ClaimAccountScreen.test.tsx`).
  - `src/features/deposit-accounts/components/DepositFeeLine.tsx:7-10`

- **#5 BUG — phantom "you already have two accounts" after a failed claim.** After a failed claim the refetched `claimable` can carry `blockedBy: 'account-limit'` (on staging the shared Bridge customer already holds accounts), and because the URL `screen` param is still `claim`, the gate branch rendered the cap screen for a user holding **zero** accounts. Guarded client-side: the account-limit screen only renders when the client can actually see held accounts; otherwise the honest "we cannot open an account → support" screen shows. New reason `'blocked'` routes that to the same support door.
  - `src/features/deposit-accounts/components/DepositAccountsFlow.tsx` (gate branch + `onContactSupport` type)
  - `src/features/deposit-accounts/components/DepositAccountsFlowContainer.tsx` (support message for `blocked`)
  - Tests: `depositBlocks.test.tsx` — cap with a genuinely-held account still shows; cap with zero held accounts shows the generic support screen and calls `onContactSupport(corridor, 'blocked')`.

- **#3 Bank above crypto.** In production standing accounts are dark, so the hub renders the crypto row *above* the "Other countries" (bank) section — i.e. crypto-first. Moved the crypto row to render **below** the countries section in that state. When accounts are shown, bank corridors were already above crypto (unchanged).
  - `src/features/deposit-accounts/components/DepositAccountsListScreen.tsx`
  - Test: `DepositAccountsListScreen.test.tsx` — "renders the bank options above crypto while accounts are dark".

[...]

PR #3260, opened 2026-09-18
fix(deposit-accounts): VA QA copy polish — family bullet, share text, es copy

Hugo0:
## Summary
Three copy/DS fixes from virtual-accounts (deposit-accounts) QA. Copy only — no behavior change, no new corridors, no money paths touched.

1. **Missing bullet — family-surname exemption.** In the USD "who can pay you" list, the rule *"Family who share your surname: any amount"* rendered with no bullet: it was concatenated onto the *"From another person: less than $4,000 each time"* line. `individualLines` now resolves the per-payment cap and the family exemption as **two** rule lines, so the exemption gets its own bullet on screen (and its own catalog entry in every locale).
2. **Trim the shared/copied bank-details text.** Sharing or copying the account details pasted the full who-can-pay rules paragraph into the message. The copied string is now **account fields + footer only** (`Sent from Peanut · peanut.me`). The rules still render on screen for the holder — only the copied/shared string changed. `ShareTextCopy.rules` and the payer-voice computation in `DepositShareActions` are removed.
3. **es-419 / es-AR wording.**
   - es-419 `rules.ownAccount.line`: *"Desde tu propia cuenta: sí"* → *"Desde tu propia cuenta: permitido"* (also inherited by es-AR, which does not override that line).
   - es-AR `claim.benefitBalance`: *"La plata llega a tu saldo…"* → *"El dinero llega a tu saldo…"*. See the note below — the "plata" string QA saw lives in the **es-AR** overlay, not es-419.

## Task
VA QA follow-up.

## Risks / breaking changes
- No money/identity behavior change; no API contract touched; corridor set unchanged.
- i18n: en / es-419 / pt-BR catalogs each gain the split `individualCapFamily` string and keep full key parity; es-AR is a deltas overlay and its now-redundant `individualCapFamily.why` override was removed so it inherits es-419. `messages.test.ts` parity + all deposit-accounts suites pass.

## Design notes / accepted trade-offs
- After #2, `ResolvedRuleLine.payer` and the `rules.*.payer` catalog strings (4 locales) have **no remaining consumer**. Left in place to keep this PR pure copy and avoid a risky 4-locale deletion + parity churn — flagging as a small follow-up, not fixing here.

## es-AR flag for Jota (item 3 — Jota confirms)
[...]

PR #3267, opened 2026-09-18
fix(deposit-accounts): VA QA hub polish — gated accounts, EUR IBAN honesty, clean rows, limit + minimum copy

Hugo0:
## Summary

Five polish items from Kushagra's QA of the deposit-accounts (virtual accounts) feature, on the add-money hub and the claim screen. UI + copy only — no money, provider, or shared-state logic.

1. **Show VA options to unverified users.** A new/unverified user (no KYC) saw none of the standing-account corridors — the hub filters by the rails the capability block returns, and an unverified user has none. The hub now also shows the *claimable* corridors with a disabled **"Verify to unlock"** badge, but only where that corridor's gate is `needs-identity` (identity not cleared). A verified user whose region simply has no rail resolves to `needs-enrollment` and is still spared rows they can never open — this is what keeps a German user from seeing a USD/ARS row.
2. **France→Lithuania IBAN honesty.** The EUR corridor is a shared SEPA account and Bridge issues its IBAN in Lithuania; picking "France" implied a French account. The claim "good to know" now says plainly that a euro (SEPA) account works across the Eurozone and the IBAN may be issued in another EU country. (The hub already labels the corridor currency-first, "EUR · SEPA" with an EU flag.)
3. **Row subtitles removed.** The per-row subtitle was inconsistent — arrival speed for EUR/GBP/USD/MXN, an eligibility caveat for the residence-gated rows. Removed for one clean `currency · rail` + status-badge column (Aleks' call, Hugo confirmed). Status is on the badge; the residence rule is on the screen behind the tap.
4. **Signal the 2-account limit.** A short note in the accounts section — the cap is **two** ("Two bank accounts is the default" — the existing gate copy; support opens more).
5. **Minimum in the FAQ/help copy.** The corridor minimum (e.g. €1.00) now also renders in the claim "good to know" aside, not only in the payer rules.

## Task

VA QA follow-up — no dedicated task yet, same as the sibling QA PRs (#3259, #3260, #3263). Relates to the deposit-accounts feature (TASK-22589, ui#3134). Task/assignee link is Hugo's call.

## Scope boundaries (three VA-QA PRs in flight)

Stayed out of the other two PRs' scope: **#3263 wave 2** owns the crypto row icon, bank-above-crypto ordering, merchant-logo, and the phantom-account-cap fix; **#3260 copy polish** owns the BRL exclusion, the "who can pay you / not in your own name" resolver, the missing bullet, the share trim, and es-419 existing strings. New i18n keys here are additive only.

## Risks / notes

- `DepositAccountsListScreen.tsx` and the i18n JSONs are also touched by #3263 / #3260 — additive changes, but a rebase-time conflict on those files is possible. Trivial union merges.
- Item 4 (limit note) is adjacent to #3263's phantom-account-cap fix — worth a glance that the copy and the fix agree.
- `list.rowRevoked` / `list.rowBlocked` catalog keys are now unused (item 3 removed their only reader). Left in place to avoid editing es-419 strings the copy PR owns; safe to delete in a later sweep.

## QA

- `npm run typecheck`, full `npx jest` (8346 tests), `npm run build`, `pnpm prettier --check .` — all green locally.
- New tests in `DepositAccountsListScreen.test.tsx`: unverified user sees the claimable corridors disabled with the verify badge; a verified/needs-enrollment user does not; the account-limit note renders. Existing subtitle/caveat assertions updated to the new badge-only rows.


Hugo0:

PR #3254, opened 2026-09-18
feat(rewards): delete the three reward drawers, invite friends back to a modal (TASK-22680)

kushagrasarathe:
## Summary

Celebration moments carry the App Store review prompt and the invite-friends prompt. A compact drawer weakens the moment and leaves those prompts no room. The September sweep turned these surfaces into drawers; this reverses that for the reward surfaces specifically.

Three reward drawers go instead of getting a facelift, and the invite sheet returns to a modal.

**Deleted outright**
- `PerkClaimDrawer` — with it, the Card Pioneer claim path on `/home`
- `PerkClaimSuccessDrawer`
- `WelcomeUnlockDrawer` — the KYC-unlock celebration
- orphans that fell out: `usePerkClaimFlow` + its test, `perkClaim.consts.ts`

**Converted back**
- `InviteFriendsDrawer` → `InviteFriendsModal`, on `ActionModal` — the same primitive `MigrationDownloadModal` uses, which went drawer→modal one day after its own conversion (`efb93f9ce` → `4c9e7528f`). This generalises that revert.

**Also here**
- `/dev/reward-surfaces` — a dev-only inventory of every reward overlay, its presentation, its real importers and its gating. Built to make this decision with evidence; kept because the task asks for the list. Blocked on peanut.me.

Unchanged and still drawers: `NoMoreJailDrawer`, `BadgeDetailDrawer`, `BadgeStatusDrawer`, `EasterEggDrawer` — detail views, not the moment.

## Task

TASK-22680 — Audit reward and celebration overlays; remove deprecated screens.

The rule change is documented separately in mono PR #205 (`design/design.md`).

## Risks / breaking changes

- **The `reward_claimed` app-review trigger is gone** with `HomeCarouselCTA`'s perk path. Three triggers remain: `payment_completed`, `qr_payment_completed`, `money_received`. Accepted deliberately.
- **Nothing celebrates activation any more.** `KycVerificationInProgressModal`'s terminal copy was flattened to "All set" *because* home owned that celebration. That reason is now void. Comment updated to say so; copy unchanged.
- **One behaviour change in `HomeModals`:** `showKycModal` was also a suppressor in the balance-warning condition. A high-balance user who just passed KYC can now see the balance warning on that visit. Priority order otherwise holds: migration > notifications > post-signup > balance warning.
- No backend change. No cross-repo deploy order.

## Design notes / accepted trade-offs

[...]

PR #3269, opened 2026-09-18
chore(perks): retire the client perk claim plumbing (TASK-22680)

kushagrasarathe:
## Summary

Follow-up to #3254, from Chip's review of it. That PR removed the home-carousel perk claim surface. It left the client plumbing behind it half-deleted: a live socket handler with no subscriber, and an API client nothing calls. No runtime defect — the failure is that the next reader assumes real-time perk pushes still drive a UI.

- `useWebSocket.ts` — drop the `onPendingPerk` option, the `pending_perk` listener, its dispatch branch and cleanup.
- `services/perks.ts` — drop `perksApi` (`getPendingPerks`, `claimPerk`) and its response types.
- Tests — drop the `perksApi` mock and the two assertions that only checked a now-absent API was not called.

## Task

TASK-22680.

## Why perks.ts is kept, not deleted

The `PendingPerk` type has two live consumers — `services/websocket.ts` (message union) and `PerkClaimGiftBox.tsx` (prop). The file stays as that type plus a tombstone comment naming what replaced the claim call: server-side auto-claim on card activation, `CARD_SPEND` and `FUND_WITHDRAW`, and the QR payout retry job. Deleting it would have cascaded past the point this PR is worth.

## Risks

Low. No user-facing behaviour changes and no UI is touched — a hook option, a service module and test scaffolding only.

Production evidence that the removed claim call was already dead: the `Card Pioneer Inviter Reward` perk has **no perk row and zero usages** in the prod database, so no user had a pending reward routed through it. Every live issuance path auto-claims server-side or auto-pays via the QR payout job.

## Deliberately out of scope

- `PerkClaimGiftBox.tsx` and `'reward_claimed'` in the `AppReviewTrigger` union — open decisions, not this PR's call.
- `usePerkHoldToClaim` — the QR-pay reveal gesture. It makes no server call and is untouched.
- **peanut-api-ts still emits `pending_perk`** (`src/routes/charges-ws/index.ts`, `src/events.ts`). Nothing subscribes to it now. That is a backend change, so it needs its own PR — never bundled here.

## QA

No user-visible surface. Gate: prettier, typecheck, jest 679/679, build — all green locally.


kushagrasarathe:
/chip review

Follow-up to #3254, raised by your own review of it. Focus on whether retiring the pending_perk socket handler and perksApi can strand anything: confirm no production path still needs a client-side claim call, and that keeping PendingPerk (for websocket.ts and PerkClaimGiftBox) rather than deleting perks.ts outright is the right cut.

PR #3273, opened 2026-09-18
feat(dev): token selector — DS proposals for the picker, its rows and its selected state

kushagrasarathe:
## What this is

A proposal page, not a migration. `Global/TokenSelector/*` is **untouched** — the diff is one new `/dev` route plus a six-line entry in the dev index.

Open it at **`/dev/token-selector`**.

The TokenSelector is the last large surface the design system never reached. Rebuilding it is blocked on three rulings, and this page asks for all three on real components with real data, with every option rendered side by side.

## Q1 — what does a SELECTED list row look like? (the main one)

`ListItem` has no `selected` prop. `grep selected src/styles/globals.css` returns nothing. There is no selected-state token, and `bg-background-disabled` is already spent twice in `ListItem.tsx:77-78` — on `active:` (pressed) **and** on `disabled` — so it cannot also mean selected.

`Tabs.tsx` says this itself after TASK-22707, verbatim in its own docblock:

> Because there is no fill, this look says NOTHING about how a selected list row should look. The app-wide selected-surface question is still open — see ui#3232's pink NetworkListItem.

**Six visual languages for "selected" ship today, carried by four different ARIA attributes:**

| look | where |
|---|---|
| solid `bg-action-primary` | `NetworkListItem.tsx:50`, `CountryCombobox.tsx:226`, `BaseSelect.tsx:121`, `BalanceSection.tsx:93` |
| `bg-action-primary/10` | `TokenListItem.tsx:63` **and** `:74` (twice, nested), `CardCountryConfirmScreen.tsx:81`, `RelationshipTable.tsx:118` |
| `bg-action-primary/20` | `LocaleSwitcher.tsx:126`, `ContentLanding.tsx:158` |
| solid pink + `foreground-inverse` + check | `CountryCombobox.tsx:226`, `BaseSelect.tsx:121` |
| border only | `CurrencySelect.tsx:251`, `AvatarPicker.tsx:164` |
| trailing check, no fill | `LanguageView.tsx:72` — the only `ListItem`-native one |

ARIA: `aria-pressed`, `aria-selected`, `aria-checked`, `aria-current` — and `TokenListItem` has **none at all**, so a screen reader cannot tell which token is picked.

ui#3232 (merged 2026-09-17) made `NetworkListItem` solid pink and explicitly deferred `TokenListItem`. So the network list and the token list **inside one drawer** disagree right now.

**Four skins on the page**, identical rows and data, differing only in the selected-state classes. `role="listbox"` / `role="option"` / `aria-selected` are on all four — that is the floor, not a choice.

1. **Fill** — `bg-action-primary` + `text-foreground-over-color-primary`
2. **Check** — trailing 20px `check`, no fill
3. **Border** — unselected rows drop to `border-border-subtle`
4. **Fill + check (recommended)** — both channels

**Why 4.** WCAG 1.4.1 says colour must never be the only signal, and the DS's own ruled Tabs look already carries two channels (colour + weight) for exactly that reason. 4 is a strict addition to what ui#3232 already shipped — the fill stays, the second channel goes on top.

**One tension the page names rather than hides:** `text-foreground-over-color-primary` exists precisely for text on a brand fill and has zero product call sites — but the DS audit (`audit-data.ts:5613`) lists it among four tokens to **delete** as exact hex duplicates. It is the same `#000000` as `foreground-primary`, so the pixels are identical either way. Ruling this skin also rules whether that token lives.

## Q2 — the leading slot

design.md, ListItem section: *"never a hand-rolled composite — no mini-badge overlaid on a logo."* `TokenListItem.tsx:97-111` and `TokenSelector.tsx:545-555` both do exactly that — a 24px token logo with a 16px chain logo pinned at `-right-1 -bottom-1` inside a `border-2 border-white … dark:bg-gray-600` ring. (Those `dark:` classes are dead — `layout.tsx:160` hardcodes `data-theme="light"`.)

- **L1 (recommended)** — token logo alone, chain on the body line. The tabs above already filter by chain, so the badge was repeating it.
- **L2** — chain as a trailing `StatusBadge` (badge is in the sanctioned trailing set).
- **L3** — today's composite, shown so the comparison is honest.

## Q3 — the trigger

Today a `Button variant="stroke"` overridden into a card: shape, height, padding and its own hover state all respelled at the call site, which law 7 puts in `.btn-*` / `Button` and nowhere else.

- **T1 (recommended)** — a real `ListItem position="single" chevron`. A row that opens a picker is a list row.
- **T2** — today's fake button-card.

## Section 7 — the composed picker, interactive

Built only from shipped DS parts. Three structural changes independent of the skin ruling:

- **the search field sits above the tabs.** Today it lives inside the tab panel (`TokenSelector.tsx:411`), so switching a tab unmounts it and the caret is lost mid-typing.
- **the rows are one grouped card** (`getCardPosition`), not N solo cards in a `gap-3` stack.
- **search is substring.** Today's is exact-match (`TokenSelector.tsx:307-314`) — typing `usd` returns nothing.

## Base branch — read this

This PR is based on **`fix/22707-tabs-absorb-segmented-control` (ui#3251)**, not `dev`, so the network row renders the ruled **Weight** tabs. Because `content` is optional there now, the tabs are a **contentless trigger row** with the list rendered beneath — instead of the same list element handed to all five tabs, which is what ships today.

Measured on the running page:

```
tabs:        All | Arbitrum | Base | Optimism | Ethereum
active:      font-weight 600, 16px, rgb(0,0,0)     → text-body-m-semibold + foreground-primary
border-bottom: 0px      background: transparent    → no rule, no fill
role=tabpanel count:   0                           → contentless trigger row
listboxes / options on the page:  8 / 42
```

**Retarget to `dev` once #3251 merges.** Basing on `dev` instead would have swallowed all of #3251 into this diff.

## Section 8 — conformance

20 rows, each `file:line` → what the proposal does → the design.md rule. Highlights beyond the above: a local `Section` duplicating `0_Bruddle/Section`; `chevron-up` rotated 90° at size 24 where the board says a 20px `chevron-right`; `font-medium` stacked on type tokens in two places; a hand-rolled attention banner instead of `Notification`; a page-level `NavHeader` rendered inside a `DrawerContent`; `space-y-2` + `gap-3` on one column (20px, off-scale); soft `shadow-sm` under hard `shadow-4`; no skeleton anywhere.

**One is a live defect, measured in a browser at 375×667:**

```
NetworkListItem row — .btn h-11 = 43.99px, Card inside = 65.10px
card starts 10.55px ABOVE its own button, overflow: visible
```

`Button` always applies `.btn` (`Button.tsx:190`) and the call site's `p-0` kills the padding, not the height. 65px of content in a 44px control, spilling ~10.5px each side. The pink fill is the element that overflows, so it spills past the control it belongs to.

## Flagged, not invented (law 6)

- `ListItem` forwards no `role` / `aria-selected` and has no `selected` prop, so the listbox semantics ride a wrapper element and the row loses ListItem's own click haptic. The promotion PR should put `selected` + ARIA on `ListItem` itself rather than repeat that wrapper.
- `fullWidth` on the network tabs is deliberately **not** used — five chain labels stretched across 375px either truncate or push the labels off-scale. The scrollable `w-max min-w-full` row is the right shape here.
- Conformance rows 19 (no skeleton) and 20 (a hardcoded `assets.coingecko.com` URL baked into the component) are flagged, not answered — both need a board.

## QA

| gate | result |
|---|---|
| `pnpm prettier --check .` | clean |
| `npm run typecheck` | exit 0 |
| `npm test` | 8350 passed, 3 failed |
| `node scripts/ds-lint-counts.mjs --check` | ratchet ok — no metric increased |
| `npm run build` | compiled, `/dev/token-selector` static |

The 3 failures are the known timezone ones — `AddWithdrawCountriesList` ×2 and `UnlockPayments` ×1, all asserting `/Sep 8/` against a date that renders `Sep 9` locally. Neither file is in this diff.

No screenshots: element boxes resolved correctly in the headless session but the compositor returned stale pixels, so anything I attached would have misled. The page is verified by measured computed styles above, and it is a `/dev` route — open it.

## Next

Pick a skin, a leading option and a trigger. The winner is applied to the real `TokenSelector` in a separate PR, with a figma board per design.md, and this directory is deleted then.


kushagrasarathe:
Superseded by #3274 — same page, correct base. This one was cut from the newer `dev` and then merged with #3251's branch, so the diff against #3251 swallowed every newer dev commit (41k lines). #3274 is branched directly off `fix/22707-tabs-absorb-segmented-control`, so its diff is the 3 files this change actually adds.

PR #3277, opened 2026-09-19
fix(kyc): stand the hosted task down beside a native step, and settle after the return (TASK-22818)

abalinda:
## Summary

Client half of TASK-22818. A Bridge user who owed a proof-of-address document saw two tasks for one requirement, took the hosted one (Persona identity, which cannot collect the document), came back to an app that looked unchanged, and ran it again several times. The resolver fix is peanutprotocol/peanut-api-ts#1639; this PR closes the client side of the loop, in three parts:

- **The hosted task stands down while a Bridge rail carries a native `sumsub` step** (`selectBridgeTasks`, now rail-aware). The API no longer emits the pair; this keeps older API responses honest on Home and on the prep screen, which now read the same selection. Advisory (future-dated) hosted tasks stay.
- **Coming back opens a settle window** (`useHostedVerification`): refetch at once, ask the API to re-read the provider (`POST /users/kyc/refresh`, peanut-api-ts#1640, best effort), and keep asking every 5s for a minute until the task clears or the window ends. The provider's page never hands the user back (Bridge issues no redirect for most customers), the ~4s auto-refresh only runs while a rail is `pending`, and a webhook can arrive hours later or never. Same handling for every return signal: in-app browser close (iOS and Android), the programmatic close of the universal-link leg, tab focus on web, BFCache restore.
- **The prep screen says what is happening.** While the window runs the CTA is held with "Checking with our payment partner…"; if the task is still pending when it ends, the screen says there is nothing more to do and that starting again creates a new check, instead of re-offering the same button in silence. Copy in en, es-419, pt-BR.

Not in this PR, deliberately: the tab-launch `SecurityError` on Chrome Mobile and Mobile Safari (Sentry PEANUT-UI-T07, T4W). Both engines allow the current order in a headless probe, so a reorder would be a guess; it needs a device-level look.

## Task

TASK-22818: https://app.notion.com/p/3df8381175798102ac1de417db7f6629

## Risks / breaking changes

- Cross-repo: `refreshKycState` calls a route that lands with peanut-api-ts#1640. Until it deploys the call answers 404 and reads as "not refreshed"; the window then only refetches. Safe in either deploy order.
- Native: no layout, positioning, input or viewport change. Two new `Notification` rows on the prep screen. Reaches installed apps through the normal OTA path.
- The settle window issues up to 13 refresh calls per return; the route is rate-limited server-side.

## QA

- Unit: 55 tests across the four touched suites, 8 new (selector stand-down and its non-triggers; card hides beside a native step; settle window rounds, early stop when the task clears, still-pending state, reset on a new start; the return leg now also asks the API). Full suite green locally.
- Screenshots: ⚠️ the two new prep-screen states are transient (they exist only after a return from the vendor) and no fixture can stage a return; the Home change is a removal for the double-task cohort. Visual review needs a sandbox Bridge customer with a proof-of-address document due; the fixture-covered screens are unchanged and stay under `ds-shots`.


abalinda:
Closing: this change goes to dev first, as peanutprotocol/peanut-ui#3278 (the same four commits cherry-picked onto dev, identical files, currently a draft). It reaches main with the next release train instead of as a hotfix. Review history stays here: Chip clean at 06a55711 with one advisory, fixed in the two commits after it; Chip's automatic re-review of 2b336eb7 failed on its side six times.

